PR #45, opened 2018-12-17
Correcting the reference to googlepay in Proguard configurations

tinsukE:
The current Proguard configurations are using the old "googlewallet" package instead of "googlepay".


timon-langlotz:
Hi @tinsukE,

Nicely spotted, and thanks a lot for pointing it out! We currently do not accept pull requests, since this repository here is just a mirror of our internal repository. But we'll make sure to fix this for our next release!

Regards,
Timon

Issue #48, opened 2019-02-12
Google Pay doesn't work in the example app unless your language is set to English


lbenedetto:
1. Set your phones language to a locale that uses commas in prices (ex $1,99)
2. Try to buy with Google Pay in the example app

"Request failed. Unexpected developer error, please try again later"

Most likely due to GooglePayUtil.java:156 where you are formatting the display amount. The code looks like it would format it as $1.99 but it is actually getting formatted to $1,99. If I edit the code in that method to replace the , with a . then Google Pay works.

caiofaustino:
Hi Lars,

Thanks for catching this. Unfortunately we do not accept pull requests, but I made a similar fix and it should be part of the next release.

Cheers!

lbenedetto:
Any idea when the next release is? 

Was planning on releasing an update to my app tomorrow, and would like for it to include this fix.
If its only a few days away, I might be able to delay my release.

caiofaustino:
Sorry for the delay. The new release has just been published! Check for 2.3.2

Cheers :)

riddhee:
@lbenedetto can you please help me to integrate G Pay

PR #49, opened 2019-02-12
Fix Google Pay bug

lbenedetto:
Fixes bug from [Issue#48](https://github.com/Adyen/adyen-android/issues/48)


caiofaustino:
Unfortunately we do not accept pull requests, but a similar fix will be available on the next release.

PR #63, opened 2019-04-18
Added Thai localization

alex-sukharev-cs:
Hello,

As discussed with Adyen support (request #636787), we would like to suggest this Thai localization made by professional translators.

Thanks.

### DIFF
--- a/checkout-ui/src/main/res/values-th-rTH/strings.xml
+++ b/checkout-ui/src/main/res/values-th-rTH/strings.xml
@@ -1,0 +1,107 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="checkout_title">เช็คเอาท์</string>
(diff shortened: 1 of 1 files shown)


caiofaustino:
Hey Alex,

Sorry for the delayed response, I think there was a bit of miscommunication. Maybe support was talking about the Web SDK, anyway.

We cannot accept your pull request as this would imply we provide future support to Thai in new strings that come up in the future, which we might not be able to do.

But you are more than welcome to add the translation file to your own project, which will have the same effect, as Android will be able to get the translated strings correctly. If you have any issues doing that let me know and we can try to help you.

I think iOS has a similar feature where you ca provide your own translated strings, they will reply to you there.

Cheers,

alex-sukharev-cs:
@caiofaustino
Got it, thank you!

PR #71, opened 2019-05-24
Fixed code example

igorwojda:


caiofaustino:
Sorry, we don't accept PR. And this change is not correct, you receive the `onRequestPaymentSession` callback from the SDK and need to forward that to your server. The server, talks to Adyen and gets the response, you then put that response back on the SDK in the `handlePaymentSessionResponse`.
Only then you will receive the `onPaymentInitialized` callback to start the payment flow.

igorwojda:
Something still does not look right
both 
`PaymentController.handlePaymentSessionResponse`
and
`CheckoutController.handlePaymentSessionResponse`

accept `com.adyen.checkout.core.handler.StartPaymentParametersHandler` as last parameter. `StartPaymentParametersHandler` does not have `onStartPaymentParameters ` method

image

caiofaustino:
Ah I'm sorry, you are correct. I got confused with another part of the documentation.
I'll apply this change and it will be available on the next release.
Thanks for spotting it :)

PR #97, opened 2019-08-21
Make Adyen3DS2Component return error instead of crash in case transaction missing on challengeShopper

moia-zsolt:
Please see related iOS SDK pull request https://github.com/Adyen/adyen-ios/pull/101

We ran into a situation where we received ChallengeShopper after we responded for ChallengeShopper, which makes the app crash. The iOS SDK has detailed steps.
We are currently in talks with Adyen support to find the issue in this "shouldn't happen" scenario.

iOS SDK developers agreed to handle this problem regardless, so I'd like to offer the matching fix for Android SDK.

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.java
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.java
@@ -175,6 +175,11 @@ public final class Adyen3DS2Component extends BaseActionComponent implements Cha
         closeTransaction(getApplication());
     }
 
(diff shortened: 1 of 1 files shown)


caiofaustino:
Hey, thanks for the catch. I'll also update this on our side to give an error on the `observeErrors()` observer instead of crashing. 

We don't receive pull requests but I'll put the fix in out next release.

Thanks

moia-zsolt:
@caiofaustino Thanks a lot!

moia-zsolt:
For the record: fixed in Adyen Android SDK 3.2.1.

Issue #112, opened 2019-09-25
NullPointerException if view just added to layout


Kolyall:
just add to your layout
```
    <com.adyen.checkout.card.CardView
        android:id="@+id/adyenCardView"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintLeft_toLeftOf="parent"
        app:layout_constraintRight_toRightOf="parent"
        app:layout_constraintTop_toTopOf="parent" />
```
and start app

Error occurs due to https://github.com/Adyen/adyen-android/blob/c7602983d6c1e8de4a50d4d62322277cc410ebbd/card-ui/src/main/java/com/adyen/checkout/card/CardView.java#L101
```
Process: com.github.adyenexample, PID: 489
java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.adyen.checkout.card.CardComponent.isStoredPaymentMethod()' on a null object reference
    at com.adyen.checkout.card.CardView$2.onFocusChange(CardView.java:101)
    at android.view.View.onFocusChanged(View.java:7573)
    at android.widget.TextView.onFocusChanged(TextView.java:10942)
    at android.widget.EditText.onFocusChanged(EditText.java:277)
    at android.view.View.handleFocusGainInternal(View.java:7245)
```

caiofaustino:
We will release a fix for this issue on the next version.

ArmanSo:
@Kolyall We released version 3.4.1 that includes fix for this issue.
Can you please confirm it?

PR #113, opened 2019-09-26
Update CardView.java

Kolyall:
Added check null for mComponent


Kolyall:
fix of https://github.com/Adyen/adyen-android/issues/112

caiofaustino:
Thanks for the implementation, but unfortunatelly we don't take external pull requests. We will discuss internally and make a similar fix soon.

PR #207, opened 2020-08-20
feat: Added new languages.

cernfr1993:


caiofaustino:
Hey @cernfr1993,

Thanks for the PR but unfortunately we are not yet accepting external ones. The added languages are nice, but at the same time we can't commit to adding languages we might not maintain support in the future, as we would have to rely on your contributions to maintain those, while the current ones have an internal maintenance structure.

PR #258, opened 2020-10-28
Release workflow opens PR to merge back to develop.

caiofaustino:
## Summary
- Create PR back to `develop` after a release. 
  - Could be just the merge commit or a hotfix branch.
- Allow manual trigger of check_pr
  - PRs from a workflow don't trigger other Actions.


caiofaustino:
Apparently there is no way to manually trigger the status check from the action required for the branch protection.
Unless we use a different credential on the Action, but that might have other implications, so parking this idea for now.
https://github.community/t/status-check-for-manual-action-workflow-dispatch/140333/2

PR #276, opened 2020-11-17
Fix clientKey validation for BaseConfiguration on DropIn.

caiofaustino:
## Summary
When no Configuration is defined for a payment method on the Drop-in, it creates a base generic one if possible. 
It sets the `clientKey` from the `DropInConfiguration` to propagate it automatically, but if there is no `clientKey` defined it tries to set the default emtpy value and throws the validation exception.

Fix: Only set the `clientKey` on generic configurations if it's not empty.

## Tested scenarios
- Reproduced issue and confirmed the problem
- Tested flow with and without the clientKey


jreij (approved):
LGTM 👍

PR #286, opened 2020-11-27
Refactor payment methods adapter

caiofaustino:
## Summary
- Create view data models and refactor payment methods list generation on the DropInViewModel.
- Refactor PaymentMethodAdapter to separate regular and stored PaymentMethods.
- Remove expanded logic from PaymentMethodAdapter as it will be replaced with it's own fragment.

This if part of a bigger refactor, one click functionality still needs to be refactored.


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -10,84 +10,184 @@ package com.adyen.checkout.dropin.ui
...
+            }
+        }
+    }
+
+    private fun makeStoredModel(storedPaymentMethod: StoredPaymentMethod): StoredPaymentMethodModel {
+        return when (storedPaymentMethod.type) {
+            PaymentMethodTypes.SCHEME -> {
+                StoredPaymentMethodModel(
+                    storedPaymentMethod.id ?: "",
+                    "•••• ${storedPaymentMethod.lastFour}",
```
jreij, line 102: Should we do the formatting on the UI here instead? I can see we have `card_number_4digit` in the string resources and `CardView.setStoredCardInterface` already uses it

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -10,84 +10,184 @@ package com.adyen.checkout.dropin.ui
...
+                    "${storedPaymentMethod.expiryMonth}/${DateUtils.removeFirstTwoDigitFromYear(storedPaymentMethod.expiryYear)}",
```
jreij, line 103: Do we have a formatting util for this one? If not maybe we should extract one, or ideally let the UI do the formatting also. What do you think?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -10,84 +10,184 @@ package com.adyen.checkout.dropin.ui
...
+            when {
+                type == null -> {
+                    throw CheckoutException("PaymentMethod type is null")
+                }
+                PaymentMethodTypes.SUPPORTED_PAYMENT_METHODS.contains(type) -> {
+                    Logger.v(TAG, "Supported payment method: $type")
+                    // We assume payment method is available and remove it later when the callback comes
+                    // this is the overwhelming majority of cases, and we keep the list ordered this way.
+                    paymentMethodsList.add(
+                        PaymentMethodModel(type, paymentMethod.name ?: nullNameFallback)
```
jreij, line 139: What about localization here? The UI will show "Unknown" for all languages. Or is this an edge case that shouldn't practically happen?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -10,84 +10,184 @@ package com.adyen.checkout.dropin.ui
...
-                } else {
-                    Logger.e(TAG, "PaymentMethod not yet supported - $type")
+    override fun onAvailabilityResult(isAvailable: Boolean, paymentMethod: PaymentMethod, config: Configuration?) {
+        Logger.d(TAG, "onAvailabilityResult - ${paymentMethod.type}: $isAvailable")
+
+        availabilitySum++
+
+        if (!isAvailable) {
+            Logger.e(TAG, "${paymentMethod.type} NOT AVAILABLE")
+            for ((index, paymentMethodModel) in paymentMethodsList.withIndex()) {
```
jreij, line 167: We can do something like this: `paymentMethodsList.removeAll {  it.type == paymentMethod.type }` ?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
```diff
@@ -8,191 +8,150 @@
...
         }
     }
 
     override fun getItemViewType(position: Int): Int {
-        return when (getPaymentMethod(position)) {
-            HEADER_PLACEHOLDER_ITEM_P, HEADER_PLACEHOLDER_ITEM_SP -> PAYMENT_METHODS_HEADER
-            FOOTER_PLACEHOLDER_ITEM -> PAYMENT_METHODS_FOOTER
-            else -> PAYMENT_METHOD_ITEMS
+        // Top is always a header
+        return if (position == 0) {
```
jreij, line 45: We can replace this with one `when` block?

```
    override fun getItemViewType(position: Int): Int {
        // Top is always a header
        return when {
            (position == 0) -> PAYMENT_METHODS_HEADER
            noStored() -> PAYMENT_METHOD // only regular payment methods left
            position <= paymentMethodsListModel.storedPaymentMethods.size -> STORED_PAYMENT_METHOD // 0 header - stored payments
            position == (paymentMethodsListModel.storedPaymentMethods.size + 1) -> PAYMENT_METHODS_HEADER //2nd header - regular payments
            else -> PAYMENT_METHOD
        }
    }
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
```diff
@@ -8,191 +8,150 @@
...
+                    // 0 header - stored payments - 2nd header - regular payments
+                    position <= paymentMethodsListModel.storedPaymentMethods.size -> STORED_PAYMENT_METHOD
+                    position == (paymentMethodsListModel.storedPaymentMethods.size + 1) -> PAYMENT_METHODS_HEADER
+                    else -> PAYMENT_METHOD
+                }
+            }
         }
     }
 
     override fun onBindViewHolder(holder: BaseViewHolder, position: Int) {
```
jreij, line 62: Can we extract 3 methods from this? Something like `bindPaymentMethodsHeaderVH` - `bindStoredPaymentMethodVH` - `bindPaymentMethodVH` ?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -10,84 +10,184 @@ package com.adyen.checkout.dropin.ui
...
+    private var availabilitySkipSum = 0
+    private var availabilityChecksum = 0
+
+    private val storedPaymentMethodsList = mutableListOf<StoredPaymentMethodModel>()
+    private val paymentMethodsList = mutableListOf<PaymentMethodModel>()
 
-        if (isAvailable) {
-            addPaymentMethod(paymentMethod)
+    fun getStoredPaymentMethod(id: String): StoredPaymentMethod {
+        for (storedPaymentMethod in paymentMethodsApiResponse.storedPaymentMethods.orEmpty()) {
```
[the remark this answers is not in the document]
caiofaustino: Awesome, agreed!

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
caiofaustino: Good point, this shouldn't really happen, I was also considering to just leave "" for the fallback. Maybe that's better and the View can have the localized "Unknown".

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
caiofaustino: Yeah I was thinking the same thing, but I thought of having the `StoredPaymentMethodModel` generic for other payments that are not Card. Let's discuss offline.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
caiofaustino: <3

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
caiofaustino: 👍

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
caiofaustino: yeah I was considering that too, thanks

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
caiofaustino: Yep, let's do it.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
caiofaustino: As we discussed it's probably better to have typed `StoredPaymentMethodModel` sub classes and to parse the things in the View. I'll refactor this bit :)

jreij (approved):
👍

PR #288, opened 2020-12-01
Separate DropInViewModel and the fragment specific ViewModels.

caiofaustino:
## Summary
- Move the logic of the Payment Methods list out of DropInViewModel to it's own.
- Add extension functions to initialize ViewModels with Factory (thx @jreij)


jreij (approved):
LGTM

PR #289, opened 2020-12-04
Re-add OneClick functionality

caiofaustino:
## Summary
- Create Fragment to handle stored payment flow
- Remove SynteticAccessor check on Lint


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.java
```diff
@@ -61,7 +60,6 @@
 import java.util.List;
 
 public final class Adyen3DS2Component extends BaseActionComponent<Adyen3DS2Configuration> implements ChallengeStatusReceiver {
-    @SuppressWarnings(Lint.SYNTHETIC)
```
jreij, line 64: I have a few of those in my other PR, I'll remove them also 👍

jreij (approved):

gmorales:
@caiofaustino dou you have an ETA for this change to be released. It's the only part that is not configurable yet in the stored payments method and we would need to take advantage of this change.

caiofaustino:
This exists in all available releases already, can you explain a bit better. What do you mean by not configurable?

gmorales:
Sorry about misunderstanding, we're using version drop-in version 3.8, and what we need is this specifically: **Fix secondary button layout**.

This [style](https://github.com/Adyen/adyen-android/pull/289/commits/88b7430b6cd4e6c2ec97da39c4630739e92105f6#diff-336df784bcc3a2835fb3e4bd9ce34678e7a262bfcb433194eb9ad01bec306fe1R39) has been included here, but is not available on version 3 https://github.com/Adyen/adyen-android/blob/v3/drop-in/src/main/res/layout/payment_methods_list_footer.xml#L21.

PR #296, opened 2020-12-09
Change 3DS2 Action type structure

jreij:
## Summary
- Merge both 3DS2 fingerprint and 3DS2 challenge actions into one 3DS2 action with subtypes

### DIFF
--- a/base-v3/src/main/java/com/adyen/checkout/base/model/payments/response/Threeds2Action.java
+++ b/base-v3/src/main/java/com/adyen/checkout/base/model/payments/response/Threeds2Action.java
@@ -15,7 +15,6 @@ import androidx.annotation.Nullable;
 
 import com.adyen.checkout.base.util.ActionTypes;
 import com.adyen.checkout.core.exception.ModelSerializationException;
-import com.adyen.checkout.core.exception.NoConstructorException;
 import com.adyen.checkout.core.model.JsonUtils;
 
 import org.json.JSONException;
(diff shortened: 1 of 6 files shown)


base-v3/src/main/java/com/adyen/checkout/base/model/payments/response/Threeds2Action.java
```diff
@@ -79,4 +84,48 @@ public String getToken() {
...
+            case SubTypes.FINGERPRINT:
+                return SubType.FINGERPRINT;
+            case SubTypes.CHALLENGE:
+                return SubType.CHALLENGE;
+            default:
+                return null;
+        }
+    }
+
+    private static final class SubTypes {
```
caiofaustino, line 117: What do you think about passing the token name as an argument in the `enum` itself instead of having this separate class for it?

caiofaustino:
Apparently this is not as simple as I thought, let's wait to discuss the API changes with the team to make surel

base-v3/src/main/java/com/adyen/checkout/base/model/payments/response/Threeds2Action.java
jreij: Hmmm we would still need to have theses 2 constants (`FINGERPRINT` and `CHALLENGE `) somewhere though right?

caiofaustino: Yeah but they can be on the top level class, I just think this `SubTypes` class is a bit unnecessary.

caiofaustino (approved):
Just need to fix the code checks, but looks good to me.

caiofaustino (approved):

PR #297, opened 2020-12-11
Frictionless flow for stored payments.

caiofaustino:
## Summary
- When payment method doesn't need input, we make the One Click flow more direct.
- Added `requiresInput()` method to check if this flow can be used.


jreij (dismissed):
Looks good :+1: Just left one minor comment

jreij (approved):

PR #308, opened 2020-12-18
Merge component modules

caiofaustino:
## Summary
Merging and renaming the base and UI modules of Components.
- Separation of using logic individually without UI didn't turn out to be used as initially intended.
- Decreasing the number of modules should improve build times.
- Naming of modules is now more straight forward and easy to understand.
- Using plugin block in gradle files.


jreij (dismissed):
🎉

jreij (approved):

PR #311, opened 2020-12-23
Refactor Component initialization

caiofaustino:
## Summary
- Removed `PaymentComponentViewModelFactory` in favor of inline function with explicit constructor call.
- Removed `StoredPaymentMethodDelegate` in favor of explicit initialization.
- Moved some logic on the `CardComponent` to the new Delegate classes.
- Refactored the BCMC, GooglePay, WeChatPay Providers to Kotlin to use new factory.
- Refactored `GenericPaymentComponentProvider` to use old reflection initialization.


card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -79,13 +78,19 @@ class CardComponent private constructor(
...
+        val firstCardType: CardType? = if (filteredSupportedCards.isNotEmpty()) filteredSupportedCards[0] else null
+
         return CardOutputData(
-            validateCardNumber(inputData.cardNumber),
-            validateExpiryDate(inputData.expiryDate),
-            validateSecurityCode(inputData.securityCode),
+            cardDelegate.validateCardNumber(inputData.cardNumber),
+            cardDelegate.validateExpiryDate(inputData.expiryDate),
+            cardDelegate.validateSecurityCode(inputData.securityCode, configuration, firstCardType),
             validateHolderName(inputData.holderName),
```
jreij, line 90: Can this be `cardDelegate.validateHolderName` instead?

caiofaustino: Yeah, this was not supposed to be part of this PR 😬 . Maybe I can finish this part and add it as well.

card/src/main/java/com/adyen/checkout/card/NewCardDelegate.kt
```diff
@@ -0,0 +1,53 @@
...
+        cardType: CardType?
+    ): ValidatedField<String> {
+        return if (cardConfiguration.isHideCvc) {
+            ValidatedField(securityCode, ValidatedField.Validation.VALID)
+        } else {
+            CardValidationUtils.validateSecurityCode(securityCode, cardType)
+        }
+    }
+
+    override fun isCvcHidden(cardConfiguration: CardConfiguration): Boolean {
```
caiofaustino, line 42: @jreij do you think we should add the `CardConfiguration` to the constructor instead?

jreij: Hmm yeah why not

jreij (approved):
👍

PR #312, opened 2020-12-24
Revert support for issuers in details object on old API

jreij:
## Summary
- Support issuers in details for older api versions

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/model/paymentmethods/PaymentMethod.java
+++ b/components-core/src/main/java/com/adyen/checkout/components/model/paymentmethods/PaymentMethod.java
@@ -37,6 +37,7 @@ public class PaymentMethod extends ModelObject {
     private static final String FUNDING_SOURCE = "fundingSource";
     private static final String ISSUERS = "issuers";
     private static final String CONFIGURATION = "configuration";
+    // This field is returned in older API versions, only used to retrieve the issuers list
     private static final String DETAILS = "details";
 
     @NonNull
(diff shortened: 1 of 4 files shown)


issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListComponent.java
```diff
@@ -35,23 +38,45 @@
...
+    private void initIssuers(@NonNull PaymentMethod paymentMethod) {
+        final List<Issuer> issuersList = paymentMethod.getIssuers();
+        if (issuersList != null) {
+            initIssuers(issuersList);
+        } else {
+            initLegacyIssuers(paymentMethod.getDetails());
+        }
+    }
+
+    private void initIssuers(@NonNull List<Issuer> issuerList) {
```
caiofaustino, line 57: Maybe rename one of the 2 methods? I think it's a bit confusing to have 2 `initIssuers` even though they take different parameters. Your choice.

caiofaustino (dismissed):

components-core/src/main/java/com/adyen/checkout/components/model/paymentmethods/PaymentMethod.java
```diff
@@ -37,6 +37,7 @@
     private static final String FUNDING_SOURCE = "fundingSource";
     private static final String ISSUERS = "issuers";
     private static final String CONFIGURATION = "configuration";
+    private static final String DETAILS = "details";
```
caiofaustino, line 40: Maybe add a comment saying this is from the old API versions?

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListComponent.java
jreij: 👍

components-core/src/main/java/com/adyen/checkout/components/model/paymentmethods/PaymentMethod.java
jreij: 👍

caiofaustino (approved):

PR #314, opened 2021-01-04
Migrate from Kotlin synthetics to Jetpack view binding

caiofaustino:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/component/CardComponentDialogFragment.kt
```diff
@@ -53,31 +54,32 @@ class CardComponentDialogFragment : BaseComponentDialogFragment() {
 
         if (!dropInConfiguration.amount.isEmpty) {
             val value = CurrencyUtils.formatAmount(dropInConfiguration.amount, dropInConfiguration.shopperLocale)
-            dropInCardView.payButton.text = String.format(resources.getString(R.string.pay_button_with_value), value)
+            binding.dropInCardView.binding.payButton.text = String.format(resources.getString(R.string.pay_button_with_value), value)
```
jreij, line 57: What if we create a method `setPayButtonText` inside `DropInCardView` instead of exposing its binding and accessing it from other views? Same goes for the other 3 usages of `dropInCardView.binding` here. Do you think it would be cleaner?

caiofaustino: Yeah I was thinking about that, but then it seems weird to try to "protect" access to the views that way when they can still use `findViewById()` to get the view if they wanted to, and we would have to maintain those setters if the layout changes, while this way they get generated automatically. But I agree it looks ugly and weird.

jreij (approved):

drop-in/src/main/java/com/adyen/checkout/dropin/ui/component/CardComponentDialogFragment.kt
caiofaustino: Like we discussed this is probably a symptom of a poor structure to begin with, the separation between `DropInCardView` and `CardComponentDialogFragment` should be improved in a new PR.

PR #317, opened 2021-01-07
Fetch public key for Card Component

caiofaustino:
## Summary
- Use the `clientKey` to fetch the `publicKey` for card encryption when not present.
- Introduce coroutines to handle asynchronous code.
- CardConfiguration now requires a `clientKey` to be initialized.


card/src/main/java/com/adyen/checkout/card/CardDelegate.kt
```diff
@@ -30,4 +38,23 @@ abstract class CardDelegate(protected val cardConfiguration: CardConfiguration)
...
     abstract fun requiresInput(): Boolean
     abstract fun isHolderNameRequired(): Boolean
+
+    suspend fun fetchPublicKey(): String {
+        return if (cardConfiguration.publicKey.isNotEmpty() && CardValidationUtils.isPublicKeyValid(cardConfiguration.publicKey)) {
+            Logger.d(TAG, "returning configuration publicKey")
+            cardConfiguration.publicKey
+        } else {
+            Logger.d(TAG, "fetching publicKey from API")
+            for (retry in 1..CONNECTION_RETRIES) {
```
jreij, line 48: Minor: you can use `repeat(CONNECTION_RETRIES) { ... }` since you don't need the index

card/src/main/java/com/adyen/checkout/card/api/PublicKeyConnection.kt
```diff
@@ -0,0 +1,34 @@
...
+private const val ENDPOINT = "v1/clientKeys/"
+private const val PUBLIC_KEY_JSON_KEY = "publicKey"
+
+class PublicKeyConnection(environment: Environment, clientKey: String) : SuspendConnection<String>(
+    "${environment.baseUrl}$ENDPOINT$clientKey"
+) {
+    @Throws(IOException::class, JSONException::class)
+    override fun call(): String {
+        Logger.v(TAG, "call - $url")
+        val result = String(get())
```
jreij, line 29: Minor: `Charsets.UTF_8` will be used by default here, maybe double check if this is the encoding used by the backend too (if you didn't already 😅).

components-core/src/main/java/com/adyen/checkout/components/api/SuspendConnection.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.adyen.checkout.components.api
+
+import com.adyen.checkout.core.api.Connection
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.withContext
+
+abstract class SuspendConnection<T>(url: String) : Connection<T>(url) {
```
jreij, line 16: If you don't want to create a new class just for this function, you can replace it with an extension function, something like this:

```
suspend inline fun <reified T> Connection<T>.suspendedCall(): T {
    return supervisorScope {
        withContext(Dispatchers.IO) {
            call()
        }
    }
}
```

card/src/main/java/com/adyen/checkout/card/CardDelegate.kt
caiofaustino: thanks 👍

card/src/main/java/com/adyen/checkout/card/api/PublicKeyConnection.kt
caiofaustino: Oh yeah it is, but good pointing out, this could slip by unnoticed. Maybe I can set it explicitly anyway.

components-core/src/main/java/com/adyen/checkout/components/api/SuspendConnection.kt
caiofaustino: Good idea, thanks.

jreij (approved):

PR #321, opened 2021-01-14
Refactor drop in ComponentState

jreij:


components-core/src/main/java/com/adyen/checkout/components/PaymentComponentState.java
```diff
@@ -8,15 +8,30 @@
...
+import com.adyen.checkout.core.util.ParcelUtils;
 
 /**
  * The current state of a PaymentComponent.
  */
-public class PaymentComponentState<PaymentMethodDetailsT extends PaymentMethodDetails> {
+public class PaymentComponentState<PaymentMethodDetailsT extends PaymentMethodDetails> implements Parcelable {
+
+    @NonNull
+    public static final Parcelable.Creator<PaymentComponentState> CREATOR = new Parcelable.Creator<PaymentComponentState>() {
```
caiofaustino, line 26: Android Studio is complaining about the raw type parameter, so maybe make it `PaymentComponentData<? extends PaymentMethodDetails>`

components-core/src/main/java/com/adyen/checkout/components/PaymentComponentState.java
```diff
@@ -8,15 +8,30 @@
...
 
 import com.adyen.checkout.components.model.payments.request.PaymentComponentData;
 import com.adyen.checkout.components.model.payments.request.PaymentMethodDetails;
+import com.adyen.checkout.core.util.ParcelUtils;
 
 /**
  * The current state of a PaymentComponent.
  */
-public class PaymentComponentState<PaymentMethodDetailsT extends PaymentMethodDetails> {
+public class PaymentComponentState<PaymentMethodDetailsT extends PaymentMethodDetails> implements Parcelable {
```
caiofaustino, line 23: Hmm I forgot we had to make this Parcelable. You need to implement this on the subclasses too right?

components-core/src/main/java/com/adyen/checkout/components/PaymentComponentState.java
jreij: Hmm right. But I will still get this warning in `return new PaymentComponentState(in);` and `return new PaymentComponentState[size];`

components-core/src/main/java/com/adyen/checkout/components/PaymentComponentState.java
jreij: Yeah I missed that. IMO the ability to use `PaymentComponentState` directly is not very ideal because we could easily miss implementing `Parcelable` and `CREATOR` on the sub classes (and if we add a new sub class later) and we will also have to initialize `PaymentComponentState` with raw parameterized type in the base class.

What if we make `PaymentComponentState` abstract, keep the `Parcel` constructor but remove the `CREATOR` from it, and create a `GenericPaymentComponentState extends PaymentComponentState<PaymentMethodDetails>` ? This way we are forced to add a `CREATOR` and a constructor with `Parcel` for each of the subclasses and we are sure we will initialize the right class?

caiofaustino: Yeah, making it abstract sounds good to me.

PR #324, opened 2021-01-19
Bin lookup

caiofaustino:
## Summary
- Implement Bin Lookup call.
  - Created `BinLookupConnection` to make the API call
  - Created model objects to map the request and response JSON
- Added `encryptBin` encryption method to CSE.
- Basic call structure in CardComponent


card/src/main/java/com/adyen/checkout/card/model/Brand.kt
```diff
@@ -0,0 +1,89 @@
...
+                        cvcPolicy = jsonObject.getStringOrNull(CVC_POLICY),
+                    )
+                } catch (e: JSONException) {
+                    throw ModelSerializationException(Brand::class.java, e)
+                }
+            }
+        }
+    }
+
+    enum class CvcPolicy(val value: String) {
```
caiofaustino, line 75: @jreij was this the way you did that other enum? I couldn't find it.

jreij: We ended up doing it a bit differently, check `Threeds2Action.SubType`

caiofaustino: I think for kotlin this is a good approach :)

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -101,6 +114,55 @@ class CardComponent private constructor(
...
+                Logger.d(TAG, "Card brand not found.")
+                // TODO: 19/01/2021 Keep regexes prediction and don't apply business rules
+            }
+            brands.size > 1 -> {
+                Logger.d(TAG, "Multiple brands found.")
+                // TODO: 19/01/2021 use first brand
+            }
+            else -> {
+                Logger.d(TAG, "Card brand: ${brands[0].brand}")
+                val cardType = CardType.getByBrandName(brands[0].brand ?: "")
```
jreij, line 151: Minor: you can also use `brands[0].brand.orEmpty()` or `brands.first().brand.orEmpty()`

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -101,6 +114,55 @@ class CardComponent private constructor(
...
+                Logger.d(TAG, "CardType: ${cardType?.name}")
+                // TODO: 19/01/2021 trigger brand specific business logic
+            }
+        }
+    }
+
+    private fun getCardTypes(): List<String> {
```
jreij, line 158: You can just `return configuration.supportedCardTypes.map { it.txVariant }`

checkout-core/src/main/java/com/adyen/checkout/core/model/JsonUtils.kt
```diff
@@ -0,0 +1,124 @@
...
+    } catch (e: JSONException) {
+        PARSING_ERROR
+    }
+}
+
+fun JSONObject.optStringList(key: String): List<String>? {
+    return JsonUtils.parseOptStringList(this.optJSONArray(key))
+}
+
+class JsonUtils private constructor() {
```
jreij, line 44: This could just be:
```
object JsonUtils {

    @JvmStatic
    fun writeToParcel(parcel: Parcel, jsonObject: JSONObject?) {
...
```

checkout-core/src/main/java/com/adyen/checkout/core/model/JsonUtils.kt
```diff
@@ -0,0 +1,124 @@
...
+import java.util.ArrayList
+import java.util.Collections
+
+private const val INDENTATION_SPACES = 4
+private const val PARSING_ERROR = "PARSING_ERROR"
+private const val FLAG_NULL = 0
+private const val FLAG_NON_NULL = FLAG_NULL + 1
+
+fun JSONObject.getStringOrNull(key: String): String? {
+    return if (this.has(key)) this.getString(key) else null
```
jreij, line 25: `this.` can be removed (same for the other extension methods)

checkout-core/src/main/java/com/adyen/checkout/core/model/JsonUtils.kt
```diff
@@ -0,0 +1,124 @@
...
+         *
+         * @param stringList The [List] of Strings to be serialized.
+         * @return The populated [JSONArray]. Could be null.
+         */
+        @JvmStatic
+        fun serializeOptStringList(stringList: List<String?>?): JSONArray? {
+            if (stringList == null) {
+                return null
+            }
+            val jsonArray = JSONArray()
```
jreij, line 115: you can do something like this:
```
return JSONArray().apply {
    stringList.filter { !it.isNullOrEmpty() }.forEach {
        put(it)
    }
}
```

example-app/src/main/java/com/adyen/checkout/example/CheckoutExampleApplication.kt
```diff
@@ -19,7 +19,7 @@ class CheckoutExampleApplication : MultiDexApplication() {
 
     companion object {
         init {
-            Logger.setLogcatLevel(Log.DEBUG)
+            Logger.setLogcatLevel(Log.VERBOSE)
```
jreij, line 22: Is this changed on purpose?

jreij (changes requested):

checkout-core/src/main/java/com/adyen/checkout/core/model/JsonUtils.kt
caiofaustino: Oh yeah, I love how we can just replace all `TextUtils.isEmpty` with `isNullOrEmpty()` ❤️

example-app/src/main/java/com/adyen/checkout/example/CheckoutExampleApplication.kt
caiofaustino: good point, see how that SENSITIVE could come in hand if we actually use it 😅

jreij: hahaha yes 💯

jreij (approved):
✨

PR #325, opened 2021-01-20
Make DropInService a bound Service

jreij:
## Summary
- Change `DropInService` from a `JobIntentService` to a bound `Service`.
- Provide a way to handle `DropInService` callbacks asynchronously, using `onPaymentsCallRequested`, `onDetailsCallRequested` and `sendResult`.
- Pass the component state and action data to the service, instead of just `JSONObject`.


example-app/src/main/java/com/adyen/checkout/example/service/ExampleDropInService.kt
```diff
@@ -40,45 +41,49 @@ class ExampleDropInService : DropInService() {
...
-            keyValueStorage.getShopperReference(),
-            keyValueStorage.getAmount(),
-            keyValueStorage.getCountry(),
-            keyValueStorage.getMerchantAccount(),
-            RedirectComponent.getReturnUrl(applicationContext),
-            AdditionalData(
-                allow3DS2 = keyValueStorage.isThreeds2Enable().toString(),
-                executeThreeD = keyValueStorage.isExecuteThreeD().toString()
+    override fun makePaymentsCall(paymentComponentData: JSONObject) {
+        ThreadManager.EXECUTOR.submit {
```
jreij, line 45: @caiofaustino as opposed to an `IntentService`, this `Service` will run on the main thread by default, and it's up to the merchants to manage that. But I suppose this wouldn't be a problem?

caiofaustino: Hmm, that's a good point. I think this is something we can also provide as an option inside the class implementation.
Because if we maintain the abstraction from v3 for compatibility, I think that should be on the background thread by default as well to not break compatibility. 

We can discuss a way to provide this with a better approach than `AWAIT`.

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
```diff
@@ -31,151 +32,56 @@ import org.json.JSONObject
...
-        if (dropInServiceResult == null) {
-            // Make sure people don't return Null from Java code
-            throw CheckoutException("DropInServiceResult result from DropInService cannot be null.")
-        }
-        Logger.d(TAG, "handleDropInServiceResult - ${dropInServiceResult::class.simpleName}")
-
-        // if type is WAIT do nothing and wait for async callback.
-        if (dropInServiceResult !is DropInServiceResult.Wait) {
+    protected fun sendResult(result: DropInServiceResult) {
+        ThreadManager.MAIN_HANDLER.post {
```
jreij, line 81: @caiofaustino the `DropInActivity` should handle the result on the main thread to be able to send an intent and finish and so on. So I wondered where we should do the switch:
* If we leave it up to the merchant to switch back to the main thread, we need to add a check and throw an exception
* We could do the switch in the `DropInActivity` since it is the component that needs to handle the result on the main thread.
* But by doing the switch in the `DropInService` we make sure the `DropInActivity` does not handle any threading logic since it's a task that seems more suitable with a `Service` than an `Activity`.
* Also am I doing the switch the right way or do we have a better way?

WDYT?

caiofaustino: All right so just to summarize the discussion so we don't forget. Something like this:
```kotlin
internal open fun requestPaymentsCall(
    paymentComponentState: paymentComponentState,
    paymentDetails: JSONObject
) {
    ThreadManager.EXECUTOR.submit {
        val callResult = makeDetailsCall(details)
        handleResult()
    }
}
```

1. We will keep the current approach of having the default method being called on a background thread.
2. They can override `requestPaymentsCall` to handle the call asynchronously, and then they can manually call `handleResult()
3. `makeDetailsCall` default implementation throws an exception.
4. The DropInActivity can observe a `resultLiveData` on the Service. `handleResult()` will post to the livedata.
5. Calling `handleResult()` with an Error can prompt the Drop-in to terminate.

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
```diff
@@ -31,152 +41,79 @@ import org.json.JSONObject
...
-        const val dropInJobId = 11
-
-        // callback to handle sending the DropInServiceResult inside the callbackFlow below
-        private var callback: DropInFlowResult = object : DropInFlowResult {
-            override fun dispatchDropInServiceResult(dropInServiceResult: DropInServiceResult) {
-                Logger.e(TAG, "dispatchDropInServiceResult - callback called before flow")
-            }
-        }
+    private val coroutineJob: Job = Job()
+    override val coroutineContext: CoroutineContext get() = Dispatchers.Main + coroutineJob
```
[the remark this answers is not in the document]
jreij: It's an abstract method in `CoroutineScope`. I made `DropInService` implement `CoroutineScope` so I can use coroutines inside it.

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
```diff
@@ -31,152 +41,79 @@ import org.json.JSONObject
...
-
-        // Merchant makes network call
-        val paymentsDropInServiceResult = makePaymentsCall(PaymentComponentData.SERIALIZER.serialize(paymentComponentData))
-
-        handleDropInServiceResult(paymentsDropInServiceResult)
+    protected open fun onPaymentsCallRequested(
+        paymentComponentState: PaymentComponentState<*>,
+        paymentComponentJson: JSONObject
+    ) {
+        launch(Dispatchers.IO) {
```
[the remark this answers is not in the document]
jreij: Yep, both the old and new async implementations worked in java

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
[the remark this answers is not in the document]
caiofaustino: Aaahh right. Thx.

caiofaustino (approved):

PR #332, opened 2021-01-28
DropIn initialization with activityResult

jreij:
## Summary
Currently the only way to handle the drop in successful response is with the `resultHandlerIntent`. We want to allow the option to handle the drop in solely using `onActivityResult`:
* Make `resultHandlerIntent` optional.
* Return the result using `setResult` if `resultHandlerIntent` is not set.
* Allow fragments to use `startActivityForResult` also.

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
@@ -52,14 +52,16 @@ object DropIn {
     fun startPayment(
         activity: Activity,
         paymentMethodsApiResponse: PaymentMethodsApiResponse,
-        dropInConfiguration: DropInConfiguration
+        dropInConfiguration: DropInConfiguration,
+        resultHandlerIntent: Intent? = null
     ) {
         Logger.d(TAG, "startPayment from Activity")
 
         val intent = preparePayment(
             activity,
             paymentMethodsApiResponse,
-            dropInConfiguration
+            dropInConfiguration,
+            resultHandlerIntent
         )
         activity.startActivityForResult(intent, DROP_IN_REQUEST_CODE)
     }
@@ -77,14 +79,16 @@ object DropIn {
     fun startPayment(
         fragment: Fragment,
         paymentMethodsApiResponse: PaymentMethodsApiResponse,
-        dropInConfiguration: DropInConfiguration
+        dropInConfiguration: DropInConfiguration,
+        resultHandlerIntent: Intent? = null
     ) {
         Logger.d(TAG, "startPayment from Fragment")
 
         val intent = preparePayment(
             fragment.requireContext(),
             paymentMethodsApiResponse,
-            dropInConfiguration
+            dropInConfiguration,
+            resultHandlerIntent
         )
         fragment.startActivityForResult(intent, DROP_IN_REQUEST_CODE)
     }
@@ -92,13 +96,19 @@ object DropIn {
     private fun preparePayment(
         context: Context,
         paymentMethodsApiResponse: PaymentMethodsApiResponse,
-        dropInConfiguration: DropInConfiguration
+        dropInConfiguration: DropInConfiguration,
+        resultHandlerIntent: Intent?
     ): Intent {
         // Add locale to prefs
         context.getSharedPreferences(DROP_IN_PREFS, Context.MODE_PRIVATE).edit()
             .putString(LOCALE_PREF, dropInConfiguration.shopperLocale.toString())
             .apply()
 
-        return DropInActivity.createIntent(context, dropInConfiguration, paymentMethodsApiResponse)
+        return DropInActivity.createIntent(
+            context,
+            dropInConfiguration,
+            paymentMethodsApiResponse,
+            resultHandlerIntent
+        )
     }
 }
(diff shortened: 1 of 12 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -31,6 +33,7 @@ class DropIn private constructor() {
         private val TAG = LogUtil.getTag()
 
         const val RESULT_KEY = "payment_result"
+        const val RESULT_ERROR_REASON = "error_reason"
```
caiofaustino, line 36: Maybe rename to `ERROR_REASON_KEY` to be consistent with the other value name?

drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -41,32 +44,63 @@ class DropIn private constructor() {
          * Starts the checkout flow to be handled by the Drop-In solution. Make sure you have [DropInService] set up before calling this.
          * We suggest that you set up the resultHandlerIntent with the appropriate flags to clear the stack of the checkout activities.
```
caiofaustino, line 45: I think we can explain the result options here. Something like:

"Starts the checkout flow to be handled by the Drop-In. Make sure you have [DropInService] set up before calling this.

You will receive the the result of the DropIn in `onActivityResult`, either payment finished, canceled or error.
You can use `RESULT_KEY` and `RESULT_ERROR_REASON` to get the contents.

Anternatively, you can specify a `resultHandlerIntent` in the `DropInConfiguration` pointing to one of your activities, which will be called in case the payment finishes successfully. We suggest that you set up the resultHandlerIntent with the appropriate flags to clear the stack of the checkout activities."

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -120,22 +120,23 @@ class DropInConfiguration : Configuration, Parcelable {
...
-        private var resultHandlerIntent: Intent
+        private var resultHandlerIntent: Intent? = null
         private var amount: Amount = Amount.EMPTY
 
         private val packageName: String
         private val serviceClassName: String
 
         /**
+         *
+         * Create a [DropInConfiguration] with a [resultHandlerIntent] which will be launched after the Drop-In is finished
```
caiofaustino, line 131: You don't have the `resultHandlerIntent` in the constructor anymore :)

drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -31,6 +33,7 @@ class DropIn private constructor() {
         private val TAG = LogUtil.getTag()
 
         const val RESULT_KEY = "payment_result"
```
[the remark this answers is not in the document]
jreij: No you can call `Dropin. RESULT_KEY` from java

caiofaustino (approved):
We just need to work on the documentation right?

drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -25,48 +29,201 @@ import com.adyen.checkout.dropin.ui.DropInActivity
...
+    /**
+     * Starts the checkout flow to be handled by the Drop-in solution.
+     * Make sure you have [DropInService] set up before calling this.
+     * You can pass a [resultHandlerIntent] that will be launched after the Drop-in has completed
+     * without any errors.
+     * We suggest that you set up the [resultHandlerIntent] with the appropriate flags to clear
+     * the stack of the checkout activities.
+     *
+     * 3 states can occur from this operation:
+     * * Cancelled by user: the user dismissed the Drop-in before it has completed.
```
caiofaustino, line 53: Minor thing, the double `* *` is a bit hard to read, maybe do `* - ` instead?

drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -25,48 +29,201 @@ import com.adyen.checkout.dropin.ui.DropInActivity
...
+     * If you prefer to handle the activity result manually, you should expect an
+     * [Activity.RESULT_CANCELED] result code (for both error and cancellation). The data
+     * intent will contain the error reason extra with key [ERROR_REASON_KEY]. Its value will be
+     * [ERROR_REASON_USER_CANCELED] in case of user cancellation or the error reason otherwise.
+     *
+     * As for the Drop-in finished case, if you did not specify a [resultHandlerIntent], you will
+     * also receive the result in [Activity.onActivityResult]. You can make use of the
+     * [handleActivityResult] helper method. If you prefer to handle the activity result
+     * manually, you should expect an [Activity.RESULT_OK] result code. The data intent will
+     * contain the result string extra with key [RESULT_KEY].
```
caiofaustino, line 70: Maybe also explain that the content of `RESULT_KEY` will be what they specified and sent back in the `DropInService`?

caiofaustino (dismissed):

caiofaustino (approved):

PR #336, opened 2021-02-02
Use Bin Lookup result to handle card type detection.

caiofaustino:
## Summary
Bin Lookup will be triggered and used as a reliable way to identify the Card brand.
- Default regex detection is the first used
- Hitting required digits triggers bin lookup API
- Bin lookup result is emitted to CardComponent and cashed for subsequent detections. 

Future implementations will rely on this result to implement brand specific business logic.


card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -94,102 +98,55 @@ class CardComponent private constructor(
...
     override fun createComponentState(): CardComponentState {
         Logger.v(TAG, "createComponentState")
 
+        // TODO: 29/01/2021 pass outputData as non null parameter
+        val stateOutputData = outputData ?: throw CheckoutException("Cannot create state with null outputData")
+
         val cardPaymentMethod = CardPaymentMethod()
         cardPaymentMethod.type = CardPaymentMethod.PAYMENT_METHOD_TYPE
 
         val unenctryptedCardBuilder = UnencryptedCard.Builder()
```
jreij, line 126: Typo here, should be `unencryptedCardBuilder`

card/src/main/java/com/adyen/checkout/card/NewCardDelegate.kt
```diff
@@ -52,4 +82,110 @@ class NewCardDelegate(
...
+            isReliable = false,
+            showExpiryDate = true,
+            enableLuhnCheck = true,
+            cvcPolicy = getCvcPolicy(cardType.txVariant)
+        )
+    }
+
+    override fun getCvcPolicy(brand: String): Brand.CvcPolicy {
+        return when {
+            cardConfiguration.isHideCvc || noCvcBrands.contains(brand) -> Brand.CvcPolicy.HIDDEN
```
jreij, line 144: `noCvcBrands` is a `Set<CardType>` while `brand` is `String`. Did you mean to do something like `noCvcBrands.any { it.txVariant == brand }`?

card/src/main/java/com/adyen/checkout/card/NewCardDelegate.kt
```diff
@@ -52,4 +82,110 @@ class NewCardDelegate(
...
     }
+
+    override fun detectCardType(
+        cardNumber: String,
+        publicKey: String,
+        coroutineScope: CoroutineScope
+    ): List<DetectedCardType> {
+        Logger.d(TAG, "detectCardType")
+        if (cardNumber.length >= BinLookupConnection.REQUIRED_BIN_SIZE) {
+            val bin = cardNumber.substring(0, BinLookupConnection.REQUIRED_BIN_SIZE)
```
jreij, line 93: You can also use `cardNumber.take(BinLookupConnection.REQUIRED_BIN_SIZE)`

jreij (changes requested):

card/src/main/java/com/adyen/checkout/card/NewCardDelegate.kt
caiofaustino: Yeah I noticed this on the refactor, thanks for catching it :)

card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -8,21 +8,42 @@
...
+
+    private val cardType = CardType.getByBrandName(storedPaymentMethod.brand.orEmpty())
+    private val storedDetectedCardTypes = if (cardType != null) {
+        listOf(
+            DetectedCardType(
+                cardType,
+                isReliable = true,
+                showExpiryDate = true,
+                enableLuhnCheck = true,
+                cvcPolicy = when {
```
jreij, line 38: Can this `cvcPolicy` logic be extracted to `CardDelegate` since it's also used in `NewCardDelegate. localDetectedCard`?

caiofaustino: No because they check different flags in the Configuration, we have `cardConfiguration.isHideCvc` and `cardConfiguration.isHideCvcStoredCard`

jreij (approved):

PR #344, opened 2021-02-18
Make client key required

jreij:
## Summary
- Make client key required for all configurations.
- Remove public key from CardConfiguration

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.java
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.java
@@ -74,9 +74,10 @@ public class Adyen3DS2Configuration extends Configuration {
          * Constructor for Builder with default values.
          *
          * @param context   A context
+         * @param clientKey Your Client Key used for network calls from the SDK to Adyen.
          */
-        public Builder(@NonNull Context context) {
-            super(context);
+        public Builder(@NonNull Context context, @NonNull String clientKey) {
+            super(context, clientKey);
         }
 
         /**
@@ -84,9 +85,10 @@ public class Adyen3DS2Configuration extends Configuration {
          *
          * @param shopperLocale The Locale of the shopper.
          * @param environment   The {@link Environment} to be used for network calls to Adyen.
+         * @param clientKey Your Client Key used for network calls from the SDK to Adyen.
          */
-        public Builder(@NonNull Locale shopperLocale, @NonNull Environment environment) {
-            super(shopperLocale, environment);
+        public Builder(@NonNull Locale shopperLocale, @NonNull Environment environment, @NonNull String clientKey) {
+            super(shopperLocale, environment, clientKey);
         }
 
         @Override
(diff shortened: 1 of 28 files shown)


bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcConfiguration.java
```diff
@@ -28,6 +28,7 @@
  */
 public class BcmcConfiguration extends Configuration {
 
+    // TODO remove this when BcmcComponent starts fetching the public key from the client key
```
jreij, line 31: This is blocked until we can fetch the public key in the `BcmcComponent`

caiofaustino: Looking at it again, I would say maybe just add this to this PR. Move the method from the CardDelegate to somewhere the BCMC can also access it since it already depends on the Card module. Otherwise we lose the consistency and this might go unnoticed later.

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.java
```diff
@@ -85,8 +85,8 @@ public Builder(@NonNull Context context) {
          * @param shopperLocale The Locale of the shopper.
          * @param environment   The {@link Environment} to be used for network calls to Adyen.
```
caiofaustino, line 86: You forgot to update the javadoc, for all of them 😅

card/src/main/java/com/adyen/checkout/card/CardValidationUtils.java
```diff
@@ -44,17 +38,6 @@
     private static final int MAXIMUM_YEARS_IN_FUTURE = 30;
     private static final int MAXIMUM_EXPIRED_MONTHS = 3;
 
-    /**
-     * Checks if the public key for encryption is valid.
-     * @param publicKey The public key string
-     * @return True if valid, False if not.
-     */
-    public static boolean isPublicKeyValid(@Nullable String publicKey) {
```
jreij, line 52: @caiofaustino I removed this assuming we won't need to validate public keys anymore, is that right?

example-app/default.local.gradle
```diff
@@ -11,7 +11,6 @@ android {
             buildConfigField "String", "MERCHANT_SERVER_URL", "\"<YOUR_SERVER_URL>\""
             buildConfigField "String", "API_KEY_HEADER_NAME", "\"<API_KEY_HEADER>\""
             buildConfigField "String", "CHECKOUT_API_KEY", "\"<YOUR_API_KEY>\""
-            buildConfigField "String", "PUBLIC_KEY", "\"<PUBLIC_KEY>\""
```
jreij, line 14: @caiofaustino same here

card/src/main/java/com/adyen/checkout/card/CardValidationUtils.java
caiofaustino: yeah that's right

example-app/default.local.gradle
caiofaustino: Yep, now we should have a default for the client key right? `<CLIENT_KEY>`

caiofaustino (approved):

example-app/default.local.gradle
caiofaustino: nvm we already have that

PR #345, opened 2021-02-19
Refactor CardView

caiofaustino:
## Summary
- Refactor CardView to Kotlin
- Use ViewBinding


card/src/main/java/com/adyen/checkout/card/CardView.kt
```diff
@@ -0,0 +1,318 @@
...
+        binding.switchStorePaymentMethod.setOnCheckedChangeListener { _, isChecked ->
+            mCardInputData.isStorePaymentSelected = isChecked
+            notifyInputDataChanged()
+        }
+        if (component.isStoredPaymentMethod()) {
+            component.getStoredPaymentInputData()?.let {
+                setStoredCardInterface(it)
+            }
+        } else {
+            binding.textInputLayoutCardHolder.visibility = if (component.isHolderNameRequire()) VISIBLE else GONE
```
jreij, line 79: What do you think if we add an extension function like this:

```kotlin
fun View.setVisible(visible: Boolean) {
    visibility = if (visible) View.VISIBLE else View.GONE
}
```

This way we can just use `binding.textInputLayoutCardHolder.setVisible(component.isHolderNameRequire()))`

If you like the idea I can open a new PR and do a refactor for the whole project in it.

card/src/main/java/com/adyen/checkout/card/CardView.kt
```diff
@@ -0,0 +1,318 @@
...
+        }
+        if (detectedCardTypes.isEmpty()) {
+            binding.cardBrandLogoImageView.setStrokeWidth(0f)
+            binding.cardBrandLogoImageView.setImageResource(R.drawable.ic_card)
+            binding.editTextCardNumber.setAmexCardFormat(false)
+        } else {
+            binding.cardBrandLogoImageView.setStrokeWidth(RoundCornerImageView.DEFAULT_STROKE_WIDTH)
+            mImageLoader?.load(detectedCardTypes[0].cardType.txVariant, binding.cardBrandLogoImageView)
+            // TODO: 29/01/2021 get this logic from OutputData
+            var isAmex = false
```
jreij, line 191: You can use `val isAmex = detectedCardTypes.any { it.cardType == CardType.AMERICAN_EXPRESS }`

card/src/main/java/com/adyen/checkout/card/CardView.kt
caiofaustino: What about this?
```kotlin
var View.visible: Boolean
    get() = this.visibility == VISIBLE
    set(value) {
        if (value) this.visibility = VISIBLE
        else this.visibility = GONE
    }
```

jreij: Even better! Btw you can remove `this` (I think)

caiofaustino: Yeah, OK, I'll add this, I was thinking if we wanted to make the view `INVISIVLE` instead, but those are less likely and we can always just set it manually.

jreij (dismissed):

jreij (approved):

PR #347, opened 2021-02-22
Fetch public key in BCMC component

jreij:
## Summary
Fetch the public key using the client key in the BCMC component, just as we do in the card component.


card/src/main/java/com/adyen/checkout/card/CardDelegate.kt
```diff
@@ -8,27 +8,21 @@
...
-
-private val TAG = LogUtil.getTag()
-private const val CONNECTION_RETRIES = 3
 
 @Suppress("TooManyFunctions")
 abstract class CardDelegate(protected val cardConfiguration: CardConfiguration) : PaymentMethodDelegate {
 
     protected val noCvcBrands: Set<CardType> = hashSetOf(CardType.BCMC)
 
+    private val publicKeyRepository: PublicKeyRepository = PublicKeyRepository()
```
caiofaustino, line 24: should we pass the repository in the constructor to allow for manual DI and testing?

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponent.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    companion object {
+        @JvmField
+        val PROVIDER: PaymentComponentProvider<BcmcComponent, BcmcConfiguration> = BcmcComponentProvider()
+
+        @JvmField
+        val SUPPORTED_CARD_TYPE = CardType.BCMC
+    }
+
+    private val publicKeyRepository: PublicKeyRepository = PublicKeyRepository()
```
caiofaustino, line 53: should we pass the repository in the constructor to allow for manual DI and testing?

caiofaustino (approved):

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponent.kt
jreij: Fixed 👍

caiofaustino (approved):

PR #350, opened 2021-02-25
Set last 4 digits in CardComponentState

jreij:
## Summary
Set the last 4 digits of the card number when the CardComponentState is valid.


card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -201,6 +204,10 @@ class CardComponent private constructor(
         return if (cardNumber.length < BIN_VALUE_LENGTH) cardNumber else cardNumber.substring(0..BIN_VALUE_LENGTH)
     }
 
+    private fun getLastFourDigitsFromCardNumber(cardNumber: String): String {
```
caiofaustino, line 207: do you think this needs a separate method?

card/src/main/java/com/adyen/checkout/card/CardComponentState.java
```diff
@@ -20,6 +20,7 @@
 
     private final CardType mCardType;
     private final String mBinValue;
+    private final String mLastFourDigits;
```
caiofaustino, line 23: I wonder what this class would look like in Kotlin 👀

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
jreij: Hmm maybe not, i had more logic inside it first then i noticed it's not needed, will fix 👍

card/src/main/java/com/adyen/checkout/card/CardComponentState.java
jreij: Check it out!

caiofaustino (approved):

PR #359, opened 2021-03-19
Handle component initialization

jreij:
## Summary
We need a way to handle the component initialization state:
- Add `isReady` flag to components to indicate if the initialization is done
- Handle flag in Drop in fragments


components-core/src/main/java/com/adyen/checkout/components/GenericComponentState.java
```diff
@@ -20,7 +20,9 @@
      */
     public GenericComponentState(
             @NonNull PaymentComponentData<PaymentMethodDetailsT> paymentComponentData,
-            boolean isValid) {
-        super(paymentComponentData, isValid);
+            boolean isInputValid,
+            boolean isReady
```
caiofaustino, line 24: Maybe refactor this to Kotlin and make `isReady = true` by default?

caiofaustino (approved):

PR #368, opened 2021-03-25
Handle card component error in frictionless payment

jreij:
## Summary
Handle errors from `CardComponent` when directly paying from `PreselectedStoredPaymentMethodFragment`.

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/stored/PreselectedStoredPaymentViewModel.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/stored/PreselectedStoredPaymentViewModel.kt
@@ -69,6 +69,7 @@ class PreselectedStoredPaymentViewModel(
             "payButtonClicked - componentState.isReady: ${componentState?.isReady} - " +
                 "fragmentState: $fragmentState"
         )
+        // check whether an error has occurred already before Pay was clicked
         val componentError = lastComponentError
         val state = when {
             componentRequiresInput -> ShowStoredPaymentDialog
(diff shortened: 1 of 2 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/stored/PreselectedStoredPaymentViewModel.kt
```diff
@@ -66,21 +69,40 @@ class PreselectedStoredPaymentViewModel(
             "payButtonClicked - componentState.isReady: ${componentState?.isReady} - " +
                 "fragmentState: $fragmentState"
         )
+        val componentError = lastComponentError
```
caiofaustino, line 72: Why do we want to store the last error? In case the dialog is being shown when the Fragment gets destroyed?

jreij: No, the error might happen before the user taps Pay, so we need to store it in order to notify about it if Pay is clicked.

caiofaustino: Then maybe add a comment or rename the variable to something like `errorBeforePay` to make the use case more clear.

jreij: Added a comment, let me know if it's still unclear.

caiofaustino (approved):

PR #381, opened 2021-04-26
Handle Redirect returnUrlQueryString param

jreij:


googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayComponent.java
```diff
@@ -87,12 +87,7 @@ private GooglePayParams getGooglePayParams() {
         final Configuration configuration = ((GenericPaymentMethodDelegate) mPaymentMethodDelegate)
                 .getPaymentMethod()
                 .getConfiguration();
-        final String serverGatewayMerchantId;
-        if (configuration == null) {
-            serverGatewayMerchantId = null;
-        } else {
-            serverGatewayMerchantId = configuration.getGatewayMerchantId();
-        }
+        final String serverGatewayMerchantId = (configuration != null) ? configuration.getGatewayMerchantId() : null;
```
caiofaustino, line 90: HAHAHAHAHAHA, I agree but I think this and the Release notes thing should be it's own PR

redirect/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     testImplementation "junit:junit:$junit_version"
     androidTestImplementation "androidx.test.ext:junit:$test_ext_version"
     androidTestImplementation "androidx.test.espresso:espresso-core:$espresso_version"
+    testImplementation "org.robolectric:robolectric:$robolectric_version"
```
caiofaustino, line 52: Is it only me that gets triggered by the order here? 😛

redirect/src/test/java/com/adyen/checkout/redirect/RedirectUtilTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import com.google.common.collect.Iterators
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+
+@RunWith(RobolectricTestRunner::class)
+@Config(manifest = Config.NONE)
+class RedirectUtilTest {
```
caiofaustino, line 22: 💪

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -0,0 +1,118 @@
...
+     */
+    @JvmStatic
+    @Throws(CheckoutException::class)
+    fun parseRedirectResult(data: Uri): JSONObject {
+        Logger.d(TAG, "parseRedirectResult - $data")
+
+        val extractedParams = HashMap<String, String>().apply {
+            data.getQueryParameter(PAYLOAD_PARAMETER)?.let { put(PAYLOAD_PARAMETER, it) }
+            data.getQueryParameter(REDIRECT_RESULT_PARAMETER)?.let { put(REDIRECT_RESULT_PARAMETER, it) }
+            data.getQueryParameter(QUERY_STRING_PARAMETER)?.let { put(QUERY_STRING_RESULT, it) }
```
caiofaustino, line 77: My understanding of the `returnUrlQueryString` was a bit different. Let me check with the team.

redirect/build.gradle
jreij: You're right I didn't notice it

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -0,0 +1,117 @@
...
+        val extractedParams = HashMap<String, String>().apply {
+            data.getQueryParameter(PAYLOAD_PARAMETER)?.let { put(PAYLOAD_PARAMETER, it) }
+            data.getQueryParameter(REDIRECT_RESULT_PARAMETER)?.let { put(REDIRECT_RESULT_PARAMETER, it) }
+            data.getQueryParameter(PAYMENT_RESULT_PARAMETER)?.let { paymentResult ->
+                data.getQueryParameter(MD_PARAMETER)?.let { md ->
+                    put(PAYMENT_RESULT_PARAMETER, paymentResult)
+                    put(MD_PARAMETER, md)
+                }
+            }
+            data.encodedQuery?.let { put(QUERY_STRING_RESULT, it) }
```
caiofaustino, line 82: How is this not adding the `QUERY_STRING_RESULT` when one of the other ones is present? From the tests it seems to work but I don't get how.

RELEASE_NOTES.md
```diff
@@ -9,14 +9,4 @@
...
-- 'isValid' now checks both 'isInputValid' and 'isReady' to be true.
-
-## Fixed
-- Handle Intent results if DropInActivity got destroyed.
-- Queue API request if DropInService is not yet bound to DropInActivity
-
-
-
-
+- Return `returnUrlQueryString` from redirect URL.
```
caiofaustino, line 12: Maybe it's nice to add "for some redirect payment methods like Swish", otherwise this change doesn't mean anything to the merchants as this is mostly an internal change.

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
jreij: It does add `QUERY_STRING_RESULT` as long as there are any query params, even if we don't recognize them. However if there are no query params at all, `data.encodedQuery` is null and `QUERY_STRING_RESULT` won't be added.

RELEASE_NOTES.md
jreij: Or is it better to remove this completely from the release notes?

caiofaustino (approved):

PR #383, opened 2021-04-28
Don't pass threeDSRequestorAppURL on challenges in version 2.1.0

caiofaustino:
This field was introduced in version 2.2.0 so older protocols don't expect it to be present and might throw an error.

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
@@ -271,7 +271,10 @@ class Adyen3DS2Component(application: Application, configuration: Adyen3DS2Confi
             acsTransactionID = challenge.acsTransID
             acsRefNumber = challenge.acsReferenceNumber
             acsSignedContent = challenge.acsSignedContent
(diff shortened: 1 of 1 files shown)


jreij (approved):

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
```diff
@@ -271,7 +271,10 @@ class Adyen3DS2Component(application: Application, configuration: Adyen3DS2Confi
             acsTransactionID = challenge.acsTransID
             acsRefNumber = challenge.acsReferenceNumber
             acsSignedContent = challenge.acsSignedContent
-            threeDSRequestorAppURL = ChallengeParameters.getEmbeddedRequestorAppURL(getApplication())
+            // This field was introduced in version 2.2.0 so older protocols don't expect to be present and might throw an error.
+            if (challenge.messageVersion != PROTOCOL_VERSION_2_1_0) {
```
jreij, line 275: Is `2.1.0` the only version older than `2.2.0` ?

caiofaustino: yeah those are the only two protocol versions we have on the SDK

jreij (approved):

PR #384, opened 2021-04-29
Update Google Pay release notes

jreij:


RELEASE_NOTES.md
```diff
@@ -8,15 +8,7 @@
...
-## Fixed
-- Handle Intent results if DropInActivity got destroyed.
-- Queue API request if DropInService is not yet bound to DropInActivity
-
-
-
-
+## Changed
+- Updated Material Components dependency to 1.3.0.
+- The `merchantAccount` parameter in `GooglePayConfiguration.Builder` is now optional.
```
caiofaustino, line 13: Also add that they might need to update the code because of the new constructor, and that if not manually set, the Google environment will follow the Adyen environment automatically.

RELEASE_NOTES.md
```diff
@@ -8,15 +8,7 @@
...
+- Provided default `GooglePayConfiguration` and `CardConfiguration` in Drop-in.
```
caiofaustino, line 14: Also add that because of this, these configuration don't always need to be set manually anymore.

caiofaustino, line 14: We also:
- Removed WeChatPay Component (but still have the WeChatPayActionComponent)
- Added Pix Component

Might as well leave the release notes up to date :)

jreij: Hmm why though? I'm sure we missed more than these 2, so we can add them along with the remaining changes on release. I only added the Google Play ones here because I needed to make this quick fix and I thought I'll include the release notes anyway.

caiofaustino (approved):

PR #392, opened 2021-05-17
Improve component availability checks

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodsListViewModel.kt
```diff
@@ -84,8 +84,7 @@ class PaymentMethodsListViewModel(
                 type == null -> {
                     throw CheckoutException("PaymentMethod type is null")
                 }
-                PaymentMethodTypes.SUPPORTED_PAYMENT_METHODS.contains(type)
-                    && !PaymentMethodTypes.SUPPORTED_ACTION_ONLY_PAYMENT_METHODS.contains(type) -> {
```
caiofaustino, line 88: This means you have to check these (PIX) in the `getComponentAvailabilityCheckForType`.

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -143,38 +146,30 @@ internal fun checkComponentAvailability(
 
         val type = paymentMethod.type ?: throw CheckoutException("PaymentMethod type is null")
 
-        val provider = getProviderForType(type)
-        val configuration = dropInConfiguration.getConfigurationForPaymentMethod<Configuration>(type, application)
+        val availabilityCheck = getComponentAvailabilityCheckForType(type)
+        if (availabilityCheck == null) {
```
caiofaustino, line 150: I know this might be stupid but, I feel like using null here is not very semantic. Maybe we should have a `DefaultAvailabilityCheck` to return by  default instead?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodsListViewModel.kt
jreij: Yep, and `getComponentAvailabilityCheckForType` returns `null` for PIX which means it's always available.

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
jreij: Hmmm good idea, but what would this class be? We can't make it an interface with a default value 🤔 
And what about PIX? We have to make a provider for it also if we want to return an `ComponentAvailabilityCheck` for every supported payment method type, right?
My idea was that we only provide a `ComponentAvailabilityCheck` for payment methods that might be not always available.
What do you think?

caiofaustino (approved):

caiofaustino (approved):

Issue #338, opened 2021-02-05
Swish Payment method 


robertdahlberg:
**Is your feature request related to a problem? Please describe.**
We want swish on Android

**Describe the solution you'd like**
Add support for swish

**Describe alternatives you've considered**
-

**Additional context**

When I run android example app in test environment I get this unknown action error when selecting swish payment `Unknown Action - qrCode`

```
I/okhttp.OkHttpClient: Date: Fri, 05 Feb 2021 08:19:35 GMT
    Connection: keep-alive
I/okhttp.OkHttpClient: {"resultCode":"Pending","action":{"paymentData":"redacted","paymentMethodType":"swish","url":"swish://paymentrequest?callbackurl=adyencheckout%3A%2F%2Fcom.example%3Fpp%3DFTh9j%252FJNe34fsDWEHLMviGqF7zT1cusMbqhKl8gSIVHQ5tWMLDSb7geXNoNKuu%252Bm0GgPbqbyYIkRs7D9TOoH%252BI5gfLyP5c4tfAp94KjOeLgsriWburhmCrVbrDw8ElUEOF8uKpS%252Fm67rDp%252FK%252BAiJbQ%253D%253D&token=gs-Ld5-kQX-dJV6iG-0jgTB2aozTa7Zx","qrCodeData":"Dgs-Ld5-kQX-dJV6iG-0jgTB2aozTa7Zx","type":"qrCode"},"paymentData":"redacted","redirect":{"data":{"qrCodeData":"Dgs-Ld5-kQX-dJV6iG-0jgTB2aozTa7Zx"},"url":"swish://paymentrequest?callbackurl=adyencheckout%3A%2F%2Fcom.example%3Fpp%3DFTh9j%252FJNe34fsDWEHLMviGqF7zT1cusMbqhKl8gSIVHQ5tWMLDSb7geXNoNKuu%252Bm0GgPbqbyYIkRs7D9TOoH%252BI5gfLyP5c4tfAp94KjOeLgsriWburhmCrVbrDw8ElUEOF8uKpS%252Fm67rDp%252FK%252BAiJbQ%253D%253D&token=gs-Ld5-kQX-dJV6iG-0jgTB2aozTa7Zx"}}
    <-- END HTTP (5211-byte body)
D/CO.DropInService: handleDropInServiceResult - Action
    dropInServiceFlow - dispatchDropInServiceResult
    dropInServiceFlow - offer
D/CO.DropInActivityKt: dropInServiceFlow collect
[...]

PR #397, opened 2021-05-21
Fix/swish redirect

jesperjohansson:
## Summary
Replaces `qrCode` action with `redirect` for payment method type Swish. Couldn't find another way to do it other than replacing the action method... Swish should use redirect on mobile even though the method says its qrCode.

## Tested scenarios
Tested Swish locally with the demo app.

**Fixed issue**:  #338

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/model/payments/response/Action.java
+++ b/components-core/src/main/java/com/adyen/checkout/components/model/payments/response/Action.java
@@ -13,6 +13,7 @@ import android.text.TextUtils;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import com.adyen.checkout.components.util.PaymentMethodTypes;
 import com.adyen.checkout.core.exception.CheckoutException;
 import com.adyen.checkout.core.model.ModelObject;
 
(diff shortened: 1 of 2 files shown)


caiofaustino:
Heyyy, you are officially our first external PR request, thanks a lot for that! 🎉 

This is a great suggestion, but unfortunately the timing was a bit unlucky. We noticed this problem recently and have been discussing internally how to solve it. This inconsistency comes from the fact that some payment methods are `qrCode` on the Web but act as a redirect on Mobile, and this is not only for Swish. For consistency we decided to make a small refactor where the recently added [`QRCodeComponent`](https://github.com/Adyen/adyen-android/blob/develop/qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt) will also be able to act as the `RedirectComponent`, since we don't expect people integrating to know that an action type `qrCode` should actually be a redirect, so they will still handle it with the QrCodeComponent.

@jreij is already working on this refactor and should push a PR soon, so keep an eye out if you want to help with code review or get a sneak peak of the change, this will definitely be part of our next release.

jesperjohansson:
@caiofaustino Ah okay that sounds good! I will keep an eye out  👍

PR #400, opened 2021-05-25
3DS2 frictionless flow

jreij:


redirect/src/main/java/com/adyen/checkout/redirect/RedirectComponent.java
```diff
@@ -60,22 +54,12 @@ public static String getReturnUrl(@NonNull Context context) {
...
-            final Intent redirectIntent = RedirectUtil.createRedirectIntent(activity, redirectUri);
-            try {
-                activity.startActivity(redirectIntent);
-            } catch (ActivityNotFoundException e) {
-                throw new ComponentException("Redirect to app failed.", e);
-            }
-        } else {
-            throw new ComponentException("Redirect URL is empty.");
-        }
+        final RedirectDelegate delegate = new RedirectDelegate();
```
jreij, line 61: @caiofaustino I tried to move this field to the component itself and inject it in the provider (just like `Adyen3DS2Component.redirectDelegate`) but this method is static. I didn't change it to non static because I supposed we don't need to initialize the component to be able to make a redirect, am I right?

caiofaustino (approved):

PR #406, opened 2021-06-07
Load correct google pay and gift card logos

jreij:
## Summary
- Load alternative logo for google pay.
- Load brand instead of type for gift card logos.
- Refactor to move logic from adapter to view model.


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodModel.kt
```diff
@@ -9,6 +9,10 @@
 package com.adyen.checkout.dropin.ui.paymentmethods
 
 data class PaymentMethodModel(
-    val type: String,
-    val name: String
-)
+    val type: String?,
```
caiofaustino, line 12: Why did you decide to make this nullable when it wasn't before? Maybe the logic is not in the best place but we check for nullability in `PaymentMethodsListViewModel:85`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodsListViewModel.kt
```diff
@@ -110,6 +106,22 @@ class PaymentMethodsListViewModel(
         }
     }
 
+    private fun PaymentMethod.mapToModel(): PaymentMethodModel {
+        val icon = when (type) {
+            PaymentMethodTypes.SCHEME -> CARD_LOGO_TYPE
+            PaymentMethodTypes.GOOGLE_PAY -> GOOGLE_PAY_LOGO_TYPE
```
caiofaustino, line 112: This might be a bit tricky, I think the TxVariant itself will also change from `paywithgoogle` to `googlepay`, so this string would actually be the payment method type, at least that's what I got from COAND-397.

Maybe for this PR it's fine like this, but then when implementing COAND-397 we would probably make `PaymentMethodTypes.PAY_WITH_GOOGLE`. Do you have more info on how this will reflect on the API changes?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodModel.kt
```diff
@@ -9,6 +9,10 @@
...
+    val name: String,
+    val icon: String,
+    val isBorderDisabled: Boolean
```
caiofaustino, line 15: Although it is the exception, I think it's confusing to have a flag about it being disabled.
Could we make it something like `val drawIconBorder: Boolean = true`?

caiofaustino (approved):

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodsListViewModel.kt
jreij: Yep, I wasn't aware of the new payment method type, but I prefer to fix it in the next PR, just to keep it in context, right?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodModel.kt
jreij: I... don't know 😆  will check

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
```diff
@@ -96,40 +77,40 @@ class PaymentMethodAdapter(
...
 
         holder.text.text = paymentMethod.name
         holder.detail.visibility = View.GONE
 
-        val txVariant =
-            if (paymentMethod.type == PaymentMethodTypes.SCHEME) CARD_LOGO_TYPE
-            else paymentMethod.type
-
-        imageLoader.load(txVariant, holder.logo)
+        holder.logo.setRoundingDisabled(!paymentMethod.drawIconBorder)
```
caiofaustino, line 92: Don't you also need to invert this logic after changed the boolean to true? Then `setRoundingDisabled` would be just `setRounding`?

jreij: Haha yeah it's a weird one, cause the lack of rounding is the exception here. But it doesn't make sense to change it in one place but not the other, will fix 👍

caiofaustino (approved):

PR #410, opened 2021-06-10
Support optional cvc after BIN lookup

jreij:


card/src/main/java/com/adyen/checkout/card/CardValidationUtils.kt
```diff
@@ -94,17 +96,17 @@ object CardValidationUtils {
...
         val normalizedSecurityCode = StringUtil.normalize(securityCode)
         val length = normalizedSecurityCode.length
         var validation: Validation = Validation.Invalid(R.string.checkout_security_code_not_valid)
         if (StringUtil.isDigitsAndSeparatorsOnly(normalizedSecurityCode)) {
-            if (cardType == CardType.AMERICAN_EXPRESS && length == AMEX_SECURITY_CODE_SIZE) {
-                validation = Validation.Valid
-            } else if (length == GENERAL_CARD_SECURITY_CODE_SIZE && cardType != CardType.AMERICAN_EXPRESS) {
-                validation = Validation.Valid
+            validation = when {
+                cardType?.cvcPolicy == Brand.CvcPolicy.OPTIONAL && length == 0 -> Validation.Valid
```
jreij, line 106: Added `length == 0` check here to make sure we don't consider ANY optional CVC as valid, but only if it's empty. Otherwise it must pass through the normal validation steps.

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -151,7 +160,8 @@ class CardComponent private constructor(
                 unencryptedCardBuilder.setNumber(stateOutputData.cardNumberState.value)
             }
             if (!cardDelegate.isCvcHidden()) {
-                unencryptedCardBuilder.setCvc(stateOutputData.securityCodeState.value)
+                val cvc = stateOutputData.securityCodeState.value
+                if (cvc.isNotEmpty()) unencryptedCardBuilder.setCvc(cvc)
```
jreij, line 164: Added this check because `unencryptedCardBuilder` only supports null cvc or length 3 and 4

caiofaustino (approved):

PR #411, opened 2021-06-10
Redirect QR code actions with URL

jreij:


redirect/src/main/java/com/adyen/checkout/redirect/RedirectDelegate.kt
```diff
@@ -21,9 +21,19 @@ class RedirectDelegate {
...
+
+    /**
+     * Make a redirect from the provided Activity to the specified URL.
+     *
+     * @param activity The Activity starting the redirect.
+     * @param url The URL to redirect to.
+     */
+    fun makeRedirect(activity: Activity, url: String?) {
+        Logger.d(TAG, "makeRedirect - $url")
+        if (!TextUtils.isEmpty(url)) {
```
caiofaustino, line 35: `url.isEmptyOrNull()` please

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt
```diff
@@ -98,7 +110,7 @@ class QRCodeComponent(application: Application, configuration: QRCodeConfigurati
         // Immediately request a new status if the user resumes the app
         lifecycleOwner.lifecycle.addObserver(object : BaseLifecycleObserver() {
             override fun onResume() {
-                statusRepository.updateStatus()
+                if (isPollingEnabled) statusRepository.updateStatus()
```
caiofaustino, line 113: This seems like something that could have a better separation. Maybe the `StatusRepository` should only handle it if it's currently polling?

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt
```diff
@@ -81,6 +87,12 @@ class QRCodeComponent(application: Application, configuration: QRCodeConfigurati
...
     override fun handleActionInternal(activity: Activity, action: Action) {
         if (action !is QrCodeAction) throw ComponentException("Unsupported action")
+        if (!PROVIDER.requiresView(action)) {
+            Logger.d(TAG, "Action does not require a view, redirecting.")
+            isPollingEnabled = false
+            redirectDelegate.makeRedirect(activity, action.url)
+            return
+        }
         paymentMethodType = action.paymentMethodType
         qrCodeData = action.qrCodeData
```
caiofaustino, line 97: Could the `qrCodeData` be the URL in case of a redirect? Is there any scenario where there are both a qrCode and a URL?

wechatpay/src/main/java/com/adyen/checkout/wechatpay/WeChatPayActionComponent.java
```diff
@@ -70,27 +67,15 @@ public WeChatPayActionComponent(@NonNull Application application, @NonNull WeCha
      *
      * @param intent The intent result from WeChatPay SDK.
      */
-    public void handleResultIntent(@Nullable Intent intent) {
```
caiofaustino, line 73: Don't forget to add this to the release notes since it's a breaking change for merchants using the Component only.

caiofaustino (approved):

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt
jreij: Agreed, I'll add a status flag inside `StatusRepository`

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt
jreij: Yeah it could if the backend sends it like that, but it shouldn't. We will just try to redirect any QR code action that is not within the supported payment methods (only PIX now) and has a URL

caiofaustino: What I meant was for us to set like `qrCodeData = action.url` in the case of a redirect.
Like, this came up because I was thinking of how different the behavior of a regular QrCode and a Redirect should be here. For example, maybe we should also populate `paymentMethodType` in the redirect flow?

This current implementation is just returning and it makes me feel like it's sort of an inconsistent state. Does this make sense?

wechatpay/src/main/java/com/adyen/checkout/wechatpay/WeChatPayActionComponent.java
jreij: Done 👍  Pushed 3 new commits

components-core/src/main/java/com/adyen/checkout/components/status/StatusRepository.java
```diff
@@ -134,6 +141,11 @@ public void updateStatus() {
      */
     public void stopPolling() {
         Logger.d(TAG, "stopPolling");
+        if (!mIsPolling) {
```
caiofaustino, line 144: I'd rather `stopPolling` happen regardless of the flag value. If the flag goes to an inconsistent state for some reason, at least stop would force a reset.

caiofaustino (approved):

caiofaustino (approved):

PR #416, opened 2021-06-21
Support new "googlepay" txVariant

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -291,6 +291,7 @@ class DropInConfiguration : Configuration, Parcelable {
          */
         fun addGooglePayConfiguration(googlePayConfiguration: GooglePayConfiguration): Builder {
             availablePaymentConfigs[PaymentMethodTypes.GOOGLE_PAY] = googlePayConfiguration
+            availablePaymentConfigs[PaymentMethodTypes.GOOGLE_PAY_LEGACY] = googlePayConfiguration
```
jreij, line 294: I added this here, although it means that we are duplicating the entries, because the other way (only using `GOOGLE_PAY`) means that we'll need to make sure we always replace `GOOGLE_PAY_LEGACY` with `GOOGLE_PAY` when fetching the configuration (like in `availablePaymentConfigs` and `getConfigurationForPaymentMethod()`)

caiofaustino (dismissed):

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -112,11 +112,11 @@ class PaymentMethodListDialogFragment : DropInBottomSheetDialogFragment(), Payme
 
         // Check some specific payment methods that don't need to show a view
         when {
-            paymentMethod.type == PaymentMethodTypes.GOOGLE_PAY -> {
+            paymentMethod.type == PaymentMethodTypes.GOOGLE_PAY || paymentMethod.type == PaymentMethodTypes.GOOGLE_PAY_LEGACY -> {
```
jreij, line 115: Or should we make `GooglePayComponent.PAYMENT_METHOD_TYPES` public and use `GooglePayComponent.PAYMENT_METHOD_TYPES.contains(paymentMethod.type)` instead?

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
caiofaustino: Either way sounds good to me, I don't think it adds any memory usage.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
caiofaustino: Yeah that sounds good. I thought we had moved the logic to see if it's supported to the `PROVIDER`, so I was gonna suggest to add this check there as well, but apparently we didn't.

caiofaustino (approved):

PR #417, opened 2021-06-22
Release 4.0.0

jreij:


RELEASE_NOTES.md
```diff
@@ -10,7 +10,13 @@
 
 ## Added
 - 3 new methods to `ActionComponentProvider`: `canHandleAction`, `requiresView` and `getSupportedActionTypes`.
-- `QRCodeComponent` will now redirect qr code actions that should work as a redirect on Android (e.g. `bcmc_mobile`). 
+- `Adyen3DS2Component`, `QRCodeComponent`, `RedirectComponent` and `WeChatPayActionComponent` now implement `IntentHandlingComponent` and will have a `handleIntent` method. This method replaces `handleRedirectResponse` in `RedirectComponent` and `handleResultIntent` in `WeChatPayActionComponent`.  
+- `QRCodeComponent` will now redirect QR Code actions that should work as a redirect on Android (e.g. `bcmc_mobile`). Use the `handleIntent` method to handle the result of the redirect.
+- `Adyen3DS2Component` now supports the new 3DS2 frictionless flow (requires API v67). In some cases the component will now make a redirect, use the `handleIntent` method to handle the result of the redirect.
+- `GooglePayComponent` is now an `ActivityResultHandlingComponent`, no methods are affected by this change.
```
caiofaustino, line 16: Since no methods are affected maybe this is not necessary, but if you prefer to keep it I think it makes sense to say it supports both txVariants now.

RELEASE_NOTES.md
```diff
@@ -10,7 +10,13 @@
...
+- For certain BINs, the BIN lookup will indicate the CVC field is optional.
```
caiofaustino, line 17: Maybe to make it easier to understand. something like:
"When the CardComponent is certain of the card type (after the bin lookup is made) it will update the CVC field requirements accordingly, making it optional for some card types."

caiofaustino (approved):

PR #422, opened 2021-07-01
Initialize amount in default GooglePayConfiguration

jreij:


caiofaustino (approved):

caiofaustino (changes requested):
The empty amount thing we discussed 😅

caiofaustino (approved):

PR #430, opened 2021-07-08
Add prefix check for client key depending on the environment

ozgur00:
## Summary
Add a check if client key has the correct prefix
Convert Configuration and BaseConfigurationBuilder to kotlin

**Fixed issue**:  COAND-403


components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Base constructor with the required fields.
+ *
+ * @param shopperLocale The Locale of the shopper.
+ * @param environment   The {@link Environment} to be used for network calls to Adyen.
+ * @param clientKey Your Client Key used for network calls from the SDK to Adyen.
+ */
+abstract class BaseConfigurationBuilder<ConfigurationT: Configuration>(
+        protected var mBuilderShopperLocale: Locale,
```
caiofaustino, line 18: These variables can be public. We only had setters in Java before, but allowing getters is not an issue.

This way you can use the direct koltin assignment in the other builders instead of the setters.

components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Check if the Client Key is valid.
+     *
+     * @param clientKey A string to check if is a client key.
+     * @param environment Selected environment in the configuration.
+     * @return If we consider it a valid client key or not.
+     */
+    fun isClientKeyValid(clientKey: String, environment: Environment): Boolean {
```
caiofaustino, line 47: I thought we could split this check between Regex validation and Environment validation.
Something like `isClientKeyValid` and `doesClientkeyMatchEnvironment`.

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -91,61 +91,57 @@ object ComponentParsingProvider {
 @Suppress("ComplexMethod", "LongMethod")
 internal fun <T : Configuration> getDefaultConfigForPaymentMethod(
     paymentMethod: String,
-    context: Context,
     dropInConfiguration: DropInConfiguration
 ): T {
+    val shopperLocale = dropInConfiguration.shopperLocale
```
caiofaustino, line 96: Good catch :)

caiofaustino (commented):
Also you can run `./gradlew check` to make sure the static analysis passes, otherwise the Github Actions CI will block the PR anyway.

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcConfiguration.java
```diff
@@ -99,11 +99,11 @@ public Builder setEnvironment(@NonNull Environment builderEnvironment) {
          * @return {@link BcmcConfiguration}
          */
         @NonNull
-        public BcmcConfiguration build() {
+        protected BcmcConfiguration buildInternal() {
             return new BcmcConfiguration(
-                    mBuilderShopperLocale,
-                    mBuilderEnvironment,
-                    mBuilderClientKey
+                    getMBuilderShopperLocale(),
```
jreij, line 104: I see a lot of these `getM` fields, the reason is probably that the super classes are in java but the base class is in Kotlin. I think they look quite ugly 🙈  Maybe we should refactor the super classes before we do the base one, but that is probably better done in another PR. @caiofaustino what do you think?

jreij: Hmmm on a second thought, the base class is now in kotlin and the `m` prefix is not a Kotlin syntax. Maybe if we refactor the field names in the `BaseConfigurationBuilder` and remove the `m`, the super java classes would become more readable:
`BaseConfigurationBuilder`: `mBuilderShopperLocale` -> `builderShopperLocale`
`BcmcConfiguration`: `getMBuilderShopperLocale()` -> `getBuilderShopperLocale()`

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
caiofaustino: Oh yeah, and rename to remove the `m` prefix.

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcConfiguration.java
caiofaustino: I think my comment on `BaseConfigurationBuilder.kt` should do the trick, just having the base variables public.

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Base constructor with the required fields.
+ *
+ * @param shopperLocale The Locale of the shopper.
+ * @param environment   The {@link Environment} to be used for network calls to Adyen.
+ * @param clientKey Your Client Key used for network calls from the SDK to Adyen.
+ */
+abstract class BaseConfigurationBuilder<ConfigurationT : Configuration>(
+    var builderShopperLocale: Locale,
```
jreij, line 18: These fields should be `protected` I think.

components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
```diff
@@ -0,0 +1,56 @@
...
+        val isLiveEnvironment = environment == Environment.AUSTRALIA || environment == Environment.UNITED_STATES || environment == Environment.EUROPE
```
jreij, line 58: Maybe we should override `.equals` inside `Environment` to make sure it compares URLs and not the whole object? I'm thinking if for some reason the object gets copied and the `==` would not work because the instances are different? (happened to me once with `Amount.EMPTY`, not sure in which scenario).

There is also the case where someone might pass an environment manually like `Environment(URL("SOME_URL"))` and this `"SOME_URL"` matches one of our environments (even though this might be an overreach, not sure anyone does that).

caiofaustino: This is a fair point, I think it would be easier for us to assume that any environment that's not `Environment.TEST` is a live one.

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
caiofaustino: 👍

ozgur00: @caiofaustino's point in one of the previous comments made sense to me since it was already possible to modify them via setters I don't think it makes too much of a difference. @jreij why do you think they should be protected?

components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
ozgur00: Got it. Removing the case where it's neither test nor live. Should we update the ticket as well?

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
caiofaustino: oh yeah right, in the builder they could change the values whenever, even if they already passed them in the constructor it's fine to let them change after, so you I guess public is better indeed

jreij: It's true that they can change the values but they couldn't read them (we didn't have getters). And besides, the methods we have are not actual setters, they are builder methods (they return the Builder), so i feel like it's confusing for someone to see that they can call `Builder.serShopperLocale` and also `Builder.builderShopperLocale = ...`.

It's just a pattern thing though, not a big deal.

caiofaustino: You are totally right Joseph.

ozgur00: got it 👍

components-core/src/main/java/com/adyen/checkout/components/base/Configuration.kt
```diff
@@ -0,0 +1,30 @@
...
+
+abstract class Configuration protected constructor(
+    val shopperLocale: Locale,
+    val environment: Environment,
+    val clientKey: String
+) : Parcelable {
+
+    constructor(parcel: Parcel) : this(
+        parcel.readSerializable() as Locale,
+        parcel.readParcelable(Environment::class.java.classLoader)!!,
```
jreij, line 17: I usually try to avoid having the `!!` operator but I don't know if there are even any better alternatives in this case 🤔

ozgur00: i agree with not forcing most of the time but here I think we either need to go with nullable parameters or non-nulls with ```!!```. another idea could be we can use a fallback value, but i'm not sure if it's wise to use fallback values while creating an object from parcel. so i think out of those three this one is doable

components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
jreij: @caiofaustino the ticket does say `For custom environments don't check the clientKey.`. I think Ozgur had it right first, the only issue was that `Environment.equals` was not implemented so it was checking if the instance the exact same one.

components-core/src/main/java/com/adyen/checkout/components/base/Configuration.kt
jreij: I agree, fallback is not good in this case because it hides a problem in the parcelization, throwing an exception is better, and `!!` will do just that. The other, maybe cleaner, way is to create extension methods that would read from the parcel and throw an exception if something is null when it shouldn't be, but we can do that in another PR if we really need it 👍

components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
caiofaustino: I doubt anyone is using custom Environments, and if they are, I think it's fair to assume those are LIVE.

caiofaustino (approved):

components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
ozgur00: To be safe I added the custom environment case again.

jreij (approved):

PR #433, opened 2021-07-13
Add config field for showing preselected stored payment method

ozgur00:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -22,7 +22,8 @@ class DropInViewModel(
     val resultHandlerIntent: Intent?
 ) : ViewModel() {
 
-    val showPreselectedStored = paymentMethodsApiResponse.storedPaymentMethods?.isNotEmpty() ?: false
+    val showPreselectedStored = paymentMethodsApiResponse.storedPaymentMethods?.isNotEmpty() ?: false &&
+        dropInConfiguration.showPreselectedStoredPaymentMethod
```
jreij, line 26: Minor: I was confused whether this expression is interpreted as
```
(paymentMethodsApiResponse.storedPaymentMethods?.isNotEmpty() ?: false) && 
(dropInConfiguration.showPreselectedStoredPaymentMethod)
```
or
```
(paymentMethodsApiResponse.storedPaymentMethods?.isNotEmpty()) ?: 
(false && dropInConfiguration.showPreselectedStoredPaymentMethod)
```

Apparently it's the first because otherwise it wouldn't work or your IDE will give you a warning, but can we maybe add parentheses anyway?

jreij (approved):

PR #436, opened 2021-07-15
Fix logos not showing

jreij:
Loading the same logo multiple times in the same screen causes only one image view to load.


components-core/src/main/java/com/adyen/checkout/components/api/LogoConnectionTask.kt
```diff
@@ -0,0 +1,109 @@
...
+        } catch (e: InterruptedException) {
+            Logger.e(TAG, "Execution interrupted.", e)
+            notifyFailed()
+        } catch (e: TimeoutException) {
+            Logger.e(TAG, "Execution timed out.", e)
+            notifyFailed()
+        }
+    }
+
+    fun addCallback(callback: LogoCallback) {
```
caiofaustino, line 60: I would `synchronize` the content of `addCallback`, `notifyCallbacksReceived` and `notifyCallbacksFailed`  just to make sure there are not called at the same time.

components-core/src/main/java/com/adyen/checkout/components/api/ImageLoader.kt
```diff
@@ -0,0 +1,97 @@
...
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+import java.lang.ref.WeakReference
+import java.util.HashMap
+
+/**
+ * Loading Image from LogoApi.
+ */
+class ImageLoader(private val mLogoApi: LogoApi) {
+    private val mCallbacks: MutableMap<String, LogoCallback> = HashMap()
```
caiofaustino, line 26: you can remove the `m` prefixes in Kotlin 😉

components-core/src/main/java/com/adyen/checkout/components/api/LogoConnectionTask.kt
```diff
@@ -0,0 +1,109 @@
...
+ */
+class LogoConnectionTask(
+    private val logoApi: LogoApi,
+    val logoUrl: String,
+    callback: LogoCallback
+) : ConnectionTask<BitmapDrawable>(
+    LogoConnection(logoUrl)
+) {
+
+    var mCallbacks: MutableList<LogoCallback> = ArrayList()
```
caiofaustino, line 31: Since the amount of callbacks won't be too big maybe it doesn't matter too much, but I would make this a `HashSet` for a small optimization.

caiofaustino (approved):

PR #438, opened 2021-07-21
Add CPF/CNPJ field to card component

ozgur00:


components-core/src/main/java/com/adyen/checkout/components/ui/Validation.kt
```diff
@@ -19,7 +17,7 @@ sealed class Validation {
     /**
      * Field is not valid.
      */
-    class Invalid(@StringRes val reason: Int) : Validation()
+    object Invalid : Validation()
```
jreij, line 20: Why are we making this change?

jreij: Hmm i see it's unused, but can we make use of it somehow instead of removing it?

card/src/main/java/com/adyen/checkout/card/ui/SocialSecurityNumberInput.kt
```diff
@@ -0,0 +1,77 @@
+package com.adyen.checkout.card.ui
+
+import android.content.Context
+import android.text.Editable
+import android.text.InputType
+import android.text.method.DigitsKeyListener
+import android.util.AttributeSet
+import com.adyen.checkout.components.ui.view.AdyenTextInputEditText
+
+class SocialSecurityNumberInput constructor(
```
jreij, line 10: Can you add some unit tests for this class? There seems to be some complicated logic, especially around `formatString`

components-core/src/main/java/com/adyen/checkout/components/ui/Validation.kt
caiofaustino: Yeah the idea was that the ViewModel could pass more descriptive error messages depending on why the validation failed, but I think this wasn't implemented yet.

card/src/main/java/com/adyen/checkout/card/NewCardDelegate.kt
```diff
@@ -79,10 +79,18 @@ class NewCardDelegate(
...
 
+    override fun validateSocialSecurityNumber(socialSecurityNumber: String): FieldState<String> {
+        return CardValidationUtils.validateSocialSecurityNumber(socialSecurityNumber)
+    }
+
     override fun isCvcHidden(): Boolean {
         return cardConfiguration.isHideCvc
     }
 
+    override fun getSocialSecurityNumberVisibility(): SocialSecurityNumberVisibility? {
```
caiofaustino, line 90: Marc said this is not required in stored cards for now, and while testing I saw a weird behavior where the cursor would jump back to the cvc field.

caiofaustino (changes requested):

card/src/main/java/com/adyen/checkout/card/CardValidationUtils.kt
```diff
@@ -38,18 +39,26 @@ object CardValidationUtils {
     private const val MAXIMUM_YEARS_IN_FUTURE = 30
     private const val MAXIMUM_EXPIRED_MONTHS = 3
 
+    private const val CPF_DIGIT_LIMIT = 11
```
caiofaustino, line 42: Since Marc said this will also be added to Pix in the future, maybe it makes sense to separate the SSN validation outside of the Card to it's own utility class?

card/src/main/java/com/adyen/checkout/card/CardOutputData.kt
```diff
@@ -17,9 +17,11 @@ data class CardOutputData(
...
     val securityCodeState: FieldState<String>,
     val holderNameState: FieldState<String>,
+    val socialSecurityNumberState: FieldState<String>,
     val isStoredPaymentMethodEnable: Boolean,
     val cvcUIState: CvcUIState,
-    val detectedCardTypes: List<DetectedCardType>
+    val detectedCardTypes: List<DetectedCardType>,
+    val socialSecurityNumberVisibility: SocialSecurityNumberVisibility?
 ) : OutputData {
     override fun isValid(): Boolean {
```
jreij, line 26: We still need to add `&& socialSecurityNumberState.validation.isValid()` here

card/src/main/java/com/adyen/checkout/card/CardView.kt
```diff
@@ -163,27 +166,30 @@ class CardView @JvmOverloads constructor(context: Context, attrs: AttributeSet?
...
-                setCardNumberError(R.string.checkout_card_number_not_valid)
+                setCardNumberError((it.cardNumberState.validation as Validation.Invalid).reason)
             }
             if (!it.expiryDateState.validation.isValid()) {
                 if (!isErrorFocused) {
                     isErrorFocused = true
                     binding.textInputLayoutExpiryDate.requestFocus()
                 }
-                binding.textInputLayoutExpiryDate.error = mLocalizedContext.getString(R.string.checkout_expiry_date_not_valid)
+                val errorReasonResId = (it.expiryDateState.validation as Validation.Invalid).reason
```
jreij, line 176: I feel like this could be improved, the if condition checks `isValid()` which internally checks `is Validation.Valid`, but we have to cast again to `Validation.Invalid` here which doesn't look great IMO.

One fix could be something like this:

```kotlin
val expiryDateValidation = it.expiryDateState.validation
if (expiryDateValidation is Validation.Invalid) {
                if (!isErrorFocused) {
                    isErrorFocused = true
                    binding.textInputLayoutExpiryDate.requestFocus()
                }
                val errorReasonResId = expiryDateValidation.reason // no need to cast anymore
```

On the java side we cannot benefit from the automatic type inference like kotlin, but we can still maybe extract the variable at least:

```java
val cardNumberValidation = outputData.getCardNumberField().getValidation();
if (!cardNumberValidation.isValid()) {
            isErrorFocused = true;
            mCardNumberEditText.requestFocus();
            final int errorReasonResId = ((Validation.Invalid) cardNumberValidation).getReason();
            setCardNumberError(errorReasonResId);
        }
```

The `isValid()` method would still be useful in other places but not if we need to get the reason. What do you think?

card/src/main/java/com/adyen/checkout/card/SocialSecurityNumberUtils.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val CNPJ_MASK_GROUPING = listOf(2, 3, 3, 4, 2) // e.g 12.123.123/1234-12
+    val CNPJ_MASK_SEPARATORS = listOf('.', '.', '/', '-')
+
+    fun validateSocialSecurityNumber(socialSecurityNumber: String): FieldState<String> {
+        val digitLength = socialSecurityNumber.filter { it.isDigit() }.length
+        val invalidState = Validation.Invalid(R.string.checkout_social_security_number_not_valid)
+        val validation = when {
+            digitLength < CPF_DIGIT_LIMIT -> invalidState
+            digitLength == CPF_DIGIT_LIMIT && CPF_PATTERN.matcher(socialSecurityNumber).matches() -> Validation.Valid
+            digitLength < CNPJ_DIGIT_LIMIT -> invalidState
```
jreij, line 28: Do we need `digitLength < CPF_DIGIT_LIMIT -> invalidState` and `digitLength < CNPJ_DIGIT_LIMIT -> invalidState` or they can be removed?

card/src/main/java/com/adyen/checkout/card/SocialSecurityNumberUtils.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        val (grouping, separators) = if (trimmedInput.length <= CPF_DIGIT_LIMIT) {
+            CPF_MASK_GROUPING to CPF_MASK_SEPARATORS
+        } else {
+            CNPJ_MASK_GROUPING to CNPJ_MASK_SEPARATORS
+        }
+
+        for (index in grouping.indices) {
+            if (trimmedInput.length >= grouping[index]) {
+                result.add(trimmedInput.substring(0, grouping[index]))
```
jreij, line 48: Minor: `trimmedInput.substring(0, grouping[index])` can be replaced by `trimmedInput.take(grouping[index])`

card/src/main/java/com/adyen/checkout/card/SocialSecurityNumberUtils.kt
```diff
@@ -0,0 +1,65 @@
...
+        val result = mutableListOf<String>()
+        val resultBuilder = StringBuilder()
+
+        val (grouping, separators) = if (trimmedInput.length <= CPF_DIGIT_LIMIT) {
+            CPF_MASK_GROUPING to CPF_MASK_SEPARATORS
+        } else {
+            CNPJ_MASK_GROUPING to CNPJ_MASK_SEPARATORS
+        }
+
+        for (index in grouping.indices) {
```
jreij, line 46: What happens if the `inputString` is longer than 14 digits? I feel like to would get trimmed to 14, is that the intended behavior?

card/src/main/java/com/adyen/checkout/card/SocialSecurityNumberUtils.kt
ozgur00: you're right since regex checks length i removed them

card/src/main/java/com/adyen/checkout/card/SocialSecurityNumberUtils.kt
ozgur00: Yeah. Keeping the behavior as it is as discussed.

jreij: Then can you make sure we have a test that covers that case if we don't already?

caiofaustino (approved):

jreij (approved):

PR #439, opened 2021-07-21
Remove focus change to next field after validating card number

ozgur00:


card/src/main/java/com/adyen/checkout/card/CardView.kt
```diff
@@ -220,16 +217,6 @@ class CardView @JvmOverloads constructor(context: Context, attrs: AttributeSet?
         }
     }
 
-    private fun changeFocusOfInput(numberValue: String) {
-        val length = numberValue.length
-        if (length == CardValidationUtils.GENERAL_CARD_NUMBER_LENGTH ||
-            length == CardValidationUtils.AMEX_CARD_NUMBER_LENGTH &&
```
jreij, line 226: These 2 constants are not used anymore, maybe we could remove them.

jreij:
This same logic exists in `BcmcView` and checks whether the length of the card number is 19. What happens there? Should we also remove the focus?

ozgur00:
> This same logic exists in `BcmcView` and checks whether the length of the card number is 19. What happens there? Should we also remove the focus?

Discussed with Marc and removed it from `BcmcView` as well

jreij (approved):

PR #442, opened 2021-07-28
Add android:exported to manifest

jreij:
https://developer.android.com/about/versions/12/behavior-changes-12#exported


jreij:
We still need to update the 3DS2 dependency with the same fix to be able to raise the target SDK level to 31.

caiofaustino (approved):

PR #449, opened 2021-08-04
KCP Auth Fields

ozgur00:


cse/src/main/java/com/adyen/checkout/cse/CardEncrypter.java
```diff
@@ -159,6 +176,9 @@ public static String encrypt(
             cardJson.put(EXPIRY_YEAR_KEY, unencryptedCard.getExpiryYear());
             cardJson.put(CVC_KEY, unencryptedCard.getCvc());
             cardJson.put(HOLDER_NAME_KEY, unencryptedCard.getCardHolderName());
+            cardJson.put(PASSWORD_KEY, unencryptedCard.getCardPassword());
+            cardJson.put(TAX_NUMBER_KEY, unencryptedCard.getTaxNumber());
+            // TODO ask caio why this is necessary
```
ozgur00, line 181: @caiofaustino me and @jreij talked about this yesterday but couldn't figure out why this method is necessary. Can you explain why do we have this method?

Also we have an `UnencryptedCard` class used in this method with all the card fields and `EncryptedCard` with only the encrypted ones. However in this case I needed to add a KCP specific fields to `EncryptedCard` which only had fields that all cards need to have before. We only use fields from `UnencryptedCard` in `CardEncrypter`. Do you think it's necessary to have two different classes or should we have one class with all the fields where some of them are encrypted and some of them are not?

card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
```diff
@@ -60,36 +61,25 @@ public CardConfiguration createFromParcel(@NonNull Parcel in) {
...
             @NonNull Environment environment,
             @NonNull String clientKey,
-            boolean holderNameRequired,
-            @NonNull String shopperReference,
-            boolean showStorePaymentField,
-            @NonNull List<CardType> supportCardTypes,
-            boolean hideCvc,
-            boolean hideCvcStoredCard,
-            SocialSecurityNumberVisibility socialSecurityNumberVisibility
+            @NonNull Builder builder
```
jreij, line 72: Why are you passing the builder here instead of the fields themselves?

card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -88,6 +88,14 @@ class StoredCardDelegate(
         return SocialSecurityNumberUtils.validateSocialSecurityNumber(socialSecurityNumber)
     }
 
+    override fun validateKcpBirthDateOrTaxNumber(kcpBirthDateOrTaxNumber: String): FieldState<String> {
+        return CardValidationUtils.validateKcpBirthDateOrTaxNumber(kcpBirthDateOrTaxNumber)
```
jreij, line 92: Do we show the KCP fields in the stored component? If not then this should always return Valid

card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -88,6 +88,14 @@ class StoredCardDelegate(
...
+    }
+
+    override fun validateKcpCardPassword(kcpCardPassword: String): FieldState<String> {
+        return CardValidationUtils.validateKcpCardPassword(kcpCardPassword)
```
jreij, line 96: Same, for validation

cse/src/main/java/com/adyen/checkout/cse/CardEncrypter.java
caiofaustino: Hmm this is indeed a new logic we didn't need until now, let me check with the other teams how they approach this.

jreij: Keep in mind that if we leave these 2 fields here, we need to also add social security number.

card/src/main/res/values-cs-rCZ/strings.xml
```diff
@@ -12,11 +12,18 @@
     <string name="checkout_card_expiry_date_hint">Konec platnosti</string>
     <string name="checkout_card_security_code_hint">Kód CVC/CVV</string>
     <string name="checkout_card_security_code_optional_hint">Kód CVC/CVV (volitelné)</string>
+    <string name="checkout_social_security_number_hint"></string>
```
caiofaustino, line 15: Was the social_security_number change on purpose? Idk if our string generator project handles translatable=false correctly.

card/src/main/java/com/adyen/checkout/card/CardValidationUtils.kt
```diff
@@ -108,6 +115,24 @@ object CardValidationUtils {
         return FieldState(normalizedSecurityCode, validation)
     }
 
+    fun validateKcpBirthDateOrTaxNumber(birthDateOrTaxNumber: String): FieldState<String> {
```
caiofaustino, line 118: The same way we discussed on the CSE class, It think this validation could be in it's own class. The Card logic is gonna get pretty complicated pretty fast and I think it will be better to have some logic separation, since KCP its a specific thing I think it's better not to have it mixed up with the generic Card logic.

tldr: I think we should move this to something like a `KcpValidationUtils` class

card/src/main/java/com/adyen/checkout/card/CardValidationUtils.kt
```diff
@@ -108,6 +115,24 @@ object CardValidationUtils {
...
+        val inputLength = birthDateOrTaxNumber.length
+        val validation = when {
+            inputLength == KCP_BIRTH_DATE_LENGTH && DateUtils.matchesFormat(birthDateOrTaxNumber, KCP_BIRTH_DATE_FORMAT) -> Validation.Valid
```
caiofaustino, line 121: Did you check this validation with the other platforms? `yyMMdd` seems odd to me, just wanted to double check.

ozgur00: Yeah here you can find it's mentioned as well https://hub.is.adyen.com/our-solution/products-and-features/payment-methods/korean_cards_direct_api/#api-request-checkout-api-classic-api

card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
ozgur00: After adding the KcpAuthVisibility config there was 11 params in the constructor. We got a lint error and I thought that was an error we shouldn't suppress because it slowly starts become a mess. This looks cleaner to me.

caiofaustino: If that's the case then the Builder should be the only parameter in the constructor. 
If you want to make this change I think it's cool, but maybe it should be in a separate PR changing all Configurations at once. wdyt?

ozgur00: Makes sense, I'll revert it for this PR.

card/src/main/res/values-cs-rCZ/strings.xml
ozgur00: it wasn't, reverted it.

cse/src/main/java/com/adyen/checkout/cse/GenericEncrypter.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.adyen.checkout.cse.exception.EncryptionException
+import org.json.JSONException
+import org.json.JSONObject
+import java.text.SimpleDateFormat
+import java.util.*
+
+object GenericEncrypter {
+
+    const val ENCRYPTION_FAILED_MESSAGE = "Encryption failed."
```
caiofaustino, line 11: This can be private

cse/src/main/java/com/adyen/checkout/cse/GenericEncrypter.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    const val KCP_PASSWORD_KEY = "password"
+
+    @JvmStatic
+    fun encryptField(
+        encryptionKey: String,
+        fieldToEncrypt: Any,
+        publicKey: String,
+        generationTime: String = makeGenerationTime()
```
caiofaustino, line 23: I think `generationTime` was not a parameter in then CardEncrypter on purpose so that the encryption method will always take the current system time to make a bit harder to "fake".

caiofaustino (dismissed):
Minor comments

caiofaustino (approved):

PR #456, opened 2021-08-16
Migrate address input

ozgur00:


card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
```diff
@@ -32,6 +32,11 @@
  */
 public class CardConfiguration extends Configuration {
 
+    public enum AddressVisibility {
```
caiofaustino, line 35: Maybe move this to it's own file like `KCPAuthVisibility`? I think it could be in the base module since other components might have Address inputs as well.

ozgur00: done

caiofaustino (dismissed):

caiofaustino (approved):

PR #473, opened 2021-08-27
Dual Branded Cards

ozgur00:


caiofaustino:
The icon overlaps the error icon when there is a validation issue. I think both icons should be hidden in that case.

card/src/main/java/com/adyen/checkout/card/DualBrandedCardUtils.kt
```diff
@@ -0,0 +1,29 @@
+package com.adyen.checkout.card
+
+import com.adyen.checkout.card.data.CardType
+import com.adyen.checkout.card.data.DetectedCardType
+
+object DualBrandedCardUtils {
+
+    fun sortBrands(cards: List<DetectedCardType>): List<DetectedCardType> {
+        return when {
```
jreij, line 9: You can do something like `if(cards.size <= 1) return cards` and you would get rid of when and else

card/src/main/java/com/adyen/checkout/card/DualBrandedCardUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+            cards.size <= 1 -> cards
+            else -> {
+                val hasCarteBancaire = cards.any { it.cardType == CardType.CARTEBANCAIRE }
+                val hasVisa = cards.any { it.cardType == CardType.VISA }
+                val hasPlcc = cards.any {
+                    it.cardType == CardType.UNKNOWN &&
+                        (it.cardType.txVariant.contains("plcc") || it.cardType.txVariant.contains("cbcc"))
```
jreij, line 16: Do we have these strings ("plcc" and "cbcc") somewhere already? Can we extract them if not?

card/src/main/java/com/adyen/checkout/card/DualBrandedCardUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+                }
+
+                if (hasCarteBancaire && hasVisa) {
```
jreij, line 19: Minor: can use `when`

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -178,6 +187,20 @@ class CardComponent private constructor(
         }
     }
 
+    private fun markSelectedCard(cards: List<DetectedCardType>, selectedIndex: Int): List<DetectedCardType> {
+        return if (cards.size > SINGLE_CARD_LIST_SIZE) {
```
jreij, line 191: You can use an early return to reduce indentation:
```kotlin
if (cards.size <= SINGLE_CARD_LIST_SIZE) return cards
return cards.mapIndexed { .. }
```

card/src/test/java/com/adyen/checkout/card/DualBrandedCardUtilsTest.kt
```diff
@@ -0,0 +1,186 @@
+package com.adyen.checkout.card
+
+import com.adyen.checkout.card.api.model.Brand
+import com.adyen.checkout.card.data.CardType
+import com.adyen.checkout.card.data.DetectedCardType
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class DualBrandedCardUtilsTest {
```
jreij, line 9: ✨

card/src/main/java/com/adyen/checkout/card/DualBrandedCardUtils.kt
caiofaustino: Yeah I thought we would let this logic to the backend anyway.

ozgur00: For now I implemented it in the same way as Web does. When backend implements it we can remove it.

ozgur00:
> The icon overlaps the error icon when there is a validation issue. I think both icons should be hidden in that case.

Fixed.

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -323,7 +347,19 @@ class CardComponent private constructor(
...
+    @StringRes fun getKcpBirthDateOrTaxNumberHint(input: String): Int {
+        return when {
+            input.length > KcpValidationUtils.KCP_BIRTH_DATE_LENGTH -> R.string.checkout_kcp_tax_number_hint
+            else -> R.string.checkout_kcp_birth_date_or_tax_number_hint
+        }
+    }
+
+    fun isDualBrandedFlow(cardOutputData: CardOutputData): Boolean {
+        return cardOutputData.detectedCardTypes.filter { it.isReliable }.size > 1 &&
+            cardOutputData.detectedCardTypes.filter { it.isReliable }.any { it.isSelected }
```
jreij, line 359: Minor: can we extract `cardOutputData.detectedCardTypes.filter { it.isReliable }` to a `val` so we don't do the filtering twice?

jreij (approved):

PR #475, opened 2021-09-08
Allow removing observers from components

jreij:


ozgur00 (dismissed):

components-core/src/main/java/com/adyen/checkout/components/base/BaseActionComponent.java
```diff
@@ -65,11 +65,21 @@ public void observe(@NonNull LifecycleOwner lifecycleOwner, @NonNull Observer<Ac
         mResultLiveData.observe(lifecycleOwner, observer);
     }
 
+    @Override
+    public void removeObservers(@NonNull LifecycleOwner lifecycleOwner) {
+        mResultLiveData.removeObservers(lifecycleOwner);
```
caiofaustino, line 70: Why not use removeObserver() passing the observer reference instead? this way the merchant can have more granular control if they want.

jreij: My thinking was that most of the times you have 1 observer per LifecycleOwner and you don't keep reference of your observers:

```kotlin
cardComponent.observe(this) { paymentComponentState ->
    // handle result
}
```

But I think we could add both and let the merchants choose what suits them better?

caiofaustino: Sure

jreij: fixed

caiofaustino (approved):

PR #476, opened 2021-09-08
Add ProGuard/R8 consumer rules to the release artifacts.

caiofaustino:
This way merchants don't have to add the ProGuard rules themselves anymore.

### DIFF
--- a/example-app/build.gradle
+++ b/example-app/build.gradle
@@ -57,7 +57,7 @@ android {
 
     buildTypes {
         release {
-            minifyEnabled false
+            minifyEnabled true
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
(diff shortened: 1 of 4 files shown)


example-app/build.gradle
```diff
@@ -71,6 +71,7 @@ android {
 dependencies {
     // Checkout
     implementation project(':drop-in')
+//    implementation "com.adyen.checkout:drop-in:4.1.1"
```
jreij, line 74: Is that added on purpose?

caiofaustino: Yeah, for when we want to make a test against the release artifacts or a test artifact in `mavenLocal()`, just have it there commented to make it easier to change =P

jreij: 👍

jreij (approved):

PR #477, opened 2021-09-09
fix: Handle NOT_CURRENTLY_KNOWN flag

developerntuc:
Consulted Google as we shouldn't pass in totalPrice when totalPriceStatus is "NOT_CURRENTLY_KNOWN". This is a quick workaround.

Logs from Logcat:
W WalletMerchantError: Error in loadPaymentData: PaymentDataRequest.transactionInfo.totalPrice should not be set for totalPriceStatus = NOT_CURRENTLY_KNOWN!

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/googlepay/src/main/java/com/adyen/checkout/googlepay/util/GooglePayUtils.java
+++ b/googlepay/src/main/java/com/adyen/checkout/googlepay/util/GooglePayUtils.java
@@ -69,6 +69,8 @@ public final class GooglePayUtils {
     private static final String TOKENIZATION_DATA = "tokenizationData";
     private static final String TOKEN = "token";
 
+    private static final String NOT_CURRENTLY_KNOWN = "NOT_CURRENTLY_KNOWN";
+
     /**
      * Create a {@link com.google.android.gms.wallet.Wallet.WalletOptions} based on the component configuration.
      *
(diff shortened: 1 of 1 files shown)


googlepay/src/main/java/com/adyen/checkout/googlepay/util/GooglePayUtils.java
```diff
@@ -235,7 +237,9 @@ private static TransactionInfoModel createTransactionInfo(@NonNull GooglePayPara
         final String displayAmount = GOOGLE_PAY_DECIMAL_FORMAT.format(bigDecimal);
 
         final TransactionInfoModel transactionInfoModel = new TransactionInfoModel();
-        transactionInfoModel.setTotalPrice(displayAmount);
+        if (!params.getTotalPriceStatus().equals(NOT_CURRENTLY_KNOWN)) {
```
caiofaustino, line 240: Thanks a lot for your PR! I was gonna propose something a bit different but given the complexity I think this approach makes more sense indeed!

Can you please just add a comment above this line saying `// Google requires to not pass the price when the price status is NOT_CURRENTLY_KNOWN`?

developerntuc: Added.

caiofaustino: I just updated the branch and will merge the fix, thanks a lot for your contribution!

caiofaustino (approved):

PR #478, opened 2021-09-09
Get allowed card networks from api for Google Pay

ozgur00:


googlepay/src/main/java/com/adyen/checkout/googlepay/model/GooglePayParamUtils.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.adyen.checkout.core.exception.CheckoutException
+import com.adyen.checkout.googlepay.util.AllowedCardNetworks
+
+internal object GooglePayParamUtils {
+
+    fun mapTxVariantToGooglePayCode(txVariant: String): String {
+        return when {
+            txVariant == "mc" -> AllowedCardNetworks.MASTERCARD
+            AllowedCardNetworks.getAllAllowedCardNetworks().contains(txVariant.uppercase()) -> txVariant.uppercase()
+            else -> throw CheckoutException("txVariant $txVariant is not an allowed card network.")
```
caiofaustino, line 12: Maybe we can just ignore the value if it's not found instead of throwing the exception. An error Log should be enough imho.

googlepay/src/main/java/com/adyen/checkout/googlepay/model/GooglePayParamUtils.kt
```diff
@@ -0,0 +1,15 @@
+package com.adyen.checkout.googlepay.model
+
+import com.adyen.checkout.core.exception.CheckoutException
+import com.adyen.checkout.googlepay.util.AllowedCardNetworks
+
+internal object GooglePayParamUtils {
+
+    fun mapTxVariantToGooglePayCode(txVariant: String): String {
```
caiofaustino, line 8: I'm not sure if txVariant is the right name here, maybe brand is better.

jreij: ✔️

caiofaustino (approved):

PR #483, opened 2021-09-14
Store BCMC

caiofaustino:


jreij:
Is the BCMC component also supposed to handle payments with a stored BCMC card?

caiofaustino:
> Is the BCMC component also supposed to handle payments with a stored BCMC card?

I don't think so, we have some BCMC specific logic for stored in the CardComponent.

card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
```diff
@@ -184,7 +192,7 @@ public Builder(@NonNull CardConfiguration cardConfiguration) {
             super(cardConfiguration.getShopperLocale(), cardConfiguration.getEnvironment(), cardConfiguration.getClientKey());
             mBuilderSupportedCardTypes = cardConfiguration.getSupportedCardTypes();
             mBuilderHolderNameRequired = cardConfiguration.isHolderNameRequired();
-            mBuilderShowStorePaymentField = cardConfiguration.isShowStorePaymentFieldEnable();
+            mBuilderShowStorePaymentField = cardConfiguration.isStorePaymentFieldVisible();
```
jreij, line 195: If `isShowStorePaymentFieldEnable` is deprecated but not removed, shouldn't it still work? Right now it will break for merchants who are setting, because the builder will ignore it, right?

caiofaustino: No it won't, both geters point to the same underlying variable `mShowStorePaymentField`. Also there is no real reason why a merchant would read that field imho, it's more for the view to know if the field is visible or not.

jreij: > both geters point to the same underlying variable `mShowStorePaymentField`

I completely missed that 🤦

jreij (approved):

Issue #205, opened 2020-08-12
Make Drop-in show directly a single payment method.


PhilippeBoisney:
**Is your feature request related to a problem? Please describe.**
When the drop-in contains **only one** payment method, it could be interesting to show directly it (instead of having the view that lists only one single payment method).

We could, of course, use the [Components integration](https://docs.adyen.com/checkout/android/components), but this solution forces us to implement by ourself a lot of other things (async calls, UI loading state, etc...).

Is there any other solutions I've missed?

**Describe the solution you'd like**
When launching the Drop-in, a `Boolean` property in order to enable/disable this option.

caiofaustino:
No you are right, this is an improvement we considered, just haven't implemented it yet. We will put it in the backlog :)

MariaTod:
@caiofaustino I thought this is possible with the drop-in, I only have Credit Card and I need the user to see directly the payment form without having to tap to select "Credit Card" as this is the only option.
On iOS it's possible, so there is no way to do this on Android?

caiofaustino:
Not right now, unless they already have a stored card with us, in that case the stored card will show up separately.
But we will work on that for future releases.
![Screenshot_1600342281](https://user-images.githubusercontent.com/3094935/93465491-2a5d9300-f8eb-11ea-9d78-dafaa0c9076c.png)

KenrickV-Hexaware:
Hi, any update when this will be prioritised in the backlog? Thank you

BenedictP:
Would be really nice to have this feature!

przan:
Also waiting for this, as we wanted to use Drop-in to tokenize user's cards only. There is no point in showing payment methods bottomsheet.

BenedictP:
@przan They added this feature in the current release but not like I wanted/needed it (for example when you have Google Pay as the only Payment method it still shows a selection also when the only payment method is a stored one). Here is the information when the payment methods will NOT be skipped:

> This only applies to payment methods that require a component (user input). Which means redirect payment methods, SDK payment methods, etc will not be skipped even if this flag is set to true and a single payment method is present.

przan:
@BenedictP Thanks for reply! Now I see `showPreselectedStoredPaymentMethod ` has been added in https://github.com/Adyen/adyen-android/releases/tag/4.1.0 However, it doesn't work as I would like it to - I can still see Payment methods bottom sheet with only one option - card. I would like it skip this screen and open the one for card input. I will test it a bit more, though

[...]

PR #486, opened 2021-09-15
Skip the PaymentMethodsDialog when there is only one PaymentMethod

BenedictP:
## Summary
When there is only one Payment Method available we should skip the PaymentMethodsDialog and go directly to the corresponding input UI for this one Payment Method.

## Tested scenarios

I mocked the response in the example app of the https://checkout-test.adyen.com/v67/paymentMethods endpoint to return only one payment method. So far it worked with Credit Cards, Google Pay, SEPA and iDEAL. I couldn't test it with a stored payment method.

**Fixed issue**:  
https://github.com/Adyen/adyen-android/issues/205

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
@@ -158,7 +158,7 @@ internal fun checkPaymentMethodAvailability(
         val type = paymentMethod.type ?: throw CheckoutException("PaymentMethod type is null")
 
         val availabilityCheck = getPaymentMethodAvailabilityCheck(type)
-        val configuration = dropInConfiguration.getConfigurationForPaymentMethodOrNull<Configuration>(type, application)
+        val configuration = dropInConfiguration.getConfigurationForPaymentMethodOrNull<Configuration>(type)
 
         availabilityCheck.isAvailable(application, paymentMethod, configuration, callback)
     } catch (e: CheckoutException) {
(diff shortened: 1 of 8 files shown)


jreij:
Hi, just to confirm, are you looking to merge this PR so that we can review it? Or was it opened my mistake instead of a fork or something?

BenedictP:
Good evening, it's something you can review and if it's ok merge it. I could really need this feature and saw that other people could also need it (see linked issue). So I checked the code and thought I could implement it ;) I also thought about adding a feature flag that the user of the SDK could influence this behavior but for this I need your feedback/guidance.

jreij:
Thanks a lot for your effort ✨  We definitely will need a feature flag for this, but in all cases we'll discuss internally how this should be done, taking all payment methods into consideration, and let you know whether we can base it on this PR or we need to start over.

BenedictP:
I just added a feature flag for this.
Would also be perfectly ok if you use this PR only as "inspiration".
But i would appreciate an approximate timeline when you will implement the feature yourself. I can not keep my fork forever up-to-date ;)

jreij:
Thanks again for your contribution! I still do not have a timeline on when this will be done but I'll let you know as soon as I do.

caiofaustino:
Hey @BenedictP, thanks a lot for this PR, it's really a nice feature that I'll make sure to work on asap, but don't worry about keeping the fork updated, there are a few things that we will probably take a different approach in the implementation, but will surely use it as inspiration.

BenedictP:
Wow. That's really good news! I will close this PR and looking forward to the next release with this feature. :rocket: 
Thank you!

PR #487, opened 2021-09-15
Giftcard: initial component

jreij:
In this first PR, the gift card component is built exactly as a regular payment component. Gift card specific flow will follow in separate PRs.


cse/src/main/java/com/adyen/checkout/cse/UnencryptedCard.java
```diff
@@ -209,11 +196,5 @@ private String removeWhiteSpaces(String string) {
         private String trimAndRemoveMultipleWhiteSpaces(String string) {
             return string != null ? string.trim().replaceAll("\\s{2,}", " ") : null;
         }
-
-        private void require(boolean condition, String message) throws IllegalStateException {
```
caiofaustino, line 213: what happens when on of these values is null? If the encryption still works then this is fine.

caiofaustino (approved):

cse/src/main/java/com/adyen/checkout/cse/UnencryptedCard.java
jreij: Yeah it used to ignore validation if something is null, now it ignores validation altogether.

PR #491, opened 2021-09-20
Fix visual bug in BCMC component card icon

jreij:
|Before|After|
|---|---|
|Screenshot_1632131528|Screenshot_1632131477|

### DIFF
--- a/bcmc/src/main/res/layout/bcmc_view.xml
+++ b/bcmc/src/main/res/layout/bcmc_view.xml
@@ -32,21 +32,28 @@
                 tools:ignore="RequiredSize" />
         </com.google.android.material.textfield.TextInputLayout>
 
-        <FrameLayout
-            android:id="@+id/cardBrandLogo_container"
+        <LinearLayout
+            android:id="@+id/cardBrandLogo_container_parent"
             style="@style/AdyenCheckout.Card.BrandRecognitionLogoContainerGroup"
             android:layout_width="wrap_content"
-            android:layout_height="wrap_content">
+            android:layout_height="wrap_content"
+            android:orientation="horizontal">
 
-            <com.adyen.checkout.components.ui.view.RoundCornerImageView
-                android:id="@+id/cardBrandLogo_imageView"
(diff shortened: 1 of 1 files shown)


bcmc/src/main/res/layout/bcmc_view.xml
```diff
@@ -32,21 +32,28 @@
...
 
-            <com.adyen.checkout.components.ui.view.RoundCornerImageView
-                android:id="@+id/cardBrandLogo_imageView"
-                style="@style/AdyenCheckout.Card.BrandRecognitionLogo"
-                android:layout_width="@dimen/brand_logo_width"
-                android:layout_height="@dimen/brand_logo_height"
-                app:strokeColor="@color/stroke_color"
-                app:strokeWidth="0dp"/>
-        </FrameLayout>
+            <FrameLayout
```
caiofaustino, line 42: Why did you have to add an extra container here where there will always only be 1 icon? I don't get the fix.

jreij: This layout uses the same style structure as `card_view.xml`, specifically the styles `AdyenCheckout.Card.BrandRecognitionLogoContainerGroup`, `AdyenCheckout.Card.BrandRecognitionLogoContainer` and `AdyenCheckout.Card.BrandRecognitionLogo`.If we don't follow the same view structure we have in `card_view.xml` we cannot make the UI behave the same way therefore we need to use the same structure (my fix).

Otherwise, we could just remove the containers and add a new style, just for bcmc, but that means merchants need to override the styles for both components instead of just the card. I'm not against this solution but I thought your [commit](https://github.com/Adyen/adyen-android/pull/483/commits/9d6ca998b4221cf4403d21c14648665aab5c5aed) in the stored BCMC PR means that you prefer using the same structure for both.

caiofaustino (approved):

bcmc/src/main/res/layout/bcmc_view.xml
caiofaustino: yeah well, I thought the extra containers were not necessary but apparently I was wrong

PR #498, opened 2021-09-24
Skip to Component screen if only 1 available

caiofaustino:
If there is only 1 available payment method, and it has a Component, the Drop-in will skip to the Component screen instead of showing the payment methods list.

Merchant can choose to disable this behavior in the DropInConfiguration

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
@@ -151,10 +151,15 @@ class DropInActivity : AppCompatActivity(), DropInBottomSheetDialogFragment.Prot
         }
 
         if (noDialogPresent()) {
-            if (dropInViewModel.showPreselectedStored) {
-                showPreselectedDialog()
-            } else {
-                showPaymentMethodsDialog()
+            when {
+                dropInViewModel.shouldSkipToSinglePaymentMethod() -> {
+                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.get(0)
+                    if (firstPaymentMethod != null) {
+                        showComponentDialog(firstPaymentMethod)
+                    }
+                }
+                dropInViewModel.showPreselectedStored -> showPreselectedDialog()
+                else -> showPaymentMethodsDialog()
             }
         }
 
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
@@ -30,6 +30,7 @@ import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
 import com.adyen.checkout.components.model.paymentmethods.StoredPaymentMethod
 import com.adyen.checkout.components.model.payments.response.Action
 import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.core.exception.CheckoutException
 import com.adyen.checkout.core.log.LogUtil
 import com.adyen.checkout.core.log.Logger
 import com.adyen.checkout.core.util.LocaleUtil
@@ -153,9 +154,11 @@ class DropInActivity : AppCompatActivity(), DropInBottomSheetDialogFragment.Prot
         if (noDialogPresent()) {
             when {
                 dropInViewModel.shouldSkipToSinglePaymentMethod() -> {
-                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.get(0)
+                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.firstOrNull()
                     if (firstPaymentMethod != null) {
                         showComponentDialog(firstPaymentMethod)
+                    } else {
+                        throw CheckoutException("First payment method is null")
                     }
                 }
                 dropInViewModel.showPreselectedStored -> showPreselectedDialog()
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
@@ -14,6 +14,7 @@ import com.adyen.checkout.components.model.PaymentMethodsApiResponse
 import com.adyen.checkout.components.model.paymentmethods.StoredPaymentMethod
 import com.adyen.checkout.components.util.PaymentMethodTypes
 import com.adyen.checkout.dropin.DropInConfiguration
+import com.adyen.checkout.googlepay.GooglePayComponent
 
 class DropInViewModel(
     val paymentMethodsApiResponse: PaymentMethodsApiResponse,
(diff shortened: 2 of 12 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -151,10 +151,15 @@ class DropInActivity : AppCompatActivity(), DropInBottomSheetDialogFragment.Prot
...
-                showPreselectedDialog()
-            } else {
-                showPaymentMethodsDialog()
+            when {
+                dropInViewModel.shouldSkipToSinglePaymentMethod() -> {
+                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.get(0)
+                    if (firstPaymentMethod != null) {
+                        showComponentDialog(firstPaymentMethod)
+                    }
+                }
```
pboos, line 160: Shouldn't this also check if there are multiple (or no) payment methods?
```suggestion
                dropInViewModel.shouldSkipToSinglePaymentMethod()
                    && dropInViewModel.paymentMethodsApiResponse.paymentMethods?.size == 1  -> {
                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.get(0)
                    if (firstPaymentMethod != null) {
                        showComponentDialog(firstPaymentMethod)
                    }
                }
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -151,10 +151,15 @@ class DropInActivity : AppCompatActivity(), DropInBottomSheetDialogFragment.Prot
         }
 
         if (noDialogPresent()) {
-            if (dropInViewModel.showPreselectedStored) {
-                showPreselectedDialog()
-            } else {
-                showPaymentMethodsDialog()
+            when {
+                dropInViewModel.shouldSkipToSinglePaymentMethod() -> {
+                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.get(0)
```
jreij, line 156: I think `firstOrNull()` would be a safer call than `get(0)`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -30,4 +31,16 @@ class DropInViewModel(
     fun getStoredPaymentMethod(id: String): StoredPaymentMethod {
         return paymentMethodsApiResponse.storedPaymentMethods?.firstOrNull { it.id == id } ?: StoredPaymentMethod()
     }
+
+    fun shouldSkipToSinglePaymentMethod(): Boolean {
+        val noStored = paymentMethodsApiResponse.storedPaymentMethods.isNullOrEmpty()
+        val singlePm = paymentMethodsApiResponse.paymentMethods?.size == 1
+
+        val firstPaymentMethod = paymentMethodsApiResponse.paymentMethods?.get(0)
```
jreij, line 39: Same for `firstOrNull()`

BenedictP:
When I get an error while I try to pay with GooglePay I see an error dialog and when I dismiss this error dialog my screen still shows the loading spinner and the DropIn UI is not terminated. With CreditCart the input fields (CreditCardNumber, CVC, etc.) are still shown and I could try the payment again.

caiofaustino:
Good catch @BenedictP, Google Pay has a bit of a different flow than most other payment methods, I'll take into it.

caiofaustino:
@BenedictP weird, I can't seem to reproduce the issue and the code looks like it's taking GooglePay into consideration correclty on the flow. Can you maybe turn on logging and share it? 
`Logger.setLogcatLevel(Log.DEBUG)`

BenedictP:
Ah, sorry I was on the wrong branch. Still used my implementation of this feature. When I use this branch and I only have GooglePay as payment method the PaymentMethodsDialog is still presented.

caiofaustino:
@BenedictP so on this branch it works as expected right?

BenedictP:
Yes, except that there is a PaymentMethodsDialog with just one possible selection ;)

caiofaustino:
Yeah that's expected because it needs to be selected, otherwise the merchant would have to present the Google Pay button specifically for branding requirements.

BenedictP:
The same could be said for PayPal for example or what do you mean "for branding requirements"?

caiofaustino:
Yep, it's the same behavior for PayPal.
This is what I mean. https://developers.google.com/pay/api/android/guides/brand-guidelines

BenedictP:
But the guideline says: "All Google Pay payment buttons displayed on your site must adhere to our brand guidelines". So WHEN someone wants to show a Google Pay payment button it must follow the guidelines. My understanding is that you can still use Google Pay without a Google Pay button.

caiofaustino:
I wouldn't rely on that assumption, we had people have their apps declined from the Play Store because of inconsistencies with those branding guidelines, so I think they are pretty strict with it.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -151,10 +151,15 @@ class DropInActivity : AppCompatActivity(), DropInBottomSheetDialogFragment.Prot
...
 
         if (noDialogPresent()) {
-            if (dropInViewModel.showPreselectedStored) {
-                showPreselectedDialog()
-            } else {
-                showPaymentMethodsDialog()
+            when {
+                dropInViewModel.shouldSkipToSinglePaymentMethod() -> {
+                    val firstPaymentMethod = dropInViewModel.paymentMethodsApiResponse.paymentMethods?.firstOrNull()
+                    if (firstPaymentMethod != null) {
```
ozgur00, line 157: another option would be: `firstPaymentMethod?.let { showComponentDialog(it) }`

ozgur00 (dismissed):

BenedictP:
@caiofaustino But than the developer did something wrong and not the SDK ;) Because I do on my side all this stuff (show the correct GooglePay/PayPal/Credit Card button) and want the DropIn SDK to handle everything after the user already picked a Payment Method (enter Credit Card info, Google Pay card selection / 3ds1/3ds2). So I really don't need the PaymentMethodsDialog and I thought that's the exact use case for this feature.

caiofaustino:
Hmmm not really. What you are doing is basically taking the first step of the drop-in, the payment method selection, into your own hands and showing your own UI for that selection.

The intention of this feature is just to reduce friction for the cases where a merchant only has 1 payment method configured, but on PayPal and GooglePay the user needs to understand what they are paying with before continuing.
It's like being in your shopping card, clicking "Checkout" and going straight to PayPal or Google Pay, it would be confusing.

Also merchants could add/remove payment methods dynamically on the server side and the Drop-in would automatically know how to handle each case.

For your specific use case I think going for standalone components is more what you are looking for, I know that handling the Actions is a bit more complicated but that's where you have flexibility to control the whole UI flow.

caiofaustino:
It's not that they are randomly chosen, is just that the skip behavior is only for payment methods that need input from the user (e.g. payment methods that have a Component to display).
This is a small optimization to reduce friction, the behavior you expect requires more consideration and probably a bit of refactoring on how the Drop-in works. We could discuss adding that feature internally but that would be outside the scope of this PR.

ozgur00 (approved):

jreij (approved):

PR #500, opened 2021-09-27
Fix error icon overlapping with secondary card brand logo

ozgur00:


ozgur00:
> Sorry, I should have mentioned this before. My changes in [this PR](https://github.com/Adyen/adyen-android/pull/497) will definitely affect your fix here. Might be better to wait until I merge mine? We can also do it the other way if you prefer and merge yours before mine.

@jreij I see the PR is a draft. Do you still need to do a lot? If so we can merge this before, otherwise it can wait.

jreij (approved):

PR #503, opened 2021-09-29
Include `LIVE` in Environment values

tobias-square:
## Summary

Add `LIVE` value to Environment. 

This is identical to the EUROPE value but makes it clearer that a region specfic endpoint is not necessary. 

Brings the Android SDK in line with the iOS one https://github.com/Adyen/adyen-ios/blob/95f158b88630f023b9f3dc71766b53fe9b17477d/Adyen/Core/APIClient/Environment.swift

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/checkout-core/src/main/java/com/adyen/checkout/core/api/Environment.java
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/api/Environment.java
@@ -27,6 +27,7 @@ import java.util.Objects;
 public final class Environment implements Parcelable {
 
     public static final Environment TEST;
+    public static final Environment LIVE;
     public static final Environment EUROPE;
     public static final Environment UNITED_STATES;
     public static final Environment AUSTRALIA;
@@ -49,6 +50,7 @@ public final class Environment implements Parcelable {
             EUROPE = new Environment(new URL("https://checkoutshopper-live.adyen.com/checkoutshopper/"));
             UNITED_STATES = new Environment(new URL("https://checkoutshopper-live-us.adyen.com/checkoutshopper/"));
             AUSTRALIA = new Environment(new URL("https://checkoutshopper-live-au.adyen.com/checkoutshopper/"));
+            LIVE = EUROPE;
         } catch (MalformedURLException e) {
             throw new CheckoutException("Failed to parse Environment URL.", e);
         }


checkout-core/src/main/java/com/adyen/checkout/core/api/Environment.java
```diff
@@ -46,6 +47,7 @@ public Environment createFromParcel(@NonNull Parcel in) {
     static {
         try {
             TEST = new Environment(new URL("https://checkoutshopper-test.adyen.com/checkoutshopper/"));
+            LIVE = new Environment(new URL("https://checkoutshopper-live.adyen.com/checkoutshopper/"));
```
caiofaustino, line 50: Maybe if you do `LIVE = EUROPE` it's a bit more clear that the default environment is the european one and the validation below would still work since they are the same object. What do you think?

tobias-square: good idea

checkout-core/src/main/java/com/adyen/checkout/core/api/Environment.java
```diff
@@ -47,6 +48,7 @@ public Environment createFromParcel(@NonNull Parcel in) {
         try {
             TEST = new Environment(new URL("https://checkoutshopper-test.adyen.com/checkoutshopper/"));
             EUROPE = new Environment(new URL("https://checkoutshopper-live.adyen.com/checkoutshopper/"));
+            LIVE = EUROPE;
```
ozgur00, line 51: Can you put the changes in only one commit? Also please put `LIVE = EUROPE` below the instantiation lines, i think that's better for readability. Other than that looks good! 👍

caiofaustino (approved):

checkout-core/src/main/java/com/adyen/checkout/core/api/Environment.java
tobias-square: done 👍

jreij:
Just a small note, `ValidationUtils.doesClientKeyMatchEnvironment` has this code:
```kotlin
val isLiveEnvironment = environment == Environment.EUROPE ||
            environment == Environment.AUSTRALIA ||
            environment == Environment.UNITED_STATES
```
`LIVE` should also return true since we are setting `LIVE = EUROPE`, but what if we make a change later and `LIVE` becomes different than `EUROPE`? Should we maybe explicitly add `|| environment == Environment.LIVE` to the code above?

tobias-square:
That's what I did the first time round but removed it in response to https://github.com/Adyen/adyen-android/pull/503#discussion_r718293704
I don't mind either way

jreij:
@tobias-square feel free to go whichever way then. And thanks for the PR! ✨

jreij (approved):

tobias-square:
Thanks all 😍

PR #506, opened 2021-10-04
Google Pay cancel and failure listeners

ozgur00:


googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayProvider.kt
```diff
@@ -62,5 +62,11 @@ class GooglePayProvider :
         readyToPayTask.addOnCompleteListener { task ->
             callbackWeakReference.get()?.onAvailabilityResult(task.result == true, paymentMethod, configuration)
         }
+        readyToPayTask.addOnCanceledListener {
```
caiofaustino, line 65: can you add a Log.e here just so we have some way to track what happened in case of a bug?

jreij (approved):

PR #520, opened 2021-11-04
Validate locale on configuration build

jreij:


checkout-core/src/main/java/com/adyen/checkout/core/util/LocaleUtil.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Gets the language tag from a Locale.
+     *
+     * @param locale The locale.
+     * @return The tag of the locale, or null if not valid.
+     */
+    @JvmStatic
+    fun toLanguageTag(locale: Locale): String {
+        return locale.toLanguageTag()
```
caiofaustino, line 43: What is this method doing now? Seems redundant.

components-core/src/main/java/com/adyen/checkout/components/analytics/AnalyticEvent.java
```diff
@@ -100,7 +99,7 @@ public static AnalyticEvent create(@NonNull Context context, @NonNull Flavor fla
             default:
                 throw new CheckoutException("Unexpected flavor - " + flavor.name());
         }
-        return new AnalyticEvent(context.getPackageName(), flavorName, components, LocaleUtil.toLanguageTag(locale));
+        return new AnalyticEvent(context.getPackageName(), flavorName, components, locale.toString());
```
caiofaustino, line 102: Have you tested this in older API versions? Maybe do a quick test on API 21, because I think there was some limitation.

checkout-core/src/main/java/com/adyen/checkout/core/util/LocaleUtil.kt
jreij: It is redundant but I thought it's better to have our wrapper methods around the `Locale` methods, in case we want to change the way we serialize and deserialize locales. What do you think?

components-core/src/main/java/com/adyen/checkout/components/analytics/AnalyticEvent.java
jreij: tested, works fine 👍

checkout-core/src/main/java/com/adyen/checkout/core/util/LocaleUtil.kt
caiofaustino: OK makes sense

caiofaustino (approved):

PR #531, opened 2021-11-11
Giftcard: full payment

jreij:
This PR handles the scenario when the selected gift card has enough balance to make a full payment.


caiofaustino (approved):
I didn't see anything that stood out as being bad or wrong, architecture looks good from what we had already discussed.

PR #533, opened 2021-11-15
Check all PRs even if they don't target develop

jreij:


.github/workflows/check_pr.yml
```diff
@@ -3,7 +3,8 @@ name: Check PR
 # Every PR should be checked for static analysis
 on:
   pull_request:
-    branches: [ develop ]
+    branches: 
+      - '!master'
```
ozgur00, line 7: i think we might have to first say run for all branches and then exclude master. because this way i don't see any checks running for this pr. can you try this:
```
branches:    
      - '**'
      - '!master'
```

jreij: Tried `branches-ignore`, let's see if it works. Thanks!

ozgur00 (approved):

PR #537, opened 2021-11-19
Remove Koin and use Hilt instead.

caiofaustino:
This change only affects the Example App.


build.gradle
```diff
@@ -28,6 +29,8 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:$spotbugs_gradle_plugin_version"
         classpath "org.jetbrains.dokka:dokka-gradle-plugin:$dokka_version"
+        classpath 'com.google.dagger:hilt-android-gradle-plugin:2.38.1'
```
ozgur00, line 32: this line looks unnecessary

caiofaustino: Ahh thx, I didn't see it was duplicated.

ozgur00 (approved):

example-app/src/main/java/com/adyen/checkout/example/di/NetworkModule.kt
```diff
@@ -15,27 +15,42 @@ import com.adyen.checkout.example.data.api.CheckoutApiService
...
+object NetworkModule {
 
-    fun provideHttpClient(): OkHttpClient {
-        val builder = OkHttpClient.Builder().let {
+    private val BASE_URL = if (CheckoutApiService.isRealUrlAvailable())
+        BuildConfig.MERCHANT_SERVER_URL
+    else
+        "http://myserver.com/my/endpoint/"
+
+    @Singleton
```
jreij, line 42: Do all these methods have to be singletons?

caiofaustino: Not really, I think it's good practice for the Retrofit instance to be a singleton that's reused by the repositories, and the repos are also singletons usually, so idk if there would be any benefit of trying to have them cleared from memory at some point.

PR #543, opened 2021-11-25
Giftcard: order creation

jreij:
- Add `orders` call.
- Include order in `/payments` call.
- Allow dynamically updating drop in with order and payment methods.

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/model/payments/Amount.java
+++ b/components-core/src/main/java/com/adyen/checkout/components/model/payments/Amount.java
@@ -96,9 +96,9 @@ public class Amount extends ModelObject {
         return EMPTY_CURRENCY.equals(currency) || value == EMPTY_VALUE;
     }
 
-    @NonNull
     @Override
+    @NonNull
     public String toString() {
-        return value + " " + currency;
+        return "Amount(" + currency + ", " + value + ")";
     }
 }
--- a/components-core/src/main/java/com/adyen/checkout/components/model/payments/Amount.java
+++ b/components-core/src/main/java/com/adyen/checkout/components/model/payments/Amount.java
@@ -95,4 +95,10 @@ public class Amount extends ModelObject {
     public boolean isEmpty() {
         return EMPTY_CURRENCY.equals(currency) || value == EMPTY_VALUE;
     }
+
+    @NonNull
+    @Override
+    public String toString() {
+        return value + " " + currency;
+    }
 }
--- a/components-core/src/main/java/com/adyen/checkout/components/model/payments/response/BalanceResult.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/model/payments/response/BalanceResult.kt
@@ -8,15 +8,18 @@
 package com.adyen.checkout.components.model.payments.response
 
 import android.os.Parcel
+import com.adyen.checkout.components.model.payments.Amount
+import com.adyen.checkout.core.exception.CheckoutException
 import com.adyen.checkout.core.exception.ModelSerializationException
 import com.adyen.checkout.core.model.JsonUtils.writeToParcel
 import com.adyen.checkout.core.model.ModelObject
+import com.adyen.checkout.core.model.ModelUtils
 import org.json.JSONException
 import org.json.JSONObject
 
 data class BalanceResult(
-    val balance: String,
-    val transactionLimit: String?
+    val balance: Amount,
+    val transactionLimit: Amount?
 ) : ModelObject() {
 
     override fun writeToParcel(dest: Parcel, flags: Int) {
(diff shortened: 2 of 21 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInServiceResult.kt
```diff
@@ -31,8 +38,46 @@ sealed class DropInServiceResult : BaseDropInServiceResult() {
...
+     * Also after cancelling an order, you need to call /paymentMethods again with the original payment amount, and pass the
+     * updated payment methods list, with a null order object.
+     *
+     * Use [OrderResponse.SERIALIZER] to serialize your JSON response string.
+     *
+     * @param paymentMethodsApiResponse the updated payment methods list.
+     * @param order the order object returned from the backend, or null if an order was cancelled.
      */
-    class Action(val actionJSON: String) : DropInServiceResult()
+    class Update(val paymentMethodsApiResponse: PaymentMethodsApiResponse, val order: OrderResponse?) : DropInServiceResult()
```
ozgur00, line 80: i think we can enforce merchants to pass a non-null `OrderResponse` here. as far as i understood this is for partially paid orders and there has to be an order in the response to know we're at this step. thus we can make the `OrderResponse` non-null. i think by doing this we force them to not use `Update` when they don't have an order

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -162,6 +194,55 @@ class DropInViewModel(private val savedStateHandle: SavedStateHandle) : ViewMode
...
+
+    fun updatePaymentComponentStateForPaymentsCall(paymentComponentState: PaymentComponentState<*>) {
+        // include amount value if merchant passed it to the DropIn
+        val amount = when {
+            cachedPartialPaymentAmount != null -> cachedPartialPaymentAmount
+            !amount.isEmpty -> amount
+            else -> null
+        }
+        cachedPartialPaymentAmount = null
+        if (amount != null) {
```
ozgur00, line 219: since the rest uses `?.let` operator i would prefer using that here instead of null check just for consistency but nbd.

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInServiceResult.kt
jreij: I made it nullable so that they can call this `Update` again after order cancellation (not implemented yet) and pass new payment methods without an order.

example-app/src/main/java/com/adyen/checkout/example/service/ExampleAsyncDropInService.kt
```diff
@@ -98,29 +105,72 @@ class ExampleAsyncDropInService : DropInService() {
...
+                getPaymentMethodRequest(
+                    keyValueStorage,
+                    Order(
+                        pspReference = order.pspReference,
+                        orderData = order.orderData
+                    )
+                )
+            )
+            val result = if (paymentMethods != null) {
+                DropInServiceResult.Update(paymentMethods, order)
```
caiofaustino, line 168: Why is this in the Example class? I thought we would be the ones calling the updated `paymentMethods` response after the Order no?

Because this is triggering the DropInActivity to change the Fragment back to the Payment Methods list right?

components-core/src/main/java/com/adyen/checkout/components/model/payments/Amount.java
```diff
@@ -95,4 +95,10 @@ public void setValue(int value) {
     public boolean isEmpty() {
         return EMPTY_CURRENCY.equals(currency) || value == EMPTY_VALUE;
     }
+
+    @NonNull
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
jreij: To improve logging, we can now just write `Logger.d(TAG, "Amount: $amount")` and it will write `Amount: 1337 EUR`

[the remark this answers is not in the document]
jreij: Actually it shouldn't write `1337 EUR` but `13.37 EUR`, should i divide it by 100 maybe? Or just log something like `value = 1337 - currency = EUR` ?

components-core/src/main/java/com/adyen/checkout/components/model/payments/response/BalanceResult.kt
```diff
@@ -8,15 +8,18 @@
...
+import com.adyen.checkout.core.exception.CheckoutException
 import com.adyen.checkout.core.exception.ModelSerializationException
 import com.adyen.checkout.core.model.JsonUtils.writeToParcel
 import com.adyen.checkout.core.model.ModelObject
+import com.adyen.checkout.core.model.ModelUtils
 import org.json.JSONException
 import org.json.JSONObject
 
 data class BalanceResult(
-    val balance: String,
```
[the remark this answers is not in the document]
jreij: It was being parsed to amount in the view model but then we agreed to parse it earlier at the merchant side (when they call `BalanceResult.SERIALIZER.deserialize(jsonResponse)`). But this isn't live yet if you mean if it's a breaking change.

example-app/src/main/java/com/adyen/checkout/example/service/ExampleAsyncDropInService.kt
jreij: I thought so too but apparently it's all the merchant's responsibility. They need to check if the response of `/payments` contains an `order` which is not fully paid and then call `paymentMethods` again while passing the `order` in the request, then give us the list back. Same goes for cancelling an order.

components-core/src/main/java/com/adyen/checkout/components/model/payments/response/BalanceResult.kt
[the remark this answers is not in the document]
caiofaustino: Ah OK, just curious.

caiofaustino (approved):

caiofaustino (approved):

ozgur00 (approved):

PR #545, opened 2021-11-30
Giftcard: payment methods

jreij:
- Fetch order status from checkout shopper.
- Update payment methods screen.
- Update gift card payment confirmation screen.

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivityEvent.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivityEvent.kt
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) 2021 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 29/11/2021.
+ */
+
+package com.adyen.checkout.dropin.ui
+
+import com.adyen.checkout.components.PaymentComponentState
+
+sealed class DropInActivityEvent {
+    class MakePartialPayment(val paymentComponentState: PaymentComponentState<*>) : DropInActivityEvent()
+    object ShowPaymentMethods : DropInActivityEvent()
+}
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivityEvent.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivityEvent.kt
@@ -11,6 +11,6 @@ package com.adyen.checkout.dropin.ui
 import com.adyen.checkout.components.PaymentComponentState
 
 sealed class DropInActivityEvent {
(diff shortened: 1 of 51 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
```diff
@@ -56,12 +59,22 @@ class PaymentMethodAdapter(
...
         }
     }
 
     private fun bindHeader(holder: HeaderVH, position: Int) {
         val header = getHeaderAt(position)
         holder.title.setText(header.titleResId)
+        if (header.actionResId == null) {
+            holder.action.visibility = View.GONE
+        } else {
+            holder.action.visibility = View.VISIBLE
```
ozgur00, line 72: when there's repetitive usage of same field you can use `with()` like:
```
with(holder.action) {
    visibility = View.VISIBLE
    setText(header.actionResId)
    setOnClickListener {
        onHeaderActionClick(header)
    }
}
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
```diff
@@ -74,7 +88,7 @@ class DropInViewModel(private val savedStateHandle: SavedStateHandle) : ViewMode
             savedStateHandle[IS_WAITING_FOR_RESULT_KEY] = value
         }
 
-    var cachedGiftCardComponentState: GiftCardComponentState?
+    private var cachedGiftCardComponentState: GiftCardComponentState?
```
ozgur00, line 91: we never reset this. is there possibly a case where we need to reset this to `null`?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivityEvent.kt
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Created by josephj on 29/11/2021.
+ */
+
+package com.adyen.checkout.dropin.ui
+
+import com.adyen.checkout.components.PaymentComponentState
+
+sealed class DropInActivityEvent {
+    class MakePartialPayment(val paymentComponentState: PaymentComponentState<*>) : DropInActivityEvent()
```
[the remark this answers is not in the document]
jreij: I had no idea data classes could extend sealed classes!

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInViewModel.kt
jreij: We don't need to reset now it but I don't like how it is now so I'm refactoring this part in the next PR

ozgur00 (approved):

PR #550, opened 2021-12-07
Voucher Action Component

ozgur00:


components-core/src/main/java/com/adyen/checkout/components/ActionComponentProvider.java
```diff
@@ -87,4 +87,9 @@ ComponentT get(
      */
     @NonNull
     List<String> getSupportedActionTypes();
+
+    /**
+     * @return If the provided component provides details to make an API call to /payments/detail end point.
```
jreij, line 92: a few typos: `/payments/details` and `endpoint`

jreij: Can we also add that `observe` will not fire when this returns `false`?

components-core/src/main/java/com/adyen/checkout/components/api/ImageLoader.kt
```diff
@@ -43,10 +43,40 @@ class ImageLoader(private val logoApi: LogoApi) {
...
+     */
+    @JvmOverloads
+    fun load(
+        txVariant: String,
+        txSubVariant: String,
+        view: ImageView,
+        @DrawableRes placeholder: Int = 0,
+        @DrawableRes errorFallback: Int = 0
+    ) {
+        load(txVariant, txSubVariant, view, null, placeholder, errorFallback)
```
jreij, line 67: Should we keep the default size as null for existing usages? Just making sure we should have been passing it before and we weren't

ozgur00: I kept it as null since that was the previous implementation. But I see your point about passing the default size here rather than having to leave that logic to `LogoApi`.

jreij (approved):

PR #555, opened 2021-12-16
Giftcard: cancel order

jreij:
- Cancel order when Remove button is clicked.
- Cancel order on drop-in dismiss.

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
@@ -297,13 +297,13 @@ abstract class DropInService : Service(), CoroutineScope, DropInServiceInterface
         throw NotImplementedError("Method createOrder is not implemented")
     }
 
-    override fun requestCancelOrder(order: OrderRequest) {
+    override fun requestCancelOrder(order: OrderRequest, isDropInCancelledByUser: Boolean) {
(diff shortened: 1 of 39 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt
```diff
@@ -296,6 +297,16 @@ abstract class DropInService : Service(), CoroutineScope, DropInServiceInterface
         throw NotImplementedError("Method createOrder is not implemented")
     }
 
+    override fun requestCancelOrder(order: OrderRequest, isDropInCancelledByUser: Boolean) {
+        Logger.d(TAG, "requestCancelOrder")
+        cancelOrder(order, isDropInCancelledByUser)
+    }
+
+    // TODO docs
+    open fun cancelOrder(order: OrderRequest, isDropInCancelledByUser: Boolean) {
```
ozgur00, line 306: do you see any usages for `isDropInCancelledByUser` other than fetching the payment methods when it's `false`? if not we can rename it so method signature would be `cancelOrder(order: OrderRequest, shouldFetchPaymentMethods: Boolean)`. then we would have to invert the boolean before passing it to this method but for merchants it would be easier to understand when they're implementing it.

jreij: 👍

ozgur00 (approved):

Issue #180, opened 2020-04-23
Passing Data from the Activity to the DropInService


PhilippeBoisney:
After seeing the [demo app](https://github.com/Adyen/adyen-android/tree/master/example-app), I'm wondering how to **properly pass some data** (for example the `Amount`) to our custom `DropInService` from the `Activity` before `makePaymentsCall()` is called.

 ```
val dropInConfigurationBuilder = DropInConfiguration.Builder(
   this, 
   resultIntent, 
   ExampleSimplifiedDropInService::class.java
).addCardConfiguration(cardConfiguration)
 .setEnvironment(Environment.TEST)

// I would like to pass data at this point !
DropIn.startPayment(this, paymentMethodsApiResponse, dropInConfigurationBuilder.build())
```
 
You are using in the demo app the shared preferences to store and retrieve the `Amount` and then, using it inside the `makePaymentsCall()` method. 

I tried to use the 

Are there any other methods to do that?

caiofaustino:
Hey,

Yeah unfortunately we didn't think of this use case when designing the Drop-in flow, the only data going to the `DropInService` is the data from the Component. The `Amount` is only stored in the SharedPreferences as a test scenario because the example app doesn't have a server to rely on, and we would expect most merchants to have the purchase information on their server and complete any missing data there since [this](https://docs.adyen.com/checkout/android/drop-in#how-drop-in-works) is the expected flow.

That said, we also expect the SDK to be customizable enough for most cases, so maybe you can extend the regular `DropInService` instead of the Simplified version. It should be pretty simple to implement as well as you can see in this example.
https://github.com/Adyen/adyen-android/blob/master/example-app/src/main/java/com/adyen/checkout/example/service/ExampleDropInService.kt

So the `DropInService` is a `JobIntentService`, where we already implement the `onHandleWork()` call, but if you override calling super, I think you should be able to handle your custom requests as well using `enqueueWork()`. Although I'm not sure if the service will stay alive in order for you to store data there reliably.

https://github.com/Adyen/adyen-android/blob/master/drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInService.kt#L94

Let me know if that helps or if you have any more questions :)

PhilippeBoisney:
Thanks for your answer! However, [in the documentation](https://docs.adyen.com/checkout/android/drop-in#step-3-make-a-payment), the `amount` field is marked **as required** for the `makePaymentsCall`. 

[...]

PR #579, opened 2022-01-26
Pass Bundle with additional information to the drop in service

BenedictP:
## Summary
When making a payment request to our backend, I need additional data in the `onPaymentsCallRequested` and `onDetailsCallRequested` method. I didn't want to use shared prefs (what you are doing in the example app) or something else (e.g. inject a "data holder" class). So I added this functionality into my fork and now I want to show you what I did.
So you can merge this if you think it could be helpful or close it if this is not useful.

## Tested scenarios

**Fixed issue**: 
- https://github.com/Adyen/adyen-android/issues/180

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
@@ -11,6 +11,7 @@ package com.adyen.checkout.dropin
 import android.app.Activity
 import android.content.Context
 import android.content.Intent
+import android.os.Bundle
 import androidx.activity.result.ActivityResultCaller
 import androidx.activity.result.ActivityResultLauncher
 import androidx.fragment.app.Fragment
@@ -86,11 +87,16 @@ object DropIn {
      * [getDropInResultFromIntent] helper method to get it or you can find it in the intent
      * extras with key [RESULT_KEY].
      *
+     * When additional data is needed in the DropInService use the [additionalDataForDropInService]
+     * bundle. This bundle will be passed to the DropInService methods onPaymentsCallRequested and
+     * onDetailsCallRequested.
+     *
      * @param activity An activity to start the Checkout flow.
      * @param dropInLauncher A launcher to start Drop-in, obtained with [registerForDropInResult].
      * @param paymentMethodsApiResponse The result from the paymentMethods/ endpoint.
      * @param dropInConfiguration Additional required configuration data.
      * @param resultHandlerIntent Intent to be called after Drop-in has finished.
+     * @param additionalDataForDropInService Bundle which will passed to the DropInService
      *
      */
     @JvmStatic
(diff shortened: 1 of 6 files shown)


caiofaustino:
Thanks for your PR! And sorry for taking so long to get to review it. This looks good! I only have 2 small comments:

1 - When I initially thought about this feature, I thought of adding the extra bundle as part of the `DropInConfiguration` instead of a new parameter in `startPayment`. This way it can be just an optional field there, we don't break the public API, and the setup to make it serializable is already there. What do you think?

2 - We have CI check for linting and code checks, this is failing for a few small things.
https://github.com/Adyen/adyen-android/runs/4964481801?check_suite_focus=true#step:5:1944
You can run this yourself before pushing by running `./gradlew check`.

BenedictP:
No problem, no urgency here.
1. I changed it but I still broke public API when I added the Bundle to the `DropInServiceInterface` 
2. Now everything is fine. Thanks for the hint :)

jreij:
Hey @BenedictP, thanks for the PR, great effort and very useful feature!

I was reading through the implementation and I have some comments about it:
-  As discussed earlier, the current solution breaks 4 public methods which means we cannot include it in a minor release (4.x).
- These 4 methods will now expose an `additionalDataForDropInService: Bundle?` parameter that can most definitely confuse anyone who does not want to use this bundle or has no idea about it, which is the majority of our merchants.

However I think we can make a small adjustment to improve it:
- What if we just add a `getAdditionalData()` method, inside `DropInService`, which you can call at any time you want to retrieve the bundle?
- In addition to not breaking any existing methods, it will also not be exposed to anyone who does not need it, avoiding possible confusion.
- And since the bundle will not change anyway and is not affected at all by the flow itself, there is no problem having it in one place, accessible at any time you might need it.

As for the technical details of the implementation, instead of using `DropInServiceInterface` we can pass the `additionalDataForDropInService: Bundle?` parameter to the `bindService` method and retrieve it in `onBind`:


```kotlin

    companion object {
        private const val INTENT_EXTRA_ADDITIONAL_DATA = "ADDITIONAL_DATA"

        internal fun bindService(
            context: Context,
            connection: ServiceConnection,
            merchantService: ComponentName,
            additionalData: Bundle?
        ): Boolean {
            Logger.d(TAG, "bindService - ${context::class.simpleName}")
            val intent = Intent().apply {
                component = merchantService
                putExtra(INTENT_EXTRA_ADDITIONAL_DATA, additionalData)
            }
            return context.bindService(intent, connection, Context.BIND_AUTO_CREATE)
        }
    }

    private var additionalData: Bundle? = null

    override fun onBind(intent: Intent?): IBinder {
        Logger.d(TAG, "onBind")
        if (intent?.hasExtra(INTENT_EXTRA_ADDITIONAL_DATA) == true) {
            additionalData = intent.getBundleExtra(INTENT_EXTRA_ADDITIONAL_DATA)
        }
        return binder
    }

    protected fun getAdditionalData(): Bundle? {
        return additionalData
    }
```

Let me know your thoughts 🙂

jreij:
@BenedictP I will copy your branch (since I can't take over it) and update it a bit in a new PR, if you don't mind. Thanks again for the contribution.

BenedictP:
No problem. I didn't have the time to work on it lately so I'm quite happy that you are picking this up :)

PR #608, opened 2022-03-14
Sessions: initial setup

jreij:


sessions-core/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
OscarSpruit, line 1: Noticed that most of these files are empty. We might as well remove them, to decrease some clutter

example-app/src/main/java/com/adyen/checkout/example/data/api/CheckoutApiService.kt
```diff
@@ -30,6 +32,10 @@ interface CheckoutApiService {
         }
     }
 
+    @Headers(BuildConfig.API_KEY_HEADER_NAME + ":" + BuildConfig.CHECKOUT_API_KEY)
+    @POST("sessions")
+    fun sessionsAsync(@Body sessionRequest: SessionRequest): Deferred<Response<Session>>
```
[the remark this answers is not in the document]
jreij: Good catch 👍 
We have a bunch of other API calls like that though so maybe let's do it as a separate refactoring in another PR?

sessions-core/proguard-rules.pro
jreij: @caiofaustino are these `proguard-rules.pro` files needed for anything? Or just the `consumer-rules.pro`?

example-app/src/main/java/com/adyen/checkout/example/data/api/CheckoutApiService.kt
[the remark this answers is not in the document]
OscarSpruit: Sure, then it's at least consistent for now 😄

sessions-core/proguard-rules.pro
caiofaustino: Yeah those will only be used if we decide to ship the SDK already obfuscated, but I don't think that will ever be the case. So feel free to remove them of you want.

PR #609, opened 2022-03-15
Make sample app pretty ✨ 

OscarSpruit:
| Screen | New | Dark mode |
| - | - | - |
| Main | image | image |
| Settings | image | image |
| Checkout | image | image |


caiofaustino:
This looks great! I only have 1 question.
We try to make Drop-in be smart to automatically extend the merchant's theming by extending `AppTheme` [here](https://github.com/Adyen/adyen-android/blob/develop/ui-core/src/main/res/values/styles.xml#L14)
So if that was working, in the Drop-in the buttons and styles should all be green since that's the new primary color.
Can you try to investigate why that is not working for this change?

OscarSpruit:
@caiofaustino looks like it was already not working as the sample. In the file you linked there are references to the actual color `@color/colorPrimary` and it just picking up this color defined in `ui-core`. Changing all `@color/colorPrimary` references to `?attr/colorPrimary` makes it work

caiofaustino (approved):

OscarSpruit:
Yeah that's probably better. I've played around with it a little and it's more difficult than it sounds, because material theming is doing weird things

PR #611, opened 2022-03-16
Show available samples in a list

OscarSpruit:
Buttons are cool, but not very scalable. Soo all samples are in a list now and it should be very easy to add new entries.

image


OscarSpruit:
Card is not available yet, but just to show how it looks 🤫

example-app/src/main/java/com/adyen/checkout/example/ui/main/ComponentItem.kt
```diff
@@ -0,0 +1,13 @@
+package com.adyen.checkout.example.ui.main
+
+internal sealed class ComponentItem {
+
+    abstract val text: String
+
+    data class Title(override val text: String) : ComponentItem()
+
+    sealed class Entry(override val text: String) : ComponentItem() {
+        object DropIn : Entry("Start")
```
jreij, line 10: Minor: extract these into string resources

jreij, line 10: Minor: extract these into string resources

example-app/src/main/java/com/adyen/checkout/example/ui/main/ComponentItemProvider.kt
```diff
@@ -0,0 +1,11 @@
+package com.adyen.checkout.example.ui.main
+
+internal object ComponentItemProvider {
+
+    fun getComponentItems() = listOf(
+        ComponentItem.Title("Drop In"),
```
jreij, line 6: Same here, string resources

jreij, line 6: Same here, string resources

example-app/src/main/java/com/adyen/checkout/example/ui/main/ComponentItem.kt
OscarSpruit: Will do, hoped this would be okay in the example app 😬

jreij (approved):

PR #612, opened 2022-03-18
Clean up networking in example app

OscarSpruit:
Changes:
- Replace `Deferred` with `suspend` functions
- Simplify `safeApiCall`
- Get rid of `BaseRepository`
- Set API key header from `Interceptor`
- Make sure Kotlin coroutine version is the same as the rest of the project
- Remove Multidex dependency (not needed because min sdk is 21)


example-app/src/main/java/com/adyen/checkout/example/CheckoutExampleApplication.kt
```diff
@@ -8,13 +8,13 @@
...
 package com.adyen.checkout.example
 
+import android.app.Application
 import android.util.Log
-import androidx.multidex.MultiDexApplication
 import com.adyen.checkout.core.log.Logger
 import dagger.hilt.android.HiltAndroidApp
 
 @HiltAndroidApp
-class CheckoutExampleApplication : MultiDexApplication() {
```
caiofaustino, line 17: Can you link me some info on why this is not relevant anymore? I'm a bit outdated I guess.

caiofaustino (approved):

example-app/src/main/java/com/adyen/checkout/example/CheckoutExampleApplication.kt
OscarSpruit: https://developer.android.com/studio/build/multidex#mdex-on-l

PR #615, opened 2022-03-21
Refactor components-core to Kotlin

caiofaustino:


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ComponentProvider.kt
```diff
@@ -55,6 +55,9 @@ class Adyen3DS2ComponentProvider : ActionComponentProvider<Adyen3DS2Component, A
         return ViewModelProvider(viewModelStoreOwner, threeDS2Factory).get(Adyen3DS2Component::class.java)
     }
 
+    override val supportedActionTypes: List<String?>
```
OscarSpruit, line 58: Why did it become nullable?

components-core/src/main/java/com/adyen/checkout/components/base/BaseActionComponent.kt
```diff
@@ -0,0 +1,116 @@
...
+import com.adyen.checkout.core.log.Logger.w
+import org.json.JSONObject
+
+@Suppress("TooManyFunctions")
+abstract class BaseActionComponent<ConfigurationT : Configuration>(
+    savedStateHandle: SavedStateHandle,
+    application: Application,
+    configuration: ConfigurationT
+) : ActionComponentViewModel<ConfigurationT>(savedStateHandle, application, configuration) {
+    private val mResultLiveData = MutableLiveData<ActionComponentData>()
```
OscarSpruit, line 33: Ohh missed some 'm's

components-core/src/main/java/com/adyen/checkout/components/status/StatusRepository.kt
```diff
@@ -0,0 +1,142 @@
...
+        override fun run() {
+            Logger.d(TAG, "mStatusPollingRunnable.run()")
+            statusApi.callStatus(clientKey!!, paymentData!!, mStatusCallback)
+            updatePollingDelay()
+            handler.postDelayed(this, pollingDelay)
+        }
+    }
+
+    val statusApi: StatusApi = StatusApi.getInstance(environment)
+    private val statusResponseLiveData = MutableLiveData<StatusResponse?>()
```
OscarSpruit, line 38: The naming convention for private backing fields is usually the public field name prefixed with `_`, that way it's more clear which public and private fields belong together

Also, is nullability intended?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ComponentProvider.kt
caiofaustino: Missed this one, thx!

components-core/src/main/java/com/adyen/checkout/components/base/BaseActionComponent.kt
caiofaustino: thx

components-core/src/main/java/com/adyen/checkout/components/model/PaymentMethodsApiResponse.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.adyen.checkout.core.model.ModelUtils.deserializeOptList
+import com.adyen.checkout.core.model.ModelUtils.serializeOptList
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Object that parses and holds the response data from the paymentMethods/ endpoint.
+ */
+class PaymentMethodsApiResponse(
+    var storedPaymentMethods: List<StoredPaymentMethod>? = null,
```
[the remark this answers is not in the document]
caiofaustino: Yeah but we want to allow merchants to be able to edit these objects before passing them to us, so they can customize behavior. 
The other option would be to have some builder or copy method, but we can consider that.

components-core/src/main/java/com/adyen/checkout/components/status/StatusRepository.kt
caiofaustino: Thx, changed the name. About the nullability unfortunately yes, check the comments on line 107.
This can obviously be improved now that we can change from LiveData to Flows.

components-core/src/main/java/com/adyen/checkout/components/util/PaymentMethodTypes.kt
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Helper class with a list of all the currently supported Payment Methods on Components and Drop-In.
+ */
+object PaymentMethodTypes {
+
+    // Placeholder value if the type is not found.
+    const val UNKNOWN = "unknown"
+
+    // Type of the payment method as received by the paymentMethods/ API
+    const val IDEAL = "ideal"
```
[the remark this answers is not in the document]
caiofaustino: Hmm, good question, `IdealComponent` for example is still in Java and the access is just `PaymentMethodTypes.IDEAL`, idk exactly why.

components-core/src/test/java/com/adyen/checkout/components/base/BaseComponentTest.java
```diff
@@ -66,56 +64,57 @@ public void init() throws IOException, JSONException {
         paymentMethod = DataProvider.getPaymentMethodResponse(classLoader).getPaymentMethods().get(0);
     }
 
-    @Test(expected = IllegalArgumentException.class)
+//    @Test(expected = IllegalArgumentException.class)
```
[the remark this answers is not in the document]
caiofaustino: Ah, because I made the generic types not nullable, so I also need to change the Fakes in the test. I'll try to do that now.

components-core/src/main/java/com/adyen/checkout/components/model/PaymentMethodsApiResponse.kt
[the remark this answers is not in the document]
OscarSpruit: Alright, let's keep it like this. Creating builders for all these classes is a lot of work, which may return it's value

components-core/src/main/java/com/adyen/checkout/components/util/PaymentMethodTypes.kt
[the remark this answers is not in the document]
OscarSpruit: Should be okay then 👍

OscarSpruit (approved):

PR #617, opened 2022-03-23
Refactor Await Component to Kotlin

caiofaustino:


await/src/main/java/com/adyen/checkout/await/AwaitOutputData.kt
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by caiof on 25/8/2020.
+ */
+package com.adyen.checkout.await
+
+import com.adyen.checkout.components.base.OutputData
+
+class AwaitOutputData(override val isValid: Boolean, val paymentMethodType: String?) : OutputData
```
OscarSpruit, line 12: A `data class` would be better here

OscarSpruit (dismissed):

await/src/main/java/com/adyen/checkout/await/AwaitOutputData.kt
caiofaustino: 👍

OscarSpruit (approved):

OscarSpruit (approved):

PR #618, opened 2022-03-23
Handle additional actions in card component example

OscarSpruit:


example-app/src/main/java/com/adyen/checkout/example/ui/main/MainActivity.kt
```diff
@@ -108,8 +110,14 @@ class MainActivity : AppCompatActivity(), DropInCallback {
         super.onNewIntent(intent)
         Logger.d(TAG, "onNewIntent")
         if (intent == null) return
-        val result = DropIn.getDropInResultFromIntent(intent) ?: return
-        Toast.makeText(this, result, Toast.LENGTH_SHORT).show()
+        val result = DropIn.getDropInResultFromIntent(intent)
```
caiofaustino, line 113: I think it would be better to have this logic in it's own Activity, just to have a separation from the code we use to manage the example app (like the settings and the API) to the code that is actually an example that merchants can follow. What do you think?

caiofaustino (approved):

PR #623, opened 2022-03-24
Refactor Blik component to Kotlin

caiofaustino:


blik/src/main/java/com/adyen/checkout/blik/BlikView.kt
```diff
@@ -0,0 +1,108 @@
...
+    override fun observeComponentChanges(lifecycleOwner: LifecycleOwner) {
+        component.observeOutputData(lifecycleOwner, this)
+    }
+
+    override val isConfirmationRequired: Boolean
+        get() = true
+
+    override fun highlightValidationErrors() {
+        Logger.d(TAG, "highlightValidationErrors")
+        val outputData: BlikOutputData? = if (component.outputData != null) {
```
OscarSpruit, line 88: `val outputData = component.outputData ?: return` is shorter and also removes the need for the `!!` below

caiofaustino: Yep, missed this one, thx!

PR #624, opened 2022-03-24
Fix BCMC supportedPaymentMethodTypes

caiofaustino:


OscarSpruit (approved):

jreij:
Hi @waqar1992, this issue was introduced very recently so it was never released, it doesn't affect you regardless of which version you are using 🙂

caiofaustino:
This issues is not in any version, was just something that went by unnoticed in a refactor we are doing, but it's not an issue on any of the released versions.

caiofaustino:
This is working as intended, we return type `scheme` so that it can be processed as a regular credit card, either via Bcmc or via Mastercard and so has a higher chance of authorization. If you want to force it to be routed as Bcmc you can simply change that string to bcmc, but this is a logic specific to your needs.

We might implement a way for the shopper to choose that in the future.

waqar1992:
I understand what you are saying just wanted to confirm because in case of iOS we were getting it as BCMC. Thanks for the feedback though.

PR #627, opened 2022-03-25
Refactor CSE module to Kotlin

caiofaustino:


config/detekt/detekt-config.yml
```diff
@@ -555,7 +555,7 @@ style:
     ignoreNumbers: ['-1', '0', '1', '2']
     ignoreHashCodeFunction: true
     ignorePropertyDeclaration: false
-    ignoreLocalVariableDeclaration: false
+    ignoreLocalVariableDeclaration: true
```
OscarSpruit, line 558: Not sure if this is a good idea. I think it's better to suppress it on the classes themselves, if you don't feel like fixing the warnings now

caiofaustino: It's not really about that. How would you define a number that has a specific meaning only for that method? 
Do you think it should also be declared in the companion object?
I just thought declaring a local variable should be enough to give the number meaning.

OscarSpruit: If you want to go full best practices then declaring it in a companion object is the way. But I do agree that it feels a bit tedious if the "magic number" is only used in one place. This check is useful in general I find. And for places where the magic number is used once, the meaning is clear and you're pretty sure you won't use the same number any time soon I would suppress the warning

cse/src/main/java/com/adyen/checkout/cse/ClientSideEncrypter.kt
```diff
@@ -0,0 +1,152 @@
...
+    private var rsaCipher: Cipher? = null
+    private val secureRandom: SecureRandom
+
+    @Throws(EncryptionException::class)
+    @Suppress("ThrowsCount")
+    fun encrypt(plainText: String): String {
+        val aesKey = generateAesKey()
+        val iv = generateIV()
+        val encrypted: ByteArray = try {
+            aesCipher!!.init(Cipher.ENCRYPT_MODE, aesKey, IvParameterSpec(iv))
```
jreij, line 45: Can we get rid of the `!!`? I can see a bunch of them around the project so feel free to fix them all (the fixable ones at least) in another PR 😄

caiofaustino: I've been trying to get rid of most of them, but maybe I'm not being very thorough. I'll check this one out, thx!

OscarSpruit (approved):

PR #628, opened 2022-03-25
Move card brands below input field

OscarSpruit:
| Type | Before | After |
| - | - | - |
| Visible | image | image |
| Not visible | image | image |

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/CardListAdapter.kt
+++ b/card/src/main/java/com/adyen/checkout/card/CardListAdapter.kt
@@ -10,6 +10,8 @@ package com.adyen.checkout.card
 import android.annotation.SuppressLint
 import android.view.LayoutInflater
 import android.view.ViewGroup
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
 import androidx.recyclerview.widget.RecyclerView
 import com.adyen.checkout.card.CardListAdapter.ImageViewHolder
 import com.adyen.checkout.card.data.CardType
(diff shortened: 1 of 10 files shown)


card/src/main/java/com/adyen/checkout/card/CardView.kt
```diff
@@ -577,4 +609,18 @@ class CardView @JvmOverloads constructor(context: Context, attrs: AttributeSet?
             else -> null
         }
     }
+
+    fun setSupportedCardsList(cards: List<CardType>) {
```
caiofaustino, line 613: I think this breaks a bit the idea if this change. Afaik, we want the list to show up in the standalone component as well, so the `supportedCardsList` should be sent by the component in the `CardInputData` instead of someone having to inject it into the view manually. This works well for Drop-in only flow, but I'm not sure if that's the intent here.

card/src/main/java/com/adyen/checkout/card/CardListAdapter.kt
```diff
@@ -9,46 +9,58 @@ package com.adyen.checkout.card
...
 import com.adyen.checkout.card.CardListAdapter.ImageViewHolder
 import com.adyen.checkout.card.data.CardType
 import com.adyen.checkout.components.api.ImageLoader
 import com.adyen.checkout.components.ui.view.RoundCornerImageView
 
-class CardListAdapter(private val mImageLoader: ImageLoader, private val mSupportedCards: List<CardType>) : RecyclerView.Adapter<ImageViewHolder>() {
-    private var filteredCards: List<CardType> = emptyList()
+internal class CardListAdapter(
+    private val imageLoader: ImageLoader,
+) : ListAdapter<CardType, ImageViewHolder>(CardDiffCallback) {
```
caiofaustino, line 23: Thanks for refactoring this to ListAdapter, :)

card/src/main/java/com/adyen/checkout/card/CardView.kt
OscarSpruit: Ah okay, let me try to get it working with `CardInputData `

jreij: While you're at it make sure to test the stored card flow as well, not sure how it should look after this change.

caiofaustino (approved):

PR #630, opened 2022-03-29
Fix availability check with action only payment methods

jreij:


caiofaustino (approved):

caiofaustino:
Is this something I broke in the refactors?

PR #631, opened 2022-03-29
Refactor IssuerList components to Kotlin

caiofaustino:


dotpay/src/main/java/com/adyen/checkout/dotpay/DotpayRecyclerView.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Created by arman on 11/6/2019.
+ */
+package com.adyen.checkout.dotpay
+
+import android.content.Context
+import android.util.AttributeSet
+import com.adyen.checkout.components.model.payments.request.DotpayPaymentMethod
+import com.adyen.checkout.issuerlist.IssuerListRecyclerView
+
+class DotpayRecyclerView : IssuerListRecyclerView<DotpayPaymentMethod, DotpayComponent> {
```
OscarSpruit, line 15: What is the use of these "RecyclerView"s? It looks like they are all the same, except for `EPSRecyclerView`. I think it would be better to make `IssuerListRecyclerView` a normal class and then in `ComponentParsingProvider` do something like:
```
...
PaymentMethodTypes.EPS -> IssuerListRecyclerView(context).apply { hideIssuersLogo = true }
...
```

dotpay/src/main/java/com/adyen/checkout/dotpay/DotpaySpinnerView.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Created by arman on 11/6/2019.
+ */
+package com.adyen.checkout.dotpay
+
+import android.content.Context
+import android.util.AttributeSet
+import com.adyen.checkout.components.model.payments.request.DotpayPaymentMethod
+import com.adyen.checkout.issuerlist.IssuerListSpinnerView
+
+class DotpaySpinnerView : IssuerListSpinnerView<DotpayPaymentMethod, DotpayComponent> {
```
OscarSpruit, line 15: These `...SpinnerView`s seem to be unused, is that right? Can we remove them?

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerLogoCallback.kt
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Created by caiof on 26/4/2019.
+ */
+package com.adyen.checkout.issuerlist
+
+import android.graphics.drawable.BitmapDrawable
+import android.graphics.drawable.Drawable
+import com.adyen.checkout.components.api.LogoConnectionTask.LogoCallback
+
+class IssuerLogoCallback internal constructor(private val mIssuerId: String, private val mDrawableFetchedCallback: DrawableFetchedCallback) :
```
OscarSpruit, line 14: This class is unused

ui-core/src/main/java/com/adyen/checkout/components/ui/adapter/ClickableListRecyclerAdapter.kt
```diff
@@ -17,10 +17,10 @@ abstract class ClickableListRecyclerAdapter<ViewHolderT : RecyclerView.ViewHolde
     var onItemCLickedListener: OnItemCLickedListener? = null
 
     @CallSuper
-    override fun onBindViewHolder(viewHolderT: ViewHolderT, position: Int) {
-        viewHolderT.itemView.setOnClickListener {
+    override fun onBindViewHolder(viewHolder: ViewHolderT, position: Int) {
```
OscarSpruit, line 20: This `abstract` class is only extended by 1 class. Is this abstraction really needed?

dotpay/src/main/java/com/adyen/checkout/dotpay/DotpaySpinnerView.kt
caiofaustino: So this was an idea we had initially, idk if/how it's being used (we might get some more info when we add better analytics), but it's two different types of view representation for the same Component, a merchant could choose to display this payment as a RecyclerView or as a Spinner. 
For us in the Drop-in the RecyclerView makes more sense.

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerLogoCallback.kt
caiofaustino: 👍

ui-core/src/main/java/com/adyen/checkout/components/ui/adapter/ClickableListRecyclerAdapter.kt
caiofaustino: Yeah you are right, there are probably other interfaces with the same behavior implemented in other parts of the code too. But I think this change falls outside of the scope of this PR.

PR #634, opened 2022-03-30
Refactor Sepa to Kotlin.

caiofaustino:


sepa/src/main/java/com/adyen/checkout/sepa/Iban.kt
```diff
@@ -0,0 +1,290 @@
...
+import java.util.Locale
+import java.util.regex.Pattern
+
+class Iban private constructor(val value: String) {
+
+    @Suppress("MemberVisibilityCanBePrivate")
+    val countryCode: String = value.substring(0, COUNTRY_CODE_POSITION_END)
+    val checkDigits: String = value.substring(CHECK_DIGIT_POSITION_START, CHECK_DIGIT_POSITION_END)
+    @Suppress("unused")
+    val bban: String
```
OscarSpruit, line 22: This could just be `val bban: String = value.substring(IBAN_BLOCK_SIZE)` as the value of `value` can't change

OscarSpruit (approved):

OscarSpruit (approved):

PR #641, opened 2022-03-31
Add session API calls

jreij:


sessions-core/src/main/java/com/adyen/checkout/sessions/repository/SessionRepository.kt
```diff
@@ -0,0 +1,184 @@
...
+            val request = SessionSetupRequest(session.sessionData.orEmpty(), order)
+            return SessionSetupConnection(
+                request = request,
+                environment = configuration.environment,
+                sessionId = session.id,
+                clientKey = configuration.clientKey
+            ).suspendedCall()
+        } catch (e: IOException) {
+            Logger.e(TAG, "SessionSetupConnection Failed", e)
+            throw CheckoutException("Unable to setup session")
```
caiofaustino, line 64: So I've been thinking we should start having some better error handling in general, this looks like a good place to start, can we have a chat about it?

caiofaustino:
I saw you separated the commits based on the type of call (balance, order, payments, details), do you think we should also do that with the model packages? Instead of `request` and `response` we can have them by feature, and have both the request and the response in the same package. What do you think?
`components-core` already kind of does that too

sessions-core/src/main/java/com/adyen/checkout/sessions/repository/SessionRepository.kt
jreij: Yes indeed, let's discuss it and make another PR to refactor all connections

jreij:
> I saw you separated the commits based on the type of call (balance, order, payments, details), do you think we should also do that with the model packages? Instead of `request` and `response` we can have them by feature, and have both the request and the response in the same package. What do you think? `components-core` already kind of does that too

Yeah sure, this is only the example app anyway but I agree as long as we're consistent 👍

caiofaustino (approved):

PR #643, opened 2022-03-31
Fix crash when initializing iDeal.

caiofaustino:
Fix crash from call to not initialized `supportedPaymentMethodTypes` in BasePaymentComponent when underlying val was not yet assigned.

### DIFF
--- a/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponent.kt
+++ b/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponent.kt
@@ -48,18 +48,6 @@ class BcmcComponent(
 ) : BasePaymentComponent<BcmcConfiguration, BcmcInputData, BcmcOutputData,
     PaymentComponentState<CardPaymentMethod>>(savedStateHandle, paymentMethodDelegate, configuration) {
 
-    companion object {
-        private val TAG = LogUtil.getTag()
-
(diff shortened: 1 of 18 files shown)


bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponent.kt
```diff
@@ -154,4 +141,16 @@ class BcmcComponent(
...
+    companion object {
+        private val TAG = LogUtil.getTag()
+
+        private val PAYMENT_METHOD_TYPES = arrayOf(PaymentMethodTypes.BCMC)
+
+        @JvmField
+        val PROVIDER: PaymentComponentProvider<BcmcComponent, BcmcConfiguration> = BcmcComponentProvider()
+
+        @JvmField
+        val SUPPORTED_CARD_TYPE = CardType.BCMC
```
jreij, line 154: Does this field need to be public?

jreij, line 154: Does this field need to be public?

caiofaustino: It's being used in the view, I guess it could be `internal`, but it's been like that for quite a while now.

jreij (approved):

jreij (approved):

PR #644, opened 2022-04-01
Use separate intent filters to handle action results 

OscarSpruit:


example-app/src/main/AndroidManifest.xml
```diff
@@ -38,16 +39,21 @@
...
+            confusing user experience. -->
             <intent-filter>
                 <action android:name="android.intent.action.VIEW" />
 
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
 
                 <data
-                    android:host="${applicationId}"
+                    android:host="${applicationId}.card"
```
jreij, line 56: Can you try to keep the same `android:host` but set `android:path="/card"`? IMO it looks cleaner to use the same host for the same application. The return URL would then be `adyencheckout://com.adyen.checkout.example/card`. WDYT?

OscarSpruit: I like it 👌

drop-in/src/main/AndroidManifest.xml
```diff
@@ -19,9 +20,16 @@
 
                 <data
                     android:host="${applicationId}"
+                    android:path=""
+                    android:scheme="${checkoutRedirectScheme}"
+                    tools:ignore="AppLinkUrlError" />
```
jreij, line 25: Out of curiosity, why do we 2 data tags and the tools ignore?

example-app/src/main/java/com/adyen/checkout/example/ui/card/CardViewModel.kt
```diff
@@ -89,8 +89,7 @@ internal class CardViewModel @Inject constructor(
                 amount = keyValueStorage.getAmount(),
                 countryCode = keyValueStorage.getCountry(),
                 merchantAccount = keyValueStorage.getMerchantAccount(),
-                // Should be set correctly for additional actions
-                redirectUrl = "",
+                redirectUrl = "adyencheckout://com.adyen.checkout.example/card",
```
jreij, line 92: We usually get this URL by calling `RedirectComponent.getReturnUrl(applicationContext)`. Maybe you can try "${RedirectComponent.getReturnUrl(applicationContext)}/card"? You need to make your view model an `AndroidViewModel` but that's alright I guess?

drop-in/src/main/AndroidManifest.xml
OscarSpruit: The first one with `""` is to handle deep links like `scheme://host`
The first second one with `"/"` is to handle deep links like `scheme://host/`
It's just defensive, because some libraries/people always suffix URLs with a slash

example-app/src/main/java/com/adyen/checkout/example/ui/card/CardViewModel.kt
OscarSpruit: Tbh I don't really like `AndroidViewModel`, because it makes unit testing difficult. Not that we have tests at the moment, but you get the point 😅 There are other ways to use `RedirectComponent.getReturnUrl` without using `AndroidViewModel`, but I didn't want to spent too much time on it

jreij: Fair enough 👍

jreij (approved):

PR #645, opened 2022-04-01
Fix crash in GooglePayConfiguration init

OscarSpruit:


googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayConfiguration.kt
```diff
@@ -95,9 +95,9 @@ class GooglePayConfiguration : Configuration, AmountConfiguration {
         countryCode = parcel.readString()
         merchantInfo = parcel.readParcelable(MerchantInfo::class.java.classLoader)
         @Suppress("UNCHECKED_CAST")
-        allowedAuthMethods = parcel.readArrayList(String::class.java.classLoader) as List<String>
+        allowedAuthMethods = (parcel.readArrayList(String::class.java.classLoader) as? List<String>) ?: emptyList()
```
jreij, line 98: I wasn't clear previously, i think `allowedAuthMethods` should be declared as nullable (`List<String>?`) judging by how it was before the Kotlin refactor. Sorry for being nitpicky 😬

caiofaustino: Don't judge these by the refactor.
https://developers.google.com/pay/api/android/reference/request-objects#PaymentMethod

jreij: Ahh ok that makes sense. But in that case I suggest something else, both `allowedAuthMethods` and `allowedCardNetworks` are required but `allowedAuthMethods` is non nullable and defaults to `AllowedAuthMethods.allAllowedAuthMethods` in the Builder class while `allowedCardNetworks` is nullable then it defaults to `AllowedCardNetworks.allAllowedCardNetworks` is the `GooglePayParams` class.

So maybe for consistency and better troubleshooting, let's make them both nullable and add a new method in `GooglePayParams` to get the `allowedAuthMethods` as well?

caiofaustino: Yeah I agree, we can try to strive to make the Config file reflect the merchant's intent, and we can route things with multiple defaults inside the Component.

jreij (approved):
✨

PR #650, opened 2022-04-06
Fix split in LogUtil.

caiofaustino:


ozgur00 (approved):

checkout-core/src/main/java/com/adyen/checkout/core/log/LogUtil.kt
```diff
@@ -53,8 +53,7 @@ object LogUtil {
     private val callerClassName: String
         get() {
             val stElements = Thread.currentThread().stackTrace
-            for (i in 1 until stElements.size) {
-                val ste = stElements[i]
+            for (ste in stElements.drop(1)) {
```
jreij, line 56: Can `drop(1)` throw an exception if the list is empty or something?

caiofaustino: Nope, just returns empty list.

PR #653, opened 2022-04-07
Internal networking refactor part 1: Abstract internal networking

OscarSpruit:
Removed the inheritance from `Connection` and created the `HttpClient` interface, which `Connection` implementations now use (these are called `Service`s now). The structure looks something like this:

```
                      HttpClientFactory
                     /
Repository -> Service            ConnectionHttpClient
                     \          /
                      HttpClient
                                \
                                 OkHttpClient (comes in a future PR)
```


sessions-core/src/main/java/com/adyen/checkout/sessions/api/SessionSetupService.kt
```diff
@@ -8,36 +8,43 @@
...
-    "${environment.baseUrl}$ENDPOINT$sessionId/setup?clientKey=$clientKey"
+
+internal class SessionSetupService(
+    private val environment: Environment
 ) {
-    @Throws(IOException::class, JSONException::class)
-    override fun call(): SessionSetupResponse {
-        Logger.v(TAG, "call - $url")
+
+    suspend fun setupSession(
```
jreij, line 29: Just a thought for the next refactor. Do we want every API call to be just a method inside a Service class or to have its own separate class?
- For the first option I guess we keep what we have now but merge some classes together. For example `SessionBalanceService - SessionCancelOrderService - SessionCreateOrderService - SessionDetailsService - SessionPaymentsService - SessionSetupService` should all just be `SessionSevice`.
- In addition, if each API has its own a separate class, we could maybe make things more generic, like every API should provide its URL, method type, request and response type. This way Service classes can call any API the same way basically.
WDYT?

jreij (approved):

sessions-core/src/main/java/com/adyen/checkout/sessions/api/SessionSetupService.kt
OscarSpruit: Went with the first option in the next refactor. Made a method there to generify making requests, but we can take that a step further (will explain in that PR)

PR #655, opened 2022-04-11
Use OkHttp for networking

OscarSpruit:
Removed `ConnectionHttpClient` and introduced `OkHttpClient`.
Also merged all session services into one class.


sessions-core/src/main/java/com/adyen/checkout/sessions/api/SessionService.kt
```diff
@@ -0,0 +1,132 @@
...
+    ): SessionCancelOrderResponse = withContext(Dispatchers.IO) {
+        val resultJson = makeRequest(
+            path = "v1/sessions/$sessionId/orders/cancel?clientKey=$clientKey",
+            requestJson = SessionCancelOrderRequest.SERIALIZER.serialize(request),
+        )
+
+        SessionCancelOrderResponse.SERIALIZER.deserialize(resultJson)
+    }
+
+    private fun makeRequest(
```
OscarSpruit, line 113: We could make this even more generic by creating a global extension function every service could use. Something like:
```
fun <T, R> HttpClient.post(
    baseUrl: String,
    path: String,
    request: T,
    requestSerializer: Serializer<T>,
    responseSerializer: Serializer<R>,
): R {
    Logger.v(TAG, "call - $path")
    val requestJson = requestSerializer.serialize(request)
    Logger.v(TAG, "request - ${requestJson.toStringPretty()}")
    val httpClient = HttpClientFactory.getHttpClient(baseUrl)
    val result = httpClient.post(path, requestJson.toString())
    val resultJson = JSONObject(String(result, Charsets.UTF_8))
    Logger.v(TAG, "response: ${resultJson.toStringPretty()}")
    return responseSerializer.deserialize(resultJson)
}
```
What do you think?

components-core/src/main/java/com/adyen/checkout/components/status/api/StatusService.kt
```diff
@@ -27,7 +26,7 @@ internal class StatusService {
         val body = StatusRequest.SERIALIZER.serialize(statusRequest).toString()
 
         val httpClient = HttpClientFactory.getHttpClient(url)
-        val bytes = httpClient.post("", body, CONTENT_TYPE_JSON_HEADER)
+        val bytes = httpClient.post("", body)
```
jreij, line 29: We should refactor this next, the `StatusApi` is directly building the url which is why we post with an empty string here.

sessions-core/src/main/java/com/adyen/checkout/sessions/api/SessionService.kt
jreij: I like the idea! Maybe make a new commit with it and we can discuss it there?

components-core/src/main/java/com/adyen/checkout/components/status/api/StatusService.kt
OscarSpruit: Sure, will take a look at this next

sessions-core/src/main/java/com/adyen/checkout/sessions/api/SessionService.kt
OscarSpruit: https://github.com/Adyen/adyen-android/pull/655/commits/1255ac1c51c3cb8b4f70fc0a4c0793908071844e

checkout-core/src/main/java/com/adyen/checkout/core/api/HttpClientExt.kt
```diff
@@ -0,0 +1,51 @@
...
+import com.adyen.checkout.core.model.toStringPretty
+import org.json.JSONObject
+
+private val TAG = LogUtil.getTag()
+
+fun <T : ModelObject> HttpClient.get(
+    path: String,
+    responseSerializer: ModelObject.Serializer<T>
+): T {
+    Logger.v(TAG, "GET - $path")
```
jreij, line 23: One last comment 🙈  Maybe we can make the path logs as `Logger.d` and keep the request and response bodies as `Logger.v`?

checkout-core/src/main/java/com/adyen/checkout/core/api/HttpClientExt.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    val result = this.get(path)
+    val resultJson = JSONObject(String(result, Charsets.UTF_8))
+
+    Logger.v(TAG, "response - ${resultJson.toStringPretty()}")
+
+    return responseSerializer.deserialize(resultJson)
+}
+
+fun <T : ModelObject, R : ModelObject> HttpClient.post(
```
jreij, line 33: I was thinking of passing `HttpClient` as an argument but with extensions it looks much better 💯

jreij (approved):

PR #659, opened 2022-04-12
Refactor internal networking

OscarSpruit:


caiofaustino (approved):

components-core/src/main/java/com/adyen/checkout/components/repository/OrderStatusRepository.kt
```diff
@@ -32,11 +31,10 @@ class OrderStatusRepository {
...
             val request = OrderStatusRequest(orderData)
-            return OrderStatusConnection(
+            return OrderStatusService(configuration.environment).getOrderStatus(
                 request,
-                environment = configuration.environment,
-                clientKey = configuration.clientKey
-            ).suspendedCall()
+                configuration.clientKey
+            )
         } catch (e: IOException) {
```
jreij, line 38: Instead of try catch and throwing a `CheckoutException`, should we return a wrapper (`sealed class ApiResult` or something)? Also keep in mind that we might need to parse the response body for some calls and extract values from it

OscarSpruit: Like the idea! It's a bit of a behavioural change though, so I was thinking maybe we can merge this PR already into develop and do that on a separate branch?

jreij: 👍

OscarSpruit: Was also thinking about `Result`, but a sealed class for errors wouldn't really work with it 🤔

PR #660, opened 2022-04-12
Remove generic Providers.

caiofaustino:
The goal here was to remove the Generic provider classes we have to basically:
- Avoid using reflection
- Allow each constructor to be different and make use of DI to come.


bacs/src/main/java/com/adyen/checkout/bacs/BacsProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.os.Bundle
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.ViewModelStoreOwner
+import androidx.savedstate.SavedStateRegistryOwner
+import com.adyen.checkout.components.PaymentComponentProvider
+import com.adyen.checkout.components.base.GenericPaymentMethodDelegate
+import com.adyen.checkout.components.base.lifecycle.viewModelFactory
+import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
+
+class BacsProvider : PaymentComponentProvider<BacsDirectDebitComponent, BacsDirectDebitConfiguration> {
```
OscarSpruit, line 20: Tbh I actually like the `...ComponentProvider` suffix 😅, because it exactly tells what it provides. Alternatively, the provider could be a sub class of the component, so you get `BacsComponent.Provider`

caiofaustino: That would be interesting but I don't think I can do it in another file can I?

OscarSpruit: No, unfortunately not. AFAIK there is nothing like extensions classes :P

caiofaustino: I'm adding the suffix again.

OscarSpruit (approved):

OscarSpruit (approved):

PR #661, opened 2022-04-12
Refactor repositories

OscarSpruit:
- Use Kotlin Result in repositories to handle exceptions
- Differentiate between HTTP errors and "normal" IO errors
    - Parse the error response if possible

### DIFF
--- a/checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
@@ -8,12 +8,15 @@
 
 package com.adyen.checkout.core.api
 
+import com.adyen.checkout.core.exception.ModelSerializationException
 import okhttp3.Headers.Companion.toHeaders
 import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
+import org.json.JSONException
 import org.json.JSONObject
+import java.io.IOException
 
 internal class OkHttpClient(
     private val client: OkHttpClient,
@@ -49,12 +52,15 @@ internal class OkHttpClient(
                 ?.bytes()
                 ?: ByteArray(0)
         } else {
-            @Suppress("TooGenericExceptionCaught")
             val errorBody = try {
                 response.body?.string()
                     ?.let { JSONObject(it) }
                     ?.let { ErrorResponseBody.SERIALIZER.deserialize(it) }
-            } catch (e: Throwable) {
+            } catch (e: IOException) {
+                null
+            } catch (e: JSONException) {
+                null
+            } catch (e: ModelSerializationException) {
                 null
             }
             throw HttpException(response.code, response.message, errorBody)
--- a/checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
@@ -13,7 +13,7 @@ import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import okhttp3.RequestBody.Companion.toRequestBody
-import java.io.IOException
+import org.json.JSONObject
 
 internal class OkHttpClient(
     private val client: OkHttpClient,
@@ -49,7 +49,15 @@ internal class OkHttpClient(
                 ?.bytes()
                 ?: ByteArray(0)
         } else {
-            throw IOException(response.body?.string())
+            @Suppress("TooGenericExceptionCaught")
+            val errorBody = try {
+                response.body?.string()
+                    ?.let { JSONObject(it) }
+                    ?.let { ErrorResponseBody.SERIALIZER.deserialize(it) }
+            } catch (e: Throwable) {
+                null
+            }
+            throw HttpException(response.code, response.message, errorBody)
         }
     }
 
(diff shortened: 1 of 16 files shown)


sessions-core/src/main/java/com/adyen/checkout/sessions/repository/SessionRepository.kt
```diff
@@ -43,135 +40,93 @@ internal class SessionRepository(
         configuration: Configuration,
         session: Session,
         order: OrderRequest?
-    ): SessionSetupResponse {
+    ): Result<SessionSetupResponse> = runCatching {
```
caiofaustino, line 43: A small note about `runCatching`, it might swallow `CancellationException`
https://github.com/Kotlin/kotlinx.coroutines/issues/1814

checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
```diff
@@ -49,7 +49,14 @@ internal class OkHttpClient(
                 ?.bytes()
                 ?: ByteArray(0)
         } else {
-            throw IOException(response.body?.string())
+            val errorBody = try {
+                response.body?.string()
+                    ?.let { JSONObject(it) }
+                    ?.let { ErrorResponseBody.SERIALIZER.deserialize(it) }
+            } catch (e: Throwable) {
```
caiofaustino, line 56: Why `Throwable`? Shouldn't it be `IOException`?

sessions-core/src/main/java/com/adyen/checkout/sessions/repository/SessionRepository.kt
OscarSpruit: Ohh didn't know that. Should we create a `runSuspendCatching` like some suggest in that thread?

checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
OscarSpruit: There are 3 different exceptions that should be handled in the same way, so I thought this would be okay

sessions-core/src/main/java/com/adyen/checkout/sessions/repository/SessionRepository.kt
caiofaustino: Yeah I think that would be best.

checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
caiofaustino: Can we do something like this in Kotlin? `catch (ExceptionType1 | Exceptiontype2 ex)`

OscarSpruit: Nope each exception needs it's own catch block :(
```
try { }
catch(e: SomeException) { }
catch(e: SomeOtherException) { }
```

checkout-core/src/main/java/com/adyen/checkout/core/api/OkHttpClient.kt
```diff
@@ -49,7 +52,18 @@ internal class OkHttpClient(
                 ?.bytes()
                 ?: ByteArray(0)
         } else {
-            throw IOException(response.body?.string())
+            val errorBody = try {
```
jreij, line 55: Were you able to make an API call that returns an error to test it? If not I can maybe suggest something

OscarSpruit: Yes, I tested it by playing with the bin lookup call

components-core/src/main/java/com/adyen/checkout/components/repository/PublicKeyRepository.kt
```diff
@@ -10,28 +10,40 @@ package com.adyen.checkout.components.repository
...
-        repeat(CONNECTION_RETRIES) {
+
+        retryOnFailure(CONNECTION_RETRIES) {
+            PublicKeyService(environment).getPublicKey(clientKey).publicKey
+        }
+    }
+
+    @Suppress("SameParameterValue")
+    private inline fun <T> retryOnFailure(times: Int, block: () -> T): T {
+        repeat(times - 1) {
```
jreij, line 32: Why should this be `times - 1` instead of `times`?

jreij: Never mind, just saw the extra `return block()`, I'm not sure if this is obvious and I missed it or we should improve it somehow

OscarSpruit: Let me explain with an example. You do `retryOnFailure(3) {}`. The repeat statement will try the block `3-1=2` times ignoring any exceptions. If it fails both times, then the last line of the method `return block()` will try one last time, throwing an exception if it fails. So in total the block has been executed 3 times

jreij: Does this look better or worse? 😅 
```kotlin
    @Suppress("SameParameterValue")
    private inline fun <T> retryOnFailure(times: Int, block: () -> T): T {
        var throwable = Throwable()
        repeat(times) { index ->
            @Suppress("TooGenericExceptionCaught")
            try {
                return block()
            } catch (e: Throwable) {
                throwable = e
            }
        }
        throw  throwable
    }
```

jreij: Your logic is sound and definitely works, I'm only wondering if `repeat(times - 1)` is confusing because it made me think `repeat` is zero based or something. Still I can't write it in a "cleaner" way because the compiler will complain that we aren't returning a value in all paths.

OscarSpruit: It's easier to understand, so I like it 👍 

Small adjustment: `var throwable = Throwable()` -> `lateinit var throwable: Throwable`

jreij: Haha amazing, that's actually one of the few times I see `lateinit` used correctly outside of the usual init during lifecycle use case 👏

jreij (approved):

PR #662, opened 2022-04-13
Dependency Injection

caiofaustino:


jreij:
Closing this for now but keeping the branch, we are unsure how to implement DI yet and the components refactor might change what we need

PR #663, opened 2022-04-13
Improve theming

OscarSpruit:
Build upon Material theme to provide more customisability (dark theme, shape appearance, etc.)

Projects using the sdk will have to do a migration, but it should be pretty easy


jreij:
I think something went wrong with the 3rd commit, check the commit description

ui-core/src/main/res/values/styles.xml
```diff
@@ -22,42 +25,44 @@
         <item name="android:textColorPrimary">?attr/colorOnBackground</item>
         <item name="android:textColorSecondary">?attr/colorOnBackground</item>
         <item name="android:textColorTertiary">?attr/colorOnBackground</item>
+
+        <item name="bottomSheetDialogTheme">@style/Adyen.Checkout.BottomSheetDialogTheme</item>
     </style>
 
     <!-- Base theme for Checkout -->
-    <style name="AdyenCheckout" parent="AppTheme" />
+    <style name="Adyen.Checkout" parent="Adyen.AppTheme" />
```
jreij, line 33: Is there a benefit from renaming `AdyenCheckout` to `Adyen.Checkout`? I'm worried it will add more migration steps for merchants and (correct me if i'm wrong) if they were already overriding our styles they will not get a build error when they update so it won't be obvious that they need to rename `AdyenCheckout` to `Adyen.Checkout` 🤔

OscarSpruit:
> I think something went wrong with the 3rd commit, check the commit description

You don't like the message? 😂 That happens when you rebase with Android Studio, there is a merge conflict, you fix it and just continue

ui-core/src/main/res/values/styles.xml
OscarSpruit: Yeah I agree with you, let me revert it. I was experimenting with theme inheritance, but in the end what I wanted wasn't possible. Thought it looked kinda nice, so left it in

ui-core/src/main/java/com/adyen/checkout/components/ui/view/AdyenSwipeToRevealLayout.kt
```diff
@@ -346,6 +349,20 @@ class AdyenSwipeToRevealLayout @JvmOverloads constructor(
         }
     }
 
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+        fixElevationOverlay()
+    }
+
+    private fun fixElevationOverlay() {
```
jreij, line 357: What exactly was this issue here?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -43,6 +43,9 @@ class PaymentMethodListDialogFragment :
     PaymentMethodAdapter.OnPaymentMethodSelectedCallback,
     PaymentMethodAdapter.OnStoredPaymentRemovedCallback {
 
+    private var _binding: FragmentPaymentMethodsListBinding? = null
```
jreij, line 46: We have this use case in many other places as well, we should fix the rest in another PR 👍

jreij (approved):

ui-core/src/main/java/com/adyen/checkout/components/ui/view/AdyenSwipeToRevealLayout.kt
OscarSpruit: In dark mode elevation overlay makes surfaces lighter the higher the elevation (instead of casting a bigger shadow). For some reason the surface was way to light here, so this makes sure the surface has the same color as the surface behind this layout

jreij (approved):

PR #664, opened 2022-04-13
Fix remove stored PM state issues

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -53,6 +55,7 @@ class PaymentMethodListDialogFragment :
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
         Logger.d(TAG, "onCreateView")
+        binding = FragmentPaymentMethodsListBinding.inflate(inflater, container, false)
```
OscarSpruit, line 58: Lol I did the same thing here https://github.com/Adyen/adyen-android/pull/663/files#diff-7692dd53b9bd81677993aa49490e4f269cf9b9ef8431769fa63daf1ac77a5dab

Note that this will cause a memory, because fragments live longer than their views. So you need to clear all references to the binding in `onDestroyView` as described here https://developer.android.com/topic/libraries/view-binding#fragments

jreij: Fixed 👍

OscarSpruit (approved):

PR #665, opened 2022-04-14
Fix remove stored PM visual bug

jreij:


OscarSpruit (approved):
What was the actual fix in the end? The `child == mainView` in `AdyenSwipeToRevealLayout`?

jreij:
@OscarSpruit yep! So inside `tryCaptureView` you're supposed to call `dragHelper.captureChildView` on the child you want to allow being dragged. We were doing that without checking which view triggered the call. So when you click on the button to delete the stored PM, a drag was being triggered by mistake and the button jumps to the start of the layout and therefore looks like it disappeared.

PR #666, opened 2022-04-15
Refactor example request/response objects in example app

OscarSpruit:


example-app/src/main/java/com/adyen/checkout/example/data/api/CheckoutApiService.kt
```diff
@@ -39,26 +37,24 @@ internal interface CheckoutApiService {
...
-    suspend fun paymentsAsync(@Body paymentsRequest: PaymentsRequest): ResponseBody
+    suspend fun paymentsAsync(@Body paymentsRequest: JSONObject): JSONObject
 
     @POST("payments/details")
-    fun details(@Body detailsRequest: JSONObject): Call<ResponseBody>
+    fun details(@Body detailsRequest: JSONObject): Call<JSONObject>
 
     @POST("payments/details")
-    suspend fun detailsAsync(@Body detailsRequest: JSONObject): ResponseBody
+    suspend fun detailsAsync(@Body detailsRequest: JSONObject): JSONObject
```
OscarSpruit, line 50: Is it okay to return `JSONObject`s or should we use data classes?

jreij: `JSONObject` is good enough for the example app 👍  I think it's even better if it "just works" instead of using proper classes since we don't want anyone to replicate this behavior and make API calls form the client app.

example-app/src/main/java/com/adyen/checkout/example/repositories/paymentMethods/PaymentsRepository.kt
```diff
@@ -47,15 +50,34 @@ internal class PaymentsRepositoryImpl(private val checkoutApiService: CheckoutAp
...
+            call = { checkoutApiService.paymentsAsync(paymentsRequest.combineToJSONObject()) }
         )
     }
 
+    private fun PaymentsRequest.combineToJSONObject(): JSONObject {
+        val moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+        val adapter = moshi.adapter(PaymentsRequestData::class.java)
+        val requestDataJson = JSONObject(adapter.toJson(this.requestData))
+
+        return this.paymentComponentData
```
jreij, line 66: I had this comment in `PaymentRequestAdapter`:
`paymentComponentData should be merged second to override any possible duplicated fields`
So I think we should `return this.requestDataJson.putAll(paymentComponentData)`?

OscarSpruit: Good catch!

example-app/src/main/java/com/adyen/checkout/example/repositories/paymentMethods/PaymentsRepository.kt
```diff
@@ -15,25 +15,27 @@ import com.adyen.checkout.example.data.api.model.CancelOrderRequest
...
+    fun paymentsRequest(paymentsRequest: PaymentsRequest): JSONObject?
+    suspend fun paymentsRequestAsync(paymentsRequest: PaymentsRequest): JSONObject?
+    fun detailsRequest(detailsRequest: JSONObject): JSONObject?
+    suspend fun detailsRequestAsync(detailsRequest: JSONObject): JSONObject?
+    suspend fun balanceRequestAsync(request: BalanceRequest): JSONObject?
+    suspend fun createOrderAsync(orderRequest: CreateOrderRequest): JSONObject?
+    suspend fun cancelOrderAsync(request: CancelOrderRequest): JSONObject?
 }
 
+@Suppress("TooManyFunctions")
```
jreij, line 38: This lint check annoys me the most 😆  I keep thinking of adding a global rule to ignore it. Do you have any ideas why we could benefit from keeping it?

jreij (approved):

example-app/src/main/java/com/adyen/checkout/example/repositories/paymentMethods/PaymentsRepository.kt
OscarSpruit: Most of the times this check is a good indicator that your class is doing too many different things. For this repository for example it could be split up in session, payment methods, payment request, details, balance and order repositories. For Android classes (Fragment, Activity and maybe ViewModel) this check doesn't really make sense, but I believe it's possible to disable it for those

PR #672, opened 2022-04-19
Provide a method to start a payment with a session

OscarSpruit:


drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -34,8 +33,6 @@ import com.adyen.checkout.dropin.ui.DropInActivity
 object DropIn {
     private val TAG = LogUtil.getTag()
 
-    const val DROP_IN_REQUEST_CODE = 529
-
     const val RESULT_KEY = "payment_result"
```
jreij, line 36: These fields can now be `internal` I think

jreij, line 36: These fields can now be `internal` I think

jreij (approved):

PR #682, opened 2022-04-29
Create script to trigger manual release of v4

jreij:


jreij:
Hahaha 😅  We had this script on v3 already so it should work the same way on v4

PR #690, opened 2022-05-02
Create `SessionDropInService`

OscarSpruit:


drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -133,9 +135,13 @@ object DropIn {
 
         DropInPrefs.setShopperLocale(context, dropInConfiguration.shopperLocale)
 
+        val configuration = DropInConfiguration.Builder(dropInConfiguration)
```
jreij, line 138: We shouldn't commit this right? 🤔

OscarSpruit: Whoops my bad 👀

drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
```diff
@@ -0,0 +1,211 @@
...
+                .fold(
+                    onSuccess = { response ->
+                        if (response.balance.value <= 0) {
+                            onFailure("Not enough balance")
+                        } else {
+                            val balanceResult = BalanceResult(response.balance, response.transactionLimit)
+                            sendBalanceResult(BalanceDropInServiceResult.Balance(balanceResult))
+                        }
+                    },
+                    onFailure = ::onFailure
```
jreij, line 132: The common `onFailure` method will always return `DropInServiceResult.Error` but that only applies for payments and details calls. For example in this method we should be returning `BalanceDropInServiceResult.Error`. Internally it will work because we handle all errors the same way but if we change this handling later we might face unexpected issues.

OscarSpruit: Ah right, I'll change it to the specific types

jreij (approved):

PR #691, opened 2022-05-03
Refactor MainActivity ViewModel

OscarSpruit:


example-app/src/main/java/com/adyen/checkout/example/ui/main/MainNavigation.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import com.adyen.checkout.components.model.PaymentMethodsApiResponse
+import com.adyen.checkout.sessions.model.Session
+
+internal sealed class MainNavigation {
+    object Card : MainNavigation()
+    data class DropIn(val paymentMethodsApiResponse: PaymentMethodsApiResponse) : MainNavigation()
+    data class DropInWithSession(
+        val session: Session,
+        val paymentMethodsApiResponse: PaymentMethodsApiResponse,
```
jreij, line 19: This one `val paymentMethodsApiResponse: PaymentMethodsApiResponse` is also just for temporary testing, normally it should be only session here

jreij, line 19: This one `val paymentMethodsApiResponse: PaymentMethodsApiResponse` is also just for temporary testing, normally it should be only session here

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+        getSessionRequest(
+            merchantAccount = keyValueStorage.getMerchantAccount(),
+            shopperReference = keyValueStorage.getShopperReference(),
+            amount = keyValueStorage.getAmount(),
+            countryCode = keyValueStorage.getCountry(),
+            shopperLocale = keyValueStorage.getShopperLocale(),
+            splitCardFundingSources = keyValueStorage.isSplitCardFundingSources(),
+            isExecuteThreeD = keyValueStorage.isExecuteThreeD(),
+            isThreeds2Enabled = keyValueStorage.isThreeds2Enable(),
+            // RedirectComponent.getReturnUrl(applicationContext) should be used here
```
jreij, line 102: Will we solve this later or in this PR?

jreij, line 102: Will we solve this later or in this PR?

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainNavigation.kt
OscarSpruit: Yeah I know, I thought it might come in handy while we are working on sessions

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainViewModel.kt
OscarSpruit: Hmm let me try it in this one. If it gets too big I will open another PR

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainNavigation.kt
jreij: Feel free to keep it until we have a proper fix

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainViewModel.kt
OscarSpruit: Turns out to be quite easy with `SavedStateHandle`, so also did it for `CardActivity`

OscarSpruit: I wanted to use assisted injection (kind of similar to what you're suggesting), but this is the better approach according to this thread https://github.com/google/dagger/issues/2287

PR #698, opened 2022-05-13
Support launching DropInActivity with a session

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -173,18 +173,7 @@ class DropInActivity :
...
-                    if (firstPaymentMethod != null) {
-                        showComponentDialog(firstPaymentMethod)
-                    } else {
-                        throw CheckoutException("First payment method is null")
-                    }
-                }
-                dropInViewModel.showPreselectedStored -> showPreselectedDialog()
-                else -> showPaymentMethodsDialog()
-            }
+            dropInViewModel.activityCreated()
```
OscarSpruit, line 176: `onCreated` would be a better name imo. That way it doesn't matter if the view model is used by an activity or fragment and it follows the android naming (e.g. onThis, onThat, onWhateverHappened, etc)

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInActivityEvent.kt
```diff
@@ -9,11 +9,23 @@
...
 import com.adyen.checkout.components.model.payments.request.OrderRequest
+import com.adyen.checkout.components.model.payments.response.Action
+import com.adyen.checkout.dropin.ui.giftcard.GiftCardPaymentConfirmationData
 
 sealed class DropInActivityEvent {
     data class MakePartialPayment(val paymentComponentState: PaymentComponentState<*>) : DropInActivityEvent()
     object ShowPaymentMethods : DropInActivityEvent()
     class CancelOrder(val order: OrderRequest, val isDropInCancelledByUser: Boolean) : DropInActivityEvent()
     object CancelDropIn : DropInActivityEvent()
+    class LoadFragment(val fragment: DropInFragmentToLoad) : DropInActivityEvent()
```
OscarSpruit, line 22: I would do something like this:
```
...
    class NavigateTo(val destination: DropInDestination) : DropInActivityEvent()
}

internal sealed class DropInDestination {
     object PreselectedStored : DropInDestination()
     object PaymentMethods : DropInDestination()
     class PaymentComponent(val paymentMethod: PaymentMethod) : DropInDestination()
     class ActionComponent(val action: Action) : DropInDestination()
     class GiftCardPaymentConfirmation(val data: GiftCardPaymentConfirmationData) : DropInDestination()
 }
```
That way it's also independent of wether a fragment or activity uses it and it tells a little better what is going to happen imo

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -112,23 +116,34 @@ class DropInViewModel(
...
+        return paymentMethodsApiResponse?.paymentMethods.orEmpty()
+    }
+
+    fun getStoredPaymentMethods(): List<StoredPaymentMethod> {
+        return paymentMethodsApiResponse?.storedPaymentMethods.orEmpty()
+    }
 
-    val preselectedStoredPayment
-        get() = paymentMethodsApiResponse.storedPaymentMethods?.firstOrNull {
+    fun showPreselectedStored(): Boolean {
```
OscarSpruit, line 127: This name implies that something will be shown, so I would call it `shouldShowPreselectedStored`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -137,6 +152,41 @@ class DropInViewModel(
...
+        } else {
+            loadDropIn()
+        }
+    }
+
+    private fun isInitializedWithSession(): Boolean {
+        return initialSession != null
+    }
+
+    fun sessionSetupSuccessful(sessionSetupResponse: SessionSetupResponse) {
```
OscarSpruit, line 169: `onSessionSetupSuccessful `

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -137,6 +152,41 @@ class DropInViewModel(
...
+        paymentMethodsApiResponse = sessionSetupResponse.paymentMethods
+        loadDropIn()
+    }
+
+    private fun loadDropIn() {
```
OscarSpruit, line 174: Hmm maybe `showInitialScreen` would be better

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
jreij: `onViewCreated` maybe? 🤔

OscarSpruit: Hmm that kinda implies fragment, what do you think about `onReady`?

jreij: I went with  `onCreated` eventually

OscarSpruit (approved):

PR #702, opened 2022-05-18
CardValidation: Fix expiry year validation

ozgur00:


jreij (dismissed):
I suggest adding some tests but feel free to merge 👍

jreij (approved):

PR #713, opened 2022-05-31
Initialize session service

OscarSpruit:
## Tested scenarios

**Fixed issue**:


components-core/src/main/java/com/adyen/checkout/components/model/connection/OrderPaymentMethod.kt
```diff
@@ -38,7 +38,7 @@ data class OrderPaymentMethod(
         @JvmField
         val CREATOR: Parcelable.Creator<OrderPaymentMethod> = Creator(OrderPaymentMethod::class.java)
 
-        @JvmStatic
+        @JvmField
```
jreij, line 41: Were you able to double check if other classes have the same issue? If not we can open a separate ticket for it

OscarSpruit: I opened a separate ticket for it

jreij (approved):

PR #717, opened 2022-06-08
Persist session data

OscarSpruit:
- Pass session data around and persist it in the ViewModel

Bonus:
- Updated Mockito version and added Mockito Kotlin and Mockito Inline
- Added Turbine for testing flows

Thinking about migrating to JUnit 5 as well, because it makes tests so much prettier, but that is for another time

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,7 +86,9 @@ allprojects {
     ext.test_ext_version = "1.1.3"
     ext.espresso_version = "3.4.0"
     ext.arch_core_testing_version = "2.1.0"
-    ext.mockito_version = "3.11.2"
+    ext.mockito_version = "4.6.1"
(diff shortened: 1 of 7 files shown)


ozgur00 (approved):

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -168,6 +168,18 @@ class DropInViewModel(
         }
     }
 
+    fun onDropInServiceConnected() {
+        session?.let { session ->
```
jreij, line 172: Can we add a log in case the session is null? I'm afraid we might miss out on any potential bugs if this method is called at the wrong time. A bit too late now but maybe we can do this in another PR.

PR #719, opened 2022-06-09
Simplify dependencies

OscarSpruit:
Put all libraries and versions in one file, so we have a better overview of all dependencies and makes it easier to maintain


jreij (approved):
✨

PR #721, opened 2022-06-17
Handle refused in partial payment flow correctly

OscarSpruit:
Added the log line in this one as well for you @jreij

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
@@ -14,10 +14,12 @@ import com.adyen.checkout.components.model.payments.request.OrderRequest
 import com.adyen.checkout.components.model.payments.request.PaymentMethodDetails
 import com.adyen.checkout.components.model.payments.response.BalanceResult
 import com.adyen.checkout.components.model.payments.response.OrderResponse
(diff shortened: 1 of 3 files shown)


jreij (dismissed):

drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
```diff
@@ -79,6 +81,7 @@ open class SessionDropInService : DropInService(), SessionDropInServiceInterface
                     onSuccess = { response ->
                         val action = response.action
                         val result = when {
+                            response.isRefused() -> DropInServiceResult.Error(reason = response.resultCode)
```
jreij, line 84: Does this terminate drop-in or not? I'm curious as to how we should handle this case

OscarSpruit: No, it does not. With `DropInServiceResult.Finished` it would terminate, but I don't think that's nice UX. In this case, with `Error`, we show a dialog with a generic error message

jreij: `Error` can also terminate drop-in but you should pass a `terminateDropIn = true` param to it. In all cases i agree that we shouldn't terminate it here.

jreij (approved):

PR #724, opened 2022-06-21
Sessions Merchant Take Over

ozgur00:


drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
```diff
@@ -99,11 +115,48 @@ open class SessionDropInService : DropInService(), SessionDropInServiceInterface
...
 
+    protected open fun makePaymentsCallMerchant(paymentComponentState: PaymentComponentState<*>, paymentComponentJson: JSONObject) : Boolean {
+        return false
+    }
+
+    override fun onPaymentsCallRequested(
+        paymentComponentState: PaymentComponentState<*>,
+        paymentComponentJson: JSONObject
+    ) {
+        val callWasHandled = makePaymentsCallMerchant(paymentComponentState, paymentComponentJson)
```
OscarSpruit, line 126: I see you are doing the same in all these methods, so maybe this logic can be extracted into something like this
```
private fun cantComeUpWithAGoodName(actionComponentData, actionComponentJson, merchantCall: (actionComponentData, actionComponentJson) -> Boolean, internalCall: (actionComponentData) -> Unit) {
         val callWasHandled = merchantCall(actionComponentData, actionComponentJson)
         if (!callWasHandled) {
             if (isFlowTakenOverFlow.value) {
                 throw CheckoutException("")
             } else {
                 internalCall(actionComponentData)
             }
         } else {
             if (!isFlowTakenOverFlow.value) {
                 isFlowTakenOverFlow.value = true
             }
         }
}
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -124,7 +124,10 @@ class DropInActivity :
 
             lifecycleScope.launch {
                 repeatOnLifecycle(Lifecycle.State.STARTED) {
-                    dropInService?.observeResult { handleDropInServiceResult(it) }
+                    dropInService?.observeResult {
+                        Logger.d(TAG, "handle called baby")
```
OscarSpruit, line 128: 😂

drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
```diff
@@ -30,14 +33,24 @@ open class SessionDropInService : DropInService(), SessionDropInServiceInterface
 
     private lateinit var sessionRepository: SessionRepository
 
-    override fun initialize(session: Session, clientKey: String, baseUrl: String, shouldFetchPaymentMethods: Boolean) {
+    private val _isFlowTakenOverFlow = MutableStateFlow(false)
+    private val isFlowTakenOverFlow: Flow<Boolean> = _isFlowTakenOverFlow
```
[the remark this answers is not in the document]
jreij: I'm a bit confused as to why do we need a flow here? It seems like it's only used inside this class right?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
jreij: 👶

jreij:
There's something we totally missed here. The merchant can still override all the other methods (`onPaymentsCallRequested`, `onDetailsCallRequested`, etc.) which they shouldn't do and which will definitely break our flow and cause issues 😓

drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
ozgur00: i wrote this function exactly, however all of these methods take different params. i can have a typed function then there are calls with 0 or 1 params instead of 2 so that means i can't use a single typed function. so unfortunately this is the way to go

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
ozgur00: oups 😅

drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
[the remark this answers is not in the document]
ozgur00: good point @jreij. removing it

drop-in/src/main/java/com/adyen/checkout/dropin/service/SessionDropInService.kt
jreij: What if we use lambdas?

```kotlin
    private fun cantComeUpWithAGoodNameEither(
        makeMerchantCall: () -> Boolean,
        makeInternalCall: () -> Unit,
        merchantMethodName: String
    ) {
        val callWasHandled = makeMerchantCall()
        if (!callWasHandled) {
            if (_isFlowTakenOverFlow.value) {
                throw CheckoutException("Sessions flow was already taken over in a previous call, $merchantMethodName should be implemented")
            } else {
                makeInternalCall()
            }
        } else {
            if (!_isFlowTakenOverFlow.value) {
                _isFlowTakenOverFlow.value = true
            }
        }
    }



    override fun onPaymentsCallRequested(
        paymentComponentState: PaymentComponentState<*>,
        paymentComponentJson: JSONObject
    ) {
        cantComeUpWithAGoodNameEither(
            makeMerchantCall = {
                makePaymentsCallMerchant(paymentComponentState, paymentComponentJson)
            },
            makeInternalCall = {
                makePaymentsCallInternal(paymentComponentState)
            },
            merchantMethodName = ::makePaymentsCallMerchant.name //not sure if this will return the method name
        )
    }

```

OscarSpruit (approved):

jreij (approved):

PR #726, opened 2022-06-29
Fix Redirects for Android 11

etman55:
## Summary
Fix Redirect issue on Android 11

## Tested scenarios

- Launching a redirect to custom tabs:
Works on AFTER and BEFORE android 11.
- Launching a redirect directly to a browser (maybe if custom tabs is not available on the device?):
Disabled chrome and installed firefox works on AFTER and BEFORE android 11.
- Launching a deep link redirect to an app that exists on the device:
Using generic youtube links on BEFORE android 11, like `vnd.youtube://4KnNVK-udTU` and `https://m.youtube.com/watch?v=4KnNVK-udTU` 
it opens youtube application and also google chrome, but when I tried something like this `https://youtu.be/GLNtbkv4pX0z` 
it just opened youtube app and didn't open the browser in that case.
Something to keep in mind this is the implementation from google custom tabs for android 11 which is pretty the same in our sdk implementation for redirect.
- Launching a deep link redirect to an app that exists on the device:
When disabling youtube application the links redirect to custom tabs using this link `https://youtu.be/GLNtbkv4pX0z` 
works on AFTER android 11 and BEFORE android 11.


redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -107,28 +83,79 @@ object RedirectUtil {
...
+                .setToolbarColor(ThemeUtil.getPrimaryThemeColor(context))
+                .build()
+            try {
+                CustomTabsIntent.Builder()
+                    .setShowTitle(true)
+                    .setDefaultColorSchemeParams(defaultColors)
+                    .build().launchUrl(context, uri)
+            } catch (e: ActivityNotFoundException) {
+                // in case the device doesn't support custom tabs or doesn't support google services.
+                val browserActivityIntent = Intent()
```
jreij, line 103: Same comment here, can we extract this to a separate method with its own try catch, etc. ?

etman55: This 'catch' block is just for handling if there's no custom tabs support or google play services or Huawei devices. 
The 'try' and 'catch' block for it would be just to throw an exception if there's something wrong, okay?

etman55: it will be something like this

```
private fun launchWithCustomTabs(context: Context, uri: Uri): Boolean {
        //open in custom tabs if there's no native app for the target uri
        val defaultColors = CustomTabColorSchemeParams.Builder()
            .setToolbarColor(ThemeUtil.getPrimaryThemeColor(context))
            .build()
        return try {
            CustomTabsIntent.Builder()
                .setShowTitle(true)
                .setDefaultColorSchemeParams(defaultColors)
                .build()
                .launchUrl(context, uri)
            true
        } catch (e: ActivityNotFoundException) {
            false
        }
    }

    /**
     * in case the device doesn't support custom tabs or doesn't support google services (Huawei device).
     */
    private fun launchBrowser(context: Context, uri: Uri) {
        try {
            val browserActivityIntent = Intent()
                .setAction(Intent.ACTION_VIEW)
                .addCategory(Intent.CATEGORY_BROWSABLE)
                .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
                .setData(uri)
            context.startActivity(browserActivityIntent)
        } catch (e: ActivityNotFoundException) {
            throw ComponentException("Redirect to app failed.", e)
        }
    }
```

```
   /**
     * Creates the intent that will start the redirect.
     * @param context Any context.
     * @param uri The Uri to redirect to.
     * @return And intent that targets either another app or a Web page.
     */
    @JvmStatic
    fun launchUri(context: Context, uri: Uri) {
        val launched = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
            launchNativeApi30(context, uri)
        else launchNativeBeforeApi30(context, uri)

        if (!launched) {
            val customTabsLaunch = launchWithCustomTabs(context, uri)
            if (!customTabsLaunch) {
                launchBrowser(context, uri)
            }
        }
    }
```

jreij: 👌

example-app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,8 @@
     package="com.adyen.checkout.example">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <!--  add this permission when tried on api level 25 emulator for redirect app crashed for not granting this permission so added it  -->
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
```
jreij, line 8: I'm a bit skeptical about this 🤔 Did you find any resource that says `WAKE_LOCK` is required for redirects?

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -107,28 +84,98 @@ object RedirectUtil {
...
-    private fun createCustomTabIntent(context: Context, uri: Uri): Intent {
+    private fun launchNativeBeforeApi30(context: Context, uri: Uri): Boolean {
+        val pm = context.packageManager
+
+        // Get all Apps that resolve a generic url
+        val browserActivityIntent = Intent()
+            .setAction(Intent.ACTION_VIEW)
+            .addCategory(Intent.CATEGORY_BROWSABLE)
+            .setData(Uri.fromParts("http", "", null))
+        val genericPackageNameList: MutableList<String> =
```
jreij, line 108: Can we just use `val genericResolvedList = pm.queryIntentActivities(browserActivityIntent, 0).map { it.resolvePackageName }` here?

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -107,28 +84,98 @@ object RedirectUtil {
...
+            pm.queryIntentActivities(browserActivityIntent, 0).map { it.resolvePackageName } as MutableList<String>
+        val genericResolvedList: Set<String> = genericPackageNameList.toSet()
+
+        // Get all apps that resolve the specific Url
+        val specializedActivityIntent = Intent(Intent.ACTION_VIEW, uri)
+            .addCategory(Intent.CATEGORY_BROWSABLE)
+        val specializedPackageNameList: MutableList<String> =
+            pm.queryIntentActivities(specializedActivityIntent, 0).map { it.resolvePackageName } as MutableList<String>
+        val resolvedSpecializedList: MutableSet<String> = specializedPackageNameList.toMutableSet()
```
jreij, line 117: Can we use `val resolvedSpecializedList = pm.queryIntentActivities(specializedActivityIntent, 0).map { it.resolvePackageName }.toMutableList()` ?

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -107,28 +84,98 @@ object RedirectUtil {
...
+            .setData(Uri.fromParts("http", "", null))
+        val genericPackageNameList: MutableList<String> =
+            pm.queryIntentActivities(browserActivityIntent, 0).map { it.resolvePackageName } as MutableList<String>
+        val genericResolvedList: Set<String> = genericPackageNameList.toSet()
+
+        // Get all apps that resolve the specific Url
+        val specializedActivityIntent = Intent(Intent.ACTION_VIEW, uri)
+            .addCategory(Intent.CATEGORY_BROWSABLE)
+        val specializedPackageNameList: MutableList<String> =
+            pm.queryIntentActivities(specializedActivityIntent, 0).map { it.resolvePackageName } as MutableList<String>
```
jreij, line 116: `queryIntentActivities` shows me a warning:
>Consider adding a <queries> declaration to your manifest when calling this method; see https://g.co/dev/packagevisibility for details

Can you investigate this? 🙏

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -107,28 +84,98 @@ object RedirectUtil {
...
-    fun createRedirectIntent(context: Context, uri: Uri): Intent {
-        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
-            createCustomTabIntent(context, uri)
-        } else {
-            if (determineResolveResult(context, uri).type == ResolveResult.Type.APPLICATION) {
-                Intent(Intent.ACTION_VIEW, uri)
-            } else {
-                createCustomTabIntent(context, uri)
+    fun launchUri(context: Context, uri: Uri) {
+        val launched = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R)
```
jreij, line 88: Sorry for being nitpicky but I have an idea. If we extract another `launchNative` method and we also make `launchBrowser` return a boolean, we can do this:

```kotlin
    fun launchUri(context: Context, uri: Uri) {
        if (launchNative(context, uri)) return
        if (launchWithCustomTabs(context, uri)) return
        if (launchBrowser(context, uri)) return
        throw ComponentException("Redirect to app failed.")
    }

    private fun launchNative(context: Context, uri: Uri): Boolean {
        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) launchNativeApi30(context, uri)
        else launchNativeBeforeApi30(context, uri)
    }
```

What do you think?

etman55: you're so picky 😂

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
etman55: sure, but will do it as set for performance.

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
etman55: hmm, I guess the package visibility limitation is for android 11 and above and this logic is for before android 11 so we're good to go.

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
etman55: whoops sorry I got a build error for that 
```
ReturnCount - [launchUriRedirect] at /Users/atef/android-projects/adyen-android/redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt:80:9
```
I'll just do it like this
```
fun launchUriRedirect(context: Context, uri: Uri) {
        val launchNative = launchNative(context, uri)
        if (!launchNative) {
            val launchCustomWithTabs = launchWithCustomTabs(context, uri)
            if (!launchCustomWithTabs)
                launchBrowser(context, uri)
        }
    }
```
```
 /**
     * in case the device doesn't support custom tabs or doesn't support google services (Huawei device).
     */
    private fun launchBrowser(context: Context, uri: Uri) {
        try {
            val browserActivityIntent = Intent()
                .setAction(Intent.ACTION_VIEW)
                .addCategory(Intent.CATEGORY_BROWSABLE)
                .addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
                .setData(uri)
            context.startActivity(browserActivityIntent)

        } catch (e: ActivityNotFoundException) {
            throw ComponentException("Redirect to app failed.")
        }
    }
```
let me know what do you think @jreij ?

jreij: There are 2 points here:
- I'm trying to avoid throwing an exception in `launchBrowser`. While it works like this right now because `launchBrowser` is the last method we call, it might cause issues later if we decide to switch the order and call something else after `launchBrowser`. Which is why I suggested making it return a boolean like the other methods, and throwing the exception in the parent method.
- The `if ... return` syntax i suggested was just to make the code more readable and avoid having nested `if`s and boolean `val`s. Feel free to add `@Suppress("ReturnCount")` above `launchUriRedirect` or keep it as you have it now 🙂

etman55: good points

example-app/src/main/AndroidManifest.xml
jreij: Does this happen with the old method as well?

redirect/src/main/java/com/adyen/checkout/redirect/RedirectUtil.kt
```diff
@@ -100,35 +77,99 @@ object RedirectUtil {
...
+            return false
+        }
+
+        // We found native handlers. Launch the Intent.
+        specializedActivityIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
+        context.startActivity(specializedActivityIntent)
+        return true
+    }
+
+    private fun launchNativeApi30(context: Context, uri: Uri?): Boolean {
```
jreij, line 127: Minor: you can add `@RequiresApi(Build.VERSION_CODES.R)` here to remove the warning.

example-app/src/main/AndroidManifest.xml
etman55: yes it happens when you just click on `start` button in dropIn

jreij (approved):
🚀 ✨

PR #729, opened 2022-07-01
Add BlikDelegate

jreij:


blik/src/main/java/com/adyen/checkout/blik/BlikComponent.kt
```diff
@@ -8,68 +8,74 @@
...
+        storedBlikDelegate: StoredBlikDelegate,
         configuration: BlikConfiguration
     ) : super(savedStateHandle, storedPaymentDelegate, configuration) {
-        // TODO: 09/12/2020 move this logic to base component, maybe create the InputData from the delegate?
-        inputDataChanged(BlikInputData())
+        this.blikDelegate = storedBlikDelegate
+        init()
+    }
+
+    private fun init() {
```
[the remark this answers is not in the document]
jreij: Yep exactly, the `init` block gets called before the `constructor`s so it crashes because the `private val blikDelegate` isn't initialised yet.

OscarSpruit (dismissed):

blik/src/main/java/com/adyen/checkout/blik/NewBlikDelegate.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
+import com.adyen.checkout.components.model.payments.request.BlikPaymentMethod
+import com.adyen.checkout.components.model.payments.request.PaymentComponentData
+import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class NewBlikDelegate(val paymentMethod: PaymentMethod) : BlikDelegate {
```
OscarSpruit, line 29: Both this and `StoredBlikDelegate` could be `internal`

blik/src/main/java/com/adyen/checkout/blik/NewBlikDelegate.kt
```diff
@@ -0,0 +1,78 @@
...
+        outputDataChanged(outputData)
+        createComponentState(outputData)
+    }
+
+    private fun outputDataChanged(outputData: BlikOutputData) {
+        _outputDataFlow.tryEmit(outputData)
+    }
+
+    override fun createComponentState(outputData: BlikOutputData) {
+        val paymentMethod = BlikPaymentMethod().apply {
```
OscarSpruit, line 53: Constructor should be used instead of `apply`

blik/src/main/java/com/adyen/checkout/blik/NewBlikDelegate.kt
```diff
@@ -0,0 +1,78 @@
...
+            type = BlikPaymentMethod.PAYMENT_METHOD_TYPE
+            blikCode = outputData.blikCodeField.value
+        }
+
+        val paymentComponentData = PaymentComponentData<BlikPaymentMethod>().apply {
```
OscarSpruit, line 58: Constructor should be used instead of `apply`

blik/src/main/java/com/adyen/checkout/blik/StoredBlikDelegate.kt
```diff
@@ -0,0 +1,73 @@
...
+    override fun onInputDataChanged(inputData: BlikInputData) {
+        Logger.e(TAG, "onInputDataChanged should not be called in StoredBlikDelegate")
+    }
+
+    private fun outputDataChanged(outputData: BlikOutputData) {
+        _outputDataFlow.tryEmit(outputData)
+    }
+
+    override fun createComponentState(outputData: BlikOutputData) {
+        val paymentMethod = BlikPaymentMethod().apply {
```
OscarSpruit, line 48: Constructor should be used instead of `apply`

blik/src/main/java/com/adyen/checkout/blik/StoredBlikDelegate.kt
```diff
@@ -0,0 +1,73 @@
...
+            type = BlikPaymentMethod.PAYMENT_METHOD_TYPE
+            storedPaymentMethodId = storedPaymentMethod.id
+        }
+
+        val paymentComponentData = PaymentComponentData<BlikPaymentMethod>().apply {
```
OscarSpruit, line 53: Constructor should be used instead of `apply`

blik/src/main/java/com/adyen/checkout/blik/NewBlikDelegate.kt
OscarSpruit, line 29: Also I prefer a name like `DefaultBlikDelegate`, so we don't have to rename this class when we remove the old delegate. WDYT?

jreij: Fixed

OscarSpruit (approved):

PR #732, opened 2022-07-07
Add India and APSE environments

OscarSpruit:


components-core/src/main/java/com/adyen/checkout/components/util/ValidationUtils.kt
```diff
@@ -56,9 +64,7 @@ object ValidationUtils {
      */
     fun doesClientKeyMatchEnvironment(clientKey: String, environment: Environment): Boolean {
         val isTestEnvironment = environment == Environment.TEST
-        val isLiveEnvironment = environment == Environment.EUROPE ||
-            environment == Environment.AUSTRALIA ||
-            environment == Environment.UNITED_STATES
+        val isLiveEnvironment = LIVE_ENVIRONMENTS.contains(environment)
```
jreij, line 67: Can you quickly check whether passing something like `environment = Environment(URL("https://checkoutshopper-live-us.adyen.com/checkoutshopper/")` would return `isLiveEnvironment = true`? The idea is that merchants should be able to create the environment manually, and if the URL is the same as one of the preset environments it should match it.

We already implement `equals` in `Environment` for that but I'm wondering if changing from `==` to `contains` will break that.

jreij (approved):

PR #736, opened 2022-07-14
Fix BACS Amount 

ozgur00:


bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitInputView.kt
```diff
@@ -232,12 +228,14 @@ class BacsDirectDebitInputView @JvmOverloads constructor(context: Context, attrs
...
 
-    private fun initButton() {
+    private fun setAmountConsentSwitchText() {
         if (!component.configuration.amount.isEmpty) {
             val formattedAmount =
                 CurrencyUtils.formatAmount(component.configuration.amount, component.configuration.shopperLocale)
             binding.switchConsentAmount.text =
                 localizedContext.getString(R.string.bacs_consent_amount_specified, formattedAmount)
+        } else {
+            binding.switchConsentAmount.text = localizedContext.getString(R.string.bacs_consent_amount)
```
jreij, line 238: Can't we use the same previous code here instead?
```kotlin
binding.switchConsentAmount.setLocalizedTextFromStyle(
    R.style.AdyenCheckout_Bacs_Switch_Amount,
    localizedContext
)
```

OscarSpruit:
I think removing the style causes the color of the switches to not match up anymore. Also the switches are not aligned, text size is different and I feel like there should be more white space between the texts 
image

ozgur00:
> I think removing the style causes the color of the switches to not match up anymore. Also the switches are not aligned, text size is different and I feel like there should be more white space between the texts image

done

OscarSpruit (approved):

PR #738, opened 2022-07-18
Add gift card delegate

OscarSpruit:
- Move logic from `GiftCardComponent` into a delegate
- Convert `CardEncrypter` from an object to a normal class
- Add a test implementation of `CardEncrypter`

### DIFF
--- a/bcmc/src/main/java/com/adyen/checkout/bcmc/DefaultBcmcDelegate.kt
+++ b/bcmc/src/main/java/com/adyen/checkout/bcmc/DefaultBcmcDelegate.kt
@@ -29,6 +29,7 @@ import com.adyen.checkout.cse.EncryptedCard
 import com.adyen.checkout.cse.UnencryptedCard
 import com.adyen.checkout.cse.exception.EncryptionException
 import com.adyen.threeds2.ThreeDS2Service
+import kotlinx.coroutines.channels.BufferOverflow
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -46,7 +47,7 @@ internal class DefaultBcmcDelegate(
     private val _componentStateFlow = MutableStateFlow<PaymentComponentState<CardPaymentMethod>?>(null)
     override val componentStateFlow: Flow<PaymentComponentState<CardPaymentMethod>?> = _componentStateFlow
 
-    private val _exceptionFlow = MutableSharedFlow<CheckoutException>()
+    private val _exceptionFlow = MutableSharedFlow<CheckoutException>(0, 1, BufferOverflow.DROP_OLDEST)
     override val exceptionFlow: Flow<CheckoutException> = _exceptionFlow
 
     private var publicKey: String? = null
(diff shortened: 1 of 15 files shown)


ozgur00 (approved):

bcmc/src/main/java/com/adyen/checkout/bcmc/DefaultBcmcDelegate.kt
```diff
@@ -38,6 +39,7 @@ internal class DefaultBcmcDelegate(
     private val publicKeyRepository: PublicKeyRepository,
     private val configuration: BcmcConfiguration,
     private val cardValidationMapper: CardValidationMapper,
+    private val cardEncrypter: DefaultCardEncrypter,
```
jreij, line 42: Should this be `CardEncrypter` or `DefaultCardEncrypter`?

OscarSpruit: Good catch!

bcmc/src/main/java/com/adyen/checkout/bcmc/DefaultBcmcDelegate.kt
```diff
@@ -46,7 +48,7 @@ internal class DefaultBcmcDelegate(
     private val _componentStateFlow = MutableStateFlow<PaymentComponentState<CardPaymentMethod>?>(null)
     override val componentStateFlow: Flow<PaymentComponentState<CardPaymentMethod>?> = _componentStateFlow
 
-    private val _exceptionFlow = MutableSharedFlow<CheckoutException>()
+    private val _exceptionFlow = MutableSharedFlow<CheckoutException>(0, 1, BufferOverflow.DROP_OLDEST)
```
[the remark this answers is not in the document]
OscarSpruit: Yes I think so. I already changed it for all `MutableSharedFlow`s I could find

giftcard/src/main/java/com/adyen/checkout/giftcard/GiftCardComponent.kt
```diff
@@ -37,98 +27,41 @@ private const val LAST_FOUR_LENGTH = 4
...
-            ).fold(
-                onSuccess = { key ->
-                    publicKey = key
-                    createComponentState()
-                },
-                onFailure = { e ->
-                    notifyException(ComponentException("Unable to fetch publicKey.", e))
-                }
-            )
+            giftCardDelegate.fetchPublicKey()
```
jreij, line 56: Is there a way to make the delegate `fetchPublicKey()` by itself on init? And If so, is there any advantage? 🤔

OscarSpruit: Can't think of a good way to do this 🤔 We would have the pass a `CoroutineScope` somehow. We can't pass `viewModelScope` in the constructor, because the `ViewModel` is created after the delegate. So we would have to pass the scope after both are created, which is less elegant than this imo

jreij: 👍

jreij (approved):

PR #742, opened 2022-07-21
Release 4.7.1

ozgur00:


jreij:
Can we also make another PR to fix the versions on develop?

ozgur00:
> Can we also make another PR to fix the versions on develop?

I will add the fixes on develop as well, there I'll update the versions

jreij (approved):

PR #745, opened 2022-07-25
Remove TextUtils

jreij:
`TextUtils` was used in the java classes but it's not necessary anymore with Kotlin and it creates problems with testing.

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/ui/CardNumberInput.kt
+++ b/card/src/main/java/com/adyen/checkout/card/ui/CardNumberInput.kt
@@ -11,7 +11,6 @@ import android.content.Context
 import android.os.Build
 import android.text.Editable
 import android.text.InputType
-import android.text.TextUtils
 import android.text.method.DigitsKeyListener
 import android.util.AttributeSet
 import com.adyen.checkout.card.util.CardValidationUtils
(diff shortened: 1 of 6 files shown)


card/src/main/java/com/adyen/checkout/card/ui/CardNumberInput.kt
```diff
@@ -68,7 +67,7 @@ open class CardNumberInput @JvmOverloads constructor(
     private fun formatProcessedString(processedValue: String): String {
         val result =
             splitStringWithMask(processedValue, *if (isAmexCard) AMEX_CARD_NUMBER_MASK else DEFAULT_CARD_NUMBER_MASK)
-        return TextUtils.join(DIGIT_SEPARATOR.toString() + "", result).trim { it <= ' ' }
```
ozgur00, line 71: Not related to this change but I think it would be nice to extract this formatting logic to somewhere else and unit test it.

ozgur00 (approved):

card/src/main/java/com/adyen/checkout/card/ui/CardNumberInput.kt
jreij: Agreed, can you create a ticket for it? 🙏

PR #753, opened 2022-08-11
Refactor StatusRepository

OscarSpruit:
Refactor `StatusRepository` to use flows and remove the complex polling mechanism.


drop-in/src/main/java/com/adyen/checkout/dropin/ActionHandler.kt
```diff
@@ -83,10 +83,10 @@ class ActionHandler(
         activity: FragmentActivity,
         provider: ActionComponentProvider<out BaseActionComponent<out Configuration>, out Configuration>
     ) {
-        getActionComponentFor(activity, provider, dropInConfiguration).apply {
+        getActionComponentFor(activity, activity.application, provider, dropInConfiguration).apply {
```
etman55, line 86: I guess this has nothing to do with StatusRepository refactor, so it's better to be done in a separate PR.

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -444,51 +445,45 @@ internal fun getActionProviderFor(
...
- * Provides a [BaseActionComponent] from an [ActionComponentProvider].
- *
- * @param activity The Action which the ActionComponent lifecycle will be bound to.
- * @param provider The component provider.
- * @throws CheckoutException In case a component cannot be created.
- */
 @Suppress("ComplexMethod", "LongMethod")
-internal fun getActionComponentFor(
-    activity: FragmentActivity,
+internal fun <T> getActionComponentFor(
```
etman55, line 449: I guess this has nothing to do with StatusRepository refactor, so it's better to be done in a separate PR.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -147,13 +147,21 @@ class ActionComponentDialogFragment : DropInBottomSheetDialogFragment(), Observe
...
+
         if (!provider.requiresView(action)) {
             throw ComponentException(
                 "Action is not viewable - action: ${action.type} - " +
                     "paymentMethod: ${action.paymentMethodType}"
             )
         }
-        val component = getActionComponentFor(requireActivity(), provider, dropInViewModel.dropInConfiguration)
+
+        val component = getActionComponentFor(
```
etman55, line 158: I guess this has nothing to do with StatusRepository refactor, so it's better to be done in a separate PR.

drop-in/src/main/java/com/adyen/checkout/dropin/ActionHandler.kt
OscarSpruit: Fair enough, will keep that in mind for next time. It does however solve a big issue with the `StatusRepository`. The component wasn't destroyed after the fragment was, causing the polling to go on as long as the `DropInActivity` is alive

components-core/src/main/java/com/adyen/checkout/components/status/StatusRepository.kt
```diff
@@ -1,142 +1,99 @@
...
-            elapsedTime <= MAX_POLLING_DURATION_MILLIS -> pollingDelay = POLLING_DELAY_SLOW
-            else -> _errorLiveData.setValue(ComponentException("Status requesting timed out with no result"))
-        }
+    fun refreshStatus(paymentData: String) {
+        refreshFlow.tryEmit(paymentData)
     }
 
     companion object {
-        val TAG = getTag()
         val MAX_POLLING_DURATION_MILLIS = TimeUnit.MINUTES.toMillis(15)
```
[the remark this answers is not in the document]
OscarSpruit: Also noticed that, but now that I know how the payment method works it makes sense. For QR codes for example, they have to finish the flow in another app. 15 minutes gives people plenty of time to complete the stuff in that app and come back to us

drop-in/src/main/java/com/adyen/checkout/dropin/ActionHandler.kt
etman55: good catch, I didn't notice that.

components-core/src/main/java/com/adyen/checkout/components/status/StatusRepository.kt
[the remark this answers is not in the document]
etman55: Okay, but what happen when the user comes back to us does it start counting again or continue counting from last time?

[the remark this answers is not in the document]
OscarSpruit: It just continues and then when the time is over an error message is displayed

etman55 (approved):

jreij (approved):
LGTM 👍
As we discussed it would be nice if we can find a way to get the timer status from the repository itself but it's not so necessary

PR #755, opened 2022-08-12
Refactor card delegate

jreij:


card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -88,86 +208,90 @@ class StoredCardDelegate(
...
+                encryptedSecurityCode = encryptedCard.encryptedSecurityCode
+            }
 
-    override fun isAddressRequired(addressFormUIState: AddressFormUIState): Boolean {
-        return false
-    }
+            try {
+                threeDS2SdkVersion = ThreeDS2Service.INSTANCE.sdkVersion
+            } catch (e: ClassNotFoundException) {
+                Logger.e(TAG, "threeDS2SdkVersion not set because 3DS2 SDK is not present in project.")
```
OscarSpruit, line 260: We're swallowing the error here - wouldn't it be better to crash?

card/src/main/java/com/adyen/checkout/card/repository/DefaultAddressRepository.kt
```diff
@@ -0,0 +1,149 @@
...
+internal class DefaultAddressRepository : AddressRepository {
+
+    private val _statesFlow: MutableSharedFlow<List<AddressItem>> = MutableSharedFlow(0, 1, BufferOverflow.DROP_OLDEST)
+    override val statesFlow: Flow<List<AddressItem>> = _statesFlow
+
+    private val _countriesFlow: MutableSharedFlow<List<AddressItem>> =
+        MutableSharedFlow(0, 1, BufferOverflow.DROP_OLDEST)
+    override val countriesFlow: Flow<List<AddressItem>> = _countriesFlow
+
+    private val cache: LruCache<String, List<AddressItem>> = LruCache<String, List<AddressItem>>(CACHE_ENTRY_SIZE)
```
OscarSpruit, line 37: Isn't this cache able to basically able to cache every possible country? Because that would defeat the purpose of `LruCache`. Also, since there are not a lot of entries in this cache I think using a simple `HashMap` is more efficient

card/src/test/java/com/adyen/checkout/card/DefaultCardDelegateTest.kt
```diff
@@ -0,0 +1,804 @@
...
+    }
+
+    @Nested
+    @DisplayName("when input data changes and")
+    inner class InputDataChangedTest {
+        @Test
+        fun `address configuration is none, then countries and states should not be fetched`() = runTest {
+            addressRepository.countriesFlow.test {
+                addressRepository.statesFlow.test {
+                    delegate.initialize(CoroutineScope(UnconfinedTestDispatcher()))
```
OscarSpruit, line 95: It would be better to pass the `TestScope` here to make sure all coroutines are correctly canceled when the test finishes. So you can do `delegate.initialize(this@runTest)`

card/src/test/java/com/adyen/checkout/card/DefaultCardDelegateTest.kt
```diff
@@ -0,0 +1,804 @@
...
+        }
+    }
+
+    @Nested
+    @DisplayName("when input data changes and")
+    inner class InputDataChangedTest {
+        @Test
+        fun `address configuration is none, then countries and states should not be fetched`() = runTest {
+            addressRepository.countriesFlow.test {
+                addressRepository.statesFlow.test {
```
OscarSpruit, line 94: Instead of nesting this I think it's nicer to do it like they do [here](https://github.com/cashapp/turbine#multiple-turbines)

card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -0,0 +1,597 @@
...
+                outputData,
+                cardNumber,
+                firstCardType,
+                binValue
+            )
+        )
+    }
+
+    // Validation
+    private fun validateCardNumber(
```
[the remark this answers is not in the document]
jreij: Will do in another PR 👍

card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -0,0 +1,597 @@
...
+
+        val cardNumber = outputData.cardNumberState.value
+
+        val firstCardType = getDetectedCardType(outputData.detectedCardTypes)?.cardType
+
+        val binValue = cardNumber.take(BIN_VALUE_LENGTH)
+
+        val publicKey = publicKey
+
+        // If data is not valid we just return empty object, encryption would fail and we don't pass unencrypted data.
```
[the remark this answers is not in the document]
jreij: We do already in `fetchPublicKey()`

card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
jreij: Adding a comment 👍

card/src/test/java/com/adyen/checkout/card/DefaultCardDelegateTest.kt
jreij: This doesn't work :/

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
```diff
@@ -11,534 +11,78 @@ package com.adyen.checkout.card
...
-        cardPaymentMethod.type = CardPaymentMethod.PAYMENT_METHOD_TYPE
-
-        if (!isStoredPaymentMethod()) {
-            cardPaymentMethod.encryptedCardNumber = encryptedCard.encryptedCardNumber
-            cardPaymentMethod.encryptedExpiryMonth = encryptedCard.encryptedExpiryMonth
-            cardPaymentMethod.encryptedExpiryYear = encryptedCard.encryptedExpiryYear
-        } else {
-            cardPaymentMethod.storedPaymentMethodId = (paymentMethodDelegate as StoredCardDelegate).getId()
-        }
+    // FIXME remove
```
OscarSpruit, line 65: When should this be fixed?

card/src/main/java/com/adyen/checkout/card/CardDelegate.kt
```diff
@@ -1,86 +1,42 @@
...
 
-    abstract fun getSupportedCardTypes(): List<CardType>
-
-    suspend fun fetchPublicKey(): Result<String> {
-        return publicKeyRepository.fetchPublicKey(
-            environment = cardConfiguration.environment,
-            clientKey = cardConfiguration.clientKey
-        )
-    }
+    fun clear()
```
OscarSpruit, line 41: Nit: I would stay consistent with the naming as in `ViewModel`s, so this would be `onCleared`

card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -0,0 +1,597 @@
...
+    private val outputData
+        get() = _outputDataFlow.value
+
+    private val _componentStateFlow = MutableStateFlow<CardComponentState?>(null)
+    override val componentStateFlow: Flow<CardComponentState?> = _componentStateFlow
+
+    private val _exceptionFlow = MutableSharedFlow<CheckoutException>(0, 1, BufferOverflow.DROP_OLDEST)
+    override val exceptionFlow: Flow<CheckoutException> = _exceptionFlow
+
+    private var coroutineScope: CoroutineScope? = null
```
OscarSpruit, line 81: I don't like all the null checks for this field in the methods, so maybe we can do something like this:
```
private var _coroutineScope: CoroutineScope? = null
private val coroutineScope: CoroutineScope get() = requireNotNull(_coroutineScope)
```
This should never go wrong and if it does something is implemented incorrectly

card/src/main/java/com/adyen/checkout/card/CardComponent.kt
jreij: I can pick it up next

OscarSpruit (approved):

PR #758, opened 2022-08-16
QRCodeDelegate: Add QRCodeDelegate

OscarSpruit:
- Move logic inside `QRCodeComponent` to `DefaultQRCodeDelegate`
- Add AndroidX Lifecycle LiveData dependency, so it's easy to convert flows to LiveData
- Add test implementation of `StatusRepository`

### DIFF
--- a/qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
+++ b/qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
@@ -8,6 +8,7 @@
 
 package com.adyen.checkout.qrcode
 
+import androidx.annotation.VisibleForTesting
 import com.adyen.checkout.components.model.payments.response.Action
 import com.adyen.checkout.components.model.payments.response.QrCodeAction
 import com.adyen.checkout.components.status.DefaultStatusRepository
@@ -64,7 +65,8 @@ internal class DefaultQRCodeDelegate(
         ) { millisUntilFinished -> onTimerTick(millisUntilFinished) }
     }
 
-    private fun onTimerTick(millisUntilFinished: Long) {
+    @VisibleForTesting
+    internal fun onTimerTick(millisUntilFinished: Long) {
         val progressPercentage =
             (HUNDRED * millisUntilFinished / DefaultStatusRepository.MAX_POLLING_DURATION_MILLIS).toInt()
         _timerFlow.tryEmit(TimerData(millisUntilFinished, progressPercentage))
@@ -148,7 +150,8 @@ internal class DefaultQRCodeDelegate(
 
     companion object {
         private val TAG = LogUtil.getTag()
-        private const val PAYLOAD_DETAILS_KEY = "payload"
+        @VisibleForTesting
+        internal const val PAYLOAD_DETAILS_KEY = "payload"
         private val STATUS_POLLING_INTERVAL_MILLIS = TimeUnit.SECONDS.toMillis(1L)
         private const val HUNDRED = 100
     }
--- a/qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
+++ b/qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
@@ -8,9 +8,9 @@
 
 package com.adyen.checkout.qrcode
 
-import android.os.CountDownTimer
 import com.adyen.checkout.components.model.payments.response.Action
 import com.adyen.checkout.components.model.payments.response.QrCodeAction
+import com.adyen.checkout.components.status.DefaultStatusRepository
 import com.adyen.checkout.components.status.StatusRepository
 import com.adyen.checkout.components.status.api.StatusResponseUtils
 import com.adyen.checkout.components.status.model.StatusResponse
@@ -33,6 +33,7 @@ import java.util.concurrent.TimeUnit
 @Suppress("TooManyFunctions")
 internal class DefaultQRCodeDelegate(
     private val statusRepository: StatusRepository,
+    private val statusCountDownTimer: QRCodeCountDownTimer,
 ) : QRCodeDelegate {
 
     private val _outputDataFlow = MutableStateFlow<QRCodeOutputData?>(null)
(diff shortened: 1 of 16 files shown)


qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -0,0 +1,158 @@
...
+        statusCountDownTimer.attach(
+            DefaultStatusRepository.MAX_POLLING_DURATION_MILLIS,
+            STATUS_POLLING_INTERVAL_MILLIS
+        ) { millisUntilFinished -> onTimerTick(millisUntilFinished) }
+    }
+
+    @VisibleForTesting
+    internal fun onTimerTick(millisUntilFinished: Long) {
+        val progressPercentage =
+            (HUNDRED * millisUntilFinished / DefaultStatusRepository.MAX_POLLING_DURATION_MILLIS).toInt()
```
jreij, line 71: Should these constants be in the `StatusRepository` instead of the `DefaultStatusRepository`

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    override val outputData: QRCodeOutputData? get() = _outputDataFlow.value
+
+    private val _exceptionFlow = MutableSharedFlow<CheckoutException>(0, 1, BufferOverflow.DROP_OLDEST)
+    override val exceptionFlow: Flow<CheckoutException> = _exceptionFlow
+
+    private val _detailsFlow = MutableStateFlow<JSONObject?>(null)
+    override val detailsFlow: Flow<JSONObject?> = _detailsFlow
+
+    private val _timerFlow = MutableStateFlow(TimerData(0, 0))
```
jreij, line 51: Nice 😄

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt
```diff
@@ -84,36 +60,24 @@ class QRCodeComponent(
...
-                    onPollingSuccessful(response)
-                }
-            },
-            onFailure = {
-                Logger.e(TAG, "Error while polling status", it)
-                notifyException(ComponentException("Error while polling status", it))
-            }
-        )
+    /**
+     * Call this method when receiving the return URL from the 3DS redirect with the result data.
```
jreij, line 69: Oops 😅 This should not have the `3DS` part, you can just remove this word.

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -0,0 +1,138 @@
...
+) {
+
+    private lateinit var statusRepository: TestStatusRepository
+    private lateinit var delegate: DefaultQRCodeDelegate
+
+    @BeforeEach
+    fun beforeEach() {
+        statusRepository = TestStatusRepository()
+        delegate = DefaultQRCodeDelegate(statusRepository, countDownTimer)
+        Logger.setLogcatLevel(Logger.NONE)
```
jreij, line 45: Did you need that for the Android method not mocked thing? I see that we used a different was to solve it in different modules:
```groovy
    testOptions {
        unitTests.returnDefaultValues = true
    }
```

But your solution seems interesting 🤔 I wonder which is better

OscarSpruit: lol I didn't even think about that. The best is to actually have an interface for the logger, so you can have different implementations. But that to much work for this PR

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -0,0 +1,138 @@
...
+            Result.success(StatusResponse(resultCode = "finished")),
+        )
+        delegate.initialize(CoroutineScope(UnconfinedTestDispatcher()))
+
+        delegate.outputDataFlow.test {
+            delegate.handleAction(QrCodeAction(paymentMethodType = "test"), "paymentData")
+
+            skipItems(1)
+
+            awaitItem()!!.apply {
```
jreij, line 76: Nit: to me `apply` implies that your changing something in the object you called it on (not sure if it's a misconception from my side)
Which is why I prefer `with(requireNotNull(awaitItem()))` or `requireNotNull(awaitItem())?.run`

OscarSpruit: agreed 👍

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
jreij: Agreed, but `DefaultQRCodeDelegate` doesn't know that we're injecting `DefaultStatusRepository` inside it and not another implementation of `StatusRepository`.

What if we add a method `getMaxPollingDurationMillis` to `StatusRepository` then? (this is all very minor btw, just discussing it)

OscarSpruit: That's a good idea actually!

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val statusCountDownTimer: QRCodeCountDownTimer,
+) : QRCodeDelegate {
+
+    private val _outputDataFlow = MutableStateFlow<QRCodeOutputData?>(null)
+    override val outputDataFlow: Flow<QRCodeOutputData?> = _outputDataFlow
+
+    override val outputData: QRCodeOutputData? get() = _outputDataFlow.value
+
+    private val _exceptionFlow = MutableSharedFlow<CheckoutException>(0, 1, BufferOverflow.DROP_OLDEST)
+    override val exceptionFlow: Flow<CheckoutException> = _exceptionFlow
```
jreij, line 46: This flow is not observed in the component right?

OscarSpruit: Very sharp! Apparently, I also missed the details flow 👀

jreij: Minor detail 😂

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeComponent.kt
```diff
@@ -84,36 +71,24 @@ class QRCodeComponent(
             redirectDelegate.makeRedirect(activity, action.url)
```
jreij, line 71: I am rethinking this part here, the logic is that the QR code component might sometime launch a redirect instead of the status polling. In this instance it's the component that decides which action to make so it can be done here.
However when we go to the 3DS2 component a redirect can be done mid flow and surely the 3DS2 delegate is the one that decides if and when a redirect should happen. Maybe we'll figure this out when we work on the 3DS2 delegate but just leaving the comment here in case we want to come back later and change the logic here as well.

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeDelegate.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    val timerFlow: Flow<TimerData>
+
+    fun initialize(coroutineScope: CoroutineScope)
+
+    fun handleAction(action: QrCodeAction, paymentData: String)
+
+    fun refreshStatus(paymentData: String)
+
+    fun getCodeString(): String?
```
jreij, line 35: Can we make this part of output data and remove the direct method call from the view to component?

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.mockito.Mock
+import org.mockito.junit.jupiter.MockitoExtension
+import java.io.IOException
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(MockitoExtension::class)
+internal class DefaultQRCodeDelegateTest(
```
jreij, line 34: Can we include `qrCodeData` in the tests?

jreij (approved):

PR #760, opened 2022-08-17
add test cases to DefaultIssuerListDelegate

etman55:
## Summary
add test cases to DefaultIssuerListDelegate

### DIFF
--- a/issuer-list/src/test/java/com/adyen/checkout/issuerlist/DefaultIssuerListDelegateTest.kt
+++ b/issuer-list/src/test/java/com/adyen/checkout/issuerlist/DefaultIssuerListDelegateTest.kt
@@ -1,0 +1,115 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 17/8/2022.
+ */
+
+package com.adyen.checkout.issuerlist
+
+import app.cash.turbine.test
+import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
+import com.adyen.checkout.issuerlist.utils.TestIssuerPaymentMethod
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.mockito.junit.jupiter.MockitoExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(MockitoExtension::class)
+class DefaultIssuerListDelegateTest {
+
+    private val delegate = DefaultIssuerListDelegate(
+        paymentMethod = PaymentMethod(),
+        typedPaymentMethodFactory = { TestIssuerPaymentMethod() }
+    )
+
+    @Nested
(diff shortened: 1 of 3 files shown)


issuer-list/src/test/java/com/adyen/checkout/issuerlist/DefaultIssuerListDelegateTest.kt
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.Assert
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.mockito.junit.jupiter.MockitoExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(MockitoExtension::class)
+class DefaultIssuerListDelegateTest {
```
OscarSpruit, line 25: Test classes should be `internal` to avoid name clashing with other modules

issuer-list/src/test/java/com/adyen/checkout/issuerlist/DefaultIssuerListDelegateTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    @Nested
+    @DisplayName("when input data changes and")
+    inner class InputDataChangedTest {
+
+        @Test
+        fun `selectedIssuer is null, then output should be null`() = runTest {
+            delegate.outputDataFlow.test {
+                delegate.onInputDataChanged(IssuerListInputData(null))
+                val outputData = expectMostRecentItem()
+                Assert.assertNull(outputData?.selectedIssuer)
```
OscarSpruit, line 41: Nit: importing `assertNull` directly makes tests more readable

issuer-list/src/test/java/com/adyen/checkout/issuerlist/DefaultIssuerListDelegateTest.kt
```diff
@@ -0,0 +1,115 @@
...
+                Assert.assertEquals(false, outputData?.isValid == true)
+            }
+        }
+
+        @Test
+        fun `selectedIssuer is valid, then output should be valid`() = runTest {
+            delegate.outputDataFlow.test {
+                delegate.onInputDataChanged(IssuerListInputData(IssuerModel(id = "id", name = "test")))
+                val outputData = expectMostRecentItem()
```
jreij, line 59: Small comment: you can get rid of the null operator and you don't need to repeat `outputData` every assert if you do something like this:

```kotlin
    with(requireNotNull(expectMostRecentItem())) {
        Assert.assertEquals("test", selectedIssuer?.name)
        Assert.assertEquals("id", selectedIssuer?.id)
        Assert.assertTrue(isValid)
    }
```

It's totally fine how it is now but I just thought I'd leave this comment in case you find it better this way.

issuer-list/src/test/java/com/adyen/checkout/issuerlist/utils/TestIssuerPaymentMethod.kt
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Created by atef on 17/8/2022.
+ */
+
+package com.adyen.checkout.issuerlist.utils
+
+import android.os.Parcel
+import com.adyen.checkout.components.model.payments.request.IssuerListPaymentMethod
+
+class TestIssuerPaymentMethod(
```
jreij, line 14: Same comment as Oscar, should be `internal`

jreij (approved):

PR #763, opened 2022-08-18
Refactor redirect delegate

jreij:


redirect/src/main/java/com/adyen/checkout/redirect/RedirectComponent.kt
```diff
@@ -44,26 +59,27 @@ class RedirectComponent(
...
+         * This value should be the beginning of the `returnUrl` sent on the payments/ call.
+         */
+        const val REDIRECT_RESULT_SCHEME = BuildConfig.checkoutRedirectScheme + "://"
+
         /**
          * Returns the suggested value to be used as the `returnUrl` value in the payments/ call.
          *
          * @param context The context provides the package name which constitutes part of the ReturnUrl
-         * @return The suggested `returnUrl` to be used. Consists of [RedirectUtil.REDIRECT_RESULT_SCHEME] + App package name.
+         * @return The suggested `returnUrl` to be used. Consists of [RedirectHandler.REDIRECT_RESULT_SCHEME] + App package name.
```
OscarSpruit, line 79: Is this really coming from `RedirectHandler`?

redirect/src/test/java/com/adyen/checkout/redirect/DefaultRedirectHandlerTest.kt
```diff
@@ -19,12 +20,14 @@ import org.robolectric.annotation.Config
 
 @RunWith(RobolectricTestRunner::class)
 @Config(manifest = Config.NONE)
-class RedirectUtilTest {
+class DefaultRedirectHandlerTest {
```
OscarSpruit, line 23: This has nothing to do with your PR, but I hate Robolectric tests. What if we use Java's URL, can we then have a normal unit test?

jreij: Unfortunately java `URL` is not the best class out there 😅

OscarSpruit (approved):

OscarSpruit (approved):

PR #766, opened 2022-08-19
WeChatDelegate: Create WeChatDelegate

OscarSpruit:
- Move logic from component to delegate
- Extract as much logic from `WeChatPayUtils` to classes that use it
- Wrap generation of `PayReq` in a class, so it's testable

### DIFF
--- a/wechatpay/src/main/java/com/adyen/checkout/wechatpay/WeChatDelegate.kt
+++ b/wechatpay/src/main/java/com/adyen/checkout/wechatpay/WeChatDelegate.kt
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 19/8/2022.
+ */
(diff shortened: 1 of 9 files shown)


wechatpay/src/main/java/com/adyen/checkout/wechatpay/WeChatDelegate.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import android.content.Intent
+import com.adyen.checkout.components.model.payments.response.Action
+import com.adyen.checkout.core.exception.CheckoutException
+import kotlinx.coroutines.flow.Flow
+import org.json.JSONObject
+
+interface WeChatDelegate {
+
+    val detailsFlow: Flow<JSONObject>
```
jreij, line 19: Nit: you might want to observe these flows somewhere 😆

OscarSpruit: Lol, can't believe I missed it again...

jreij (approved):

PR #772, opened 2022-08-26
3DS2Delegate: Create 3DS2Delegate

OscarSpruit:
- Move logic from component to delegate
- Split `Base64Encoder` into an interface and two implementations (for testing)
- Small tweaks to make testing easier

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
@@ -9,88 +9,65 @@ package com.adyen.checkout.adyen3ds2
 
 import android.app.Activity
 import android.app.Application
-import android.content.Context
 import android.content.Intent
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.Observer
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.viewModelScope
-import com.adyen.checkout.adyen3ds2.exception.Authentication3DS2Exception
-import com.adyen.checkout.adyen3ds2.exception.Cancelled3DS2Exception
-import com.adyen.checkout.adyen3ds2.model.ChallengeToken
-import com.adyen.checkout.adyen3ds2.model.FingerprintToken
-import com.adyen.checkout.adyen3ds2.repository.SubmitFingerprintRepository
-import com.adyen.checkout.adyen3ds2.repository.SubmitFingerprintResult
 import com.adyen.checkout.components.ActionComponentData
 import com.adyen.checkout.components.ActionComponentProvider
 import com.adyen.checkout.components.base.BaseActionComponent
 import com.adyen.checkout.components.base.IntentHandlingComponent
-import com.adyen.checkout.components.encoding.Base64Encoder
 import com.adyen.checkout.components.model.payments.response.Action
-import com.adyen.checkout.components.model.payments.response.RedirectAction
-import com.adyen.checkout.components.model.payments.response.Threeds2Action
-import com.adyen.checkout.components.model.payments.response.Threeds2Action.SubType
-import com.adyen.checkout.components.model.payments.response.Threeds2ChallengeAction
-import com.adyen.checkout.components.model.payments.response.Threeds2FingerprintAction
-import com.adyen.checkout.core.exception.CheckoutException
 import com.adyen.checkout.core.exception.ComponentException
 import com.adyen.checkout.core.log.LogUtil
 import com.adyen.checkout.core.log.Logger
-import com.adyen.checkout.redirect.handler.RedirectHandler
-import com.adyen.threeds2.AuthenticationRequestParameters
-import com.adyen.threeds2.ChallengeStatusReceiver
-import com.adyen.threeds2.CompletionEvent
-import com.adyen.threeds2.ProtocolErrorEvent
-import com.adyen.threeds2.RuntimeErrorEvent
 import com.adyen.threeds2.ThreeDS2Service
-import com.adyen.threeds2.Transaction
 import com.adyen.threeds2.customization.UiCustomization
-import com.adyen.threeds2.exception.InvalidInputException
-import com.adyen.threeds2.exception.SDKAlreadyInitializedException
 import com.adyen.threeds2.exception.SDKNotInitializedException
-import com.adyen.threeds2.exception.SDKRuntimeException
-import com.adyen.threeds2.parameters.ChallengeParameters
-import com.adyen.threeds2.util.AdyenConfigParameters
-import kotlinx.coroutines.CoroutineExceptionHandler
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
-import org.json.JSONException
-import org.json.JSONObject
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
 
-@Suppress("TooManyFunctions")
 class Adyen3DS2Component(
     savedStateHandle: SavedStateHandle,
     application: Application,
     configuration: Adyen3DS2Configuration,
-    private val submitFingerprintRepository: SubmitFingerprintRepository,
-    private val adyen3DS2Serializer: Adyen3DS2Serializer,
-    private val redirectHandler: RedirectHandler,
+    private val adyen3DS2Delegate: Adyen3DS2Delegate,
 ) : BaseActionComponent<Adyen3DS2Configuration>(savedStateHandle, application, configuration),
-    ChallengeStatusReceiver,
     IntentHandlingComponent {
 
-    private var mTransaction: Transaction? = null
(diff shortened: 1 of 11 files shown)


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Delegate.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    val detailsFlow: Flow<JSONObject>
+
+    val exceptionFlow: Flow<CheckoutException>
+
+    val eventFlow: Flow<Adyen3DS2Event>
+
+    var uiCustomization: UiCustomization?
+
+    val gotDestroyedWhileChallenging: Boolean
```
jreij, line 30: Wondering if this will even work, `gotDestroyedWhileChallenging` was a static field before because the whole component would be destroyed and recreated which now the value will be set to true but the delegate reference will be lost anyway so it will be `false` again with the new delegate.

I think we can just remove it completely, all it did was log a line and we already have a check for transaction availability in `challengeShopper` with an exception. WDYT?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Event.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Created by oscars on 23/8/2022.
+ */
+
+package com.adyen.checkout.adyen3ds2
+
+sealed class Adyen3DS2Event {
+
+    object CleanUp3DS2 : Adyen3DS2Event()
+
+    object ClearPaymentData : Adyen3DS2Event()
```
jreij, line 15: We can remove this in the WIP action component branch right? Maybe add a TODO for it

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
```diff
@@ -0,0 +1,404 @@
...
+    private val submitFingerprintRepository: SubmitFingerprintRepository,
+    private val adyen3DS2Serializer: Adyen3DS2Serializer,
+    private val redirectHandler: RedirectHandler,
+    private val threeDS2Service: ThreeDS2Service,
+    private val defaultDispatcher: CoroutineDispatcher,
+    private val embeddedRequestorAppUrl: String,
+    private val base64Encoder: Base64Encoder,
+) : Adyen3DS2Delegate, ChallengeStatusReceiver {
+
+    private val _detailsFlow = MutableSharedFlow<JSONObject>(1, 1, BufferOverflow.DROP_OLDEST)
```
jreij, line 69: Can be replaced with `private val _detailsFlow: MutableSharedFlow<JSONObject> = MutableSingleEventSharedFlow()`

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
```diff
@@ -0,0 +1,404 @@
...
+                submitFingerprintAutomatically(activity, encodedFingerprint, paymentData)
+            } else {
+                _detailsFlow.tryEmit(adyen3DS2Serializer.createFingerprintDetails(encodedFingerprint))
+            }
+        }
+    }
+
+    @Throws(ComponentException::class)
+    private fun createEncodedFingerprint(authenticationRequestParameters: AuthenticationRequestParameters): String {
+        val fingerprintJson = JSONObject()
```
jreij, line 232: Nit: can use `JSONObject().apply` and remove the `fingerprintJson.` calls

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
```diff
@@ -9,88 +9,65 @@ package com.adyen.checkout.adyen3ds2
...
+        adyen3DS2Delegate.eventFlow
+            .onEach { onEvent(it) }
+            .launchIn(viewModelScope)
+    }
+
+    private fun onEvent(event: Adyen3DS2Event) {
+        when (event) {
+            Adyen3DS2Event.CleanUp3DS2 -> {
+                try {
+                    ThreeDS2Service.INSTANCE.cleanup(getApplication())
```
jreij, line 57: How bad is it if we inject application into the delegate and clean up from there? Trying to reduce responsibilities of the components

components-core/src/main/java/com/adyen/checkout/components/encoding/Base64Encoder.kt
```diff
@@ -9,21 +9,57 @@ package com.adyen.checkout.components.encoding
...
+            if (Charset.isSupported(UTF_8)) Charset.forName(UTF_8) else Charset.defaultCharset()
+    }
+}
+
+/**
+ * Java implementation of [AndroidBase64Encoder]. This implementations can only be used from API 26+ and is mainly used
+ * for testing.
+ */
+@Suppress("NewApi")
+class JavaBase64Encoder : Base64Encoder {
```
jreij, line 48: Would there be any benefit from using this in the main code as well for API 26+?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
```diff
@@ -0,0 +1,404 @@
...
+        submitFingerprintAutomatically: Boolean,
+        paymentData: String?,
+    ) {
+        Logger.d(TAG, "identifyShopper - submitFingerprintAutomatically: $submitFingerprintAutomatically")
+        val decodedFingerprintToken = base64Encoder.decode(encodedFingerprintToken)
+
+        val fingerprintJson: JSONObject = try {
+            JSONObject(decodedFingerprintToken)
+        } catch (e: JSONException) {
+            throw ComponentException("JSON parsing of FingerprintToken failed", e)
```
jreij, line 169: Should we replace all `throw` with flow emit in the delegate?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Delegate.kt
OscarSpruit: You're right, I think it's safe to remove

redirect/src/main/java/com/adyen/checkout/redirect/test/TestRedirectHandler.kt
```diff
@@ -24,15 +24,21 @@ class TestRedirectHandler : RedirectHandler {
 
     var exception: ComponentException? = null
 
+    private var timesLaunchRedirectCalled = 0
```
jreij, line 27: Nice! I should add some tests for this in the other components as well

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
OscarSpruit: I have been thinking about that as well, but I think it will make the code less readable. In most cases, the method that calls a method that throws an exception has to return in case of an exception. Meaning, the method that throws should return a boolean or something, so the calling method can return. It will some add extra if statements/overhead 🤔

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
OscarSpruit: I don't like having any context injected into a delegate, not even in a ViewModel. I'd rather delegate it to a fragment/activity. But I suppose we can inject the application into a delegate, that should not cause any memory leaks. But I'm not sure if it will work in tests

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
jreij: Good point

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Component.kt
jreij: Yeah maybe let's not do it, we'll add some code in the components doesn't matter

OscarSpruit: Ohh I did it already, it didn't cause any problems

jreij: 👍

jreij (approved):

PR #774, opened 2022-08-30
Online banking Czech

etman55:


online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,71 @@
...
+        } ?: IssuersUtils.getLegacyIssuers(paymentMethod.details)
+    }
+
+    override fun createComponentState(outputData: IssuerListOutputData) {
+        val issuerListPaymentMethod = onlineBankingCZPaymentMethod()
+        issuerListPaymentMethod.type = getPaymentMethodType()
+        issuerListPaymentMethod.issuer = outputData.selectedIssuer?.id ?: ""
+
+        val paymentComponentData = PaymentComponentData<OnlineBankingCZPaymentMethod>()
+        paymentComponentData.paymentMethod = issuerListPaymentMethod
```
OscarSpruit, line 48: Would be better to use the constructor to set the payment method, like this
```suggestion
        val paymentComponentData = PaymentComponentData(
            paymentMethod = issuerListPaymentMethod
        )
```

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,71 @@
...
+import com.adyen.checkout.issuerlist.IssuerListInputData
+import com.adyen.checkout.issuerlist.IssuerListOutputData
+import com.adyen.checkout.issuerlist.IssuerModel
+import com.adyen.checkout.issuerlist.IssuersUtils
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class OnlineBankingCZDelegate(
+    private val paymentMethod: PaymentMethod,
+    private val onlineBankingCZPaymentMethod: () -> OnlineBankingCZPaymentMethod
```
OscarSpruit, line 26: `paymentMethodFactory` would be a better name. Right now, I was confused why this would be a lambda, but after diving in the code I realised there is a good reason for it

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,71 @@
...
+) : IssuerListDelegate<OnlineBankingCZPaymentMethod> {
+
+    private val _outputDataFlow = MutableStateFlow<OnlineBankingCZOutputData?>(null)
+    override val outputDataFlow: Flow<OnlineBankingCZOutputData?> get() = _outputDataFlow
+
+    private val _componentStateFlow = MutableStateFlow<PaymentComponentState<OnlineBankingCZPaymentMethod>?>(null)
+    override val componentStateFlow: Flow<PaymentComponentState<OnlineBankingCZPaymentMethod>?>
+        get() = _componentStateFlow
```
OscarSpruit, line 34: There is no getter needed here
```suggestion
    override val componentStateFlow: Flow<PaymentComponentState<OnlineBankingCZPaymentMethod>?> = _componentStateFlow
```

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    override fun getPaymentMethodType(): String {
+        return paymentMethod.type ?: PaymentMethodTypes.UNKNOWN
+    }
+
+    fun getTermsAndConditionsUrl(): String = TERMS_CONDITIONS_URL
+
+    companion object {
+        private const val TERMS_CONDITIONS_URL = "https://static.payu.com/sites/terms/files/payu_privacy_policy_cs.pdf"
```
OscarSpruit, line 69: Why is the URL hardcoded? I don't think it's a good idea. What if this URL ever changes? In that case we have to change the SDK, publish a new release, merchants have to update and release a new version of their app. If users can't view terms and conditions merchants might be in trouble, because of laws. 🤔  It would be better to have this coming from the backend

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinner.kt
```diff
@@ -0,0 +1,40 @@
...
+
+class OnlineBankingCZSpinner @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : IssuerListSpinnerView<OnlineBankingCZPaymentMethod, OnlineBankingCZComponent>(context, attrs, defStyleAttr) {
+
+    override fun onComponentAttached() {
+        super.onComponentAttached()
+        termsAndConditionsTextView?.visibility = View.VISIBLE
```
OscarSpruit, line 28: `IssuerListSpinnerView`  now has these terms and conditions views, but they do not fully function there. I would either make it fully working for `IssuerListSpinnerView` or have a special layout for `OnlineBankingCZSpinner`, so only this spinner does have terms and conditions. The last option has my preference as it makes sure existing spinners are not affected

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinner.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.content.Context
+import android.net.Uri
+import android.util.AttributeSet
+import android.view.View
+import androidx.browser.customtabs.CustomTabsIntent
+import com.adyen.checkout.components.model.payments.request.OnlineBankingCZPaymentMethod
+import com.adyen.checkout.components.ui.util.ThemeUtil
+import com.adyen.checkout.issuerlist.IssuerListSpinnerView
+
+class OnlineBankingCZSpinner @JvmOverloads constructor(
```
OscarSpruit, line 20: All views are suffixed with `View`, so it should be `OnlineBankingCZSpinnerView`

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
```diff
@@ -38,17 +40,30 @@ abstract class IssuerListSpinnerView<
...
 
     override fun initView() {
-        issuersSpinner = findViewById<AppCompatSpinner?>(R.id.spinner_issuers).apply {
-            adapter = issuersAdapter
-            onItemSelectedListener = this@IssuerListSpinnerView
+        autoCompleteTextViewIssuers.apply {
+            inputType = 0
+            setAdapter(issuersAdapter)
+            onItemClickListener = AdapterView.OnItemClickListener { _, _, position, _ ->
+                d(TAG, "onItemSelected - " + issuersAdapter.getItem(position).name)
```
OscarSpruit, line 63: If functions do not have a descriptive name I would not import them directly. What is `d()`? No idea. `Logger.d()`? Ohhh debug data is being logged

components-core/src/main/java/com/adyen/checkout/components/model/payments/request/OnlineBankingCZPaymentMethod.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import android.os.Parcel
+import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.core.exception.ModelSerializationException
+import com.adyen.checkout.core.model.JsonUtils
+import com.adyen.checkout.core.model.getStringOrNull
+import org.json.JSONException
+import org.json.JSONObject
+
+class OnlineBankingCZPaymentMethod(
```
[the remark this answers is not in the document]
etman55: Yes would be better, but all other PaymentMethods aren't data classes so I guess there's no use of data class in this context

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZComponent.kt
```diff
@@ -0,0 +1,44 @@
...
+import androidx.lifecycle.SavedStateHandle
+import com.adyen.checkout.components.PaymentComponentProvider
+import com.adyen.checkout.components.model.payments.request.OnlineBankingCZPaymentMethod
+import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.issuerlist.IssuerListComponent
+import com.adyen.checkout.issuerlist.IssuerListDelegate
+
+class OnlineBankingCZComponent(
+    savedStateHandle: SavedStateHandle,
+    issuerListDelegate: IssuerListDelegate<OnlineBankingCZPaymentMethod>,
```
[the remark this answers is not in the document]
etman55: True, but I just did this for code consistency with other IssuerListDelegate

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZDelegate.kt
etman55: good point and I asked same questions before and had the same thoughts about better to come from backend 😣 and they just told me the link will stay the same maybe the file will be updated on same link

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinner.kt
etman55: I tried to do another view for `OnlineBankingCZSpinner ` but I have to create a whole new component that implements a `ViewableComponent` and there's no use of `IssuerListSpinnerView` anywhere I mean every other Issuer component is using recyclerView instead.

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
etman55: I guess this happen because of copy and paste and android studio decided it will be like that 😅

components-core/src/main/java/com/adyen/checkout/components/model/payments/request/OnlineBankingCZPaymentMethod.kt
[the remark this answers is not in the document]
OscarSpruit: That is not really true, take `DotpayPaymentMethod` for example. Data classes make comparisons a lot easier, which comes in handy with for example unit tests. In general, I would say always use data class for objects that only hold data

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinner.kt
OscarSpruit: From what I understood, the other implementations of `IssuerListSpinnerView` can be used by merchants if they implement standalone components and don't want to use the `RecyclerView`s. It might be more work, but I still think it's a good idea to separate the terms and conditions stuff from the other spinners

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
OscarSpruit: Haha yeah sometimes it does that

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinner.kt
etman55: It's a lot of work I'll be redoing the whole thing 😅 because `IssuerListComponent` is not extending `ViewableComponent` so I'll redo the whole component and delegate

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -271,6 +272,14 @@ class DropInConfiguration : Configuration, Parcelable {
             return this
         }
 
+        /**
+         * Add configuration for Online Banking Poland payment method.
```
ozgur00, line 276: Poland :)

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
```diff
@@ -80,20 +92,26 @@ abstract class IssuerListSpinnerView<
...
+        textInputLayoutIssuers.isEnabled = enabled
     }
 
     override fun onNothingSelected(parent: AdapterView<*>) {
         // nothing changed
     }
 
+    private fun setupTermsAndConditionsHyperLink() {
+        termsAndConditionsTextView?.movementMethod = LinkMovementMethod.getInstance()
+        termsAndConditionsTextView?.setLinkTextColor(Color.BLUE)
```
ozgur00, line 112: are we sure about this blue? shouldn't it be a custom blue?

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
```diff
@@ -80,20 +92,26 @@ abstract class IssuerListSpinnerView<
...
+        autoCompleteTextViewIssuers.isEnabled = enabled
+        textInputLayoutIssuers.isEnabled = enabled
     }
 
     override fun onNothingSelected(parent: AdapterView<*>) {
         // nothing changed
     }
 
+    private fun setupTermsAndConditionsHyperLink() {
+        termsAndConditionsTextView?.movementMethod = LinkMovementMethod.getInstance()
```
ozgur00, line 111: when there's multiple usages of the same field you can use `.apply`

```
termsAndConditionsTextView?.apply {
    // setting stuff
}

File: issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt [hunk 13]
```diff
@@ -38,32 +42,40 @@ abstract class IssuerListSpinnerView<
...
-    private lateinit var issuersSpinner: AppCompatSpinner
-    private lateinit var issuersAdapter: IssuerListSpinnerAdapter
+    protected val termsAndConditionsTextView: TextView?
+        get() = rootView.findViewById<TextView>(R.id.textview_termsAndConditions)
+    private val autoCompleteTextViewIssuers: AutoCompleteTextView
+        get() = rootView.findViewById(R.id.autoCompleteTextView_issuers)
+    private val textInputLayoutIssuers: TextInputLayout
+        get() = rootView.findViewById(R.id.textInputLayout_issuers)
+
+    private val issuersAdapter: SimpleTextListAdapter<IssuerModel> = SimpleTextListAdapter(context)
```
ozgur00, line 52: Why did we change this from `IssuerListSpinnerAdapter` to `SimpleTextListAdapter`? `IssuerListSpinnerAdapter` items has an image and `SimpleTextListAdapter` items don't. And it would break payment methods like `iDeal`. Did you test that?

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerModel.kt
```diff
@@ -8,7 +8,9 @@
 
 package com.adyen.checkout.issuerlist
 
+import com.adyen.checkout.components.ui.adapter.SimpleTextListItem
+
 data class IssuerModel(
     val id: String,
     val name: String
-)
+) : SimpleTextListItem(name)
```
ozgur00, line 16: This shouldn't be the case either. For issuers we shouldn't use `SimpleTextListAdapter`

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
etman55: I will revert all changes in `IssuerListSpinner` and will create a new `OnlineBankingSpinnerView` as yeah maybe this will break stuff.

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerView.kt
etman55: good catch I am not really sure.

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
etman55: haha a lot of copy pasting as they're the same in more than one thing 😅

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/IssuersExt.kt
```diff
@@ -0,0 +1,34 @@
+/*
```
OscarSpruit, line 1: This file is duplicated now, one in issuer list as well, right?

online-banking-cz/src/main/res/values/syles.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ Copyright (c) 2022 Adyen N.V.
+  ~
+  ~ This file is open source and available under the MIT license. See the LICENSE file for more info.
+  ~
+  ~ Created by atef on 25/8/2022.
+  -->
+
+<resources>
```
OscarSpruit, line 9: `syles.xml` -> `styles.xml`

issuer-list/build.gradle
```diff
@@ -40,6 +40,7 @@ dependencies {
     // Dependencies
     implementation libraries.androidx.appcompat
     implementation libraries.androidx.recyclerview
+    implementation libraries.material
```
OscarSpruit, line 43: I guess this is no longer needed now the terms and conditions is separated

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinnerView.kt
```diff
@@ -0,0 +1,113 @@
...
+import com.adyen.checkout.components.PaymentComponentState
+import com.adyen.checkout.components.model.payments.request.OnlineBankingCZPaymentMethod
+import com.adyen.checkout.components.ui.adapter.SimpleTextListAdapter
+import com.adyen.checkout.components.ui.util.ThemeUtil
+import com.adyen.checkout.components.ui.view.AdyenLinearLayout
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+import com.adyen.checkout.onlinebankingcz.databinding.OnlineBankingCzSpinnerLayoutBinding
+
+class OnlineBankingCZSpinnerView @JvmOverloads constructor(
```
OscarSpruit, line 29: I would leave out `Spinner` of this name. The issuer list "spinner" view is already a weird name, but at least it's got a spinner, this view doesn't have a spinner

OscarSpruit:
Opening the terms and conditions forces you to download a file now. It would be nicer UX if we could just show the file in a browser or chrome tab 🤔

OscarSpruit:
What is this empty entry in the bank selection? 
image

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/IssuersExt.kt
etman55: no it's for another model as I created a new `OnlineBankingModel` data class that extends `SimpleTextListItem`, but yeah it's the same functionality

OscarSpruit:
According the design the terms and conditions text should have a little more margin left and right, so it's not fully aligned with the selector and the button 
image

Actual design:
image

etman55:
about force downloading we can ask for UX opinion about it. Also there's nothing about it in the ticket I just did it the same way like `VoucherView`

etman55:
> What is this empty entry in the bank selection? image

it's empty from backend side I double checked this when I saw it the first time

etman55:
> Also the text should be gray, but maybe it's like this because of our theming?
yeah it is because of our theming.

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/IssuersExt.kt
OscarSpruit: Hmm in that case these methods can be `internal`, also the ones in issuer list

OscarSpruit:
Things that are clickable should have a ripple, so the terms and conditions should have one. This can easily be done with `android:background="?selectableItemBackground"`. Also, according to design guidelines clickable things should have a height and width of at least 48dp. This means the terms and conditions TextView should have a `minHeight` of 48dp and to make the ripple look nicer the width should be equal to its parent (full screen width). You'll have to play around with other view paddings and margins to make it look like the design

online-banking-cz/src/main/res/layout/online_banking_cz_spinner_layout.xml
```diff
@@ -0,0 +1,37 @@
...
+            android:layout_height="wrap_content"
+            android:dropDownAnchor="@id/textInputLayout_issuers" />
+    </com.google.android.material.textfield.TextInputLayout>
+
+    <TextView
+        android:id="@+id/textview_termsAndConditions"
+        style="@style/AdyenCheckout.Issuer.TermsAndConditions"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        tools:text="A.Etman" />
```
OscarSpruit, line 35: It's more useful to have the actual text here, so the preview looks more like how it would like when you run the app

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZSpinnerView.kt
```diff
@@ -0,0 +1,113 @@
...
+            inputType = 0
+            setAdapter(issuersAdapter)
+            onItemClickListener = AdapterView.OnItemClickListener { _, _, position, _ ->
+                Logger.d(TAG, "onItemSelected - " + issuersAdapter.getItem(position).name)
+                component.inputData.selectedIssuer = issuersAdapter.getItem(position)
+                component.notifyInputDataChanged()
+            }
+        }
+        binding.textviewTermsAndConditions.apply {
+            movementMethod = LinkMovementMethod.getInstance()
```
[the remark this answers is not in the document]
etman55: hmmm I was kinda not sure which one should be clickable the link or the whole text view so that's why I added this, but I guess since we have different opinions about it I should ask Arjen about it

online-banking-cz/src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,18 @@
...
+  ~ Copyright (c) 2022 Adyen N.V.
+  ~
+  ~ This file is open source and available under the MIT license. See the LICENSE file for more info.
+  ~
+  ~ Created by atef on 25/8/2022.
+  -->
+
+<resources>
+
+    <dimen name="termsAndConditions_marginStart">17dp</dimen>
```
OscarSpruit, line 11: Please stick to the 8 pixel grid, so it should be 16dp. Figma is sometimes wrong in this

online-banking-cz/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,29 @@
...
+
+<resources>
+
+
+    <style name="AdyenCheckout.Issuer" />
+
+    <style name="AdyenCheckout.Issuer.TermsAndConditions">
+        <item name="android:text">@string/checkout_issuers_list_terms_and_conditions_text</item>
+        <item name="android:layout_marginStart">@dimen/termsAndConditions_marginStart</item>
+        <item name="android:textColorLink">@color/termsAndConditionsColor</item>
```
OscarSpruit, line 17: It would be better to define a generic `textColorLink` in the base theme. That would make it easier for merchants to override it

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -0,0 +1,97 @@
...
+        OnlineBankingConfiguration,
+        PaymentComponentState<OnlineBankingCZPaymentMethod>,
+        OnlineBankingCZComponent
+        >(context, attrs, defStyleAttr),
+    AdapterView.OnItemSelectedListener {
+
+    private val binding: OnlineBankingCzSpinnerLayoutBinding =
+        OnlineBankingCzSpinnerLayoutBinding.inflate(LayoutInflater.from(context), this)
+
+    private val issuersAdapter: SimpleTextListAdapter<OnlineBankingModel> = SimpleTextListAdapter(context)
```
jreij, line 41: This adapter means that the icons of the banks won't be appearing right? Should we make a custom adapter that inflates custom views and shows the icon as well, as per the design?
Screenshot_1663078837

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -271,6 +272,14 @@ class DropInConfiguration : Configuration, Parcelable {
             return this
         }
 
+        /**
+         * Add configuration for Online Banking Czech payment method.
```
jreij, line 276: Nitpick: should this be called `Czech Republic` instead of just `Czech`?

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuersUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Created by atef on 25/8/2022.
+ */
+
+package com.adyen.checkout.issuerlist
+
+import com.adyen.checkout.components.model.paymentmethods.InputDetail
+import com.adyen.checkout.components.model.paymentmethods.Issuer
+
+fun getIssuers(issuerList: List<Issuer>): List<IssuerModel> =
```
jreij, line 14: My issue with this extension method is that you can call `getIssuers` from anywhere in the project (this module or anything that extends it) without much context, you start writing `get` and you see this method that is 99% not relevant to you.

I suggest we make these methods `internal` to begin with and change the name to maybe `mapToModelList` to better explain what it does.
And also change them from `fun getIssuers(issuerList: List<Issuer>): List<IssuerModel>` to `fun List<Issuer>.mapToModel(): List<IssuerModel>` so that you can only call the method on a list of issuers and not just anywhere.

WDYT?

online-banking-cz/build.gradle
```diff
@@ -0,0 +1,51 @@
...
+ * Created by atef on 23/8/2022.
+ */
+
+plugins {
+    id 'com.android.library'
+    id 'kotlin-android'
+}
+
+ext.mavenArtifactId = "online-banking-cz"
+ext.mavenArtifactName = "Adyen checkout Online Banking Czech"
```
jreij, line 15: Maybe we can ask product about `Online Banking Czech` vs `Online Banking Czech Republic` and change it everywhere if needed.

issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuersUtils.kt
etman55: good point actually

online-banking-cz/build.gradle
etman55: it's "Czech republic" inside the ticket actually, so it's just small change will do it now

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
jreij: Oops I was looking at PL 🤦

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/DefaultOnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    companion object {
+        private val TAG = LogUtil.getTag()
+        private const val TERMS_CONDITIONS_URL = "https://static.payu.com/sites/terms/files/payu_privacy_policy_cs.pdf"
+    }
+
+    @Suppress("ReturnCount")
+    override fun launchOpenPdf(context: Context) {
+        val uri = Uri.parse(TERMS_CONDITIONS_URL)
+        if (OpenPdfUtils.launchNative(context, uri)) return
```
OscarSpruit, line 80: The flow here is a bit inefficient. `launchWithCustomTabs` can be called twice in some cases.

`launchNative` will call `launchWithCustomTabs` if the api level is lower than 30. Let's say that fails, thus returns false. Then `launchWithCustomTabs` will be called again on the next line (and it will probably fail again). 
If you flatten this logic, you should be able to prevent this from happening

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/DefaultOnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+        if (OpenPdfUtils.launchWithCustomTabs(context, uri)) return
+        if (OpenPdfUtils.launchBrowser(context, uri)) return
+        Logger.e(TAG, "Could not launch url")
+        throw ComponentException("failed to open terms and conditions pdf.")
```
OscarSpruit, line 84: Instead of throwing the exception, it should be emitted in the `exceptionFlow`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/DefaultOnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.core.exception.CheckoutException
+import com.adyen.checkout.core.exception.ComponentException
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class DefaultOnlineBankingCZDelegate(
```
OscarSpruit, line 27: Can be `internal`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/DefaultOnlineBankingCZDelegate.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val paymentMethod: PaymentMethod,
+    private val paymentMethodFactory: () -> OnlineBankingCZPaymentMethod
+) : OnlineBankingDelegate<OnlineBankingCZPaymentMethod> {
+
+    private val _outputDataFlow = MutableStateFlow<OnlineBankingOutputData?>(null)
```
OscarSpruit, line 32: Oh and speaking of testing, I just noticed this class doesn't have tests yet ;)

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZComponent.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.adyen.checkout.components.model.payments.request.OnlineBankingCZPaymentMethod
+import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.core.exception.CheckoutException
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+
+class OnlineBankingCZComponent(
+    savedStateHandle: SavedStateHandle,
+    private val delegate: DefaultOnlineBankingCZDelegate,
```
jreij, line 26: The delegate should be from the base type not the implementation.

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OpenPdfUtils.kt
```diff
@@ -0,0 +1,86 @@
...
+import android.net.Uri
+import android.os.Build
+import androidx.annotation.RequiresApi
+import androidx.browser.customtabs.CustomTabColorSchemeParams
+import androidx.browser.customtabs.CustomTabsIntent
+import com.adyen.checkout.components.ui.util.ThemeUtil
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+
+object OpenPdfUtils {
```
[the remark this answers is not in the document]
etman55: image

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/DefaultOnlineBankingCZDelegate.kt
etman55: good catch

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OpenPdfUtils.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    fun launchNative(context: Context, uri: Uri): Boolean {
```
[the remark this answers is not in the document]
etman55: Yeah makes sense.

etman55:
still working on unit testing

online-banking-cz/src/test/java/com/adyen/checkout/onlinebankingcz/DefaultOnlineBankingCZDelegateTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    )
+
+    @Nested
+    @DisplayName("when input data changes and")
+    inner class InputDataChangedTest {
+
+        @Test
+        fun `selectedIssuer is null, then output should be null`() = runTest {
+            delegate.outputDataFlow.test {
+                delegate.onInputDataChanged(OnlineBankingInputData(null))
```
OscarSpruit, line 39: Try to follow the following structure when writing tests:
```
// GIVEN
// Perform any setup needed for the test

// WHEN
Execute the code that needs to be tested

// THEN
Do assertions
```

Pay attention here to how these 3 blocks are separated by new lines. Let's see how that would translate into this test:
```Kotlin
@Test
fun `selectedIssuer is null, then output should be null`() = runTest {
    delegate.outputDataFlow.test {
        val inputData = OnlineBankingInputData(null)
        
        delegate.onInputDataChanged(inputData)
        
        with(requireNotNull(expectMostRecentItem())) {
            assertNull(selectedIssuer)
        }
    }
}
```

Side note: `Assert.assertWhatever()` doesn't read very nicely. It is better to import the methods directly, so you just have `assertWhatever()`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    override val isConfirmationRequired: Boolean
+        get() = true
+
+    override fun highlightValidationErrors() {
+        // no implementation
+    }
+
+    override fun initLocalizedStrings(localizedContext: Context) {
+        // no embedded localized strings on this view
```
jreij, line 73: This method should not be empty because we do have strings set in the xml file, namely the "Select your bank" and the terms and conditions strings. These strings will follow the system locale by default but we want to override them by code to make sure they follow the merchant provided locale.

The best way to understand what I mean is this:
- Go to `CheckoutConfigurationProvider` in the example app
- Create a custom `OnlineBankingConfiguration` and set the shopper locale to something like `nl-NL` but keep the drop-in locale to the default locale
- Set the custom locale by using `. addOnlineBankingCZConfiguration`
- You should see that this component displays in Dutch while the rest of drop-in in English, but you won't because `initLocalizedStrings` is empty.

Take a look at other components to see how it's done there.

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+        binding.textviewTermsAndConditions.setOnClickListener { component.openTermsAndConditionsPdf(context) }
+    }
+
+    override val isConfirmationRequired: Boolean
+        get() = true
+
+    override fun highlightValidationErrors() {
+        // no implementation
```
jreij, line 69: What happens if you click Pay without selecting a bank? We should check this here and show an error or select the first bank by default, first option is probably better.

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingConfiguration.kt
```diff
@@ -0,0 +1,83 @@
...
+package com.adyen.checkout.onlinebankingcz
+
+import android.content.Context
+import android.os.Parcel
+import android.os.Parcelable
+import com.adyen.checkout.core.api.Environment
+import com.adyen.checkout.issuerlist.IssuerListConfiguration
+import java.util.Locale
+
+class OnlineBankingConfiguration : IssuerListConfiguration {
```
jreij, line 18: This class should be called `OnlineBankingCZConfiguration` right? Same for the delegate, input data, output data, and model.

online-banking-cz/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,31 @@
...
+        <item name="android:text">@string/checkout_issuers_list_terms_and_conditions_text</item>
+        <item name="android:minHeight">@dimen/termsAndConditions_minHeight</item>
+        <item name="android:paddingStart">@dimen/termsAndConditions_paddingStart</item>
+        <item name="android:paddingEnd">@dimen/termsAndConditions_paddingEnd</item>
+        <item name="android:textSize">@dimen/termsAndConditions_text_size</item>
+        <item name="android:gravity">center_vertical</item>
+        <item name="android:background">?selectableItemBackground</item>
+    </style>
+
+    <style name="AdyenCheckout.DropdownTextInputLayout.Issuer" parent="AdyenCheckout.DropdownTextInputLayout">
```
jreij, line 24: All styles here should be prefixed by `AdyenCheckout.Issuer`, so `AdyenCheckout.Issuer.DropdownTextInputLayout` not `AdyenCheckout.DropdownTextInputLayout.Issuer`

Also maybe it shouldn't be called `AdyenCheckout.Issuer`, it's only for the online banking CZ component, you can just call it `AdyenCheckout.OnlineBankingCZ`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingConfiguration.kt
etman55: I'll redo this after making `online-banking-core` module in the next PR

online-banking-cz/src/main/res/values/styles.xml
etman55: I guess I should have put it like `AdyenCheckout.DropdownTextInputLayout.TermsAndCondition` WDYT?

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
etman55: there's no implementation so I guess nothing happens, I guess yeah we can pop up an error message indicates that you should select your bank or something.

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
etman55: will do it thanks.

online-banking-cz/src/main/res/values/styles.xml
OscarSpruit: I agree with Atef here. Style names in Android go from global to more specific after each dot. Because `AdyenCheckout.DropdownTextInputLayout` is the base style we want to build upon, we should go more specific from there, so you get something like `AdyenCheckout.DropdownTextInputLayout.OnlineBanking.TermsAndConditions` (or you could choose to omit `OnlineBanking`). This is mostly because of how extending styles works, which also means that if you follow this way you could remove the `parent=...`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/PdfOpener.kt
```diff
@@ -0,0 +1,98 @@
...
+import com.adyen.checkout.components.ui.util.ThemeUtil
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+
+internal class PdfOpener {
+
+    @Throws(IllegalStateException::class)
+    fun open(context: Context, url: String) {
+        val uri = Uri.parse(url)
+        if (open(context, uri)) return
```
OscarSpruit, line 27: As I was saying yesterday if `openInExternalApp` fails then `openInBrowser` browser will be called. It is better to first fallback to `openInCustomTab` before opening the browser, so there is one app less in this flow. To achieve this with this structure, this method:
```Kotlin
private fun open(context: Context, uri: Uri): Boolean {
    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) openInExternalApp(context, uri)
    // because custom tabs pdf viewer is working before api 30 on chrome
    else openInCustomTab(context, uri)
}
```
Has to become something like this:
```Kotlin
private fun open(context: Context, uri: Uri): Boolean {
    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
        if (openInExternalApp(context, uri)) true
        else openInCustomTab(context, uri)
    }
    else openInCustomTab(context, uri)
}
```
Does it become more clear now why I wrote that if statement like that in the example I gave before? In this way the flow is in multiple methods and nested if statements, which is more difficult to follow. In my example the whole flow is in just this one if statement (I do agree it's not the prettiest, but it's just 3 lines you have to follow)

etman55: yeah I guess that's why I added an extra if statement for `openInCustomTab` in the first place like this
```
    @Throws(IllegalStateException::class)
    fun open(context: Context, url: String) {
        val uri = Uri.parse(url)
        if (open(context, uri)) return
        if (openInCustomTab(context, uri)) return
        if (openInBrowser(context, uri)) return

        Logger.e(TAG, "Couldn't open pdf with url: $uri")

        throw IllegalStateException("Couldn't open pdf with url: $uri")
    }
```
I guess now I am gonna go with you're suggestion because `openInCustomTab(context, uri)` will be called twice for api level 30

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -109,6 +114,29 @@ class OnlineBankingCZView @JvmOverloads constructor(
         binding.textInputLayoutOnlineBanking.isEnabled = enabled
     }
 
+    private fun styleTermsAndConditions(string: String) {
```
jreij, line 117: Can we extract this method to a a common module in case we need it later? Also can you check what happens if the merchant overrides this string with something that doesn't have `%#` `#%`? Maybe add some tests around it to be safe?

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -109,6 +114,29 @@ class OnlineBankingCZView @JvmOverloads constructor(
...
+        val firstIndex = string.indexOf("%#", 0, ignoreCase = true)
+        val secondIndex = string.indexOf("%#", firstIndex + 2, ignoreCase = true)
+
+        val newString = string.replace("%#", "", ignoreCase = true)
```
jreij, line 121: We're only replacing `%#` with `""`, should we also replace `#%`?

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -122,7 +122,7 @@ class OnlineBankingCZView @JvmOverloads constructor(
 
         val spannable = SpannableString(newString).apply {
             setSpan(
-                ForegroundColorSpan(resources.getColor(R.color.textColorLink, null)),
+                ForegroundColorSpan(resources.getColor(R.color.textColorLink)),
```
jreij, line 125: Isn't this method deprecated?

etman55: it's, but I got an error because we're using min api level 21 and it's deprecated in api level 23 I think not sure though when it was deprecated

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
etman55: the string doesn't contain #% so I don't think we need that

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
jreij: You can use `ResourcesCompat.getColor`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
etman55: about extracting the method I agree with you on this, but if they override this strings I guess they will reimplement their own `initLocalizedStrings` right?

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
jreij: My bad 🤦 I assumed they open with `%#` and close with `#%`

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -0,0 +1,145 @@
...
+        // nothing changed
+    }
+
+    override fun setEnabled(enabled: Boolean) {
+        super.setEnabled(enabled)
+        binding.autoCompleteTextViewOnlineBanking.isEnabled = enabled
+        binding.textInputLayoutOnlineBanking.isEnabled = enabled
+    }
+
+    private fun styleTermsAndConditions(string: String) {
```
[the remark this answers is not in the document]
etman55: Will do it in an extension function and yeah URLSpan make more sense, but does it use the same color that we're using in our design? I am not sure will try it first.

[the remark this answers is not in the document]
etman55: > Biggest advantage of this way is that it uses the link text color that is set in the theme
yeah you're right I didn't see this last line 😅

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZView.kt
```diff
@@ -115,31 +112,8 @@ class OnlineBankingCZView @JvmOverloads constructor(
...
-                secondIndex - 2,
-                Spannable.SPAN_EXCLUSIVE_EXCLUSIVE
-            )
-        }
-
-        binding.textviewTermsAndConditions.text = spannable
-    }
-
     companion object {
+        private const val REPLACEMENT_TOKEN = "%#"
```
jreij, line 116: This should probably be in the same file as `formatStringWithHyperlink` since we will use the same token everywhere.

ui-core/src/main/java/com/adyen/checkout/components/extensions/ViewExtensions.kt
```diff
@@ -36,13 +36,10 @@ fun TextView.setLocalizedTextFromStyle(
...
     val stringResValue = typedArray.getString(0)
-    text = if (formatHyperLink) {
-        // check if the string contains the replacement token twice
-        val counter = stringResValue?.split(replacementToken)?.size?.minus(1)
-        if (counter != 2) return
-        val spannable = stringResValue.formatStringWithHyperlink(replacementToken)
-        spannable
-    } else stringResValue
+    // check if the string contains the replacement token twice
+    val counter = stringResValue?.split(replacementToken)?.size?.minus(1)
```
jreij, line 40: Should we move this whole logic to inside `formatStringWithHyperlink`? `setLocalizedTextFromStyle` shouldn't really know about the logic of hyperlink formatting, including counting how many times did the token appear. What do you think?

jreij (approved):

PR #782, opened 2022-09-15
Bind actions views to delegates

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -585,6 +565,22 @@ class DropInActivity :
         }
     }
 
+    private fun handleWeChatPayResponse(intent: Intent) {
+        val actionFragment = getActionFragment() ?: return
```
OscarSpruit, line 569: In what instances can `getActionFragment()` return null? Just returning would result in nothing happening. Would it be better to throw an error or show some kind of error dialog?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -91,12 +92,8 @@ class ActionComponentDialogFragment : DropInBottomSheetDialogFragment(), Observe
...
             }
 
-            if (!isHandled) {
-                (actionComponent as ActionComponent<*>).handleAction(requireActivity(), action)
-                isHandled = true
-            } else {
-                Logger.d(TAG, "action already handled")
-            }
+            actionComponent.handleAction(requireActivity(), action)
+            attachComponent(actionComponent, binding.componentView)
```
OscarSpruit, line 96: Wouldn't it be better to turn these 2 lines around? That way the view doesn't miss out on anything that happens in `handleAction`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -197,10 +154,24 @@ class ActionComponentDialogFragment : DropInBottomSheetDialogFragment(), Observe
...
 
+    fun handleRedirectResponse(intent: Intent) {
+        handleIntent(intent)
+    }
+
+    fun handleWeChatPayResponse(intent: Intent) {
+        handleIntent(intent)
+    }
+
+    private fun handleIntent(intent: Intent) {
```
OscarSpruit, line 170: If we make this public, then we don't need the `handleRedirectResponse` and `handleWeChatPayResponse` methods

qr-code/src/main/java/com/adyen/checkout/qrcode/QrCodeViewNew.kt
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+        binding.textViewTimer.text = localizedContext.getString(
+            R.string.checkout_qr_code_timer_text,
+            minutesSecondsString
+        )
+        binding.progressIndicator.progress = timerData.progress
+    }
+
+    private fun copyCode(delegate: QRCodeDelegate) {
```
OscarSpruit, line 129: Makes more sense to pass the `qrCodeData` directly to this method, instead of the whole delegate

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -166,11 +164,16 @@ internal fun <T : Configuration> getDefaultConfigForPaymentMethod(
...
-    dropInConfiguration: DropInConfiguration
-): T {
-    return dropInConfiguration.getConfigurationForAction()
-        ?: ActionDelegateProvider.getDefaultConfiguration(dropInConfiguration)
+internal fun createGenericActionConfiguration(dropInConfiguration: DropInConfiguration): GenericActionConfiguration {
+    return GenericActionConfiguration.Builder(
+        dropInConfiguration.shopperLocale,
+        dropInConfiguration.environment,
+        dropInConfiguration.clientKey
+    ).apply {
```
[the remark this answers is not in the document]
jreij: Well the problem is that I don't want to add a `setAvailableActionConfigs` method because `availableActionConfigs` should be an internal map only managed by the `GenericActionConfiguration`. But I needed to access it from drop-in just for that specific case. Maybe I should mark it with `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)`? Or maybe there's a flaw in what i'm trying to do to begin with 🤔

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
jreij: It would happen if the user went back to the app and dismissed drop-in then went back to the browser and completed the payment, something very annoying indeed and we have an open issue about but I'm not sure how to deal with it now.

qr-code/src/main/java/com/adyen/checkout/qrcode/QrCodeViewNew.kt
jreij: Well, you wrote that method...

image

OscarSpruit: I know 😂 Just commented that as a reminder for myself

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
jreij: It doesn't work with the generic component, because until we call `handleAction` the delegate is not created yet, and `attachComponent` requires that the delegate is already created. Even creating an `ActionDelegate` might not solve this :/

OscarSpruit (dismissed):

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -571,7 +571,7 @@ class DropInActivity :
     }
 
     private fun getActionFragment(): ActionComponentDialogFragment? {
-        val fragment = getFragmentByTag(COMPONENT_FRAGMENT_TAG) as? ActionComponentDialogFragment
+        val fragment = getFragmentByTag(ACTION_FRAGMENT_TAG) as? ActionComponentDialogFragment
```
jreij, line 574: Good catch 😬

OscarSpruit (approved):

PR #784, opened 2022-09-15
UpdateKotlin: Update Kotlin and a bunch of other dependencies

OscarSpruit:
Dependencies updated:
- Kotlin > 1.7.10
- Kotlin Coroutines > 1.6.3
- Ktlint > 0.47.1
    - Locally ktlint became a bit unstable after updating Robolectric, so I decided to update it
    - Fixed all new errors
    - Disabled new check that don't add value imo (`multiline-if-else`, `trailing-comma-on-call-site`, `trailing-comma-on-declaration-site`)
- Android Gradle plugin > 7.2.0
- Dagger Hilt > 2.43.2
- Gradle wrapper > 7.5.1

With all these updates compilation should have improved and everything should be better than ever


.editorconfig
```diff
@@ -472,7 +471,8 @@ ij_groovy_while_on_new_line = false
 ij_groovy_wrap_long_lines = false
 
 [{*.kt,*.kts}]
-disabled_rules=import-ordering
+# noinspection EditorConfigKeyCorrectness
+ktlint_disabled_rules=import-ordering,multiline-if-else,trailing-comma-on-call-site,trailing-comma-on-declaration-site
```
jreij, line 475: Should we disable `import-ordering`?

OscarSpruit: That was already disabled. Auto reformatting in Android Studio uses a different ordering than ktlint wants, so that results in some annoying errors. I couldn't figure out how to align them back then, so decided to disable that check

jreij (approved):

PR #791, opened 2022-09-20
Online banking core module

etman55:


online-banking-core/src/main/java/com/adyen/checkout/onlinebankingcore/OnlineBankingInputData.kt
```diff
@@ -1,3 +1,11 @@
+/*
```
jreij, line 1: Duplicate copyright

online-banking-core/src/main/java/com/adyen/checkout/onlinebankingcore/OnlineBankingOutputData.kt
```diff
@@ -1,3 +1,11 @@
+/*
```
jreij, line 1: Duplicate copyright

online-banking-core/src/main/java/com/adyen/checkout/onlinebankingcore/OnlineBankingView.kt
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
```
jreij, line 2: Duplicate copyright

online-banking-core/src/main/res/values/styles.xml
```diff
@@ -1,4 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?><!--
+  ~ Copyright (c) 2022 Adyen N.V.
```
jreij, line 2: Duplicate copyright

online-banking-cz/src/main/java/com/adyen/checkout/onlinebankingcz/OnlineBankingCZComponent.kt
```diff
@@ -3,72 +3,41 @@
...
-import kotlinx.coroutines.flow.launchIn
-import kotlinx.coroutines.flow.onEach
+import com.adyen.checkout.onlinebankingcore.OnlineBankingComponent
+import com.adyen.checkout.onlinebankingcore.DefaultOnlineBankingDelegate
+import com.adyen.checkout.onlinebankingcore.OnlineBankingConfiguration
 
 class OnlineBankingCZComponent(
     savedStateHandle: SavedStateHandle,
-    private val delegate: OnlineBankingDelegate<OnlineBankingCZPaymentMethod>,
+    delegate: DefaultOnlineBankingDelegate<OnlineBankingCZPaymentMethod>,
```
jreij, line 21: This should be `OnlineBankingDelegate` not `DefaultOnlineBankingDelegate`

OscarSpruit, line 21: I think this should use `OnlineBankingDelegate`

ui-core/src/main/java/com/adyen/checkout/components/ui/util/PdfOpener.kt
```diff
@@ -15,11 +15,10 @@ import android.net.Uri
 import android.os.Build
 import androidx.browser.customtabs.CustomTabColorSchemeParams
 import androidx.browser.customtabs.CustomTabsIntent
-import com.adyen.checkout.components.ui.util.ThemeUtil
 import com.adyen.checkout.core.log.LogUtil
 import com.adyen.checkout.core.log.Logger
 
-internal class PdfOpener {
+class PdfOpener {
```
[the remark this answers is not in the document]
jreij: We had a discussion about that actually, any class that need to launch custom tabs needs to rely on `ThemeUtil` which is currently a class in `ui-core` because it tries to fetch the primary theme color. You could argue that `PdfOpener` (and even `RedirectHandler`) are not UI related classes but then what about `ThemeUtil`? 🤔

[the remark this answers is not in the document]
etman55: Joseph and I discussed that yesterday and I wanted to put it actually inside `components-core` but I would depend on `ui-core` and that would make a circular dependency between the two modules as we need to use `ThemeUtil` from `ui-core` so I put it in `ui-core` as there's something related to the ui which is `ThemeUtil` to get primary color for the custom tabs. 
PS: If you have other suggestions please let me know.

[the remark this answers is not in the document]
OscarSpruit: There are multiple solutions here:
1. Move `PdfOpener` to `online-banking-core`. For the time being only online banking is using it. We'll have to find a solution for these dependencies once we have to move it
2. Move `ThemeUtil` with `PdfOpener` to `components-core` or `checkout-core`. Doesn't really solve anything, because now `ThemeUtil` is in a wrong module
3. Ideally, `ui-core` shouldn't have a dependency on `components-core`. It should be the other way around imo. This will require a lot of work because of `AdyenLinearLayout`

Option 1 is best for now I think, but we should go for 3 on the long term if everyone agrees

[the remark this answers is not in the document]
etman55: will move it to `online-banking-core` for now as you suggested and eventually will go with your 3rd option

etman55:
converted to draft till we change the text for terms and conditions from `By  clicking continue.....` to `By continuing...`

jreij (dismissed):

PR #792, opened 2022-09-20
Check bundle on DropInActivity launch and terminate gracefully

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -164,8 +164,7 @@ class DropInActivity : AppCompatActivity(), DropInBottomSheetDialogFragment.Prot
 
         val bundle = savedInstanceState ?: intent.extras
 
-        val initializationSuccessful = assertBundleExists(bundle)
-        if (!initializationSuccessful) {
+        if (!DropInViewModel.assertBundleExists(bundle)) {
```
OscarSpruit, line 167: Why is this a static call?

jreij: The crash happens right when `DropInViewModel` is being initialised, so it would have to be a method that gets called before the view model is initialised.

I decided to put in `DropInViewModel.Companion` because we also have `putIntentExtras` there which is very related to `assertBundleExists`.

How do you suggest we improve this?

OscarSpruit: This is in `onCreate`, right? If so, that means the ViewModel is already available. `assertBundleExists` doesn't need anything else from the ViewModel, so I think we can make it a normal, non-static method

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -345,5 +346,22 @@ class DropInViewModel(
                 putExtra(DROP_IN_RESULT_INTENT_KEY, resultHandlerIntent)
             }
         }
+
+        fun assertBundleExists(bundle: Bundle?): Boolean {
+            if (bundle == null) {
```
OscarSpruit, line 351: I see the check is failing because of the returnCount, you can probably avoid it by doing something like:
```Kotlin
return when {
    bundle == null -> { ... }
    !bundle.containsKey(PAYMENT_METHODS_RESPONSE_KEY) ||
    !bundle.containsKey(DROP_IN_CONFIGURATION_KEY) ||
    !bundle.containsKey(DROP_IN_RESULT_INTENT_KEY) -> { ... }
    else -> true
}
```

jreij: 👍

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
jreij: The view model is lazily initialised meaning even inside `onCreate` it's not created yet. And the crash happens on init because some `val`s inside it depend on the saved state handle having these keys present there.

OscarSpruit: Hmm I see now

OscarSpruit (approved):

PR #800, opened 2022-09-28
IntermediateActionScreen: Show payment in progress screen for redirect and 3ds2

OscarSpruit:
- Add `PaymentInProgressView`. It looks a bit like the `AwaitView`, but it might change once we have a final design
- Show this new view for redirect and 3ds2 payment flows

Bonus:
- Make `DefaultStatusRepositoryTest` more stable
- Add extension method to lazily initialise arguments:
```Kotlin
class MyFragment : Fragment() {

    // Not so cool way to do it
    private val someArgument: ParcelableClass
        get() = arguments?.getParcelable(ARG_KEY) ?: throw IllegalStateException("OH NO")

    // Cool way to do it 😎 
    private val someArgument: ParcelableClass by arguments(ARG_KEY)

    ...
}
```

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ViewProvider.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ViewProvider.kt
@@ -6,7 +6,7 @@
  * Created by oscars on 23/9/2022.
  */
 
-package com.adyen.checkout.redirect
+package com.adyen.checkout.adyen3ds2
 
 import android.content.Context
 import android.util.AttributeSet
@@ -15,7 +15,7 @@ import com.adyen.checkout.components.ui.PaymentInProgressView
 import com.adyen.checkout.components.ui.ViewProvider
 import com.adyen.checkout.components.ui.view.ComponentViewType
 
-internal object RedirectViewProvider : ViewProvider {
+object Adyen3DS2ViewProvider : ViewProvider {
 
     override fun getView(
         viewType: ComponentViewType,
@@ -24,9 +24,9 @@ internal object RedirectViewProvider : ViewProvider {
         defStyleAttr: Int
     ): ComponentViewNew = when (viewType) {
         // TODO find out why passing attrs crashes
-        RedirectComponentViewType -> PaymentInProgressView(context, null, defStyleAttr)
-        else -> throw IllegalArgumentException("Unsupported view type")
+        Adyen3DS2ComponentViewType -> PaymentInProgressView(context, null, defStyleAttr)
+        else -> throw IllegalStateException("Unsupported view type")
     }
 }
 
-object RedirectComponentViewType : ComponentViewType
+object Adyen3DS2ComponentViewType : ComponentViewType
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ViewProvider.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ViewProvider.kt
@@ -23,8 +23,7 @@ object Adyen3DS2ViewProvider : ViewProvider {
         attrs: AttributeSet?,
         defStyleAttr: Int
     ): ComponentViewNew = when (viewType) {
-        // TODO find out why passing attrs crashes
-        Adyen3DS2ComponentViewType -> PaymentInProgressView(context, null, defStyleAttr)
+        Adyen3DS2ComponentViewType -> PaymentInProgressView(context, attrs, defStyleAttr)
         else -> throw IllegalStateException("Unsupported view type")
     }
 }
--- a/components-core/src/main/java/com/adyen/checkout/components/base/ActionDelegate.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/base/ActionDelegate.kt
@@ -26,6 +26,11 @@ interface ActionDelegate<ActionT : Action> : ComponentDelegate {
     fun initialize(coroutineScope: CoroutineScope) = Unit
 
     /**
+     * Override this method if you need to emit to the [exceptionFlow] from outside of this class.
+     */
+    fun onError(e: CheckoutException) = Unit
+
+    /**
      * Override this method if you used [initialize] to clear any local reference to [CoroutineScope].
      */
     fun onCleared() = Unit
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/LazyArguments.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/LazyArguments.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 20/9/2022.
+ */
+
+package com.adyen.checkout.dropin.ui
+
+import android.app.Activity
+import android.os.Bundle
+import androidx.fragment.app.Fragment
+import kotlin.reflect.KProperty
+
+internal inline fun <reified T> Activity.arguments(key: String): LazyProvider<Fragment, T> =
+    argumentDelegate(key) { intent?.extras }
+
+internal inline fun <reified T> Fragment.arguments(key: String): LazyProvider<Fragment, T> =
+    argumentDelegate(key) { arguments }
+
+internal inline fun <A, reified T> argumentDelegate(
+    key: String,
+    crossinline provideArguments: (A) -> Bundle?
+): LazyProvider<A, T> = object : LazyProvider<A, T> {
+    override fun provideDelegate(argumentsFactory: A, prop: KProperty<*>): Lazy<T> = lazy {
+        provideArguments(argumentsFactory)?.get(key) as T
+    }
+}
+
+internal interface LazyProvider<A, T> {
+    operator fun provideDelegate(argumentsFactory: A, prop: KProperty<*>): Lazy<T>
+}
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
@@ -20,6 +20,7 @@ import com.adyen.checkout.action.GenericActionConfiguration
 import com.adyen.checkout.components.ActionComponentData
 import com.adyen.checkout.components.ComponentError
 import com.adyen.checkout.components.model.payments.response.Action
+import com.adyen.checkout.core.exception.CancellationException
 import com.adyen.checkout.core.exception.CheckoutException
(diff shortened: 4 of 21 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -33,18 +34,13 @@ class ActionComponentDialogFragment : DropInBottomSheetDialogFragment() {
     private var _binding: FragmentGenericActionComponentBinding? = null
```
etman55, line 34: I have another cool way to do this if you're interested you can fing it [here](https://github.com/etman55/android-house/blob/main/app/src/main/java/com/atef/clubhouse/base/extension/ViewBinding.kt) 
and you will be using it like this 
```
private val binding by viewBinding(FragmentMainBinding::bind)
```

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ViewProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+        context: Context,
+        attrs: AttributeSet?,
+        defStyleAttr: Int
+    ): ComponentViewNew = when (viewType) {
+        Adyen3DS2ComponentViewType -> PaymentInProgressView(context, attrs, defStyleAttr)
+        else -> throw IllegalStateException("Unsupported view type")
+    }
+}
+
+object Adyen3DS2ComponentViewType : ComponentViewType
```
jreij, line 31: Are there any advantages of having a common `ViewType`/`ViewProvider` for all components who need it, in the same module as `PaymentInProgressView`?

jreij:
Do we need this view for WeChat Pay? Or does it come with a cancellation callback?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/LazyArguments.kt
```diff
@@ -0,0 +1,33 @@
...
+    argumentDelegate(key) { intent?.extras }
+
+internal inline fun <reified T> Fragment.arguments(key: String): LazyProvider<Fragment, T> =
+    argumentDelegate(key) { arguments }
+
+internal inline fun <A, reified T> argumentDelegate(
+    key: String,
+    crossinline provideArguments: (A) -> Bundle?
+): LazyProvider<A, T> = object : LazyProvider<A, T> {
+    override fun provideDelegate(argumentsFactory: A, prop: KProperty<*>): Lazy<T> = lazy {
```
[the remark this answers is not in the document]
OscarSpruit: Although it seems there is no use, it is actually needed in order for Kotlin's property delegation to work (just take a look at the `getValue` method in the ViewBinding magic you linked)

components-core/src/main/java/com/adyen/checkout/components/base/ActionDelegate.kt
```diff
@@ -25,6 +25,11 @@ interface ActionDelegate<ActionT : Action> : ComponentDelegate {
      */
     fun initialize(coroutineScope: CoroutineScope) = Unit
 
+    /**
+     * Override this method if you need to emit to the [exceptionFlow] from outside of this class.
+     */
+    fun onError(e: CheckoutException) = Unit
```
jreij, line 31: Would it be better to use the `InputData` for this?

ui-core/src/main/java/com/adyen/checkout/components/ui/PaymentInProgressView.kt
```diff
@@ -0,0 +1,66 @@
...
+        context,
+        attrs,
+        defStyleAttr
+    ),
+    ComponentViewNew {
+
+    private val binding = ViewPaymentInProgressBinding.inflate(LayoutInflater.from(context), this)
+
+    init {
+        layoutParams = LayoutParams(MATCH_PARENT, WRAP_CONTENT)
```
jreij, line 39: Is this still necessary with the fix I added in the parent component view itself?

ui-core/src/main/java/com/adyen/checkout/components/ui/PaymentInProgressView.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    override fun initView(delegate: ComponentDelegate, coroutineScope: CoroutineScope, localizedContext: Context) {
+        if (delegate !is ActionDelegate<*>) throw IllegalStateException("Unsupported delegate type")
```
jreij, line 43: What if we make an interface just for the error throwing delegates? We might use this in other places, like payment components and the only thing we need the `delegate` for is to throw the error, WDYT?

ui-core/src/main/java/com/adyen/checkout/components/ui/PaymentInProgressView.kt
```diff
@@ -0,0 +1,66 @@
...
+            delegate.onError(CancellationException("Payment in progress was cancelled"))
+        }
+    }
+
+    @Suppress("SetTextI18n", "UNUSED_PARAMETER")
+    private fun initLocalizedStrings(localizedContext: Context) {
+        with(binding) {
+            // TODO: Use string resources when we have final design
+            textviewPaymentinprogressDescription.text = "Waiting for confirmation"
+            buttonPaymentinprogressAbort.text = "I'm too broke"
```
jreij, line 57: 😭

ui-core/src/main/res/layout/view_payment_in_progress.xml
```diff
@@ -0,0 +1,52 @@
...
+        android:id="@+id/button_paymentinprogress_abort"
+        style="@style/AdyenCheckout.Button.Secondary"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_margin="16dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/progressbar_paymentinprogress"
+        tools:text="Abandon payment" />
```
jreij, line 50: Please change to `I'm too broke` for consistency 😆

ui-core/src/main/res/layout/view_payment_in_progress.xml
```diff
@@ -0,0 +1,52 @@
...
+
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    tools:parentTag="androidx.constraintlayout.widget.ConstraintLayout">
+
+    <ProgressBar
+        android:id="@+id/progressbar_paymentinprogress"
```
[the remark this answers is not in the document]
OscarSpruit: I know, but I don't like that we use camel case and snake case combined. But looking at it now, combining the two makes ViewBinding names better readable (`textviewPaymentinprogressDescription` vs `textViewPaymentInProgressDescription`). Will change it

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
OscarSpruit: That's pretty cool! I know how to make what you linked even better. That's for another PR though

jreij:
After the user clicks on the cancel payment button, we show an error dialog (check `ActionComponentDialogFragment.handleError`) with a confusing error message. What if we check if it's a cancellation exception and call `onBackPressed()` instead of `protocol.showError`?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ViewProvider.kt
OscarSpruit: Hmm I have been thinking about that. In the end I decided to go with a specific type per component, because it gives more fine grained control and makes it easier to change the view (if that's ever needed)

ui-core/build.gradle
```diff
@@ -38,6 +42,7 @@ dependencies {
 
     // Dependencies
     implementation libraries.androidx.appcompat
+    api libraries.androidx.constraintlayout
```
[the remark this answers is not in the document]
OscarSpruit: `PaymentInProgressView` is a `ConstraintLayout` that is exposed to other modules. If modules use this view and we wouldn't use `api` those modules would have to do `implementation libraries.androidx.constraintlayout` themselves. 

Use `api` if the insides of a dependency are publicly exposed. If not, use `implementation`. Always favour `implementation` though, because it speeds up build times quite a bit.

ui-core/src/main/java/com/adyen/checkout/components/ui/PaymentInProgressView.kt
OscarSpruit: Maybe not, let me check

components-core/src/main/java/com/adyen/checkout/components/base/ActionDelegate.kt
OscarSpruit: Let me look into that

OscarSpruit:
@jreij handling clicks and navigating back is quite difficult with our setup. I would prefer not to use exceptions for this, but I can't really think of a nice way to do this

jreij:
> @jreij handling clicks and navigating back is quite difficult with our setup. I would prefer not to use exceptions for this, but I can't really think of a nice way to do this

@OscarSpruit we have the `onBackPressed()` method already with some custom (and awful) logic, I'm just suggesting we call it when we have a cancellation as well.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
etman55: I'll work on this PR as I don't have much of stuff to do right now actually.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/LazyArguments.kt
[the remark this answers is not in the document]
etman55: Can you elaborate more about the kotlin property delegation? as the link I sent it was extending `KotlinProperty` and we're passing it here as an argument.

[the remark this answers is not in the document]
OscarSpruit: Take a look at this method (screenshot taken from the link as example). This `getValue` method needs the `KProperty` parameter. Instead of `ReadOnlyProperty` I used `Lazy`. It hides the implementation of `getValue`, but essentially works the same plus some extra benefits like thread safety
image

jreij (approved):

jreij (approved):

PR #801, opened 2022-09-28
Bind BACS views to delegate

jreij:


bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponent.kt
```diff
@@ -23,20 +26,24 @@ import kotlinx.coroutines.flow.onEach
...
-    private val bacsDirectDebitDelegate: BacsDirectDebitDelegate,
+    override val delegate: BacsDirectDebitDelegate,
     configuration: BacsDirectDebitConfiguration
 ) : BasePaymentComponent<BacsDirectDebitConfiguration, BacsDirectDebitInputData, BacsDirectDebitOutputData,
-    BacsDirectDebitComponentState>(savedStateHandle, bacsDirectDebitDelegate, configuration) {
+    BacsDirectDebitComponentState>(savedStateHandle, delegate, configuration),
+    ViewProvidingComponent {
 
-    override val inputData: BacsDirectDebitInputData = BacsDirectDebitInputData()
+    override val viewFlow: Flow<ComponentViewType?> get() = delegate.viewFlow
```
OscarSpruit, line 35: It's a bit better to not use a custom getter, like this 👇 It prevents some overhead code generation. I always think of it in this way: is the value (in this case `delegate.viewFlow`) subject to change (a `var` or calling a method)? If so, use a custom getter. If not, directly assign the value
```Kotlin
override val viewFlow: Flow<ComponentViewType?> = delegate.viewFlow
```

bacs/src/main/java/com/adyen/checkout/bacs/DefaultBacsDirectDebitDelegate.kt
```diff
@@ -8,25 +8,61 @@
...
     override val outputDataFlow: Flow<BacsDirectDebitOutputData?> = _outputDataFlow
 
+    override val outputData
+        get() = _outputDataFlow.value
+
     private val _componentStateFlow = MutableStateFlow<BacsDirectDebitComponentState?>(null)
     override val componentStateFlow: Flow<BacsDirectDebitComponentState?> = _componentStateFlow
 
+    @VisibleForTesting
+    @Suppress("VariableNaming")
```
OscarSpruit, line 40: Just curious: why did you have to suppress this? We use this naming convention everywhere 🤔

jreij: It doesn't like this naming when the variable is not private :/

OscarSpruit: Hmm I get why you did it like this, because it makes testing easier of course. But have you considered just calling `onInputDataChanged` and make this `private`? Or does that have too many side effects?

bacs/src/main/java/com/adyen/checkout/bacs/DefaultBacsDirectDebitDelegate.kt
```diff
@@ -39,11 +75,24 @@ internal class DefaultBacsDirectDebitDelegate(
...
 
         createComponentState(outputData)
     }
 
+    private fun updateViewType(mode: BacsDirectDebitMode) {
+        val viewType = when (mode) {
+            BacsDirectDebitMode.INPUT -> BacsComponentViewType.INPUT
+            BacsDirectDebitMode.CONFIRMATION -> BacsComponentViewType.CONFIRMATION
+        }
+        if (_viewFlow.value != viewType) {
```
OscarSpruit, line 90: This check is not really needed, because `StateFlow`s basically have a built in `distinctUntilChanged`

bacs/src/main/java/com/adyen/checkout/bacs/DefaultBacsDirectDebitDelegate.kt
jreij: I need to setup `_viewFlow` with a certain value before the test, but to do that with only `onInputDataChanged` means i need to call it with very specific values that will make the test confusing.

bacs/src/main/java/com/adyen/checkout/bacs/DefaultBacsDirectDebitDelegate.kt
jreij: Hmm yeah but I'd still like to log a line only when it's different

OscarSpruit (approved):

PR #802, opened 2022-09-29
Bind blik view to delegate

jreij:


blik/src/main/java/com/adyen/checkout/blik/BlikViewNew.kt
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private fun observeDelegate(delegate: BlikDelegate, coroutineScope: CoroutineScope) {
+        delegate.outputDataFlow
+            .onEach { outputDataChanged(it) }
+            .launchIn(coroutineScope)
+    }
+
+    private fun outputDataChanged(blikOutputData: BlikOutputData?) {
+        blikOutputData ?: return
```
OscarSpruit, line 82: What is this supposed to do?

blik/src/main/java/com/adyen/checkout/blik/BlikViewNew.kt
```diff
@@ -0,0 +1,128 @@
...
+            if (hasFocus) {
+                binding.textInputLayoutBlikCode.error = null
+            } else if (blikCodeValidation != null && !blikCodeValidation.isValid()) {
+                val errorReasonResId = (blikCodeValidation as Validation.Invalid).reason
+                binding.textInputLayoutBlikCode.error = localizedContext.getString(errorReasonResId)
+            }
+        }
+    }
+
+    override val isConfirmationRequired: Boolean
```
OscarSpruit, line 105: This could be `override val isConfirmationRequired: Boolean = true`

blik/src/main/java/com/adyen/checkout/blik/BlikViewNew.kt
jreij: For now nothing, but if someone needs it later then can just use it right away

blik/src/main/java/com/adyen/checkout/blik/BlikViewNew.kt
jreij: 👍

blik/src/main/java/com/adyen/checkout/blik/BlikViewNew.kt
OscarSpruit: Ah okay, I would leave it out next time, because it doesn't have a function now and could confuse others. People will add it again when they need it

OscarSpruit (approved):

PR #803, opened 2022-09-29
BindBcmcViewToDelegate: Bind BCMC view to delegate

OscarSpruit:


await/src/main/java/com/adyen/checkout/await/AwaitViewProvider.kt
```diff
@@ -14,7 +14,7 @@ import com.adyen.checkout.components.ui.ComponentViewNew
 import com.adyen.checkout.components.ui.ViewProvider
 import com.adyen.checkout.components.ui.view.ComponentViewType
 
-object AwaitViewProvider : ViewProvider {
+internal object AwaitViewProvider : ViewProvider {
```
jreij, line 17: Nice catch 👍

jreij (approved):

PR #804, opened 2022-09-30
Bind gift card view to delegate

jreij:


giftcard/src/main/java/com/adyen/checkout/giftcard/GiftCardViewNew.kt
```diff
@@ -0,0 +1,149 @@
...
+import com.adyen.checkout.components.ui.ComponentViewNew
+import com.adyen.checkout.components.ui.Validation
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+import com.adyen.checkout.giftcard.databinding.GiftcardViewBinding
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+
+private val TAG = LogUtil.getTag()
```
OscarSpruit, line 29: For consistency I think it's better to have this in a companion object

giftcard/src/main/java/com/adyen/checkout/giftcard/GiftCardViewNew.kt
```diff
@@ -0,0 +1,149 @@
...
+    LinearLayout(
+        context,
+        attrs,
+        defStyleAttr
+    ),
+    ComponentViewNew {
+
+    private val binding: GiftcardViewBinding = GiftcardViewBinding.inflate(LayoutInflater.from(context), this)
+
+    private var imageLoader: ImageLoader? = null
```
OscarSpruit, line 45: Why do we need a `ImageLoader`? This field is only assigned, but never used

PR #805, opened 2022-09-30
Bind sepa view to delegate

jreij:


sepa/src/main/res/layout/sepa_view.xml
```diff
@@ -22,8 +20,9 @@
         android:layout_height="wrap_content">
 
         <com.adyen.checkout.components.ui.view.AdyenTextInputEditText
-            tools:ignore="RequiredSize"
-            style="@style/AdyenCheckout.Sepa.HolderNameInput" />
+            android:id="@+id/editText_holderName"
+            style="@style/AdyenCheckout.Sepa.HolderNameInput"
+            tools:ignore="RequiredSize" />
```
OscarSpruit, line 25: Why don't we specify the size? This could cause some issues

sepa/src/main/res/layout/sepa_view.xml
```diff
@@ -33,8 +32,9 @@
         android:layout_height="wrap_content">
 
         <com.adyen.checkout.sepa.ui.IbanInput
-            tools:ignore="RequiredSize"
-            style="@style/AdyenCheckout.Sepa.AccountNumberInput" />
+            android:id="@+id/editText_ibanNumber"
+            style="@style/AdyenCheckout.Sepa.AccountNumberInput"
+            tools:ignore="RequiredSize" />
```
OscarSpruit, line 37: Same here

sepa/src/main/res/layout/sepa_view.xml
jreij: We have this attribute in 58 places 😅
I don't know what's the thought behind it, maybe for merchants to be able to override width and height using `styles.xml`? 🤔
In all cases I think it needs its own PR if we were to remove it.

OscarSpruit: Right, let's look at it some other time 👍

OscarSpruit (approved):

PR #806, opened 2022-09-30
BindMBWayViewToDelegate: Bind MbWayView to delegate

ozgur00:


mbway/src/main/java/com/adyen/checkout/mbway/MbWayViewNew.kt
```diff
@@ -0,0 +1,133 @@
...
+        if (delegate !is MBWayDelegate) throw IllegalArgumentException("Unsupported delegate type")
+        this.delegate = delegate
+        this.localizedContext = localizedContext
+
+        initMobileNumberInput()
+        initCountryInput()
+    }
+
+    private fun initMobileNumberInput() {
+        val mobileNumberEditText = binding.textInputLayoutMobileNumber.editText as? AdyenTextInputEditText
```
jreij, line 62: Let's add an `id` to the edit text and use it directly?

mbway/src/main/java/com/adyen/checkout/mbway/DefaultMBWayDelegate.kt
```diff
@@ -61,11 +75,20 @@ internal class DefaultMBWayDelegate(val paymentMethod: PaymentMethod) : MBWayDel
         componentStateChanged(paymentComponentState)
     }
 
+    override fun getSupportedCountries(): List<CountryInfo> = CountryUtils.getCountries(SUPPORTED_COUNTRIES)
```
jreij, line 78: We have some redundant stuff in `MBWayComponent` now, like `getSupportedCountries` and the companion `val`s. Should we remove them from there? (maybe just use `emptyList()` in `MBWayView` where we call `getSupportedCountries`, since that view will be removed soon anyway.

jreij (approved):

PR #810, opened 2022-09-30
Bind issuer list views to delegate

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -397,18 +390,9 @@ internal fun getViewFor(
     @Suppress("UNCHECKED_CAST")
     return when (paymentType) {
         PaymentMethodTypes.BCMC -> BcmcView(context)
-        PaymentMethodTypes.DOTPAY -> DotpayRecyclerView(context)
```
OscarSpruit, line 400: Shouldn't we return `IssuerListRecyclerView` for these types? Or is that not useful since this method will probably disappear soon?

jreij: Yeah exactly, method will be removed soon

PR #815, opened 2022-10-05
Add textColor value in night colors

etman55:


example-app/src/main/res/values-night/colors.xml
```diff
@@ -10,7 +10,7 @@
     <color name="color_divider">#1AFFFFFF</color>
 
     <color name="text_color">@color/color_white</color>
-
+    <color name="textColor">@color/color_white</color>
```
OscarSpruit, line 13: This does not really fix the problem, but hides it. `@color/textColor` should not be referenced directly by styles and views. Instead they should use `?android:attr/textColor`

jreij: Correct me if I'm wrong but these styes are in the example app, the SDK styles still do not reference `@color/textColor` 🤔

OscarSpruit: `AdyenCheckout.SimpleTextItemTextView` and `AdyenCheckout.AddressForm.HeaderTextAppearance` are referencing `@color/textColor` and shouldn't

OscarSpruit (approved):

PR #816, opened 2022-10-05
Bind online banking view to delegate

jreij:


online-banking-core/src/main/java/com/adyen/checkout/onlinebankingcore/OnlineBankingComponent.kt
```diff
@@ -7,34 +7,34 @@
...
+import com.adyen.checkout.components.ui.view.ComponentViewType
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 
 abstract class OnlineBankingComponent<IssuerListPaymentMethodT : IssuerListPaymentMethod>(
     savedStateHandle: SavedStateHandle,
-    private val delegate: OnlineBankingDelegate<IssuerListPaymentMethodT>,
+    final override val delegate: OnlineBankingDelegate<IssuerListPaymentMethodT>,
```
OscarSpruit, line 24: It's final by default, right?

OscarSpruit: Ahh because it's an abstract class of course 👍

OscarSpruit (approved):

PR #818, opened 2022-10-06
ImproveCardListAdapter: Improve CardListAdapter

OscarSpruit:
- Extract filtering of detected cards out of the adapter
- Use a data class as the adapters items
- Wrap the card brand ImageView in a FrameLayout

This makes updating the list of items more efficient and enables some nice animations

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/CardListAdapter.kt
+++ b/card/src/main/java/com/adyen/checkout/card/CardListAdapter.kt
@@ -7,7 +7,6 @@
  */
 package com.adyen.checkout.card
 
-import android.annotation.SuppressLint
 import android.view.LayoutInflater
 import android.view.ViewGroup
 import androidx.recyclerview.widget.DiffUtil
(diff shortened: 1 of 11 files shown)


card/src/main/java/com/adyen/checkout/card/CardListAdapter.kt
```diff
@@ -7,47 +7,38 @@
...
-            .inflate(R.layout.brand_logo, parent, false) as RoundCornerImageView
-        return ImageViewHolder(imageView)
+        val binding = BrandLogoBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+        return ImageViewHolder(binding)
     }
 
     override fun onBindViewHolder(holder: ImageViewHolder, position: Int) {
         val card = currentList[position]
-        val alpha = if (filteredCards.isEmpty() || filteredCards.contains(card)) ACTIVE else NOT_ACTIVE
+        val alpha = if (card.isDetected) ACTIVE else NOT_ACTIVE
```
jreij, line 31: Can we rename `ACTIVE` and `NOT_ACTIVE` to something meaningful? 😆

card/src/main/java/com/adyen/checkout/card/CardView.kt
```diff
@@ -688,20 +687,20 @@ class CardView @JvmOverloads constructor(
...
-        binding.recyclerViewCardList.isVisible = cards.isNotEmpty()
+    private fun setCardList(originalCards: List<CardType>, detectedCards: List<DetectedCardType>) {
+        binding.recyclerViewCardList.isVisible = originalCards.isNotEmpty()
+
         if (cardListAdapter == null) {
             cardListAdapter = CardListAdapter(requireNotNull(imageLoader))
             binding.recyclerViewCardList.adapter = cardListAdapter
         }
 
+        val cards = originalCards.map { cardType ->
```
jreij, line 698: Is it possible to move this mapping to the delegate?

card/src/main/java/com/adyen/checkout/card/CardViewProvider.kt
```diff
@@ -23,7 +23,7 @@ internal object CardViewProvider : ViewProvider {
         defStyleAttr: Int
     ): ComponentViewNew {
         return when (viewType) {
-            CardComponentViewType -> CardViewNew(context, attrs, defStyleAttr)
+            CardComponentViewType -> CardViewNew(context, null, defStyleAttr)
```
jreij, line 26: Is this `null` by mistake?

OscarSpruit: Ohhh my bad, I didn't mean to commit this... I have to do this to be able to render the view, because of this weird bug with view attribute inspection :(

card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -595,6 +596,17 @@ class DefaultCardDelegate(
         return cardOutputData.installmentOptions.isNotEmpty()
     }
 
+    override fun getCardListItems(): List<CardListItem> {
```
jreij, line 599: Sorry I meant to do in it the output data not as a separate method. Also it seems that `supportedCardTypes` is only used for this adapter so maybe we can simply replace it with the mapped list? Does that work?

OscarSpruit: This `CardListItem` class is just a representation of how we display the data, so I'm not sure if we should expose it to merchants in the output data 🤔 WDYT?

jreij: Merchants can't see the output data anymore, it's strictly for the view now

OscarSpruit: Oh nice! Then I'll put it in there

jreij (dismissed):

jreij (approved):

PR #819, opened 2022-10-06
Clean up components

jreij:
* Remove `AdyenLinearLayout`
* Remove all old component views and remove the `New` prefix from the new views
* Remove old `ComponentView` and replace it with `ComponentViewNew`
* Make all component views internal
* Use the new `AdyenComponentView` in drop-in and example app
* Remove input data and output data from components


OscarSpruit (approved):
Really nice clean up!

PR #821, opened 2022-10-06
AddressFormConfig: Make card address form configurable

OscarSpruit:
Add option to enable address form for card component from the configuration menu

image


jreij (approved):
👏

PR #823, opened 2022-10-07
NightThemeConfig: Make theme (dark mode) configurable from the settings

OscarSpruit:
https://user-images.githubusercontent.com/17701279/194528646-e9b53546-1b92-4236-b349-6d6ef51467b7.mp4

### DIFF
--- a/example-app/src/main/res/values/strings.xml
+++ b/example-app/src/main/res/values/strings.xml
@@ -13,6 +13,7 @@
     <string name="merchant_account_key">merchant_account</string>
     <string name="merchant_account_title">Merchant Account</string>
     <string name="additional_data_title">Additional Data</string>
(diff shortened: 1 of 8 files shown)


example-app/src/main/res/values/strings.xml
```diff
@@ -13,6 +13,7 @@
     <string name="merchant_account_key">merchant_account</string>
     <string name="merchant_account_title">Merchant Account</string>
     <string name="additional_data_title">Additional Data</string>
+    <string name="app_title">App</string>
```
etman55, line 16: hmm it's not important, but can we change this `app_title` to something like `app_title_key` because it's confusing.

etman55 (approved):

PR #825, opened 2022-10-11
Convert remaining RecyclerView adapters to ListAdapter

OscarSpruit:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/giftcard/GiftCardPaymentConfirmationDialogFragment.kt
```diff
@@ -102,7 +102,9 @@ class GiftCardPaymentConfirmationDialogFragment : DropInBottomSheetDialogFragmen
         )
 
         binding.recyclerViewGiftCards.layoutManager = LinearLayoutManager(requireContext())
-        binding.recyclerViewGiftCards.adapter = PaymentMethodAdapter(paymentMethods, imageLoader)
+        binding.recyclerViewGiftCards.adapter = PaymentMethodAdapter(imageLoader).apply {
```
etman55, line 105: I guess it would be better to do it like this
```
        val paymentAdapter = PaymentMethodAdapter(imageLoader)
        binding.recyclerViewGiftCards.apply {
            layoutManager = LinearLayoutManager(requireContext())
            adapter = paymentAdapter
        }
        paymentAdapter.submitList(paymentMethods)
```

OscarSpruit: 👍 I moved setting the layout manager to xml, so that saves even a few more lines

etman55 (approved):

PR #829, opened 2022-10-14
BcmcCardHolder: Add card holder name input to bcmc

OscarSpruit:


bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcView.kt
```diff
@@ -111,6 +117,24 @@ internal class BcmcView @JvmOverloads constructor(
...
+        binding.editTextCardHolder.setOnChangeListener {
+            delegate.updateInputData { cardHolderName = binding.editTextCardHolder.rawValue }
+            binding.textInputLayoutCardHolder.hideError()
+        }
+
+        binding.editTextCardHolder.onFocusChangeListener = OnFocusChangeListener { _, hasFocus ->
+            val cardHolderValidation = delegate.outputData.cardHolderNameField.validation
+            if (hasFocus) {
+                binding.textInputLayoutCardHolder.hideError()
+            } else if (!cardHolderValidation.isValid()) {
```
ozgur00, line 131: we can use `cardHolderValidation is Validation.Invalid` so we won't have to cast it below. there are examples in `CardView`. I think this PR is a good place to refactor all the other places that uses `isValid()` for this purpose.

OscarSpruit: Good one! I indeed got this from `CardView`, so will change it everywhere I can find

ozgur00 (approved):

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcConfiguration.kt
```diff
@@ -98,6 +106,17 @@ class BcmcConfiguration : Configuration {
             return super.setEnvironment(environment) as Builder
         }
 
+        /**
+         * Set if the holder name is required and should be shown as an input field.
+         *
+         * @param isHolderNameRequired [Boolean]
+         * @return [CardConfiguration.Builder]
```
jreij, line 113: Should be `BcmcConfiguration.Builder`

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcView.kt
```diff
@@ -70,6 +72,10 @@ internal class BcmcView @JvmOverloads constructor(
                 R.style.AdyenCheckout_Card_ExpiryDateInput,
                 localizedContext
             )
+            binding.textInputLayoutCardHolder.setLocalizedHintFromStyle(
+                com.adyen.checkout.card.R.style.AdyenCheckout_Card_HolderNameInput,
```
jreij, line 76: Do we need the full `com.adyen.checkout.card.R` or can we just use `R`?

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -74,6 +74,7 @@ internal class CheckoutConfigurationProvider @Inject constructor(
     private fun getBcmcConfiguration(): BcmcConfiguration =
         BcmcConfiguration.Builder(shopperLocale, environment, clientKey)
             .setShopperReference(keyValueStorage.getShopperReference())
+            .setHolderNameRequired(true)
```
jreij, line 77: Maybe let's not enable it by default in the example app, to be consistent with card

PR #833, opened 2022-10-17
Fix stored card disabled fields text color

etman55:


ui-core/src/main/res/drawable/edit_text_color_selector.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
```
OscarSpruit, line 2: Few things that can be improved here:

1. This file is not really a drawable, but actually a color. It should be located in ui-core/src/main/res/color
2. A better name would be `text_color_primary`
3. The disabled color is not nicely customisable and it forces a certain color that might not match the actual text color used. I'd suggest doing something like this:
```XML
<selector xmlns:android="http://schemas.android.com/apk/res/android">
    <!-- 0.32 is the alpha value that material design uses for disabled text -->
    <item android:alpha="0.32" android:color="@color/textColor" android:state_enabled="false" />
    <item android:color="@color/textColor" />
</selector>
```

ui-core/src/main/res/values/styles.xml
```diff
@@ -72,6 +72,7 @@
         <item name="android:textDirection">ltr</item>
         <item name="android:textAlignment">viewStart</item>
         <item name="android:inputType">text</item>
+        <item name="android:textColor">@drawable/edit_text_color_selector</item>
```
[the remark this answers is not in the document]
etman55: Nice idea 👍

example-app/src/main/res/values/styles.xml
```diff
@@ -16,7 +16,7 @@
         <item name="colorControlNormal">@color/color_control_normal</item>
         <item name="colorControlHighlight">@color/color_control_highlight</item>
 
-        <item name="android:textColor">?attr/colorOnBackground</item>
+        <item name="android:textColor">@color/text_color_primary</item>
         <item name="android:textColorPrimary">?attr/colorOnBackground</item>
```
OscarSpruit, line 20: To make sure the correct text color is used everywhere we should also override `textColorPrimary`, `textColorSecondary` and `textColorTertiary` with `@color/text_color_primary`

OscarSpruit (approved):

PR #837, opened 2022-10-18
Refactor google pay component

jreij:
- Remove input and output data
- Move logic to delegate


mbway/src/main/java/com/adyen/checkout/mbway/MbWayView.kt
```diff
@@ -44,8 +44,6 @@ internal class MbWayView @JvmOverloads constructor(
     init {
         orientation = VERTICAL
 
-        LayoutInflater.from(context).inflate(R.layout.mbway_view, this, true)
```
OscarSpruit, line 47: Oh wow I'm amazed that didn't cause any issues

jreij: It did 😂 The view was duplicated

PR #840, opened 2022-10-19
Focus on next field when card number matches pan

OscarSpruit:
https://user-images.githubusercontent.com/17701279/196677887-d9458fbc-e28d-495c-ad99-9582099aded3.mp4


jreij (approved):

checkout-core/src/main/java/com/adyen/checkout/core/model/JsonUtils.kt
```diff
@@ -23,6 +23,10 @@ fun JSONObject.getBooleanOrNull(key: String): Boolean? {
     return if (has(key)) getBoolean(key) else null
 }
 
+fun JSONObject.getIntOrNull(key: String): Int? {
```
jreij, line 26: Interesting that we never needed this before 🤔

jreij (approved):

PR #841, opened 2022-10-19
Fix drop in not showing stored payment methods when payment methods list is empty

etman55:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -94,14 +97,13 @@ class PaymentMethodListDialogFragment :
     }
 
     private fun initObservers() {
-        paymentMethodsListViewModel.paymentMethodsLiveData.observe(
-            viewLifecycleOwner
-        ) { paymentMethods ->
-            Logger.d(TAG, "paymentMethods changed")
-            if (paymentMethods == null) {
-                throw CheckoutException("List of PaymentMethodModel is null.")
+        lifecycleScope.launch {
```
OscarSpruit, line 100: A cleaner way to do this:
```Kotlin
paymentMethodsListViewModel.paymentMethodsFlow
    .flowWithLifecycle(viewLifecycleOwner.lifecycle)
    .onEach { paymentMethods ->
        Logger.d(TAG, "paymentMethods changed")
        paymentMethodAdapter?.submitList(paymentMethods)
    }
    .launchIn(lifecycleScope)
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/stored/StoredUtils.kt
```diff
@@ -14,26 +14,27 @@ import com.adyen.checkout.dropin.ui.paymentmethods.GenericStoredModel
...
                 isRemovingEnabled,
-                storedPaymentMethod.lastFour.orEmpty(),
-                storedPaymentMethod.expiryMonth.orEmpty(),
-                storedPaymentMethod.expiryYear.orEmpty()
+                lastFour.orEmpty(),
+                expiryMonth.orEmpty(),
+                expiryYear.orEmpty()
+            )
+        }
+        else -> with(this) {
```
OscarSpruit, line 31: `this` is already in scope, so why would you rescope to the same thing?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListItem
+import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodModel
+import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodNote
+import com.adyen.checkout.dropin.ui.paymentmethods.StoredPaymentMethodModel
+import com.adyen.checkout.dropin.ui.stored.mapStoredModel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+@Suppress("TooManyFunctions")
+internal class PaymentMethodsListDelegate constructor(
```
OscarSpruit, line 37: The concept of delegates was only introduced for components and this is not a component. This delegate just adds an extra layer without a real purpose. I think it's better to have all this logic in `PaymentMethodsListViewModel`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+                order?.paymentMethods?.mapToGiftCardPaymentMethodModel().orEmpty()
+            if (giftCardsList.isNotEmpty()) {
+                add(PaymentMethodHeader(PaymentMethodHeader.TYPE_GIFT_CARD_HEADER))
+                addAll(giftCardsList)
+            }
+            // payment notes
+            order?.remainingAmount?.let { remainingAmount ->
+                val value = CurrencyUtils.formatAmount(remainingAmount, dropInConfiguration.shopperLocale)
+                add(
+                    PaymentMethodNote(application.getString(R.string.checkout_giftcard_pay_remaining_amount, value))
```
OscarSpruit, line 107: It's a bad practice to use context outside of view related classes. It would be better to pass to string id int and let the fragment resolve it

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModel.kt
```diff
@@ -8,227 +8,31 @@
...
-
-    private var availabilitySum = 0
-    private var availabilitySkipSum = 0
-    private var availabilityChecksum = 0
-
-    private val storedPaymentMethodsList = mutableListOf<StoredPaymentMethodModel>()
-    private val paymentMethodsList = mutableListOf<PaymentMethodModel>()
-    private val orderPaymentMethodsList: List<GiftCardPaymentMethodModel> = setupOrderPaymentMethods(order)
+    private val paymentMethodsMutableFlow = MutableStateFlow<List<PaymentMethodListItem>>(emptyList())
+    internal val paymentMethodsFlow: StateFlow<List<PaymentMethodListItem>> = paymentMethodsMutableFlow
```
OscarSpruit, line 23: Why use this structure? It's better to directly pass the delegate's flow like this:
```Kotlin
val paymentMethodsFlow: Flow<List<PaymentMethodListItem>> = delegate.paymentMethodsFlow
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
etman55: the problem is we still need to call context for this:
```
checkPaymentMethodAvailability(application, paymentMethod, dropInConfiguration, amount, this)
```

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+    private val order: OrderModel?,
+    private val dropInConfiguration: DropInConfiguration,
+    private val amount: Amount
+) : ComponentAvailableCallback<Configuration> {
+
+    private val _paymentMethodsFlow = MutableStateFlow<List<PaymentMethodListItem>>(emptyList())
+    internal val paymentMethodsFlow: StateFlow<List<PaymentMethodListItem>> = _paymentMethodsFlow
+
+    private lateinit var storedPaymentMethodsList: MutableList<StoredPaymentMethodModel>
+    private var checkPaymentMethodsMap: HashMap<PaymentMethod, Boolean> = hashMapOf()
```
jreij, line 50: Maybe rename this to `checkedPaymentMethodsMap`?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
OscarSpruit: Ah didn't see that, that's unfortunate

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
etman55: yeah good point we don't want to mistake viewModels with components

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+                    PaymentMethodHeader.TYPE_REGULAR_HEADER_WITH_STORED
+                add(PaymentMethodHeader(headerType))
+                addAll(paymentMethodsList)
+            }
+        }
+        _paymentMethodsFlow.tryEmit(paymentMethods)
+    }
+
+    internal fun removePaymentMethodWithId(id: String) {
+        if (this::storedPaymentMethodsList.isInitialized) {
```
jreij, line 132: I'm not a big fan of this approach, it's the same as having a nullable list and checking whether it's null or not, except that now someone might call `storedPaymentMethodsList` directly somewhere else and cause a crash.
Maybe make the list nullable?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -149,7 +146,7 @@ class PaymentMethodListDialogFragment :
             }
             PaymentMethodTypes.SUPPORTED_PAYMENT_METHODS.contains(paymentMethod.type) -> {
                 Logger.d(TAG, "onPaymentMethodSelected: payment method is supported")
-                protocol.showComponentDialog(paymentMethodsListViewModel.getPaymentMethod(paymentMethod))
+                protocol.showComponentDialog(paymentMethodsListViewModel.getPaymentMethod(paymentMethod.index))
```
jreij, line 149: Why do we need to pass the index here instead of the whole `paymentMethod`?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+            storedPaymentMethodsList.remove(item)
+            populatePaymentMethods()
+        } else throw IllegalStateException("store payment methods not found")
+    }
+
+    private fun List<PaymentMethod>.setupPaymentMethodsList(): List<PaymentMethodModel> =
+        this.mapIndexedNotNull { index, paymentMethod ->
+            val isAvailable = checkPaymentMethodsMap[paymentMethod]
+                ?: throw IllegalStateException("payment method not found in map")
+            if (isAvailable) {
```
jreij, line 143: Minor: remove brackets from the if or add them to the else

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+    internal fun removePaymentMethodWithId(id: String) {
+        if (this::storedPaymentMethodsList.isInitialized) {
+            val item = storedPaymentMethodsList.first { item -> item.id == id }
+            storedPaymentMethodsList.remove(item)
+            populatePaymentMethods()
+        } else throw IllegalStateException("store payment methods not found")
+    }
+
+    private fun List<PaymentMethod>.setupPaymentMethodsList(): List<PaymentMethodModel> =
+        this.mapIndexedNotNull { index, paymentMethod ->
```
jreij, line 140: You don't need to explicitly call `this`, you can remove it. Same for most other usages of `this` in this class.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+
+    internal fun removePaymentMethodWithId(id: String) {
+        if (this::storedPaymentMethodsList.isInitialized) {
+            val item = storedPaymentMethodsList.first { item -> item.id == id }
+            storedPaymentMethodsList.remove(item)
+            populatePaymentMethods()
+        } else throw IllegalStateException("store payment methods not found")
+    }
+
+    private fun List<PaymentMethod>.setupPaymentMethodsList(): List<PaymentMethodModel> =
```
jreij, line 139: Can be called `mapTo...` for consistency

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+        this.mapIndexedNotNull { index, paymentMethod ->
+            val isAvailable = checkPaymentMethodsMap[paymentMethod]
+                ?: throw IllegalStateException("payment method not found in map")
+            if (isAvailable) {
+                paymentMethod.mapToModel(index)
+            } else
+                null
+        }
+
+    private fun List<StoredPaymentMethod>.setupStoredPaymentMethods(): List<StoredPaymentMethodModel> =
```
jreij, line 149: Can be called `mapTo...` for consistency

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
```diff
@@ -0,0 +1,198 @@
...
+        this.mapNotNull { storedPaymentMethod ->
+            if (storedPaymentMethod.isStoredPaymentSupported())
+                storedPaymentMethod.mapStoredModel(dropInConfiguration.isRemovingStoredPaymentMethodsEnabled)
+            else
+                null
+        }
+
+    private fun StoredPaymentMethod.isStoredPaymentSupported(): Boolean {
```
jreij, line 157: This can be moved to `StoredUtils`, and used in other places as well, check out `DropInViewModel.shouldShowPreselectedStored` and `DropInViewModel.getPreselectedStoredPaymentMethod`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
etman55: we don't need the whole payment method we just need the index to get the corresponding paymentMethod

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListDelegate.kt
etman55: nice idea

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
jreij: Yes but shouldn't the view model decide which part of the `paymentMethod` it needs, instead of the fragment?

etman55: tbh I hate the way this method is done 😅, but we can remove this thing and replace it with something like that 
```
protocol.showComponentDialog(dropInViewModel.getPaymentMethods()[paymentMethod.index])
```
and remove `getPaymentMethod()` from the view model

jreij: By using `dropInViewModel.getPaymentMethods()[paymentMethod.index]` you are assuming that:
- `dropInViewModel.paymentMethods` is the same as `paymentMethodsListViewModel.paymentMethods`
- The `index` inside `PaymentMethodModel` matches the index of the corresponding payment method inside the list in the view model

While both these points are true now, any change we do to the logic in either view models might break this flow. Because currently the view model is the one that maps `PaymentMethod`s to `PaymentMethodModel`s, so if the fragment calls `dropInViewModel.getPaymentMethods()[paymentMethod.index]` it's basically mapping the `PaymentMethodModel` back to a `PaymentMethod` itself which can easily cause bugs.

WDYT?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModel.kt
```diff
@@ -32,105 +30,132 @@ import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListItem
...
-import com.adyen.checkout.dropin.ui.stored.makeStoredModel
+import com.adyen.checkout.dropin.ui.stored.isStoredPaymentSupported
+import com.adyen.checkout.dropin.ui.stored.mapStoredModel
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 
 @Suppress("TooManyFunctions")
-class PaymentMethodsListViewModel(
-    application: Application,
+internal class PaymentMethodsListViewModel constructor(
```
jreij, line 39: `constructor` is not explicitly needed right?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModel.kt
```diff
@@ -32,105 +30,132 @@ import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListItem
...
+            val isAvailable = paymentMethodsAvailabilityMap[paymentMethod]
+                ?: throw IllegalStateException("payment method not found in map")
+            if (isAvailable)
+                paymentMethod.mapToModel(index)
+            else
+                null
+        }
+
+    private fun List<StoredPaymentMethod>.mapToStoredPaymentMethodsModelList(): List<StoredPaymentMethodModel> =
+        this.mapNotNull { storedPaymentMethod ->
```
jreij, line 152: Explicit this not needed

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModel.kt
```diff
@@ -148,64 +173,8 @@ class PaymentMethodsListViewModel(
...
-                add(PaymentMethodHeader(headerType))
-                addAll(paymentMethodsList)
-            }
-        }
-    }
-
-    private fun setupOrderPaymentMethods(order: OrderModel?): List<GiftCardPaymentMethodModel> {
-        return order?.paymentMethods.orEmpty().map {
+    private fun List<OrderPaymentMethod>.mapToGiftCardPaymentMethodModel(): List<GiftCardPaymentMethodModel> =
+        this.map {
```
jreij, line 177: Explicit this not needed

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModel.kt
etman55: no it's not

drop-in/src/test/java/com/adyen/checkout/util/ConfigurationProvider.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.adyen.checkout.components.model.payments.Amount
+import com.adyen.checkout.core.api.Environment
+import com.adyen.checkout.core.exception.CheckoutException
+import com.adyen.checkout.core.log.LogUtil
+import com.adyen.checkout.core.log.Logger
+import com.adyen.checkout.dropin.DropInConfiguration
+import com.adyen.checkout.googlepay.GooglePayConfiguration
+import java.util.Locale
+
+object ConfigurationProvider {
```
jreij, line 22: Minor: can be `internal`

drop-in/src/test/java/com/adyen/checkout/util/DataProvider.kt
```diff
@@ -0,0 +1,44 @@
...
+package com.adyen.checkout.util
+
+import com.adyen.checkout.components.model.PaymentMethodsApiResponse
+import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
+import com.adyen.checkout.components.model.paymentmethods.StoredPaymentMethod
+import org.json.JSONException
+import org.json.JSONObject
+import java.io.IOException
+
+object DataProvider {
```
jreij, line 18: Minor: can be `internal`

drop-in/src/test/java/com/adyen/checkout/util/Helpers.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
+import com.adyen.checkout.components.model.paymentmethods.StoredPaymentMethod
+import com.adyen.checkout.components.util.PaymentMethodTypes
+import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodModel
+import com.adyen.checkout.dropin.ui.paymentmethods.StoredPaymentMethodModel
+import com.adyen.checkout.dropin.ui.stored.isStoredPaymentSupported
+import com.adyen.checkout.dropin.ui.stored.mapStoredModel
+
+object Helpers {
```
jreij, line 19: Minor: can be `internal`

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    @Test
+    fun `test get payment method from payment method model success`() {
+        val paymentMethodModelList = paymentMethods.mapToPaymentMethodModelList()
+
+        val paymentMethod = viewModel.getPaymentMethod(paymentMethodModelList[0])
+
+        assertTrue(paymentMethod == paymentMethods[0])
```
jreij, line 91: You can use `assertEquals(paymentMethod, paymentMethods[0])`

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+        fun `has payment methods`() = runTest {
+            viewModel.paymentMethodsFlow.test {
+                with(expectMostRecentItem()) {
+                    assertTrue(filterIsInstance<PaymentMethodModel>().isNotEmpty())
+                }
+            }
+        }
+
+        @Test
+        fun `has payment method header `() = runTest {
```
jreij, line 117: Minor: unnecessary space after `header`

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+            viewModel.paymentMethodsFlow.test {
+                with(expectMostRecentItem()) {
+                    assertTrue(filterIsInstance<PaymentMethodHeader>().isNotEmpty())
+                }
+            }
+        }
+
+        @Test
+        fun `doesn't have gift card payment method model `() = runTest {
+            viewModel.paymentMethodsFlow.test {
```
jreij, line 127: Minor: unnecessary space after `model`

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+import org.mockito.junit.jupiter.MockitoExtension
+
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(MockitoExtension::class, TestDispatcherExtension::class)
+internal class PaymentMethodsListViewModelTest {
```
jreij, line 41: Good tests overall 👍 Can we also test some more scenarios, especially the ones that caused us to start this fix? Like passing only `paymentMethods` or only `storedPaymentMethods` or a non null `order`, etc?

etman55: they are already add in the inner class `PaymentMethodFLow`

jreij: Oops sorry my local version was out of date 😅

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+                with(expectMostRecentItem()) {
+                    assertTrue(filterIsInstance<GiftCardPaymentMethodModel>().isEmpty())
+                }
+            }
+        }
+    }
+
+    @Nested
+    @DisplayName("test payment method flow when")
+    inner class PaymentMethodFLow {
```
jreij, line 137: Typo in `FLow`, should be `Flow`

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+            viewModel.paymentMethodsFlow.test {
+                with(expectMostRecentItem()) {
+                    assertTrue(filterIsInstance<GiftCardPaymentMethodModel>().isEmpty())
+                }
+            }
+        }
+    }
+
+    @Nested
+    @DisplayName("test payment method flow when")
```
jreij, line 136: I don't really see the point in these 2 inner classes specifically, they don't seem like they are different as the tested scenarios are all around the `paymentMethodsFlow` and their titles are also not very descriptive.

I suggest if you can find a better way to group them or just remove the inner classes completely.

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
jreij: Then one more test with a non null `order` would make this 💯 🙏

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
etman55: hmmm yeah I see now inner class `PaymentMethodsType` can be combined in one test case 😅

drop-in/src/test/java/com/adyen/checkout/dropin/ui/viewmodel/PaymentMethodsListViewModelTest.kt
jreij: This is still not fixed btw

jreij (approved):

PR #851, opened 2022-10-25
Open Banking Component

ozgur00:


paybybank/src/main/java/com/adyen/checkout/paybybank/PayByBankViewProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+
+package com.adyen.checkout.paybybank
+
+import android.content.Context
+import android.util.AttributeSet
+import com.adyen.checkout.components.ui.ComponentView
+import com.adyen.checkout.components.ui.ViewProvider
+import com.adyen.checkout.components.ui.view.ComponentViewType
+
+object PayByBankViewProvider: ViewProvider {
```
OscarSpruit, line 17: Can be `internal`

paybybank/src/main/java/com/adyen/checkout/paybybank/DefaultPayByBankDelegate.kt
```diff
@@ -0,0 +1,135 @@
...
+            } else {
+                null
+            }
+        }
+
+    private fun List<InputDetail>?.getLegacyIssuers(): List<IssuerModel> =
+        this.orEmpty()
+            .flatMap { it.items.orEmpty() }
+            .mapNotNull { item ->
+                val (id, name) = item
```
OscarSpruit, line 128: I believe you can already do destructuring in the line above:
```Kotlin
.mapNotNull { (id, name) -> 
```

paybybank/src/main/java/com/adyen/checkout/paybybank/DefaultPayByBankDelegate.kt
```diff
@@ -0,0 +1,135 @@
...
+        )
+    }
+
+    override fun getIssuers(): List<IssuerModel> {
+        return paymentMethod.issuers?.mapToModel() ?: paymentMethod.details.getLegacyIssuers()
+    }
+
+    private fun List<Issuer>.mapToModel(): List<IssuerModel> =
+        this.mapNotNull { issuer ->
+            val (id, name, isDisabled) = issuer
```
OscarSpruit, line 116: Same here

paybybank/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
OscarSpruit, line 1: We don't really need these files if there are no rules in them. For now we can keep them for consistency, but it looks a bit cleaner if all the comments here are removed

paybybank/src/main/java/com/adyen/checkout/paybybank/DefaultPayByBankDelegate.kt
ozgur00: Tried it but didn't work.

OscarSpruit (approved):

PR #857, opened 2022-11-01
Instantiate Components by a key

ozgur00:


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ComponentProvider.kt
```diff
@@ -35,17 +36,19 @@ class Adyen3DS2ComponentProvider :
     override fun <T> get(
         owner: T,
         application: Application,
-        configuration: Adyen3DS2Configuration
+        configuration: Adyen3DS2Configuration,
+        key: String?,
```
jreij, line 40: Do we need `key: String? = null` here as well or no?

components-core/src/main/java/com/adyen/checkout/components/base/lifecycle/ViewModelExt.kt
```diff
@@ -38,3 +38,11 @@ inline fun <reified ViewModelT : ViewModel> viewModelFactory(
         }
     }
 }
+
+fun <T : ViewModel> getWithKey(provider: ViewModelProvider, key: String?, modelClass: Class<T>): T {
```
jreij, line 42: This could be `fun <T : ViewModel> ViewModelProvider.get(key: String?, modelClass: Class<T>) : T`, it will look just like the default methods provided by androidx while also not clashing with their signatures. WDYT?

ozgur00: good idea

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ComponentProvider.kt
ozgur00: It's not possible because you can only set default value in the interface

jreij (approved):

PR #867, opened 2022-11-09
OptionalAddressFields: Update translations v4

ozgur00:


card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
```diff
@@ -396,29 +376,38 @@ class AddressFormInput @JvmOverloads constructor(
...
-        val streetStyleResId = if (isOptional) spec.street.optionalStyleResId else spec.street.styleResId
-        textInputLayoutStreet?.setLocalizedHintFromStyle(streetStyleResId, localizedContext)
+        val streetStyleResId = spec.street.getStyleResId(isOptional)
+        streetStyleResId?.let {
+            textInputLayoutStreet?.setLocalizedHintFromStyle(it, localizedContext)
+        }
 
-        val houseNumberStyleResId = if (isOptional) spec.houseNumber.optionalStyleResId else spec.houseNumber.styleResId
-        textInputLayoutHouseNumber?.setLocalizedHintFromStyle(houseNumberStyleResId, localizedContext)
+        val houseNumberStyleResId = spec.houseNumber.getStyleResId(isOptional)
```
jreij, line 384: Minor: you could directly use `spec.houseNumber.getStyleResId(isOptional)?.let{` to reduce one line

ozgur00: I like it better this way because it separates the two operations we're doing, so slightly better readability.

jreij: 👍

PR #871, opened 2022-11-10
Make drop-in classes that shouldn't be public internal

jreij:


drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
```diff
@@ -93,7 +93,7 @@ internal const val GOOGLE_PAY_REQUEST_CODE = 1
  * Activity that presents the available PaymentMethods to the Shopper.
  */
 @Suppress("TooManyFunctions")
-class DropInActivity :
+internal class DropInActivity :
```
OscarSpruit, line 96: Shouldn't activities and fragments always be public?

etman55 (dismissed):

etman55 (approved):

drop-in/src/main/java/com/adyen/checkout/dropin/ui/DropInActivity.kt
jreij: Not in Kotlin apparently

PR #872, opened 2022-11-10
Provide public API for session creation

jreij:
This PR is an attempt at providing a public API allowing merchants to create and setup a session, which they can pass to drop-in/components.

### DIFF
--- a/sessions-core/src/main/java/com/adyen/checkout/sessions/provider/CheckoutSessionProvider.kt
+++ b/sessions-core/src/main/java/com/adyen/checkout/sessions/provider/CheckoutSessionProvider.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 9/11/2022.
+ */
+
+package com.adyen.checkout.sessions.provider
+
+import com.adyen.checkout.components.base.Configuration
+import com.adyen.checkout.sessions.model.SessionModel
+import kotlinx.coroutines.CoroutineScope
+
+// TODO docs
+object CheckoutSessionProvider {
+    // TODO docs
+    suspend fun createSession(
+        sessionModel: SessionModel,
+        configuration: Configuration,
+    ): CheckoutSessionResult {
+        return CheckoutSessionInitializer(sessionModel, configuration).setupSession()
+    }
+
+    // TODO docs
+    // callback alternative for the suspend function
+    fun createSession(
+        coroutineScope: CoroutineScope,
+        sessionModel: SessionModel,
+        configuration: Configuration,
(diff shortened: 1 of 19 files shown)


sessions-core/src/main/java/com/adyen/checkout/sessions/provider/CheckoutSessionProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+    suspend fun createSession(
+        sessionModel: SessionModel,
+        configuration: Configuration,
+    ): Result<CheckoutSession> {
+        return CheckoutSessionInitializer(sessionModel, configuration).setupSession()
+    }
+
+    // TODO docs
+    // callback alternative for the suspend function
+    fun createSession(
```
OscarSpruit, line 28: If we are giving a callback alternative it should also be without `CoroutineScope` IMO. Now it's basically the same thing, but we are calling 'launch'. WDYT?

sessions-core/src/main/java/com/adyen/checkout/sessions/provider/CheckoutSessionProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.adyen.checkout.sessions.model.SessionModel
+import kotlinx.coroutines.CoroutineScope
+
+// TODO docs
+object CheckoutSessionProvider {
+    // TODO docs
+    suspend fun createSession(
+        sessionModel: SessionModel,
+        configuration: Configuration,
+    ): Result<CheckoutSession> {
```
OscarSpruit, line 22: Not totally sure if we should expose `Result` publicly 🤔 Some people really dislike it

OscarSpruit (commented):
I like the idea, think we should do it like this

sessions-core/src/main/java/com/adyen/checkout/sessions/provider/CheckoutSessionProvider.kt
jreij: Yeah i had a sealed class first but I removed it for some reason, will add it back 👍

sessions-core/src/main/java/com/adyen/checkout/sessions/provider/CheckoutSessionProvider.kt
jreij: How do you suggest we do it without using any `CoroutineScope`?

OscarSpruit: With `ThreadManager` 😂 No, let's not do that. To be honest I'm not sure what the best way of doing this is

jreij: 😂 ☠️

OscarSpruit (approved):

PR #873, opened 2022-11-10
PayNow component

etman55:
## Description
New feature `PayNow`, QR based payment method

## Checklist 
- [x] Code is unit tested

COAND-630


checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 16/11/2022.
+ */
+
+package com.adyen.checkout.core.exception
+
+class PermissionException constructor(
```
jreij, line 11: Unnecessary `constructor` call

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
```diff
@@ -0,0 +1,14 @@
...
+    errorMessage: String,
+    val askedPermission: String
```
jreij, line 13: Maybe `requiredPermission` or something could be clearer here? Also just a small doc on top of the class to explains what is this exception would be nice.

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
```diff
@@ -0,0 +1,112 @@
...
+                    output?.let {
+                        input.copyTo(it, DEFAULT_BUFFER_SIZE)
+                        withContext(Dispatchers.Main) { notifyState() }
+                    }
+                }
+            }
+        }
+    }
+
+    private fun String.toULR(): URL? {
```
jreij, line 100: Typo, should be `toURL`

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    private suspend fun downloadAndSaveApi28AndBelow(
+        context: Context,
+        url: URL,
+        fileName: String,
+        filePath: String,
+        notifyState: () -> Unit
```
jreij, line 55: Instead of callbacks, have you tried using a `suspendCoroutine` which allows transforming a callback into a coroutine? [Here's an example](https://stackoverflow.com/questions/48552925/existing-3-function-callback-to-kotlin-coroutines/48562175#48562175)

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
```diff
@@ -0,0 +1,112 @@
...
+            put(MediaStore.MediaColumns.RELATIVE_PATH, filePath)
+        }
+        val resolver = context.contentResolver
+        val uri = resolver.insert(MediaStore.Downloads.EXTERNAL_CONTENT_URI, contentValues)
+        if (uri != null) {
+            url.openStream().use { input ->
+                resolver.openOutputStream(uri).use { output ->
+                    output?.let {
+                        input.copyTo(it, DEFAULT_BUFFER_SIZE)
+                        withContext(Dispatchers.Main) { notifyState() }
```
jreij, line 93: Also don't forget to always return a value, either an exception or a successful result. For example, `if (uri == null)` and `if(output == null)` are both completely ignored

components-core/src/main/java/com/adyen/checkout/components/util/PaymentMethodTypes.kt
```diff
@@ -39,6 +39,7 @@ object PaymentMethodTypes {
     const val ONLINE_BANKING_PL = "onlineBanking_PL"
     const val ONLINE_BANKING_SK = "onlineBanking_SK"
     const val PAY_BY_BANK = "paybybank"
+    const val PAY_NOW = "paynow"
```
jreij, line 42: I believe `PAY_NOW` behaves like `PIX`, so it should be somewhere else in this file, check out `PIX`

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -156,6 +167,11 @@ internal class ActionComponentDialogFragment : DropInBottomSheetDialogFragment()
                 Logger.d(TAG, "Flow was cancelled by user")
                 onBackPressed()
             }
+            is PermissionException -> {
+                val askedPermission = (componentError.exception as PermissionException).askedPermission
```
jreij, line 171: This cast shouldn't be needed is you use `when (val exception = componentError.exception) {`

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
etman55: nice catch 👍

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -198,7 +198,7 @@ internal class DefaultQRCodeDelegateTest(
                     Activity(),
                 )
 
-                assertEquals(awaitItem(), QrCodeComponentViewType.QR_CODE)
+                assertEquals(awaitItem(), QrCodeComponentViewType.SIMPLE_QR_CODE)
```
jreij, line 201: Can you add a similar test to this but passing `PAY_NOW` and asserting `FULL_QR_CODE`?

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,212 @@
...
+        updateAmount(delegate.componentParams)
+    }
+
+    private fun updateAmount(componentParams: ComponentParams) {
+        if (!componentParams.amount.isEmpty) {
+            val formattedAmount = CurrencyUtils.formatAmount(
+                componentParams.amount,
+                componentParams.shopperLocale
+            )
+            binding.textviewAmount.visibility = View.VISIBLE
```
jreij, line 122: Should use `isVisible = true`

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,212 @@
...
+            binding.textviewAmount.text = formattedAmount
+        } else {
+            binding.textviewAmount.visibility = View.GONE
+        }
```
jreij, line 126: Should use `isVisible = false`

drop-in/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,10 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="com.adyen.checkout.dropin">
 
+    <uses-permission
```
jreij, line 6: Shouldn't this be in the manifest of `qr-code`?

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    override fun getView(): View = this
+
+    override fun highlightValidationErrors() = Unit
+
+    companion object {
+        private val TAG = LogUtil.getTag()
+        private const val QR_IMAGE_BASE_PATH = "barcode.shtml?barcodeType=qrCode&fileType=png&data="
```
jreij, line 207: This REALLY should not be in the view, same as file name, mime type and `android.os.Environment.DIRECTORY_DOWNLOADS` 😬

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+        binding.textViewTimer.text = localizedContext.getString(
+            R.string.checkout_qr_code_pay_now_timer_text,
+            minutesSecondsString
+        )
+        binding.progressIndicator.progress = timerData.progress
+    }
+
+    @SuppressLint("MissingPermission")
```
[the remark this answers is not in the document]
etman55: if I tried to not suppress here it will ask me to request the permission and we can't request the permission inside the view.

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
etman55: what's the benefit of using `suspendCoroutine` here?

jreij: Your method is a suspend function but still uses a callback. One main benefit from coroutines and suspend functions is to [replace callbacks](https://levelup.gitconnected.com/avoid-callback-hell-with-kotlin-coroutines-7a72a2a8a701). With `suspendCoroutine` you can remove your callback and return a value with your method instead.

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
etman55: what's the difference?

jreij: Just for consistency, we use `isVisible` everywhere

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
```diff
@@ -0,0 +1,115 @@
...
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    suspend fun download(
+        context: Context,
+        stringUrl: String,
+        fileName: String,
+        filePath: String,
+        mimeType: String? = null
+    ): Result<Unit> {
+        val url = stringUrl.toULR() ?: return Result.failure(CheckoutException("Malformed URL"))
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
```
OscarSpruit, line 42: These two can actually be combined into one method making the class less verbose
```Kotlin
class FileDownloader {

    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
    suspend fun download(
        context: Context,
        stringUrl: String,
        fileName: String,
        filePath: String,
        mimeType: String? = null
    ): Result<Unit> = withContext(Dispatchers.IO) {
        val resolver = context.contentResolver

        val values = ContentValues().apply {
            put(MediaStore.MediaColumns.DISPLAY_NAME, fileName)
            put(MediaStore.MediaColumns.MIME_TYPE, mimeType)
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
                put(MediaStore.MediaColumns.RELATIVE_PATH, filePath)
            } else {
                put(MediaStore.MediaColumns.DATA, "${Environment.getExternalStorageDirectory()}/$filePath/$fileName")
            }
        }

        val target = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
            MediaStore.Downloads.EXTERNAL_CONTENT_URI
        } else {
            MediaStore.Files.getContentUri("external")
        }

        val uri = resolver.insert(target, values) ?: return@withContext Result.failure(CheckoutException("URI is null"))

        val url = stringUrl.toURL() ?: return@withContext Result.failure(CheckoutException("Malformed URL"))

        url.openStream().use { input ->
            resolver.openOutputStream(uri).use { output ->
                if (output == null) return@withContext Result.failure(CheckoutException("out is null"))
                input.copyTo(output, DEFAULT_BUFFER_SIZE)
                return@withContext Result.success(Unit)
            }
        }
    }

    private fun String.toURL(): URL? {
        return try {
            URL(this)
        } catch (e: MalformedURLException) {
            Logger.e(TAG, "toURL: $e")
            null
        }
    }

    companion object {
    }
}
```

OscarSpruit:
The button seems to be cut off. Can you check why this happens? 
image

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
jreij: What i meant by docs was docs for the merchants, to understand what this exception means, not internal docs for us.

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -181,7 +185,27 @@ internal class DefaultQRCodeDelegate(
 
     private fun createOutputData(statusResponse: StatusResponse?, action: QrCodeAction) {
         val isValid = statusResponse != null && StatusResponseUtils.isFinalResult(statusResponse)
-        val outputData = QRCodeOutputData(isValid, action.paymentMethodType, action.qrCodeData)
+
+        var qrParams: QRParams? = null
+        if (_viewFlow.value == QrCodeComponentViewType.FULL_QR_CODE) {
+            val date = System.currentTimeMillis()
+            val fileName = "qr-code_$date.png"
```
jreij, line 192: The whole file name should be constructor in the delegate

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,215 @@
...
+            R.string.checkout_qr_code_pay_now_timer_text,
+            minutesSecondsString
+        )
+        binding.progressIndicator.progress = timerData.progress
+    }
+
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    private fun saveQrImage(outputData: QRCodeOutputData) {
+        val baseUrl = delegate.componentParams.environment.baseUrl
+        val imageUrl = baseUrl + outputData.qrParams?.qrImagePath
```
jreij, line 185: The whole imageUrl should be constructed in the delegate not here

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
etman55: let me check again as when I tried something similar to this it wasn't working on below api 28.

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
etman55: ohh sorry about that!

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ *
+ * this exception indicates that the required runtime permission is not granted.
+ */
+class PermissionException(
+    errorMessage: String,
+    val requiredPermission: String
+) : CheckoutException(errorMessage)
+
+fun String.formatPermissionName(): String {
```
onurkaragunlu, line 20: I think that we should create a file called StringExtension and we add this extension there.

etman55: I think we don't need that it's better to be in the same file related as it's very specific to `PermissionException`

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
```diff
@@ -0,0 +1,22 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 16/11/2022.
+ */
+
+package com.adyen.checkout.core.exception
+
+/**
+ *
+ * this exception indicates that the required runtime permission is not granted.
```
jreij, line 13: `This` not `this`

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
jreij: I agree with Onur, the `PermissionException` is a merchant facing class, and `String.formatPermissionName()` is an internal method that they don't need to know about or use. I'd rather have it somewhere else.

But I also don't like having it as a string extension, the reason is that whenever we have any string in the code base, we will be able to apply `formatPermissionName()` on it, while in fact this method 99% of the time will not be applicable.

components-core/src/main/java/com/adyen/checkout/components/util/PaymentMethodTypes.kt
jreij: This is still not fixed

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -249,14 +269,36 @@ internal class DefaultQRCodeDelegate(
         qrCodeData = null
     )
 
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    override fun downloadQRImage(onSuccess: () -> Unit) {
+        val date: Long = System.currentTimeMillis()
+        val imageName = "paynow-qr-code_$date.png"
```
jreij, line 275: Try not to use `-` and `_` in the same file name, just for consistency

jreij: Also this should be a constant in the companion

checkout-core/src/main/java/com/adyen/checkout/core/exception/PermissionException.kt
etman55: Sorry I didn't get back on that after talking with Arjen about this part we decided we won't be doing this as we won't show the permission name so I'll push an updated about this when the translations are finished.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -151,11 +161,17 @@ internal class ActionComponentDialogFragment : DropInBottomSheetDialogFragment()
...
-        when (componentError.exception) {
+        when (val exception = componentError.exception) {
             is CancellationException -> {
                 Logger.d(TAG, "Flow was cancelled by user")
                 onBackPressed()
             }
+            is PermissionException -> {
+                val requiredPermission = exception.requiredPermission
+                Logger.d(TAG, "$requiredPermission permission not granted")
+                exception.message?.let { requireContext().toast(it) }
```
jreij, line 172: The toast message here should probably be a dialog, with `requestPermissionLauncher.launch(requiredPermission)` called in `OnDismissListener`. And the message should not be coming from `exception.message` but from the strings.

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeDelegate.kt
```diff
@@ -21,4 +21,7 @@ interface QRCodeDelegate :
     ViewableDelegate<QRCodeOutputData>,
     IntentHandlingDelegate,
     StatusPollingDelegate,
-    ViewProvidingDelegate
+    ViewProvidingDelegate {
+
+    fun downloadQRImage(onSuccess: () -> Unit)
```
jreij, line 26: Why are we still using a callback here?

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -249,14 +269,36 @@ internal class DefaultQRCodeDelegate(
         qrCodeData = null
     )
 
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    override fun downloadQRImage(onSuccess: () -> Unit) {
```
jreij, line 273: This method should be tested

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
etman55: I talked with Arjen about that and we will just leave the storage native dialog and will remove the toast.

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,190 @@
...
+            R.string.checkout_qr_code_pay_now_timer_text,
+            minutesSecondsString
+        )
+        binding.progressIndicator.progress = timerData.progress
+    }
+
+    @RequiresPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+    private fun saveQrImage(coroutineScope: CoroutineScope) {
+        coroutineScope.launch {
+            delegate.downloadQRImage().fold(
```
jreij, line 176: We shouldn't call suspend methods like this from the UI

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -265,6 +293,17 @@ internal class DefaultQRCodeDelegateTest(
         }
     }
 
+    @Test
+    fun `download qr image successfully`(): Unit = runBlocking {
```
jreij, line 297: Method name should be a bit more descriptive

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -198,7 +210,23 @@ internal class DefaultQRCodeDelegateTest(
                     Activity(),
                 )
 
-                assertEquals(awaitItem(), QrCodeComponentViewType.QR_CODE)
+                assertEquals(awaitItem(), QrCodeComponentViewType.SIMPLE_QR_CODE)
```
jreij, line 213: The order of the arguments here is reversed, should be `assertEquals(expected, actual)`

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -198,7 +210,23 @@ internal class DefaultQRCodeDelegateTest(
...
+
+            delegate.viewFlow.test {
+                assertNull(awaitItem())
+
+                delegate.handleAction(
+                    QrCodeAction(paymentMethodType = PaymentMethodTypes.PAY_NOW, paymentData = "paymentData"),
+                    Activity(),
+                )
+
+                assertEquals(awaitItem(), QrCodeComponentViewType.FULL_QR_CODE)
```
jreij, line 229: Should be `assertEquals(expected, actual)`

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -265,6 +293,17 @@ internal class DefaultQRCodeDelegateTest(
...
     }
 
+    @Test
+    fun `download qr image successfully`(): Unit = runBlocking {
+        val expectedResult = Result.success(Unit)
+        whenever(fileDownloader.download(anyString(), anyString(), anyString(), anyString())) doReturn expectedResult
+
+        val actualResult = delegate.downloadQRImage()
+
+        assertEquals(expectedResult.isSuccess, actualResult.isSuccess)
```
jreij, line 303: This line checks if `expectedResult.isSuccess == actualResult.isSuccess`, it seems like you only want to be sure that the success state is the same.
So for example if they are both `false` your test passes even though the test case is about download a QR image successfully.
What you want to check is that the whole result propagated as it is from `fileDownloader` to the `delegate` right?

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
```diff
@@ -265,6 +293,17 @@ internal class DefaultQRCodeDelegateTest(
...
+        verify(fileDownloader).download(anyString(), anyString(), anyString(), anyString()).isSuccess
```
jreij, line 304: You are verifying that `isSuccess` is called on the result of the `download`, why is that?

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,190 @@
...
+        observeDelegate(delegate, coroutineScope)
+
+        binding.buttonSaveImage.setOnClickListener {
+            val requiredPermission = Manifest.permission.WRITE_EXTERNAL_STORAGE
+            if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q &&
+                ContextCompat.checkSelfPermission(context, requiredPermission) != PackageManager.PERMISSION_GRANTED
+            ) {
+                delegate.onError(
+                    PermissionException(
+                        errorMessage = "storage permission is not granted",
```
jreij, line 81: `Storage`

jreij: or even better `Manifest.permission.WRITE_EXTERNAL_STORAGE` so that they know which one

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -151,11 +161,16 @@ internal class ActionComponentDialogFragment : DropInBottomSheetDialogFragment()
...
     private fun handleError(componentError: ComponentError) {
-        when (componentError.exception) {
+        when (val exception = componentError.exception) {
             is CancellationException -> {
                 Logger.d(TAG, "Flow was cancelled by user")
                 onBackPressed()
             }
+            is PermissionException -> {
+                val requiredPermission = exception.requiredPermission
+                Logger.e(TAG, "$requiredPermission permission is not granted", exception)
```
jreij, line 171: Then here you can simply log the message from the exception itself since it should contain the formatted permission name

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
etman55: same here

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
etman55: nice catch though

qr-code/src/test/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegateTest.kt
jreij: Minor: `: Unit` is not needed

jreij: Hmm, i just noticed you're using `runBlocking`, why is that?

etman55: because `downloadQRImage` is a suspended function, but after the changes I will add with the `eventFlow` it won't be a suspended anymore.

jreij: Hmm so `runTest` doesn't work with suspend then?

qr-code/src/main/java/com/adyen/checkout/qrcode/QrCodeEvent.kt
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 10/1/2023.
+ */
+
+package com.adyen.checkout.qrcode
+
+sealed class QrCodeUIEventEvent {
+    sealed class QrImageDownloaded : QrCodeUIEventEvent() {
```
jreij, line 12: Maybe use a name like `QrImageDownloadResult`?

qr-code/src/main/java/com/adyen/checkout/qrcode/QrCodeUIEvent.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2023 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 10/1/2023.
+ */
+
+package com.adyen.checkout.qrcode
+
+sealed class QrCodeUIEvent {
```
jreij, line 11: Could this be internal? If not let's annotate it with library only visibility

qr-code/src/main/java/com/adyen/checkout/qrcode/QrCodeUIEvent.kt
```diff
@@ -0,0 +1,16 @@
...
+    sealed class QrImageDownloadedResult : QrCodeUIEvent() {
```
jreij, line 12: `QrImageDownloadResult` is probably a more accurate name, because saying `Downloaded` implies it was successful

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private fun handleEventFlow(event: QrCodeUIEvent) {
+        when (event) {
+            QrCodeUIEvent.QrImageDownloadedResult.Success -> {
+                // TODO: add translation
+                context.toast("qr code downloaded successfully")
+            }
+            is QrCodeUIEvent.QrImageDownloadedResult.Failure -> {
+                Logger.d(TAG, "download file failed", event.throwable)
```
jreij, line 182: Did we consider showing anything to the user when the download fails?

qr-code/src/main/java/com/adyen/checkout/qrcode/QrCodeUIEvent.kt
etman55: It can't be internal, will annotate it with library scope.

qr-code/src/main/java/com/adyen/checkout/qrcode/FullQRCodeView.kt
etman55: Good catch, will add a message for it.

jreij:
> The button seems to be cut off. Can you check why this happens? image

@etman55 this issue is still not fixed

etman55:
This issue is a fragment issue has nothing to do with `PayNow`, @OscarSpruit and I discussed it and I guess he opened a ticket for it.

drop-in/src/main/res/template/values/strings.xml.tt
```diff
@@ -30,6 +30,9 @@
     <string name="checkout_giftcard_remove_gift_cards_positive_button">%%storedPaymentMethod.disable.confirmButton%%</string>
     <string name="checkout_giftcard_remove_gift_cards_negative_button">%%storedPaymentMethod.disable.cancelButton%%</string>
     <string name="checkout_remove_stored_payment_method_body">%%storedPaymentMethod.disable.confirmation%%</string>
+    <string name="checkout_qr_code_permission_not_granted">%%runtimePermission.permissionDeniedMessage%%</string>
```
jreij, line 33: These 3 strings are generic and can be used later for other components that need storage permission. I think we should remove the `qr_code` part from the string names

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -51,6 +55,13 @@ internal class ActionComponentDialogFragment : DropInBottomSheetDialogFragment()
     private val actionConfiguration: GenericActionConfiguration by arguments(ACTION_CONFIGURATION)
     private lateinit var actionComponent: GenericActionComponent
 
+    private val requestPermissionLauncher =
+        registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
+            if (!isGranted) {
```
jreij, line 60: Something to improve UX wise: if this is granted we should retry the download, right now the shopper needs to click on the button again. Although I don't think it's an easy task so maybe add a TODO comment to handle this scenario later

drop-in/src/main/java/com/adyen/checkout/dropin/ui/action/ActionComponentDialogFragment.kt
```diff
@@ -151,11 +162,21 @@ internal class ActionComponentDialogFragment : DropInBottomSheetDialogFragment()
...
+        when (val exception = componentError.exception) {
             is CancellationException -> {
                 Logger.d(TAG, "Flow was cancelled by user")
                 onBackPressed()
             }
+            is PermissionException -> {
+                val requiredPermission = exception.requiredPermission
+                Logger.e(TAG, exception.message.orEmpty(), exception)
+                AlertDialog.Builder(requireContext())
+                    .setTitle(R.string.checkout_qr_code_rationale_title_storage_permission)
```
jreij, line 174: These strings will be used now no matter what permission is required, for now this is okay since we only need the storage permission but it's good to add a comment to fetch these strings based on the required permission

checkout-core/src/main/java/com/adyen/checkout/core/util/FileDownloader.kt
```diff
@@ -0,0 +1,117 @@
...
+        url.openStream().use { input ->
+            resolver.openOutputStream(uri).use { output ->
+                if (output == null) return@withContext Result.failure(CheckoutException("out is null"))
+                input.copyTo(output, DEFAULT_BUFFER_SIZE)
+                return@withContext Result.success(Unit)
+            }
+        }
+    }
+
+    private fun String.toURL(): URL? {
```
jreij, line 103: Did you try `Uri.encode`?

etman55: No, will  check it.

etman55: Didn't work I got `OutOfMemoryError`

etman55: I got this error after doing the encoding
```
java.net.MalformedURLException: Protocol not found: https%3A%2F%2Fcheckoutshopper-test.adyen.com%2Fcheckoutshopper%2Fbarcode.shtml%3FbarcodeType%3DqrCode%26fileType%3Dpng%26data%3D00020126580009SG.PAYNOW010120213201210678MADY03010041420230120235031520400005303702540513.375802SG5925ADYEN%20SINGAPORE%20PTE.%20LTD.6009Singapore62200116PWRHRMBJVXSKGK8263043D50
```

etman55: I got this error after trying it `java.net.MalformedURLException: Protocol not found`

qr-code/src/main/java/com/adyen/checkout/qrcode/DefaultQRCodeDelegate.kt
```diff
@@ -147,14 +149,18 @@ internal class DefaultQRCodeDelegate(
         }
 
         val viewType = when (action.paymentMethodType) {
-            PaymentMethodTypes.PAY_NOW -> QrCodeComponentViewType.FULL_QR_CODE
+            PaymentMethodTypes.PAY_NOW -> {
+                maxPollingDurationMillis = PAY_NOW_MAX_POLLING_DURATION
+                QrCodeComponentViewType.FULL_QR_CODE
+            }
             else -> QrCodeComponentViewType.SIMPLE_QR_CODE
```
OscarSpruit, line 156: I think it would be nice to set the `maxPollingDurationMillis` back to the default value in the `else`

jreij (approved):

PR #877, opened 2022-11-16
Create card component parameters for internal use

jreij:
Our configuration classes are built directly by the developers and they are immutable. We need to be able to inject some values into them from drop-in or from the API responses.

In this PR we create a data class for the card component that shadows the card configuration and includes any extra parameters we need.

COAND-549

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
+++ b/card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
@@ -12,7 +12,6 @@ import androidx.annotation.RestrictTo
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.ViewModelStoreOwner
 import androidx.savedstate.SavedStateRegistryOwner
-import com.adyen.checkout.card.data.CardType
 import com.adyen.checkout.card.repository.DefaultAddressRepository
 import com.adyen.checkout.card.repository.DefaultDetectCardTypeRepository
 import com.adyen.checkout.components.StoredPaymentComponentProvider
@@ -24,12 +23,8 @@ import com.adyen.checkout.components.model.paymentmethods.StoredPaymentMethod
 import com.adyen.checkout.components.repository.DefaultPublicKeyRepository
 import com.adyen.checkout.components.repository.PaymentObserverRepository
 import com.adyen.checkout.core.exception.ComponentException
-import com.adyen.checkout.core.log.LogUtil
-import com.adyen.checkout.core.log.Logger
 import com.adyen.checkout.cse.DefaultCardEncrypter
 import com.adyen.checkout.cse.DefaultGenericEncrypter
-
-private val TAG = LogUtil.getTag()
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class CardComponentProvider(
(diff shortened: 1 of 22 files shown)


card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
```diff
@@ -8,28 +8,30 @@
...
 import com.adyen.checkout.components.repository.DefaultPublicKeyRepository
 import com.adyen.checkout.components.repository.PaymentObserverRepository
 import com.adyen.checkout.core.exception.ComponentException
-import com.adyen.checkout.core.log.LogUtil
-import com.adyen.checkout.core.log.Logger
 import com.adyen.checkout.cse.DefaultCardEncrypter
 import com.adyen.checkout.cse.DefaultGenericEncrypter
 
-private val TAG = LogUtil.getTag()
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
OscarSpruit, line 29: Can't we just make this class internal?

card/src/test/java/com/adyen/checkout/card/StoredCardDelegateTest.kt
```diff
@@ -297,7 +297,8 @@ internal class StoredCardDelegateTest {
             observerRepository = PaymentObserverRepository(),
             storedPaymentMethod = storedPaymentMethod,
             publicKeyRepository = publicKeyRepository,
-            configuration = configuration,
+            configuration  =configuration,
```
OscarSpruit, line 300: Oh no

card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
jreij: No because in drop-in we need to use `CardComponentProvider(dropInConfiguration)`

card/src/test/java/com/adyen/checkout/card/StoredCardDelegateTest.kt
jreij: How did that not cause a gradle check fail? 🤔

OscarSpruit: Did we disable ktlint for tests?

jreij: Yeah some rules it seems

card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
```diff
@@ -54,15 +56,16 @@ class CardComponentProvider : StoredPaymentComponentProvider<CardComponent, Card
                 DefaultCardDelegate(
                     observerRepository = PaymentObserverRepository(),
                     publicKeyRepository = publicKeyRepository,
-                    configuration = verifiedConfiguration,
+                    configuration = configuration,
```
OscarSpruit, line 59: Maybe a stupid question; but why do we need to pass both the configuration and params? Isn't only the params enough?

jreij: We only need params, but I cannot remove the configuration yet because it's in the `ComponentDelegate` interface

OscarSpruit: Ahh okay 👍

OscarSpruit (approved):

PR #878, opened 2022-11-16
Handle actions with CardComponent

OscarSpruit:
`CardComponent` will be able to handle redirect and 3DS2 actions by itself with this. Implementing `CardComponent` will take less steps.

COAND-650


card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -69,7 +71,9 @@ internal class DefaultCardDelegate(
     private val cardValidationMapper: CardValidationMapper,
     private val cardEncrypter: CardEncrypter,
     private val genericEncrypter: GenericEncrypter,
-) : CardDelegate {
+    private val actionHandlingDelegate: ActionHandlingDelegate,
+) : CardDelegate,
+    ActionHandlingDelegate by actionHandlingDelegate {
```
jreij, line 76: What sorcery is this? 😱

components-core/src/main/java/com/adyen/checkout/components/ComponentViewType.kt
```diff
@@ -6,6 +6,6 @@
  * Created by josephj on 1/9/2022.
  */
 
-package com.adyen.checkout.components.ui.view
+package com.adyen.checkout.components
```
jreij, line 9: Why can't his stay in `ui-core`?

card/src/main/java/com/adyen/checkout/card/CardViewProvider.kt
```diff
@@ -24,6 +27,8 @@ internal object CardViewProvider : ViewProvider {
     ): ComponentView {
         return when (viewType) {
             CardComponentViewType -> CardView(context, attrs, defStyleAttr)
+            Adyen3DS2ComponentViewType -> PaymentInProgressView(context, attrs, defStyleAttr)
```
jreij, line 30: Hmm we shouldn't really do that, we'll need to add the other action views and do the same for all the view providers. But I don't have an alternative yet, let's think about it

OscarSpruit: Oh yeah I wanted to discuss this, because this feels wrong indeed. Couldn't think of anything better for now

components-core/src/main/java/com/adyen/checkout/components/ActionHandlingComponent.kt
```diff
@@ -0,0 +1,32 @@
...
+import kotlinx.coroutines.CoroutineScope
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface ActionHandlingComponent {
+
+    fun handleAction(action: Action, activity: Activity)
+
+    fun handleIntent(intent: Intent)
+
+    fun observeAction(
```
jreij, line 25: I think this should not be its own method, `PaymentComponentEvent` should include a new event for the action and the merchant would only need one `observe`. WDYT?

components-core/src/main/java/com/adyen/checkout/components/ComponentViewType.kt
OscarSpruit: For `ActionHandlingDelegate` to use it and that class has to be in `components-core`, because it has dependencies on other components classes

components-core/src/main/java/com/adyen/checkout/components/ActionHandlingComponent.kt
OscarSpruit: Good suggestion 👍

ui-core/src/main/java/com/adyen/checkout/components/ui/view/AdyenComponentView.kt
```diff
@@ -58,16 +58,16 @@ class AdyenComponentView @JvmOverloads constructor(
     ) where T : ViewableComponent, T : Component<*, *> {
         component.viewFlow
             .filterNotNull()
-            .onEach {
+            .onEach { componentViewType ->
                 val delegate = component.delegate
```
jreij, line 62: I don't think this works now. The view flow will emit a new view type for the actions but the delegate will always be the CardDelegate. We need a way to return different delegates based on where we are in the flow.

The way we achieve it now in `GenericActionDelegate` is by making the `delegate` change with every new action and overriding the getter:
```kotlin
    private var _delegate: ActionDelegate? = null
    override val delegate: ActionDelegate get() = requireNotNull(_delegate)
```
Not sure if we can do something similar to payment components.

OscarSpruit: The only thing I can come up with is this, but it's not so pretty
```Kotlin
private var _delegate: ComponentDelegate = cardDelegate
override val delegate: ComponentDelegate get() = _delegate
override val viewFlow: Flow<ComponentViewType?> = merge(
        cardDelegate.viewFlow,
        genericActionDelegate.viewFlow,
    )
        .onEach {
            _delegate = when (it) {
                is CardComponentViewType -> cardDelegate
                else -> genericActionDelegate
            }
        }
```

jreij: What about this?

```kotlin
    private var _delegate: ComponentDelegate = cardDelegate
    override val delegate: ComponentDelegate get() = _delegate

        genericActionDelegate.handleAction(action, activity)
        _delegate = genericActionDelegate.delegate
    }
```

But the `handleAction` code is in the `DefaultActionHandlingComponent` so dunno how to achieve that.

jreij: Also either way, we might have a problem if the view flow emits before the value of `delegate` changes, because he new view will be created but the old delegate will be passed to it :/

OscarSpruit: With what I posted earlier that can't happen, but your solution looks cleaner. Let me test it

action/src/main/java/com/adyen/checkout/action/ActionHandlingComponent.kt
```diff
@@ -0,0 +1,33 @@
...
+package com.adyen.checkout.action
+
+import android.app.Activity
+import android.content.Intent
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.components.model.payments.response.Action
+import com.adyen.threeds2.customization.UiCustomization
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface ActionHandlingComponent {
```
jreij, line 18: Can we add the docs for the other methods as well? We can just copy them from other classes I think. Or i wonder if we should reuse this interface with the other action components to avoid duplication 🤔

action/src/test/java/com/adyen/checkout/action/TestActionDelegate.kt
```diff
@@ -38,7 +38,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
-import java.util.Locale
+import java.util.*
```
jreij, line 41: Is this your IDE messing up?

components-core/src/main/java/com/adyen/checkout/components/ActionComponentEvent.kt
```diff
@@ -8,8 +8,26 @@
...
+import com.adyen.checkout.components.model.payments.request.PaymentMethodDetails
+
 // TODO add docs
 sealed class ActionComponentEvent : ComponentEvent {
     class ActionDetails(val data: ActionComponentData) : ActionComponentEvent()
     class Error(val error: ComponentError) : ActionComponentEvent()
 }
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun <T> ((PaymentComponentEvent<T>) -> Unit).toActionCallback(): (ActionComponentEvent) -> Unit
```
jreij, line 21: Nice

ui-core/src/main/java/com/adyen/checkout/components/ui/PaymentInProgressView.kt
```diff
@@ -36,7 +36,7 @@ class PaymentInProgressView @JvmOverloads constructor(
     private val binding = ViewPaymentInProgressBinding.inflate(LayoutInflater.from(context), this)
 
     override fun initView(delegate: ComponentDelegate, coroutineScope: CoroutineScope, localizedContext: Context) {
-        if (delegate !is ActionDelegate) throw IllegalStateException("Unsupported delegate type")
+        if (delegate !is ActionDelegate) return
```
jreij, line 39: Can we have a scenario where a non `ActionDelegate`s is passed here?

action/src/main/java/com/adyen/checkout/action/ActionHandlingComponent.kt
OscarSpruit: Not sure how we would do the `observe` and `removeObserver` methods from `ActionComponent`. I think it's better to keep them separate as it feels like they have different purposes

action/src/test/java/com/adyen/checkout/action/TestActionDelegate.kt
OscarSpruit: Yes, apparently some setting was not right

ui-core/src/main/java/com/adyen/checkout/components/ui/PaymentInProgressView.kt
OscarSpruit: Whoops I changed it to return to test something. Only `ActionDelegate`s should be passed, so I think throwing an exception is better

action/src/main/java/com/adyen/checkout/action/ActionHandlingComponent.kt
OscarSpruit: Added docs for now

jreij: 👍

jreij (approved):

PR #880, opened 2022-11-16
Fix max line length

jreij:
- Set max line length as 120 in the detekt config (it was 150).
- Update detekt to `1.21.0` to benefit from `MaximumLineLength: ignoreBackTickedIdentifier` (for test method names) and `MaxLineLength: ignoreAnnotated: ['Test']` (to work with the previous one).

COAND-671

### DIFF
--- a/config/detekt/detekt-config.yml
+++ b/config/detekt/detekt-config.yml
@@ -258,6 +258,7 @@ formatting:
   MaximumLineLength:
     active: true
     maxLineLength: 120
+    ignoreBackTickedIdentifier: true
   ModifierOrdering:
     active: true
     autoCorrect: true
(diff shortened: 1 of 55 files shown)


config/detekt/detekt-config.yml
```diff
@@ -570,10 +571,11 @@ style:
     active: false
   MaxLineLength:
     active: true
-    maxLineLength: 150
+    maxLineLength: 120
     excludePackageStatements: true
     excludeImportStatements: true
     excludeCommentStatements: false
+    ignoreAnnotated: ['Test']
```
OscarSpruit, line 578: Oh nice! So there is a way to do ignore test titles

OscarSpruit (approved):

config/detekt/detekt-config.yml
jreij: Yep, there are 2 rules: [MaxLineLength](https://detekt.dev/docs/rules/style/#maxlinelength) and [MaximumLineLength](https://detekt.dev/docs/rules/formatting/#maximumlinelength).

The first one checks the line length for code but also comments and everything. This one doesn't have a way of ignoring long method names only, so I disabled it for tests.

The second one only checks the code but also has an `ignoreBackTickedIdentifier` option which is exactly what we need.

So configuring them together like this should cover all the cases we need.

PR #908, opened 2022-11-21
Remove spotbugs

jreij:
The plugin wasn't configured correctly so it wasn't even working. After enabling it, it flagged errors inside the auto generated view binding classes. Disabling it for now and will investigate later whether it's needed.

COAND-674


jreij:
Related to #907.

OscarSpruit (dismissed):

OscarSpruit (approved):

PR #915, opened 2022-11-21
Fix crash in IssuerListSpinnerAdapter

ozgur00:


issuer-list/src/main/java/com/adyen/checkout/issuerlist/IssuerListSpinnerAdapter.kt
```diff
@@ -45,6 +45,7 @@ class IssuerListSpinnerAdapter internal constructor(
             binding = SpinnerListWithImageBinding.inflate(LayoutInflater.from(context), parent, false)
             view = binding.root
             viewHolder = IssuerListSpinnerViewHolder(binding, imageLoader, paymentMethod, hideIssuerLogo)
+            view.tag = viewHolder
```
etman55, line 48: Can you make sure that this crash doesn't happen somewhere else? As I guess there's some other places that is done the same way this class is done

ozgur00: I did, this is the only place we missed adding the tag. Thanks for pointing out!

PR #920, opened 2022-11-21
Create generic component parameters for internal use

jreij:
Create a params data class for all generic components with no extra special configurations.

COAND-549

### DIFF
--- a/action/src/main/java/com/adyen/checkout/action/ActionDelegateProvider.kt
+++ b/action/src/main/java/com/adyen/checkout/action/ActionDelegateProvider.kt
@@ -10,9 +10,9 @@ package com.adyen.checkout.action
 
 import android.app.Application
 import androidx.lifecycle.SavedStateHandle
-import com.adyen.checkout.adyen3ds2.Adyen3DS2Component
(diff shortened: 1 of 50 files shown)


action/src/main/java/com/adyen/checkout/action/ActionDelegateProvider.kt
```diff
@@ -44,22 +46,42 @@ internal class ActionDelegateProvider {
     ): ActionDelegate {
         val delegate = when (action) {
             is AwaitAction -> {
-                AwaitComponent.PROVIDER.getDelegate(getConfiguration(configuration), savedStateHandle, application)
+                AwaitComponentProvider(parentConfiguration).getDelegate(
+                    getConfiguration(configuration),
```
OscarSpruit, line 50: This line is a bit confusing, I had to look up `getConfiguration` before I understood. Maybe we should rename `getConfiguration` to something like `getSpecificConfiguration` or `getActionConfiguration`?

jreij: Good point 👍

OscarSpruit (approved):

PR #925, opened 2022-11-22
Rename configurations from delegates and views

jreij:
Now that we have component params, we can completely remove configurations from delegates and views.

COAND-549


bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentParams.kt
```diff
@@ -8,15 +8,17 @@
...
+import androidx.annotation.RestrictTo
 import com.adyen.checkout.components.base.ComponentParams
 import com.adyen.checkout.components.model.payments.Amount
 import com.adyen.checkout.core.api.Environment
 import kotlinx.parcelize.Parcelize
 import java.util.Locale
 
 @Parcelize
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 // TODO check if this class is still needed once all params support [Amount]
```
OscarSpruit, line 20: So is this still needed?

jreij: Yep, we didn't add the amount to all configs yet

OscarSpruit: Whoops my bad, somehow I thought this was saying to check again after the config classes could be removed 😅

OscarSpruit (approved):

PR #926, opened 2022-11-22
Move configuration defaults to params

jreij:
Moving all the default values from the configuration builders to the mappers means that the mappers become the source of truth for all config values, since we already compute other values in there. And we need to make sure that we keep the config builder docs up to date.

COAND-549

### DIFF
--- a/bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentParamsMapper.kt
+++ b/bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentParamsMapper.kt
@@ -9,6 +9,7 @@
 package com.adyen.checkout.bacs
 
 import com.adyen.checkout.components.base.Configuration
+import com.adyen.checkout.components.model.payments.Amount
 
 internal class BacsDirectDebitComponentParamsMapper(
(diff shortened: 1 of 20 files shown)


bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentParamsMapper.kt
```diff
@@ -31,7 +32,7 @@ internal class BacsDirectDebitComponentParamsMapper(
             shopperLocale = parentConfiguration.shopperLocale,
             environment = parentConfiguration.environment,
             clientKey = parentConfiguration.clientKey,
-            amount = bacsDirectDebitConfiguration.amount,
+            amount = bacsDirectDebitConfiguration.amount ?: Amount.EMPTY,
```
OscarSpruit, line 35: Do we have tests that validate the default values? If tests validate this, then you can't accidentally change a default value

OscarSpruit: Just wanted to make sure ✅

OscarSpruit (approved):

PR #929, opened 2022-11-23
Remove setShopperLocale and setEnvironment in configuration builders

jreij:
These params are available in the constructor now so we don't need their setters anymore

COAND-549


ozgur00:
Can you remove the example usage from `README` file as well?

jreij:
> Can you remove the example usage from `README` file as well?

Good one 👍 But the `README` reflects the latest released version so maybe let's do that after v5 is released.

ozgur00 (approved):

PR #932, opened 2022-11-24
Move isCreatedByDropIn and isAnalyticsEnabled to component params

jreij:
## Description
- Move `isCreatedByDropIn` to params: we need this field in order to send the correct analytics flavor (drop-in/components). It is currently in `BasePaymentComponent` and it's set using a `setCreatedForDropIn` method which is not ideal (needs to be called after component creation) and is exposed to merchants.
- Move `isAnalyticsEnabled` to configurations and params: it's also in `BasePaymentComponent` like `isCreatedByDropIn`. Moving this field to configurations to improve the API and facilitate removing `BasePaymentComponent`.

## Checklist
- [x] Code is unit tested

COAND-549


components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -23,6 +23,8 @@ abstract class BaseConfigurationBuilder<
     protected var clientKey: String
 ) {
 
+    protected var isAnalyticsEnabled: Boolean? = null
```
OscarSpruit, line 26: A nullable boolean feels a bit wrong to me 🤔 WDYT?

jreij: I agree it looks weird but so does a nullable string or enum that will never actually be null when it reaches the component. It's just because we decided setting all the default values in the component params mappers and not in the configuration builders. I did that as well in other configurations. WDYT?

OscarSpruit: Okay let's keep it nullable 👍

OscarSpruit (approved):

PR #938, opened 2022-11-28
Refactor analytics

jreij:
## Description
Analytics are implemented in `AnalyticsDispatcher` using a `JobIntentService` which is deprecated. It also uses a context which we shouldn't hold a reference to inside components.

This refactor creates `AnalyticsRepository` which uses coroutines and is inline with our other repos. We also won't need to use a background service since we only need to fire and forget the analytics API calls and there's nothing to be handled in the response.

## Checklist 
- [x] Code is unit tested

COAND-679


components-core/src/main/java/com/adyen/checkout/components/api/AnalyticsService.kt
```diff
@@ -0,0 +1,31 @@
...
+class AnalyticsService(
+    private val httpClient: HttpClient,
+) {
+
+    suspend fun sendEvent(
+        queryParameters: Map<String, String>,
+    ) {
+        withContext(Dispatchers.IO) {
+            httpClient.get(
+                "images/analytics.png",
```
OscarSpruit, line 26: why are we getting an image for analytics? 😂

components-core/src/main/java/com/adyen/checkout/components/base/BasePaymentComponent.kt
```diff
@@ -49,19 +45,7 @@ abstract class BasePaymentComponent<
...
-                Flavor.COMPONENT
-            }
-            val type = paymentComponentDelegate.getPaymentMethodType()
-            if (type.isEmpty()) {
-                throw CheckoutException("Payment method has empty or null type")
-            }
-            val analyticEvent = create(context, flavor, type, configuration.shopperLocale)
-            dispatchEvent(context, configuration.environment, analyticEvent)
-        }
+        // whole class will be deleted soon
```
OscarSpruit, line 48: 🙌

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -121,6 +127,9 @@ internal class DropInViewModel(
             savedStateHandle[CURRENT_ORDER] = value
         }
 
+    private val orderStatusRepository: OrderStatusRepository
```
OscarSpruit, line 130: Why not put these in the constructor?

components-core/src/main/java/com/adyen/checkout/components/api/AnalyticsService.kt
jreij: hahaha 🙈

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
OscarSpruit: The default value should indeed be removed. But all constructor arguments can be provided from `DropInViewModelFactory`. Everything is already setup correctly to make that work

jreij: The default value is needed because we were getting a crash without it, since there is no zero argument constructor anymore. I will try to reproduce.

OscarSpruit (approved):

PR #940, opened 2022-11-29
Extract Bundle and SaveStateHandle logic to BundleHandler

jreij:
## Description
We use `var`s with getters and setters that access `savedStateHandle` in order to persist values in bundles.

This solution uses an implementation of `ReadWriteProperty` to delegate reading and writing to the `savedStateHandle` to a custom class.

## Checklist 

COAND-682

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
@@ -23,6 +23,8 @@ import com.adyen.checkout.adyen3ds2.repository.SubmitFingerprintResult
 import com.adyen.checkout.components.ActionComponentData
 import com.adyen.checkout.components.ActionComponentEvent
 import com.adyen.checkout.components.base.GenericComponentParams
+import com.adyen.checkout.components.bundle.SavedStateHandleContainer
+import com.adyen.checkout.components.bundle.SavedStateHandleProperty
 import com.adyen.checkout.components.channel.bufferedChannel
 import com.adyen.checkout.components.encoding.Base64Encoder
 import com.adyen.checkout.components.handler.RedirectHandler
@@ -67,7 +69,7 @@ import org.json.JSONObject
 @Suppress("TooManyFunctions", "LongParameterList")
 internal class DefaultAdyen3DS2Delegate(
     private val observerRepository: ActionObserverRepository,
-    private val savedStateHandle: SavedStateHandle,
+    override val savedStateHandle: SavedStateHandle,
     override val componentParams: GenericComponentParams,
     private val submitFingerprintRepository: SubmitFingerprintRepository,
     private val paymentDataRepository: PaymentDataRepository,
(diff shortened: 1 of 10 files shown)


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/DefaultAdyen3DS2Delegate.kt
```diff
@@ -95,11 +97,7 @@ internal class DefaultAdyen3DS2Delegate(
 
     private var currentTransaction: Transaction? = null
 
-    private var authorizationToken: String?
-        get() = savedStateHandle[AUTHORIZATION_TOKEN_KEY]
-        set(value) {
-            savedStateHandle[AUTHORIZATION_TOKEN_KEY] = value
-        }
+    private var authorizationToken: String? by SavedStateHandleProperty(AUTHORIZATION_TOKEN_KEY)
```
OscarSpruit, line 100: image

components-core/src/main/java/com/adyen/checkout/components/bundle/BundleHandler.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.annotation.RestrictTo
+import androidx.lifecycle.SavedStateHandle
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+/**
+ * Implement this in any view model or class that contains a [SavedStateHandle].
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface BundleHandler {
```
OscarSpruit, line 20: This name doesn't feel right, because it's specifically for `SavedStateHandle`. `SavedStateHandler` would be better, but imo `SavedStateHandleContainer` would be even better. The implementing class is not really handling the `SavedStateHandle` it's just having one. Naming is hard, so feel free to suggest something else 😆

jreij: You're right actually, I think the issue started in `DropInBundleHandler`. At first I only had this class and it handled the `SavedStateHandle` as well as the bundle. Then I decided to use the same approach for `SavedStateHandle` in the other parts of the project but forgot to rename it. Will fix 👍 

But as for the `DropInBundleHandler` class, do you think the name should change as well? This class is weird because it handles the `SavedStateHandle` inside it but also the bundle in its companion. I can't decide if it should be split in 2 classes or restructured/renamed 🤔

OscarSpruit: That's a difficult one indeed 🤔 According to [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) it should be split up in two. This would result in `class DropInSavedStateHandleContainer` and `DropInIntentCreator` (the first names that came to mind). If you put those two in the same file you can still share the constants and keep them private

jreij: Yeah sharing the constants is what I was thinking of, I'll try to use one file 👍

OscarSpruit (approved):

PR #941, opened 2022-11-30
Fix screen rotation issues with action handling in payment components

OscarSpruit:
## Description
Steps to reproduce:
1. Open card implementation example
2. Enter credit card details to open native 3ds2 flow
3. Rotate device
4. Complete 3ds2 flow
5. 💥

The issue was that Kotlin's `merge(flows)` function doesn't keep state and the flows in question that are merged are `StateFlow`s. When (re)subscribing to the merged flow, the last values from the `StateFlow`s were emitted again. With `mergeStateFlows` and `mergeViewFlows` added in this PR, it is possible to merge multiple flows into a `StateFlow`. `mergeViewFlows` is tailored to our special needs

## Checklist 

COAND-660

### DIFF
--- a/ui-core/src/main/java/com/adyen/checkout/components/extensions/FlowExtensions.kt
+++ b/ui-core/src/main/java/com/adyen/checkout/components/extensions/FlowExtensions.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 30/11/2022.
+ */
+
+package com.adyen.checkout.components.extensions
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.components.flow.mergeStateFlows
+import com.adyen.checkout.components.ui.view.ComponentViewType
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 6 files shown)


ui-core/src/main/java/com/adyen/checkout/components/extensions/FlowExtensions.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.adyen.checkout.components.flow.mergeStateFlows
+import com.adyen.checkout.components.ui.view.ComponentViewType
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.drop
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun mergeViewFlows(
```
jreij, line 21: Not very important but is there any way we can have useful tests for this?

OscarSpruit: Some cases are testable, but I'm not sure if we can simulate screen rotation somehow. Let me see what we can do

OscarSpruit: I was only able to test the `drop` part and emitting of values

jreij: That's good enough I think 🙏

jreij (approved):

PR #948, opened 2022-12-07
Add option to add action configs to payment method configs

OscarSpruit:
## Description
Because components will be able to handle any action it should be possible to configure actions for components. It's now possible to do that by calling `add...ActionConfiguration` on any payment method component configuration.

## Checklist 

COAND-681


jreij (approved):
💯

PR #921, opened 2022-11-21
Pay Button

ozgur00:
COAND-651

PR #961, opened 2022-12-13
Address commented out code and todos related to Pay Button changes

ozgur00:
## Description
Fix comments that's left out in [this PR](https://github.com/Adyen/adyen-android/pull/921).

## Checklist 

COAND-651

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
@@ -13,8 +13,6 @@ import com.adyen.checkout.dropin.DropInConfiguration
 import com.adyen.checkout.example.BuildConfig
 import com.adyen.checkout.example.data.storage.KeyValueStorage
 import com.adyen.checkout.googlepay.GooglePayConfiguration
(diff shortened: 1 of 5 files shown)


example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -67,6 +60,7 @@ internal class CheckoutConfigurationProvider @Inject constructor(
     fun getCardConfiguration(): CardConfiguration =
         CardConfiguration.Builder(shopperLocale, environment, clientKey)
             .setShopperReference(keyValueStorage.getShopperReference())
+            .setHideCvcStoredCard(true)
```
jreij, line 63: We shouldn't commit this

OscarSpruit, line 63: Are you sure we want to add this?

jreij:
@ozgur00 one more comment that we could fix here, `BacsDirectDebitDelegate` has a `handleBackPress` function but the component doesn't have it, meaning merchants cannot use it. Can we add it to the component as well?

jreij (approved):

PR #962, opened 2022-12-13
Support handling actions with all payment components

OscarSpruit:
## Description
Add generic action handling to all payment components.

## Checklist 

COAND-650

### DIFF
--- a/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
+++ b/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
@@ -14,7 +14,7 @@ import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.ViewModelStoreOwner
 import androidx.savedstate.SavedStateRegistryOwner
 import com.adyen.checkout.action.DefaultActionHandlingComponent
-import com.adyen.checkout.action.GenericActionComponent
+import com.adyen.checkout.action.GenericActionComponentProvider
 import com.adyen.checkout.card.CardValidationMapper
 import com.adyen.checkout.components.PaymentComponentProvider
 import com.adyen.checkout.components.analytics.AnalyticsMapper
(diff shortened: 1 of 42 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/component/BacsDirectDebitDialogFragment.kt
```diff
@@ -78,7 +79,8 @@ internal class BacsDirectDebitDialogFragment : BaseComponentDialogFragment() {
     }
 
     override fun onBackPressed(): Boolean {
-        return if (bacsDirectDebitComponent.delegate.handleBackPress()) {
+        val delegate = bacsDirectDebitComponent.delegate as? BacsDirectDebitDelegate
```
jreij, line 82: You'll be able to call `component.handleBackPress()` directly after [this commit](https://github.com/Adyen/adyen-android/pull/961/commits/1bdbdc8686a14561377eea9caf78b19bf06ee3ef)

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
```diff
@@ -65,18 +67,29 @@ class BcmcComponentProvider(
...
+                paymentMethod = paymentMethod,
+                publicKeyRepository = publicKeyRepository,
+                componentParams = componentParams,
+                cardValidationMapper = cardValidationMapper,
+                cardEncrypter = cardEncrypter,
+                analyticsRepository = analyticsRepository,
+                submitHandler = SubmitHandler()
+            )
+
+            val genericActionDelegate = GenericActionComponent.PROVIDER.getDelegate(
```
jreij, line 83: I think we should use `GenericActionComponentProvider(overrideComponentParams)` here.

My suspicion is that if we launch a component from drop-in, and set a custom config for it, the drop-in config will override the custom config for the component delegate but not for the action delegate. Can you make a quick test to see if this true?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/component/BacsDirectDebitDialogFragment.kt
OscarSpruit: Oh nice! Didn't see that

jreij (approved):

bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
OscarSpruit: You are right, let me fix it

jreij (approved):

PR #967, opened 2022-12-15
Add caching to image loader

OscarSpruit:
## Description
- Add caching to image loader
- Ensure images are set from correct thread

## Checklist 

COAND-673

### DIFF
--- a/checkout-core/src/main/java/com/adyen/checkout/core/image/ImageLoader.kt
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/image/ImageLoader.kt
@@ -27,8 +27,8 @@ interface ImageLoader {
 
     suspend fun load(
         url: String,
-        onSuccess: (Bitmap) -> Unit,
-        onError: (Throwable) -> Unit
+        onSuccess: suspend (Bitmap) -> Unit,
+        onError: suspend (Throwable) -> Unit
     )
 }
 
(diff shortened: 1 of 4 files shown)


checkout-core/src/main/java/com/adyen/checkout/core/image/ImageLoader.kt
```diff
@@ -58,16 +69,41 @@ object DefaultImageLoader : ImageLoader {
...
+    }
+
+    companion object {
+
+        private const val LOW_MEMORY_PERCENT = 0.15
+        private const val DEFAULT_MEMORY_PERCENT = 0.2
+        private const val DEFAULT_MEMORY_MEGABYTES = 256
+        private const val BYTE_CONVERSION = 1024
+
+        private fun calculateInMemoryCacheSize(context: Context): Int = try {
```
ozgur00, line 98: Is there a reason for this method to be static?

OscarSpruit: Not really, you think it's better to move it out?

ozgur00: I read a bit about it [here](https://softwareengineering.stackexchange.com/questions/215826/make-methods-that-do-not-depend-on-instance-fields-static) and don't have a strong opinion about it. Some say keeping functions where you don't access the member fields static helps extracting them later on if needed (probably because you can see immediately). I think functionally there's not much distinction, so we can keep it as is.

ozgur00 (approved):

PR #974, opened 2022-12-19
Prevent duplicate toast messages in MainActivity

OscarSpruit:
## Description
1. Turn off internet on your device
2. Click an entry in MainActivity
3. See error toast
4. Rotate device
5. See error toast again

Errors were a view state, so they would display the toast again and again on for example a configuration change. Now errors are an event to ensure the toast is displayed only once. 

## Checklist 

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/main/MainActivity.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/main/MainActivity.kt
@@ -65,7 +65,7 @@ class MainActivity : AppCompatActivity(), DropInCallback {
         lifecycleScope.launch {
             repeatOnLifecycle(Lifecycle.State.STARTED) {
                 launch { viewModel.viewState.collect(::onViewState) }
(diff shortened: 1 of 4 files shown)


example-app/src/main/java/com/adyen/checkout/example/ui/main/MainActivity.kt
```diff
@@ -85,21 +85,10 @@ class MainActivity : AppCompatActivity(), DropInCallback {
...
-    }
+    override fun onDropInResult(dropInResult: DropInResult?) = viewModel.onDropInResult(dropInResult)
 
     private fun onViewState(viewState: MainViewState) {
         when (viewState) {
-            is MainViewState.Error -> {
-                setLoading(false)
-                Toast.makeText(this, viewState.message, Toast.LENGTH_SHORT).show()
-            }
             MainViewState.Loading -> setLoading(true)
```
etman55, line 92: did you double check for if loading not dismissed when any error happens? `setLoading(false)`

etman55 (approved):

PR #977, opened 2022-12-20
Add amount to all Configurations

ozgur00:
## Description
Add amount field to Configuration interface and add it to all implementations of it.
## Checklist 
- [x] Code is unit tested

COAND-692


components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -63,6 +66,18 @@ abstract class BaseConfigurationBuilder<
         return this as BuilderT
     }
 
+    /**
+     * TODO docs
+     */
+    fun setAmount(amount: Amount): BuilderT {
+        if (!CheckoutCurrency.isSupported(amount.currency) || amount.value < 0) {
+            throw CheckoutException("Currency is not valid.")
```
OscarSpruit, line 74: Should we be more specific in this message? Cause either the currency or the value is invalid

OscarSpruit:
Did you test the gift card flow where the amount changes?

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
jreij: Maybe we should move the validation to the `Amount` class even?

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -63,6 +66,18 @@ abstract class BaseConfigurationBuilder<
         return this as BuilderT
     }
 
+    /**
+     * TODO docs
+     */
+    fun setAmount(amount: Amount): BuilderT {
+        if (!CheckoutCurrency.isSupported(amount.currency) || amount.value < 0) {
```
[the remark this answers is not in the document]
OscarSpruit: Coincidentally, I was going through this list, so I updated it in https://github.com/Adyen/adyen-android/pull/979

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
ozgur00: We can move it to `init` block of `Amount` class but this class is being used in response models too. We would be throwing exceptions while deserializing if backend doesn't send us a valid object. I think it's better to do the check for validity only for the `Amount`s merchants provide.

bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
```diff
@@ -117,6 +117,50 @@ internal class BcmcComponentParamsMapperTest {
         assertEquals(expected, params)
     }
 
+    @Test
+    fun `when parent configuration is set with empty amount and bcmc configuration is set with amount then amount shouldn't be overridden`() {
```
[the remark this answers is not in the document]
ozgur00: Makes sense. It's best to move setting the amount to default to overriding part to have that logic only in one place. Drop-in will always override the amount, then if overridden amount is empty we will set it to default.

ozgur00:
> Did you test the gift card flow where the amount changes?

Yes, tested!

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -63,6 +66,21 @@ abstract class BaseConfigurationBuilder<
         return this as BuilderT
     }
 
+    /**
+     * TODO docs
+     */
+    fun setAmount(amount: Amount): BuilderT {
```
jreij, line 72: Does calling `setAmount(Amount.EMPTY)` trigger an exception? If so should we allow it maybe?

ozgur00: It will. We didn't use to allow it though. For what use case we should allow it?

jreij: Not sure, just thinking that `Amount.EMPTY` is a valid option, so if you have a case where your `amount` might be EMPTY sometimes or positive other times, you'll have to change this:
```kotlin
return DropInConfiguration.Builder(
    shopperLocale,
    environment,
    clientKey,
)
.addCardConfiguration(cardConfiguration)
.addBcmcConfiguration(bcmcConfiguration)
.addGooglePayConfiguration(googlePayConfiguration)
.add3ds2ActionConfiguration(threeDS2Configuration)
.setAmount(amount)
.build()
```
to this:
```kotlin
val dropInConfigurationBuilder = DropInConfiguration.Builder(
    shopperLocale,
    environment,
    clientKey,
)
.addCardConfiguration(cardConfiguration)
.addBcmcConfiguration(bcmcConfiguration)
.addGooglePayConfiguration(googlePayConfiguration)
.add3ds2ActionConfiguration(threeDS2Configuration)

if(amount != Amount.EMPTY) {
    dropInConfigurationBuilder.setAmount(amount)
}

return dropInConfigurationBuilder.build()
```
Which is not terrible but not nice either. Does that make any sense?

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -67,15 +67,14 @@ abstract class BaseConfigurationBuilder<
     }
 
     /**
-     * TODO docs
+     * Sets the amount of the transaction.
+     *
+     * Default is [Amount.EMPTY] for all configurations, except it is 0 USD for Google Pay Component.
```
jreij, line 72: What do you think if we just write `Default is [Amount.EMPTY].` here, then override `setAmount` in google pay config builder and change the doc there, the method would only call super but the doc will be accurate.

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
```diff
@@ -67,15 +67,14 @@ abstract class BaseConfigurationBuilder<
...
+     *
+     * @param amount Amount of the transaction. c cxzz
```
jreij, line 74: Was your command key stuck there? 😅

ozgur00: Oups apparently. 😅

components-core/src/main/java/com/adyen/checkout/components/base/BaseConfigurationBuilder.kt
jreij: I thought we had the same case somewhere else but apparently I removed it by mistake, [re-added it here](https://github.com/Adyen/adyen-android/pull/983) if you wanna check what I mean.

jreij (approved):

PR #978, opened 2022-12-20
Move `isConfirmationRequired` check to delegates

OscarSpruit:
## Description
The view shouldn't have this logic, so it makes sense to move it to the delegates. Also this is a part needed for a bigger refactor.

## Checklist 
- [x] Code is unit tested

COAND-691

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/ButtonComponent.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/ButtonComponent.kt
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 20/12/2022.
+ */
+
+package com.adyen.checkout.components
+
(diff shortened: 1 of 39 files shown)


components-core/src/main/java/com/adyen/checkout/components/ButtonComponent.kt
```diff
@@ -0,0 +1,26 @@
...
+     * If confirmation is not required, it means the view handles input in a way that the user has already expressed the
+     * desire to continue.
+     *
+     * Each type of view always returns the same value, so if the type of view is known, there is no need to check this
+     * method.
+     *
+     * @return If an update from the component attached to this View requires further user confirmation to continue or
+     * not.
+     */
+    fun isConfirmationRequired(): Boolean
```
jreij, line 25: Let's rewrite this documentation or put a TODO for it, some stuff in there are not applicable anymore

OscarSpruit: Added TODOs for now

jreij (approved):

PR #983, opened 2022-12-22
Fix docs for EPSConfiguration.Builder.setHideIssuerLogos

jreij:
## Description
The docs for `IssuerListConfiguration.Builder.setHideIssuerLogos` state that the default value is false, while for EPS that default value is actually true.

## Checklist 

COAND-549

### DIFF
--- a/eps/src/main/java/com/adyen/checkout/eps/EPSConfiguration.kt
+++ b/eps/src/main/java/com/adyen/checkout/eps/EPSConfiguration.kt
@@ -49,6 +49,17 @@ class EPSConfiguration private constructor(
         )
 
         /**
+         * Sets whether the logos should be shown next to the issuers name.
+         *
+         * Default is true.
+         *
+         * @param hideIssuerLogos if issuer logos should be hidden or not.
+         */
+        override fun setHideIssuerLogos(hideIssuerLogos: Boolean): Builder {
+            return super.setHideIssuerLogos(hideIssuerLogos)
+        }
+
+        /**
          * Builder with required parameters.
          *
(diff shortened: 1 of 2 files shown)


eps/src/main/java/com/adyen/checkout/eps/EPSConfiguration.kt
```diff
@@ -45,6 +45,17 @@ class EPSConfiguration private constructor(
             clientKey
         )
 
+        /**
+         * Sets whether the logos should be shows next to the issuers name.
```
OscarSpruit, line 49: `shows` -> `shown`

eps/src/main/java/com/adyen/checkout/eps/EPSConfiguration.kt
```diff
@@ -45,6 +45,17 @@ class EPSConfiguration private constructor(
...
+         *
+         * Default is true.
+         *
+         * @param hideIssuerLogos if issuer logos should be hidden or not.
+         */
+        override fun setHideIssuerLogos(hideIssuerLogos: Boolean): Builder {
```
OscarSpruit, line 55: Didn't it work without overriding?

jreij: Not sure how else to do it?

OscarSpruit: I was just curious, because I thought the base class handled it

OscarSpruit (approved):

PR #985, opened 2022-12-22
Make delegates in control of showing submit button

OscarSpruit:
## Description
`ButtonDelegate`s now have this `shouldShowSubmitButton` method that `AdyenComponentView` uses to show or hide the submit button.

`PayByBankDelegate` implemented the `ButtonDelegate` and `UIStateDelegate` interfaces, which it shouldn't. Pay by bank doesn't show a button, so there is no need for this.

## Checklist 
- [x] Code is unit tested

COAND-691


paybybank/src/main/java/com/adyen/checkout/paybybank/DefaultPayByBankDelegate.kt
```diff
@@ -201,6 +201,9 @@ internal class DefaultPayByBankDelegate(
 
     override fun isConfirmationRequired(): Boolean = _viewFlow.value is ButtonComponentViewType
 
+    // Pay by bank should not show a button as the entries navigate to the next step.
+    override fun shouldShowSubmitButton(): Boolean = false
```
jreij: Why can't we also use `isConfirmationRequired() && componentParams.isSubmitButtonVisible` here? The view type can either be null or `PayByBankComponentViewType` which is not a `ButtonComponentViewType`, so the result will be always false

jreij (approved):

PR #990, opened 2022-12-28
Refactor PreselectedStoredPaymentMethodFragment

jreij:
## Description
`PreselectedStoredPaymentMethodFragment` logic should be moved to the view model, and it should not rely on `requiresInput()` anymore, but instead we can make use of the component state.

## Checklist 
- [x] Code is unit tested

COAND-698

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PreselectedStoredPaymentViewModel.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PreselectedStoredPaymentViewModel.kt
@@ -16,8 +16,6 @@ import com.adyen.checkout.components.PaymentComponentState
 import com.adyen.checkout.components.channel.bufferedChannel
 import com.adyen.checkout.components.model.paymentmethods.StoredPaymentMethod
 import com.adyen.checkout.components.model.payments.Amount
-import com.adyen.checkout.components.util.CurrencyUtils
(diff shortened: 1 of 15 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/PreselectedStoredPaymentViewModel.kt
```diff
@@ -8,51 +8,114 @@
...
 import com.adyen.checkout.dropin.ui.stored.mapStoredModel
-import com.adyen.checkout.dropin.ui.viewmodel.PreselectedStoredState.Idle
-import com.adyen.checkout.dropin.ui.viewmodel.PreselectedStoredState.ShowStoredPaymentDialog
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.receiveAsFlow
+import java.util.Locale
 
 internal class PreselectedStoredPaymentViewModel(
```
OscarSpruit, line 29: Should we already unit test this class?

jreij: Done

OscarSpruit (approved):

PR #993, opened 2022-12-29
Stop default country picker from picking first alphabetically available country

etman55:
## Description

 - Stop default country picker to first alphabetically available country Instead of choosing the first available option.
- States are not selected by default to first option you need to select them.

## Checklist 
- [x] Code is unit tested

COAND-668

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
+++ b/card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
@@ -46,6 +46,8 @@ class AddressFormInput @JvmOverloads constructor(
 
     private lateinit var delegate: AddressDelegate
 
+    private var currentSpec: AddressSpecification? = null
+
     private var countryAdapter: SimpleTextListAdapter<AddressListItem> = SimpleTextListAdapter(context)
     private var statesAdapter: SimpleTextListAdapter<AddressListItem> = SimpleTextListAdapter(context)
 
@@ -120,6 +122,7 @@ class AddressFormInput @JvmOverloads constructor(
                     }
                     populateFormFields(AddressSpecification.fromString(selectedCountryCode))
                 }
+                textInputLayoutCountry?.hideError()
             }
         }
     }
@@ -140,8 +143,17 @@ class AddressFormInput @JvmOverloads constructor(
         this.localizedContext = localizedContext
     }
 
+    @Suppress("LongMethod")
     internal fun highlightValidationErrors(isErrorFocusedPreviously: Boolean) {
         var isErrorFocused = isErrorFocusedPreviously
+        val countryValidation = delegate.addressOutputData.country.validation
+        if (countryValidation is Validation.Invalid) {
+            if (!isErrorFocused) {
+                isErrorFocused = true
+                textInputLayoutCountry?.requestFocus()
+            }
+            textInputLayoutCountry?.showError(localizedContext.getString(countryValidation.reason))
+        }
         val streetValidation = delegate.addressOutputData.street.validation
         if (streetValidation is Validation.Invalid) {
             if (!isErrorFocused) {
@@ -188,19 +200,21 @@ class AddressFormInput @JvmOverloads constructor(
                 @Suppress("UNUSED_VALUE")
                 isErrorFocused = true
                 textInputLayoutProvinceTerritory?.requestFocus()
+                textInputLayoutState?.requestFocus()
             }
             textInputLayoutProvinceTerritory?.showError(localizedContext.getString(provinceTerritoryValidation.reason))
+            textInputLayoutState?.showError(localizedContext.getString(provinceTerritoryValidation.reason))
         }
     }
 
     private fun updateCountries(countryList: List<AddressListItem>) {
         countryAdapter.setItems(countryList)
-        countryList.firstOrNull { it.selected }?.let {
-            val selectedSpecification = AddressSpecification.fromString(it.code)
-            if (formContainer.childCount == 0) {
-                autoCompleteTextViewCountry.setText(it.name)
-                populateFormFields(selectedSpecification)
-            }
+        val selectedCountry = countryList.firstOrNull { it.selected }
+        val selectedSpecification = AddressSpecification.fromString(selectedCountry?.code)
+        if (selectedSpecification != currentSpec) {
+            currentSpec = selectedSpecification
+            autoCompleteTextViewCountry.setText(selectedCountry?.name)
+            populateFormFields(selectedSpecification)
         }
     }
 
@@ -408,6 +422,7 @@ class AddressFormInput @JvmOverloads constructor(
             setAdapter(statesAdapter)
             onItemClickListener = AdapterView.OnItemClickListener { _, _, position, _ ->
                 delegate.updateAddressInputData { stateOrProvince = statesAdapter.getItem(position).code }
+                textInputLayoutState?.hideError()
             }
         }
     }
--- a/card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
+++ b/card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
@@ -6,6 +6,7 @@ import com.adyen.checkout.card.AddressParams
 import com.adyen.checkout.card.api.model.AddressItem
 import com.adyen.checkout.card.ui.model.AddressListItem
 import com.adyen.checkout.components.model.payments.request.Address
+import java.util.Locale
 
 internal object AddressFormUtils {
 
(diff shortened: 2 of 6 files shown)


card/src/test/java/com/adyen/checkout/card/AddressFormUtilsTest.kt
```diff
@@ -49,7 +49,7 @@ internal class AddressFormUtilsTest {
     }
 
     @Test
-    fun markAddressListItemSelected_CodeNotProvided_ExpectFirstItemSelected() {
+    fun markAddressListItemSelected_CodeNotProvided_ExpectNothingSelected() {
```
ozgur00, line 52: Let's rewrite this test function names to match with our current convention

card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
```diff
@@ -177,19 +189,25 @@ class AddressFormInput @JvmOverloads constructor(
...
+            textInputLayoutState?.showError(localizedContext.getString(provinceTerritoryValidation.reason))
         }
     }
 
     internal fun updateCountries(countryList: List<AddressListItem>) {
         countryAdapter.setItems(countryList)
         countryList.firstOrNull { it.selected }?.let {
             val selectedSpecification = AddressSpecification.fromString(it.code)
-            if (formContainer.childCount == 0) {
+            if (selectedSpecification != currentSpec) {
```
ozgur00, line 203: Can you clean up this code a bit? We probably don't need the `run` block

card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
```diff
@@ -57,7 +58,7 @@ internal object AddressFormUtils {
                     countryList
                 }
 
-                val defaultCountryCode = addressConfiguration.defaultCountryCode.orEmpty()
+                val defaultCountryCode = addressConfiguration.defaultCountryCode ?: shopperLocale.country.orEmpty()
```
ozgur00, line 61: Maybe extract this `initialCountryCode` to a private function. In case we need to update the logic in the future it'll be easy to find.

card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
etman55: I guess we need the run block just for initial state when output has no countries.

etman55: I'll check again if we can make it better.

card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
```diff
@@ -57,13 +58,31 @@ internal object AddressFormUtils {
...
         }
     }
 
+    /**
+     * initialize country code.
+     *
+     * @param shopperLocale Locale value from dropIn configuration or card component configuration.
+     * @param addressConfiguration which has to be FullAddress as we only get the country on FullAddress mode.
+     *
+     * @return String.
```
ozgur00, line 77: Instead of just mentioning type of return value, can you tell about what this return value means?

card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
```diff
@@ -57,13 +58,31 @@ internal object AddressFormUtils {
...
+     */
+    private fun initialCountryCode(
```
ozgur00, line 79: Maybe rename it to `getInitialCountryCode`

card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
ozgur00: We don't need the null check here. At first when there's no selected country `AddressSpecification.fromString()` will return default. And then if there's a country it's gonna work as expected.

card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
```diff
@@ -57,13 +58,31 @@ internal object AddressFormUtils {
...
+                    addressConfiguration = addressConfiguration
+                )
                 markAddressListItemSelected(mapToListItem(filteredCountryList), defaultCountryCode)
             }
             else -> emptyList()
         }
     }
 
+    /**
+     * get initial country code.
```
ozgur00, line 72: We should start these docs by capital letter.

card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
```diff
@@ -57,13 +58,31 @@ internal object AddressFormUtils {
...
+     *
+     * @param shopperLocale Locale value from dropIn configuration or card component configuration.
+     * @param addressConfiguration
```
ozgur00, line 75: Optional: On second thought we could do something like: `@param addressConfiguration Full Address Configuration`

etman55: it's self explanatory we don't need to add that.

etman55: also it's limited to `AddressConfiguration.FullAddress` type

card/src/test/java/com/adyen/checkout/card/AddressFormUtilsTest.kt
```diff
@@ -88,7 +89,7 @@ internal class AddressFormUtilsTest {
     }
 
     @Test
-    fun markAddressListItemSelected_ListNotContainingItemWithGivenCode_ExpectFirstItemSelected() {
+    fun ` when there's no item selected on address list and list not containing item with given code expect nothing selected`() {
```
jreij, line 92: extra space before `when`

card/src/test/java/com/adyen/checkout/card/AddressFormUtilsTest.kt
```diff
@@ -210,7 +211,47 @@ internal class AddressFormUtilsTest {
         )
         assertEquals(
             expected,
-            AddressFormUtils.initializeCountryOptions(addressConfiguration, inputCountryList)
+            AddressFormUtils.initializeCountryOptions(Locale.GERMANY, addressConfiguration, inputCountryList)
+        )
+    }
+
+    @Test
+    fun ` when country options address configuration is full, address without default country code and locale with country that is supported expect list with locale country selected`() {
```
jreij, line 219: extra space before `when`

card/src/test/java/com/adyen/checkout/card/DefaultCardDelegateTest.kt
```diff
@@ -172,11 +173,31 @@ internal class DefaultCardDelegateTest(
...
+
+            delegate = createCardDelegate(
+                configuration = getDefaultCardConfigurationBuilder()
+                    .setAddressConfiguration(AddressConfiguration.FullAddress())
+                    .build()
+            )
+            delegate.initialize(CoroutineScope(UnconfinedTestDispatcher()))
+
+            assertEquals(TestAddressRepository.COUNTRIES, countriesFlow.awaitItem())
+            assertEquals(Unit, statesFlow.expectNoEvents())
```
jreij, line 189: Why are you trying to test the outcome of `expectNoEvents()`?

card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
```diff
@@ -46,6 +46,8 @@ class AddressFormInput @JvmOverloads constructor(
 
     private lateinit var delegate: AddressDelegate
 
+    private var currentSpec: AddressSpecification = AddressSpecification.fromString(null)
```
ozgur00, line 49: `currentSpec` should be null initially because before we get any update from output data we don't populate the view and for that state it should be null.

card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
```diff
@@ -188,20 +200,22 @@ class AddressFormInput @JvmOverloads constructor(
...
-                autoCompleteTextViewCountry.setText(it.name)
-                populateFormFields(selectedSpecification)
-            }
+        val selectedCountry = countryList.firstOrNull { it.selected }
+        val selectedSpecification = AddressSpecification.fromString(selectedCountry?.code)
+        if (selectedSpecification != currentSpec) {
+            currentSpec = selectedSpecification
+            autoCompleteTextViewCountry.setText(selectedCountry?.name)
         }
+        populateFormFields(currentSpec)
```
ozgur00, line 218: We should populate the fields in the if block, we don't have to do it every time.

card/src/test/java/com/adyen/checkout/card/DefaultCardDelegateTest.kt
etman55: As when there's no default countries so we're not selecting any country therefore no states request happening here.

jreij: Yes but this line doesn't make sense, check how we use `expectNoEvents()` in other places

ozgur00 (approved):

PR #995, opened 2023-01-02
Use new/simple way of action handling in Blik example

OscarSpruit:
## Description
Use new/simple way of action handling in Blik example. Also, fixes a crash in actions views that expect a specific type of action delegate.

## Checklist 

COAND-700

### DIFF
--- a/action/src/main/java/com/adyen/checkout/action/DefaultActionHandlingComponent.kt
+++ b/action/src/main/java/com/adyen/checkout/action/DefaultActionHandlingComponent.kt
@@ -32,6 +32,9 @@ class DefaultActionHandlingComponent(
     override fun handleAction(action: Action, activity: Activity) {
         activeDelegate = genericActionDelegate
         genericActionDelegate.handleAction(action, activity)
+        // genericActionDelegate.delegate is set when calling genericActionDelegate.handleAction, so we set the more
+        // specific delegate here as soon as we can.
+        activeDelegate = genericActionDelegate.delegate
     }
 
     override fun handleIntent(intent: Intent) {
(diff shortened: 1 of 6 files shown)


action/src/main/java/com/adyen/checkout/action/DefaultActionHandlingComponent.kt
```diff
@@ -32,6 +32,9 @@ class DefaultActionHandlingComponent(
     override fun handleAction(action: Action, activity: Activity) {
         activeDelegate = genericActionDelegate
```
jreij, line 33: Should this be removed?

OscarSpruit: Nope, the `PaymentInProgressView` will crash if we'd do that

example-app/src/main/java/com/adyen/checkout/example/ui/blik/BlikActivity.kt
```diff
@@ -68,28 +70,20 @@ class BlikActivity : AppCompatActivity() {
...
             }
             is BlikViewState.ShowComponent -> {
                 binding.progressIndicator.isVisible = false
                 binding.errorView.isVisible = false
-                binding.blikView.isVisible = true
+                binding.componentView.isVisible = true
 
                 setupBlikView(blikViewState.paymentMethod)
             }
             is BlikViewState.Await -> {
```
jreij, line 82: This should be changed to `BlikViewState.Action` and the `Unsupported` event should be removed

jreij (approved):

PR #997, opened 2023-01-03
Add example implementation of BACS

OscarSpruit:
## Description
Add example implementation of BACS. Also cleaned up some unnecessary things in the card example and fixed a bug in `VoucherView`.

## Checklist 

COAND-701

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsEvent.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsEvent.kt
@@ -6,15 +6,15 @@
  * Created by oscars on 7/11/2022.
  */
 
-package com.adyen.checkout.example.ui.card
+package com.adyen.checkout.example.ui.bacs
 
 import com.adyen.checkout.components.model.payments.response.Action
 
-internal sealed class CardEvent {
+internal sealed class BacsEvent {
 
-    data class PaymentResult(val result: String) : CardEvent()
+    data class PaymentResult(val result: String) : BacsEvent()
 
-    data class AdditionalAction(val action: Action) : CardEvent()
+    data class AdditionalAction(val action: Action) : BacsEvent()
 
-    object Invalid : CardEvent()
+    object Invalid : BacsEvent()
 }
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsEvent.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsEvent.kt
@@ -1,9 +1,9 @@
 /*
- * Copyright (c) 2022 Adyen N.V.
+ * Copyright (c) 2023 Adyen N.V.
  *
  * This file is open source and available under the MIT license. See the LICENSE file for more info.
  *
- * Created by oscars on 7/11/2022.
+ * Created by oscars on 3/1/2023.
  */
 
 package com.adyen.checkout.example.ui.bacs
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsFragment.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsFragment.kt
@@ -101,7 +101,12 @@ class BacsFragment : BottomSheetDialogFragment() {
         when (state) {
             is BacsViewState.Error -> {
                 binding.errorView.isVisible = true
-                binding.errorView.setText(state.message)
+                val errorMessage = if (state.arg != null) {
+                    getString(state.message, state.arg)
+                } else {
+                    getString(state.message)
+                }
+                binding.errorView.text = errorMessage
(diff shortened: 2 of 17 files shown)


example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsEvent.kt
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) 2022 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 7/11/2022.
```
jreij, line 6: Update copyright date

example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsFragment.kt
```diff
@@ -0,0 +1,139 @@
...
+    private var _binding: FragmentBacsBinding? = null
+    private val binding: FragmentBacsBinding get() = requireNotNull(_binding)
+
+    private val viewModel: BacsViewModel by viewModels()
+
+    private var bacsComponent: BacsDirectDebitComponent? = null
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        // Insert return url in extras, so we can access it in the ViewModel through SavedStateHandle
+        val returnUrl = RedirectComponent.getReturnUrl(requireActivity().applicationContext)
```
jreij, line 48: Should this have `+ "/bacs"`?

example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsFragment.kt
```diff
@@ -0,0 +1,139 @@
...
+                binding.componentView.isVisible = true
+                binding.progressIndicator.isVisible = false
+            }
+        }
+    }
+
+    private fun onEvent(event: BacsEvent) {
+        when (event) {
+            is BacsEvent.AdditionalAction -> bacsComponent?.handleAction(event.action, requireActivity())
+            BacsEvent.Invalid -> binding.componentView.highlightValidationErrors()
```
jreij, line 113: Just double checking, if `BacsEvent.Invalid` was emitted multiple times in succession, all of them will be handled here because the flow is not a state flow, is that correct?

example-app/src/main/res/values/strings.xml
```diff
@@ -8,6 +8,10 @@
     <string name="components_title">Components</string>
     <string name="card_component_entry">Card</string>
 
+    <string name="bacs_component_entry">BACS (using bottom sheet)</string>
+    <string name="bacs_country_error">Your country code must be GB</string>
```
jreij, line 12: Can we change this string and blik to take the country code as a parameter? Same for the other string and currency code

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -68,8 +68,8 @@ internal class CheckoutConfigurationProvider @Inject constructor(
     fun getBlikConfiguration(): BlikConfiguration =
         BlikConfiguration.Builder(shopperLocale, environment, clientKey).build()
 
-    fun getAwaitConfiguration(): AwaitConfiguration =
-        AwaitConfiguration.Builder(shopperLocale, environment, clientKey).build()
+    fun getBacsConfiguration(): BacsDirectDebitConfiguration =
```
jreij, line 71: Do we need this?

example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsFragment.kt
OscarSpruit: Yes that's right

example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsEvent.kt
OscarSpruit: Hmm interesting

example-app/src/main/java/com/adyen/checkout/example/ui/bacs/BacsFragment.kt
OscarSpruit: Not sure, because a fragment can't really handle the return url. I figured the default one would do, so we could potentially handle it in `MainActivity` then. Also, I didn't really implement the return url flow, because I thought BACS doesn't have it (currently)

jreij: Hmm ok then we'll look at it later with some component that produces actions

jreij (approved):

PR #1001, opened 2023-01-05
The address logics were moved from the card delegate to the address delegate

onurkaragunlu:
This change was made to use AddressFormInput in other modules.
## Checklist 
- [x] Code is unit tested

COAND-368

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/AddressDelegate.kt
+++ b/card/src/main/java/com/adyen/checkout/card/AddressDelegate.kt
@@ -8,9 +8,10 @@
 
 package com.adyen.checkout.card
 
-
+import androidx.annotation.RestrictTo
 import kotlinx.coroutines.flow.Flow
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 interface AddressDelegate {
     val addressOutputData: AddressOutputData
     val addressOutputDataFlow: Flow<AddressOutputData>
--- a/card/src/main/java/com/adyen/checkout/card/AddressDelegate.kt
+++ b/card/src/main/java/com/adyen/checkout/card/AddressDelegate.kt
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2023 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by onurk on 3/1/2023.
+ */
+
+package com.adyen.checkout.card
+
+
+import kotlinx.coroutines.flow.Flow
+
+interface AddressDelegate {
+    val addressOutputData: AddressOutputData
+    val addressOutputDataFlow: Flow<AddressOutputData>
+    fun updateAddressInputData(update: AddressInputModel.() -> Unit)
+}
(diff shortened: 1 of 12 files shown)


card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -51,7 +51,7 @@ import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.launch
 
-@Suppress("TooManyFunctions", "LongParameterList")
+@Suppress("TooManyFunctions", "LongParameterList", "EmptyFunctionBlock")
```
jreij, line 54: You can put a `// no ops` comment in that empty function block, this way you won't need his this rule suppression and we'll know that it's empty on purpose.

card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -84,6 +87,15 @@ internal class DefaultCardDelegate(
     private val _outputDataFlow = MutableStateFlow(createOutputData())
     override val outputDataFlow: Flow<CardOutputData> = _outputDataFlow
 
+    override val addressOutputData: AddressOutputData
+        get() = outputData.addressState
+
+    override val addressOutputDataFlow: Flow<AddressOutputData> by lazy {
```
jreij, line 93: Why do we need `by lazy` here?

onurkaragunlu: If this variable is not lazy, it runs before we set the coroutinescope in the initialize method, so it crashes beacuse coroutineScope is null. I need to run this line after the initialize method, so I set it lazy to prevent a crash.

card/src/main/java/com/adyen/checkout/card/ui/AddressFormInput.kt
```diff
@@ -110,19 +114,27 @@ class AddressFormInput @JvmOverloads constructor(
...
     }
 
-    internal fun attachDelegate(delegate: CardDelegate) {
+    internal fun attachDelegate(delegate: AddressDelegate, coroutineScope: CoroutineScope) {
         this.delegate = delegate
+        subscribeCountryAndStateList(coroutineScope)
+    }
+
+    private fun subscribeCountryAndStateList(coroutineScope: CoroutineScope) {
+        delegate.addressOutputDataFlow.distinctUntilChanged().onEach { addressOutputData ->
```
ozgur00, line 134: Doesn't `StateFlow`s have `distinctUntilChanged()` applied by default? `distinctUntilChanged()` might be unnecessary here

card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -83,6 +83,11 @@ internal class StoredCardDelegate(
     private val _outputDataFlow = MutableStateFlow(createOutputData())
     override val outputDataFlow: Flow<CardOutputData> = _outputDataFlow
 
+    override val addressOutputData: AddressOutputData
+        get() = _outputDataFlow.value.addressState
```
ozgur00, line 87: You can use `outputData.addressState` here directly.

jreij (approved):

card/src/main/java/com/adyen/checkout/card/AddressDelegate.kt
```diff
@@ -0,0 +1,17 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by onurk on 3/1/2023.
+ */
+
+package com.adyen.checkout.card
+
+import kotlinx.coroutines.flow.Flow
+
+interface AddressDelegate {
```
ozgur00, line 13: Can we add an annotation for restricting this interface to the library scope?

ozgur00 (approved):

PR #1002, opened 2023-01-09
Components + sessions: calling APIs internally

jreij:
## Description
Initial implementation of components in sessions. In this branch we added the ability to initialize the Card Component with a session, and the logic to internally call the session APIs.

## Checklist 

COAND-699


card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
```diff
@@ -101,11 +109,97 @@ class CardComponentProvider(
...
                 genericActionDelegate = genericActionDelegate,
                 actionHandlingComponent = DefaultActionHandlingComponent(genericActionDelegate, cardDelegate),
+                savedStateHandle = savedStateHandle,
             )
         }
         return ViewModelProvider(viewModelStoreOwner, factory)[key, CardComponent::class.java]
     }
 
+    @Suppress("LongParameterList", "LongMethod")
+    fun get(
```
OscarSpruit, line 119: The original `get` method and this one share a lot of code, we should see if we can extract the similar things

OscarSpruit: Addressed it in 5e47fd8ecde09fca462588bdb0169c71c19d49e1

sessions-core/src/main/java/com/adyen/checkout/sessions/CheckoutSession.kt
```diff
@@ -8,9 +8,16 @@
...
 
+import com.adyen.checkout.components.model.paymentmethods.PaymentMethod
 import com.adyen.checkout.sessions.model.setup.SessionSetupResponse
 
 // TODO docs
 data class CheckoutSession(
     val sessionSetupResponse: SessionSetupResponse
-)
+) {
+    fun getPaymentMethod(paymentMethodType: String): PaymentMethod? {
```
ozgur00, line 18: Shall we add this helper for stored payment methods too?
```kotlin
    fun getStoredPaymentMethod(predicate: (StoredPaymentMethod) -> Boolean): StoredPaymentMethod? {
        return sessionSetupResponse.paymentMethods?.storedPaymentMethods.orEmpty().firstOrNull {
            predicate(it)
        }
    }
```

OscarSpruit: Why do we need the predicate again?

ozgur00: For stored payment methods we don't know what they're gonna use to choose the payment method they want. I think this is the best option we can provide for that so they can pass the condition they desire.

ozgur00 (approved):

PR #1004, opened 2023-01-11
Add Address Params

onurkaragunlu:
it was added address params to use address input view in other modules. it was changed from the address configuration to address params in cardDelegate
## Checklist 
- [x] Code is unit tested

COAND-368

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/AddressFormUIState.kt
+++ b/card/src/main/java/com/adyen/checkout/card/AddressFormUIState.kt
@@ -7,15 +7,15 @@ enum class AddressFormUIState {
         /**
          * Get visibility state of the address form.
          *
-         * @param addressConfiguration Configuration object for address form.
+         * @param addressParams Configuration object for address form.
          *
          * @return Visibility state of the address form.
          */
-        fun fromAddressConfiguration(addressConfiguration: AddressConfiguration): AddressFormUIState {
-            return when (addressConfiguration) {
-                is AddressConfiguration.FullAddress -> FULL_ADDRESS
-                is AddressConfiguration.PostalCode -> POSTAL_CODE
-                is AddressConfiguration.None -> NONE
+        fun fromAddressParams(addressParams: AddressParams): AddressFormUIState {
+            return when (addressParams) {
+                is AddressParams.FullAddress -> FULL_ADDRESS
+                is AddressParams.PostalCode -> POSTAL_CODE
+                is AddressParams.None -> NONE
             }
         }
     }
--- a/card/src/main/java/com/adyen/checkout/card/AddressFormUIState.kt
+++ b/card/src/main/java/com/adyen/checkout/card/AddressFormUIState.kt
@@ -1,5 +1,8 @@
 package com.adyen.checkout.card
 
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 enum class AddressFormUIState {
     NONE, POSTAL_CODE, FULL_ADDRESS;
 
--- a/card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
+++ b/card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
@@ -20,6 +20,7 @@ import com.adyen.checkout.card.ui.model.AddressListItem
 import com.adyen.checkout.card.ui.model.CardListItem
 import com.adyen.checkout.card.util.AddressFormUtils
 import com.adyen.checkout.card.util.AddressValidationUtils
+import com.adyen.checkout.card.util.CardAddressValidationUtils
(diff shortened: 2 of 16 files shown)


card/src/main/java/com/adyen/checkout/card/CardAddressConfiguration.kt
```diff
@@ -7,21 +7,21 @@ import kotlinx.parcelize.Parcelize
  * Configuration class for Address Form in Card Component. This class can be used define the
  * visibility of the address form.
  */
-sealed class AddressConfiguration : Parcelable {
+sealed class CardAddressConfiguration : Parcelable {
```
ozgur00, line 10: Can't we use `AddressConfiguration` as the name and it would get resolved because it's in the `card` package?

card/src/main/java/com/adyen/checkout/card/AddressFormUIState.kt
```diff
@@ -7,15 +7,15 @@ enum class AddressFormUIState {
...
+         * @param addressParams Configuration object for address form.
          *
          * @return Visibility state of the address form.
          */
-        fun fromAddressConfiguration(addressConfiguration: AddressConfiguration): AddressFormUIState {
-            return when (addressConfiguration) {
-                is AddressConfiguration.FullAddress -> FULL_ADDRESS
-                is AddressConfiguration.PostalCode -> POSTAL_CODE
-                is AddressConfiguration.None -> NONE
+        fun fromAddressConfiguration(addressParams: AddressParams): AddressFormUIState {
```
ozgur00, line 14: We should rename the function as well :)

card/src/main/java/com/adyen/checkout/card/CardAddressFieldPolicyParams.kt
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Created by onurk on 6/1/2023.
+ */
+
+package com.adyen.checkout.card
+
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+sealed class CardAddressFieldPolicyParams : AddressFieldPolicy {
```
ozgur00, line 14: Same thing with naming goes here as well. These names are a bit long.

card/src/main/java/com/adyen/checkout/card/DefaultCardDelegate.kt
```diff
@@ -512,16 +512,45 @@ internal class DefaultCardDelegate(
         countryOptions: List<AddressListItem>,
         stateOptions: List<AddressListItem>
     ): AddressOutputData {
+        val isOptional = when (componentParams.addressParams) {
```
ozgur00, line 515: This when logic could be part of another function. Let's keep the delegate clean and extract these segments to somewhere else. You can put this and `isAddressOptional()` in `CardValidationUtils` for now.

card/src/main/java/com/adyen/checkout/card/util/AddressFormUtils.kt
```diff
@@ -31,33 +31,33 @@ internal object AddressFormUtils {
...
      *
-     * First filter if there's [AddressConfiguration.FullAddress.supportedCountryCodes] defined in the
-     * configuration. Then mark [AddressConfiguration.FullAddress.defaultCountryCode] as selected if it
+     * First filter if there's [AddressParams.FullAddress.supportedCountryCodes] defined in the
+     * configuration. Then mark [AddressParams.FullAddress.defaultCountryCode] as selected if it
      * is defined in configuration and exists in the filtered country list. Otherwise mark first item
      * in the list as selected.
      *
-     * @param addressConfiguration Configuration object.
+     * @param addressParams Configuration object.
```
ozgur00, line 41: Update the kdoc to reflect the return type :)

card/src/test/java/com/adyen/checkout/card/AddressFormUtilsTest.kt
```diff
@@ -127,8 +127,8 @@ internal class AddressFormUtilsTest {
     }
 
     @Test
-    fun initializeCountryOptions_AddressConfigurationIsNone_ExpectEmptyList() {
-        val addressConfiguration = AddressConfiguration.None
+    fun `When Country Options are initialized that a Address Params is null ,Expect Empty List`() {
```
ozgur00, line 130: Tests should start with lower case as it's the convention throughout the whole project.

card/src/main/java/com/adyen/checkout/card/CardAddressConfiguration.kt
onurkaragunlu: Do you think we don't need to add prefix like "payment Method Name- Address Configuration?" . All payment method which use a address configuration can have the same name but they should be in a different package, right?

ozgur00 (dismissed):

ozgur00 (approved):

PR #1018, opened 2023-01-17
Remove unnecessary Suppress annotations

etman55:
## Description
Remove unnecessary Suppress annotations

## Checklist 

COAND-624

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
@@ -32,7 +32,6 @@ import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListItem.Compani
 import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListItem.Companion.PAYMENT_METHODS_NOTE
 import com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListItem.Companion.STORED_PAYMENT_METHOD
 
-@SuppressWarnings("TooManyFunctions")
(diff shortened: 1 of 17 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodAdapter.kt
```diff
@@ -117,6 +117,7 @@ internal class PaymentMethodAdapter @JvmOverloads constructor(
             }
         }
 
+        @SuppressLint("StringFormatInvalid")
```
OscarSpruit, line 120: Why are these suddenly needed?

etman55: I found an error over them that for example instead of using `context.getString(R.string.card_number_4digit, model.lastFour)` for example it recommends to change it to `String.format`

etman55: after some research for this lint highlight is looks like a lint bug and when I just cleaned project it went away. Will remove it.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/stored/PreselectedStoredPaymentMethodFragment.kt
```diff
@@ -111,6 +110,7 @@ internal class PreselectedStoredPaymentMethodFragment : DropInBottomSheetDialogF
         }.launchIn(viewLifecycleOwner.lifecycleScope)
     }
 
+    @SuppressLint("StringFormatInvalid")
```
OscarSpruit, line 113: I guess this one can also be removed?

OscarSpruit (approved):

PR #1019, opened 2023-01-18
Exclude external dependencies from components

OscarSpruit:
## Description
External dependencies (like WeChat and 3DS2) shouldn't be shipped with the standalone components, because it makes the size of the SDKs bigger and we don't want to force 3rd party dependencies.

## Checklist 

COAND-695


RELEASE_NOTES.md
```diff
@@ -13,6 +13,11 @@
   | Name                                                                                                   | Version                       |
   |--------------------------------------------------------------------------------------------------------|-------------------------------|
   | [Android Gradle plugin](https://developer.android.com/build/releases/gradle-plugin)                    | **8.0.1** (requires Java 17)  |
+- 3DS2 and WeChat are no longer shipped with standalone components by default. In case you want to handle 3DS2 and/or WeChat actions you have to add the dependencies in your `build.gradle`:
```
jreij, line 16: This needs clarification

checkout-core/src/main/java/com/adyen/checkout/core/internal/util/RunCompileOnly.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package com.adyen.checkout.core.internal.util
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+inline fun <R> runCompileOnly(block: () -> R): R? {
+    try {
+        return block()
```
jreij, line 16: One limitation of this is if the block itself can return `null` right?

jreij:
Weill be also use this method in the places where we fetch the 3DS2 SDK version?

checkout-core/src/main/java/com/adyen/checkout/core/internal/util/RunCompileOnly.kt
OscarSpruit: > The default upper bound (if there was none specified) is `Any?` ([source](https://kotlinlang.org/docs/generics.html#generic-constraints))

So by default `R` is nullable here and `block` is allowed to return `null`. If we specify `<R : Any>` instead `block` can no longer return null

jreij: Maybe it's better to force it to be non nullable? Or make `runCompileOnly` return a sealed class instead of null or not null? I'm fine with either.

jreij (approved):

PR #1022, opened 2023-01-23
Fix card examples becoming empty when re-opening the app

OscarSpruit:
## Description
Prevent from components being attached multiple times to `AdyenComponentView`.

Bonus fix: handle back press correctly in BACS

## Checklist 

COAND-711

### DIFF
--- a/ui-core/src/main/java/com/adyen/checkout/components/ui/view/AdyenComponentView.kt
+++ b/ui-core/src/main/java/com/adyen/checkout/components/ui/view/AdyenComponentView.kt
@@ -36,6 +36,7 @@ import com.adyen.checkout.core.log.Logger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
+import java.lang.ref.WeakReference
 
 /**
  * A View that can display input and fill in details for a Component.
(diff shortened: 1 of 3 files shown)


jreij (approved):

etman55:
I think we should come up with another solution as we don't want for the flow to be observed when the app is in the background as `CREATYED` gets triggered just two time after `onCreate` and before `onStop` as mentioned in the [docs](https://developer.android.com/reference/android/arch/lifecycle/Lifecycle.State#created)

ui-core/src/main/java/com/adyen/checkout/components/ui/view/AdyenComponentView.kt
```diff
@@ -59,7 +60,8 @@ class AdyenComponentView @JvmOverloads constructor(
     private var isInteractionBlocked = false
 
     private var componentView: ComponentView? = null
-    private var componentViewType: ComponentViewType? = null
+
+    private var attachedComponent = WeakReference<Component?>(null)
```
etman55, line 64: why did you remove `componentViewType`?

OscarSpruit: It wasn't used

jreij (approved):

PR #1036, opened 2023-01-31
Remove hard-coded brand support on card component/drop-in Android SDK (V4)

etman55:
## Description

- Fix supported card brands that are not part of predefined types inside `CardBrand.CardType` enum.
- Add `SupportedCardBrands` inside `CardConfiguration`
- Add new method `CardConfiguration.setSupportedCardTypes` with new argument `CardBrand` to allow merchants to add more `supported` types that are not predefined inside `CardType` enum.

## Checklist 
- [x] Code is unit tested

COAND-708

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
+++ b/card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
@@ -11,7 +11,7 @@ import android.os.Bundle
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.ViewModelStoreOwner
 import androidx.savedstate.SavedStateRegistryOwner
-import com.adyen.checkout.card.data.CardType
+import com.adyen.checkout.card.data.CardBrand
 import com.adyen.checkout.card.repository.AddressRepository
 import com.adyen.checkout.card.repository.BinLookupRepository
 import com.adyen.checkout.components.StoredPaymentComponentProvider
@@ -101,30 +101,28 @@ class CardComponentProvider : StoredPaymentComponentProvider<CardComponent, Card
      * @return The Configuration object with possibly adjusted values.
      */
     private fun checkSupportedCardTypes(paymentMethod: PaymentMethod, cardConfiguration: CardConfiguration): CardConfiguration {
-        if (cardConfiguration.supportedCardTypes.isNotEmpty()) {
+        if (cardConfiguration.supportedCardBrands.isNotEmpty()) {
             return cardConfiguration
         }
 
         val brands = paymentMethod.brands
-        var supportedCardTypes = CardConfiguration.DEFAULT_SUPPORTED_CARDS_LIST
+        var supportedCardBrands = CardConfiguration.DEFAULT_SUPPORTED_CARDS_LIST.map {
+            CardBrand(it)
+        }
 
-        // Get card types from brands in PaymentMethod object
+        // Get card brands from brands in PaymentMethod object
         if (!brands.isNullOrEmpty()) {
-            supportedCardTypes = arrayListOf()
+            supportedCardBrands = arrayListOf()
             for (brand in brands) {
-                val brandType = CardType.getByBrandName(brand)
-                if (brandType != null) {
-                    supportedCardTypes.add(brandType)
-                } else {
-                    Logger.e(TAG, "Failed to get card type for brand: $brand")
-                }
+                val cardBrand = CardBrand(brand)
+                supportedCardBrands.add(cardBrand)
             }
         } else {
             Logger.d(TAG, "Falling back to DEFAULT_SUPPORTED_CARDS_LIST")
         }
         @Suppress("SpreadOperator")
         return cardConfiguration.newBuilder()
-            .setSupportedCardTypes(*supportedCardTypes.toTypedArray())
+            .setSupportedCardTypes(*supportedCardBrands.toTypedArray())
             .build()
     }
 }
--- a/card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
+++ b/card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
@@ -15,6 +15,7 @@ import android.os.Parcelable;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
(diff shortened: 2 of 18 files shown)


card/src/main/java/com/adyen/checkout/card/repository/BinLookupRepository.kt
```diff
@@ -13,7 +13,7 @@ import com.adyen.checkout.card.api.BinLookupConnection
 import com.adyen.checkout.card.api.model.BinLookupRequest
 import com.adyen.checkout.card.api.model.BinLookupResponse
 import com.adyen.checkout.card.api.model.Brand
-import com.adyen.checkout.card.data.CardType
+import com.adyen.checkout.card.data.CardBrand.Companion.getByBrandName
```
jreij, line 16: We shouldn't have this static function import

card/src/main/java/com/adyen/checkout/card/StoredCardDelegate.kt
```diff
@@ -9,6 +9,7 @@
 package com.adyen.checkout.card
 
 import com.adyen.checkout.card.api.model.Brand
+import com.adyen.checkout.card.data.CardBrand.Companion.getByBrandName
```
jreij, line 12: We shouldn't have this static function import

card/src/main/java/com/adyen/checkout/card/CardComponentProvider.kt
```diff
@@ -11,7 +11,8 @@ import android.os.Bundle
 import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.ViewModelStoreOwner
 import androidx.savedstate.SavedStateRegistryOwner
-import com.adyen.checkout.card.data.CardType
+import com.adyen.checkout.card.data.CardBrand
+import com.adyen.checkout.card.data.CardBrand.Companion.getByBrandName
```
jreij, line 15: We shouldn't have this static function import

card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
```diff
@@ -265,6 +283,26 @@ public Builder setEnvironment(@NonNull Environment builderEnvironment) {
...
+                cardBrands.add(new CardBrand(cardType));
+            }
+            mBuilderSupportedCardBrands.clear();
+            mBuilderSupportedCardBrands.addAll(cardBrands);
+            return this;
+        }
+
+
+        /**
+         * This is the same as setSupportedCardTypes just with another type {@link CardBrand}
```
jreij, line 297: We should add the same docs from the original `setSupportedCardTypes` method here, and then explain that they are the same

card/src/main/java/com/adyen/checkout/card/CardConfiguration.java
```diff
@@ -265,6 +283,25 @@ public Builder setEnvironment(@NonNull Environment builderEnvironment) {
...
+                cardBrands.add(new CardBrand(cardType));
+            }
+            mBuilderSupportedCardBrands.clear();
+            mBuilderSupportedCardBrands.addAll(cardBrands);
+            return this;
+        }
+
+        /**
+         * Set the supported card types for this payment. Supported types will be shown as user inputs the card number.
+         * Use this method when adding supported card type that's not inside {@link CardType} enum.
```
jreij, line 297: `supported types that are not inside the...`

card/src/main/java/com/adyen/checkout/card/data/CardType.java
```diff
@@ -58,6 +58,7 @@ public enum CardType {
     VISA("visa", Pattern.compile("^4[0-9]{0,18}$")),
     VISADANKORT("visadankort", Pattern.compile("^(4571)[0-9]{0,12}$")),
     // UNKNOWN type is used for txVariants that are valid but not accounted for in this enum
+    @Deprecated
```
jreij, line 61: We should elaborate on why these fields are deprecated and what to use instead.

jreij:
You can add the release notes to this PR

card/src/main/java/com/adyen/checkout/card/CardComponentState.kt
```diff
@@ -19,7 +19,7 @@ class CardComponentState(
     paymentComponentData: PaymentComponentData<CardPaymentMethod>,
     isInputValid: Boolean,
     isReady: Boolean,
-    val cardType: CardType?,
+    val cardBrand: CardBrand?,
```
jreij, line 22: This is a breaking change

etman55: good catch

card/src/main/java/com/adyen/checkout/card/data/CardBrand.kt
```diff
@@ -0,0 +1,52 @@
...
+            override fun newArray(size: Int): Array<CardBrand?> {
+                return arrayOfNulls(size)
+            }
+        }
+
+        /**
+         * Get [CardBrand] from the brand name as it appears in the Checkout API.
+         */
+        @JvmStatic
+        fun getByBrandName(brand: String): CardBrand {
```
jreij, line 43: Why do we still have this method?

etman55: Because we still sometimes need to compare some card types with the ones inside `CardType` so we need to map brands coming from checkout api to the corresponding `CardType` if needed.

etman55: For example in `CardView` line 278

jreij: Yes but `CardBrand` has a `cardType` filed inside it and can be initialized with a string `txVariant`, shouldn't this be enough to achieve what we want?

jreij: To explain more, with the current code `CardBrand("visa").cardType == null` but  `CardBrand. getByBrandName("visa").cardType == CardType.VISA`. Can you see the issue with this method better now?

jreij (approved):

PR #1039, opened 2023-02-01
Remove hard-coded brand support on card component (V5)

etman55:
## Description
- Rename `CardType -> CardBrand` and `CardBrand -> CardType` 
- Add a new method inside `CardConfiguration` `setSupportedCardType(CardBrand)` so now we have two methods one with `CardBrand` and another one with `CardType` to make it easy for migration from V4 to V5 
PS: `CardBrand` is the source of truth internally when adding `SupportCardTypes`.

## Checklist 
- [x] Code is unit tested

COAND-708

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
+++ b/card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
@@ -34,7 +34,7 @@ class CardConfiguration private constructor(
     override val amount: Amount,
     override val isSubmitButtonVisible: Boolean?,
     val isHolderNameRequired: Boolean?,
-    val supportedCardTypes: List<CardType>?,
+    val supportedCardBrands: List<CardBrand>?,
     val shopperReference: String?,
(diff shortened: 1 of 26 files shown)


card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
```diff
@@ -91,6 +91,21 @@ class CardConfiguration private constructor(
             clientKey: String
         ) : super(shopperLocale, environment, clientKey)
 
+        /**
+         * Set the supported card types for this payment. Supported types will be shown as user inputs the card number.
+         *
+         * Defaults to [PaymentMethod.brands] if it exists, or [DEFAULT_SUPPORTED_CARDS_LIST] otherwise.
+         *
+         * Use this method when adding supported card type that's not inside [CardType] enum.
```
jreij, line 99: Same comment as the other PR

jreij (dismissed):

jreij (approved):

PR #1042, opened 2023-02-02
Cancel button in redirect view not working in components

jreij:
## Description

The cancel button in the redirect view does not work with standalone components. `delegate.onError` gets called from `PaymentInProgressView` but the delegate there is the `GenericActionDelegate` which does not implement `onError`.

## Checklist 

COAND-720

### DIFF
--- a/action/src/main/java/com/adyen/checkout/action/DefaultGenericActionDelegate.kt
+++ b/action/src/main/java/com/adyen/checkout/action/DefaultGenericActionDelegate.kt
@@ -179,6 +179,10 @@ internal class DefaultGenericActionDelegate(
         delegate.refreshStatus()
     }
 
+    override fun onError(e: CheckoutException) {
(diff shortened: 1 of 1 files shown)


jreij:
@OscarSpruit one thing I noticed while testing this is that it doesn't happen with drop-in. The reason is that we use the `GenericActionComponent` there as opposed to the standalone components in the other examples.
With the `GenericActionComponent` the views there get attached to the specific action delegate they need, which with the standalone components they get attached to the `GenericActionDelegate`.
I'm worried if this might cause any issues for us later? I saw the comment in `DefaultActionHandlingComponent.handleAction` but II don't think it's working as intended. Or at least it doesn't work for the views. WDYT?

jreij:
@OscarSpruit should we merge this one then and work on a different fix later?

OscarSpruit (approved):

PR #1043, opened 2023-02-02
Release 4.10.0 🎉

etman55:
## Release Notes

# New
* The `CardComponentState.binValue` now reports 8 digit bins in case of card numbers with 16 or more digits.
* We added `CardBrand`, along with `CardType`, to use when adding new card brands through `CardConfiguration`.
* In `CardConfiguration` when adding supported card types you can now add new supported card types which are not part of `CardType`.
  example:

```kotlin
CardConfiguration.Builder([SHOPPER_LOCALE], [ENVIRONMENT], [CLIENT_KEY])
.setSupportedCardTypes(CardBrand(txVariant = "[CARD_BRAND1]"), CardBrand(txVariant = "[CARD_BRAND2]"))
.build()
```

## Fixed
* Undefined supported card types now works correctly for card component/drop-in.
  `CardType.UNKNOWN` was not working correctly when there were supported card brands coming from `/paymentMethods`, under `scheme` payment method,
  which are not part of `CardType` predefined brands, so we had to add `CardBrand` to make these undefined supported card types work.

## Deprecated
* The `CardType.UNKNOWN` property. Use `CardBrand(txVariant = "[CARD_BRAND]")` instead.
* The `CardType.setTxVariant()` method. No longer needed as it was used with `CardType.UNKNOWN`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,6 +9,7 @@
 [//]: # ( - Configurations public constructor are deprecated, please use each Configuration's builder to make a Configuration object)
 
 # New
+* The `CardComponentState.binValue` now reports 8 digit bins in case of card numbers with 16 or more digits.
 * We added `CardBrand`, along with `CardType`, to use when adding new card brands through `CardConfiguration`.
 * In `CardConfiguration` when adding supported card types you can now add new supported card types which are not part of `CardType`.
   example:
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -26,4 +26,4 @@ CardConfiguration.Builder([SHOPPER_LOCALE], [ENVIRONMENT], [CLIENT_KEY])
 
 ## Deprecated
 * The `CardType.UNKNOWN` property. Use `CardBrand(txVariant = "[CARD_BRAND]")` instead.
-* The `CardType.setTxVariant()` method. No longer needed as it was used with `CardType.UNKNOWN`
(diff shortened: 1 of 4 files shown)


jreij:
Is the card type fix the only thing we're releasing?

etman55:
Yes, AFAIK.

RELEASE_NOTES.md
```diff
@@ -26,4 +27,4 @@ CardConfiguration.Builder([SHOPPER_LOCALE], [ENVIRONMENT], [CLIENT_KEY])
 
 ## Deprecated
 * The `CardType.UNKNOWN` property. Use `CardBrand(txVariant = "[CARD_BRAND]")` instead.
-* The `CardType.setTxVariant()` method. No longer needed as it was used with `CardType.UNKNOWN`
\ No newline at end of file
+* The `CardType.setTxVariant()` method. No longer needed as it was used with `CardType.UNKNOWN`
```
jreij, line 30: Missing`.` at the end

build.gradle
```diff
@@ -45,7 +45,7 @@ allprojects {
     // just for example app, don't need to increment
     ext.version_code = 1
     // The version_name format is "major.minor.patch(-(alpha|beta|rc)[0-9]{2}){0,1}" (e.g. 3.0.0, 3.1.1-alpha04 or 3.1.4-rc01 etc).
-    ext.version_name = "4.9.1"
+    ext.version_name = "4.9.2"
```
jreij, line 48: We should make this `4.10.0` since we introduce new functionality

jreij:
The branch is still named `release/4.9.2`, I think we should rename it

PR #1054, opened 2023-02-09
Display store card checkbox based on session setup response

etman55:
## Description
- Display store card checkbox based on session setup response.

## Checklist 
- [x] Code is unit tested

COAND-614

PS: update checkout api version to v70 to be able to test.

### DIFF
--- a/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
+++ b/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
@@ -132,7 +132,10 @@ class BcmcComponentProvider(
     ): BcmcComponent {
         assertSupported(paymentMethod)
 
-        val componentParams = componentParamsMapper.mapToParams(configuration)
+        val componentParams = componentParamsMapper.mapToParams(
+            configuration,
+            checkoutSession.sessionSetupResponse.configuration
+        )
         val httpClient = HttpClientFactory.getHttpClient(componentParams.environment)
         val publicKeyService = PublicKeyService(httpClient)
         val publicKeyRepository = DefaultPublicKeyRepository(publicKeyService)
--- a/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
+++ b/bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
@@ -44,17 +44,19 @@ import com.adyen.checkout.sessions.SessionComponentEventHandler
 import com.adyen.checkout.sessions.SessionSavedStateHandleContainer
 import com.adyen.checkout.sessions.api.SessionService
 import com.adyen.checkout.sessions.interactor.SessionInteractor
+import com.adyen.checkout.sessions.model.setup.SessionSetupConfiguration
 import com.adyen.checkout.sessions.provider.SessionPaymentComponentProvider
 import com.adyen.checkout.sessions.repository.SessionRepository
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class BcmcComponentProvider(
-    overrideComponentParams: ComponentParams? = null,
+    private val overrideComponentParams: ComponentParams? = null,
+    private val sessionSetupConfiguration: SessionSetupConfiguration? = null
 ) :
     PaymentComponentProvider<BcmcComponent, BcmcConfiguration, PaymentComponentState<CardPaymentMethod>>,
     SessionPaymentComponentProvider<BcmcComponent, BcmcConfiguration, PaymentComponentState<CardPaymentMethod>> {
 
-    private val componentParamsMapper = BcmcComponentParamsMapper(overrideComponentParams)
+    private val componentParamsMapper = BcmcComponentParamsMapper()
 
     override fun get(
         savedStateRegistryOwner: SavedStateRegistryOwner,
@@ -69,7 +71,11 @@ class BcmcComponentProvider(
     ): BcmcComponent {
         assertSupported(paymentMethod)
 
-        val componentParams = componentParamsMapper.mapToParams(configuration)
+        val componentParams = componentParamsMapper.mapToParams(
+            configuration,
+            overrideComponentParams,
+            sessionSetupConfiguration
+        )
         val httpClient = HttpClientFactory.getHttpClient(componentParams.environment)
         val publicKeyService = PublicKeyService(httpClient)
         val publicKeyRepository = DefaultPublicKeyRepository(publicKeyService)
@@ -134,6 +140,7 @@ class BcmcComponentProvider(
 
         val componentParams = componentParamsMapper.mapToParams(
             configuration,
+            overrideComponentParams,
             checkoutSession.sessionSetupResponse.configuration
         )
         val httpClient = HttpClientFactory.getHttpClient(componentParams.environment)
--- a/bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
+++ b/bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
@@ -26,7 +26,7 @@ internal class BcmcComponentParamsMapperTest {
         )
             .build()
 
-        val params = BcmcComponentParamsMapper(null).mapToParams(bcmcConfiguration)
+        val params = BcmcComponentParamsMapper().mapToParams(bcmcConfiguration)
 
         val expected = BcmcComponentParams(
             shopperLocale = Locale.US,
@@ -59,7 +59,7 @@ internal class BcmcComponentParamsMapperTest {
             .setSubmitButtonVisible(false)
             .build()
 
-        val params = BcmcComponentParamsMapper(null).mapToParams(bcmcConfiguration)
+        val params = BcmcComponentParamsMapper().mapToParams(bcmcConfiguration)
 
         val expected = BcmcComponentParams(
             shopperLocale = Locale.US,
@@ -100,7 +100,7 @@ internal class BcmcComponentParamsMapperTest {
             )
         )
 
-        val params = BcmcComponentParamsMapper(overrideParams).mapToParams(bcmcConfiguration)
+        val params = BcmcComponentParamsMapper().mapToParams(bcmcConfiguration, overrideParams)
 
         val expected = BcmcComponentParams(
             shopperLocale = Locale.GERMAN,
--- a/bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
+++ b/bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
@@ -11,35 +11,24 @@ package com.adyen.checkout.bcmc
 import com.adyen.checkout.components.base.GenericComponentParams
 import com.adyen.checkout.components.model.payments.Amount
 import com.adyen.checkout.core.api.Environment
+import com.adyen.checkout.sessions.model.setup.SessionSetupConfiguration
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments.arguments
+import org.junit.jupiter.params.provider.MethodSource
 import java.util.Locale
 
 internal class BcmcComponentParamsMapperTest {
 
     @Test
     fun `when parent configuration is null and custom bcmc configuration fields are null then all fields should match`() {
-        val bcmcConfiguration = BcmcConfiguration.Builder(
-            shopperLocale = Locale.US,
-            environment = Environment.TEST,
-            clientKey = TEST_CLIENT_KEY_1
-        )
+        val bcmcConfiguration = getBcmcConfigurationBuilder()
             .build()
 
         val params = BcmcComponentParamsMapper().mapToParams(bcmcConfiguration)
 
-        val expected = BcmcComponentParams(
-            shopperLocale = Locale.US,
-            environment = Environment.TEST,
-            clientKey = TEST_CLIENT_KEY_1,
-            isAnalyticsEnabled = true,
-            isCreatedByDropIn = false,
-            isHolderNameRequired = false,
-            shopperReference = null,
-            isStorePaymentFieldVisible = false,
-            amount = Amount.EMPTY,
-            isSubmitButtonVisible = true
-        )
+        val expected = getBcmcComponentParams()
 
         assertEquals(expected, params)
     }
@@ -48,11 +37,7 @@ internal class BcmcComponentParamsMapperTest {
     fun `when parent configuration is null and custom bcmc configuration fields are set then all fields should match`() {
         val shopperReference = "SHOPPER_REFERENCE_1"
 
-        val bcmcConfiguration = BcmcConfiguration.Builder(
-            shopperLocale = Locale.US,
-            environment = Environment.TEST,
-            clientKey = TEST_CLIENT_KEY_1
-        )
+        val bcmcConfiguration = getBcmcConfigurationBuilder()
             .setShopperReference(shopperReference)
             .setHolderNameRequired(true)
             .setShowStorePaymentField(true)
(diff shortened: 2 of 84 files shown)


bcmc/src/main/java/com/adyen/checkout/bcmc/BcmcComponentProvider.kt
```diff
@@ -49,12 +49,13 @@ import com.adyen.checkout.sessions.repository.SessionRepository
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class BcmcComponentProvider(
-    overrideComponentParams: ComponentParams? = null,
+    private val overrideComponentParams: ComponentParams? = null,
+    private val checkoutSession: CheckoutSession? = null
```
OscarSpruit, line 53: It's better to give this another name to avoid confusion in the `get()` where we pass the same thing. Maybe `dropInCheckoutSession` or `overrideCheckoutSession` is a good name 🤔

bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
```diff
@@ -100,27 +74,135 @@ internal class BcmcComponentParamsMapperTest {
...
                 value = 25_00L
             ),
-            isSubmitButtonVisible = true
         )
 
         assertEquals(expected, params)
     }
 
+    @Test
+    fun `when parent configuration is null, custom bcmc configuration fields are null and enable store details in session setup response is false then the field value should be false`() {
```
OscarSpruit, line 95: `ParameterizedTest`s would be perfect for this

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -376,10 +377,11 @@ internal fun getComponentFor(
  * @param paymentMethod The payment method to be parsed.
  * @throws CheckoutException In case a component cannot be created.
  */
-@Suppress("LongMethod", "UNCHECKED_CAST")
+@Suppress("LongMethod", "UNCHECKED_CAST", "LongParameterList")
 internal fun getComponentFor(
     fragment: Fragment,
     paymentMethod: PaymentMethod,
+    checkoutSession: CheckoutSession? = null,
```
OscarSpruit, line 384: It's nicer to have parameters with a default value at the end

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInViewModel.kt
```diff
@@ -83,6 +84,8 @@ internal class DropInViewModel(
             bundleHandler.paymentMethodsApiResponse = value
         }
 
+    val checkoutSession: CheckoutSession? = bundleHandler.checkoutSession
```
OscarSpruit, line 87: This should use a custom getter, because the value in `bundleHandler` could change

example-app/src/main/java/com/adyen/checkout/example/data/api/model/SessionRequest.kt
```diff
@@ -27,4 +27,6 @@ data class SessionRequest(
     val threeDS2RequestData: ThreeDS2RequestDataRequest?,
     val shopperEmail: String?,
     val allowedPaymentMethods: List<String>?,
+    val storePaymentMethodMode: String?,
+    val recurringProcessingModel: String?
```
OscarSpruit, line 31: What is this needed for? I don't see anything about it in the ticket

sessions-core/src/main/java/com/adyen/checkout/sessions/model/setup/Configuration.kt
```diff
@@ -0,0 +1,38 @@
+package com.adyen.checkout.sessions.model.setup
+
+import com.adyen.checkout.core.exception.ModelSerializationException
+import com.adyen.checkout.core.model.ModelObject
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+data class Configuration(val enableStoreDetails: Boolean? = null) : ModelObject() {
```
OscarSpruit, line 10: A more descriptive name would be nice as we already have 2 classes with this name. Maybe something like `SessionSetupConfiguration`

example-app/src/main/java/com/adyen/checkout/example/ui/card/SessionsCardTakenOverViewModel.kt
```diff
@@ -114,6 +116,8 @@ internal class SessionsCardTakenOverViewModel @Inject constructor(
                     ?: throw IllegalStateException("Return url should be set"),
                 shopperEmail = keyValueStorage.getShopperEmail(),
                 allowedPaymentMethods = listOf(paymentMethodType),
+                storePaymentMethodMode = StorePaymentMethodMode.ASK_FOR_CONSENT.mode,
```
jreij, line 119: Instead of manually passing these values every time we need them, how about we add them as default values to the parameters of the `getSessionRequest` method?

sessions-core/src/main/java/com/adyen/checkout/sessions/model/setup/SessionSetupResponse.kt
```diff
@@ -66,7 +68,11 @@ data class SessionSetupResponse(
                             jsonObject.optJSONObject(PAYMENT_METHODS),
                             PaymentMethodsApiResponse.SERIALIZER
                         ),
-                        returnUrl = jsonObject.optString(RETURN_URL)
+                        returnUrl = jsonObject.optString(RETURN_URL),
+                        configuration = ModelUtils.deserializeOpt(
```
jreij, line 72: You forgot to add this to the `serialize` function as well

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -419,7 +421,7 @@ internal fun getComponentFor(
         CardComponent.PROVIDER.isPaymentMethodSupported(paymentMethod) -> {
             val cardConfig: CardConfiguration =
                 getConfigurationForPaymentMethod(paymentMethod, dropInConfiguration)
-            CardComponentProvider(dropInParams).get(
+            CardComponentProvider(dropInParams, checkoutSession).get(
```
jreij, line 424: This should also be done in the stored component

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -399,7 +401,7 @@ internal fun getComponentFor(
         BcmcComponent.PROVIDER.isPaymentMethodSupported(paymentMethod) -> {
             val bcmcConfiguration: BcmcConfiguration =
                 getConfigurationForPaymentMethod(paymentMethod, dropInConfiguration)
-            BcmcComponentProvider(dropInParams).get(
+            BcmcComponentProvider(dropInParams, checkoutSession).get(
```
jreij, line 404: Let's pass it everywhere not just to card and bcmc, to keep the provider class consistent

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInSavedStateHandleContainer.kt
```diff
@@ -75,6 +76,7 @@ internal object DropInBundleHandler {
             packageName,
         )
         intent.apply {
+            putExtra(CHECKOUT_SESSION_KEY, checkoutSession)
```
jreij, line 79: Having both `checkoutSession` and `checkoutSession.sessionSetupResponse` is redundant, let's remove the previous

jreij: Or actually maybe we don't need to pass the whole `CheckoutSession` object to the providers. How about we only pass the `SessionSetupConfiguration` object, and we store it inside `SessionDetails` (just like the rest of the fields. Passing the whole session just to take one field seems overkill

bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
```diff
@@ -100,29 +77,104 @@ internal class BcmcComponentParamsMapperTest {
...
+        isStorePaymentFieldVisible = isStorePaymentFieldVisible
+    )
+
     companion object {
         private const val TEST_CLIENT_KEY_1 = "test_qwertyuiopasdfghjklzxcvbnmqwerty"
         private const val TEST_CLIENT_KEY_2 = "live_qwertyui34566776787zxcvbnmqwerty"
+
+        @JvmStatic
+        fun sessionSetupConfigurationSource() = listOf(
+            arguments(false, false, false),
```
jreij, line 174: This doesn't seem very readable, you can to check the main methods to know what each field is for. How about we create a data class with these 3 booleans and use it as a parameter for the test? This way we can used named arguments and we'll be able to read what each boolean is for

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
etman55: why do we need to do this? I mean stored component doesn't have that field.

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
etman55: If we passed this everywhere then I guess we would want to pass the whole checkoutSession object in case we needed it in the future WDYT?

example-app/src/main/java/com/adyen/checkout/example/data/api/model/SessionRequest.kt
OscarSpruit: Oh okay I see 👍

sessions-core/src/main/java/com/adyen/checkout/sessions/model/setup/SessionSetupResponse.kt
etman55: Do we use the `serialize` method for responses? I mean it's nice to have both, but do we actually use it?

bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
OscarSpruit: As you said, the test method makes it clear what each boolean means. IMO we could also just add a little comment here to tell what each boolean means

etman55: I agree with @OscarSpruit I guess just a comment should be enough.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInSavedStateHandleContainer.kt
etman55: I agree with you, I am just concerned that we might need something else inside the `ComponentParams` not just  `SessionSetupConfiguration`

sessions-core/src/main/java/com/adyen/checkout/sessions/model/setup/SessionSetupResponse.kt
jreij: We don't use it but for the sake of consistency let's do it, since we serialize the rest of the fields

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
jreij: For now there is nothing from the configuration that is required for the stored component, but if we add later it'll be a simple change

drop-in/src/main/java/com/adyen/checkout/dropin/ui/viewmodel/DropInSavedStateHandleContainer.kt
jreij: Yeah I agree but it seems like the API only wants to use this object to configure components, I guess we'll refactor it if that changes 🤷

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
jreij: Yes indeed let's pass the same param everywhere (whether it's `CheckoutSession` or `SessionSetupConfiguration`) to keep providers consistent

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -333,20 +335,21 @@ internal fun getPaymentMethodAvailabilityCheck(
...
  */
-@Suppress("UNCHECKED_CAST")
+@Suppress("UNCHECKED_CAST", "LongParameterList")
 internal fun getComponentFor(
     fragment: Fragment,
     storedPaymentMethod: StoredPaymentMethod,
     dropInConfiguration: DropInConfiguration,
     amount: Amount,
     componentCallback: ComponentCallback<*>,
+    sessionSetupConfiguration: SessionSetupConfiguration? = null,
```
jreij, line 345: Let's remove the default value here so we don't forget to pass `sessionSetupConfiguration` whenever this method is used

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -376,20 +379,21 @@ internal fun getComponentFor(
...
  */
-@Suppress("LongMethod", "UNCHECKED_CAST")
+@Suppress("LongMethod", "UNCHECKED_CAST", "LongParameterList")
 internal fun getComponentFor(
     fragment: Fragment,
     paymentMethod: PaymentMethod,
     dropInConfiguration: DropInConfiguration,
     amount: Amount,
     componentCallback: ComponentCallback<*>,
+    sessionSetupConfiguration: SessionSetupConfiguration? = null,
```
jreij, line 389: Same here with the default value

bcmc/src/test/java/com/adyen/checkout/bcmc/BcmcComponentParamsMapperTest.kt
```diff
@@ -100,29 +75,86 @@ internal class BcmcComponentParamsMapperTest {
...
             ),
-            isSubmitButtonVisible = true
         )
 
         assertEquals(expected, params)
     }
 
+    @ParameterizedTest
+    @MethodSource("sessionSetupConfigurationSource")
+    @Suppress("MaxLineLength")
```
jreij, line 97: Hmm this is strange, we shouldn't get a `MaxLineLength` error when a method is between ` `` `

etman55: It's strange, I added it after running `./gradlew check` and it gave me an error!

example-app/src/main/java/com/adyen/checkout/example/ui/card/SessionsCardTakenOverViewModel.kt
```diff
@@ -113,7 +113,7 @@ internal class SessionsCardTakenOverViewModel @Inject constructor(
                 redirectUrl = savedStateHandle.get<String>(SessionsCardTakenOverActivity.RETURN_URL_EXTRA)
                     ?: throw IllegalStateException("Return url should be set"),
                 shopperEmail = keyValueStorage.getShopperEmail(),
-                allowedPaymentMethods = listOf(paymentMethodType),
+                allowedPaymentMethods = listOf(paymentMethodType)
```
jreij, line 116: Minor: this change is unnecessary

example-app/src/main/java/com/adyen/checkout/example/ui/card/SessionsCardViewModel.kt
```diff
@@ -99,7 +99,7 @@ internal class SessionsCardViewModel @Inject constructor(
                 redirectUrl = savedStateHandle.get<String>(SessionsCardActivity.RETURN_URL_EXTRA)
                     ?: throw IllegalStateException("Return url should be set"),
                 shopperEmail = keyValueStorage.getShopperEmail(),
-                allowedPaymentMethods = listOf(paymentMethodType),
+                allowedPaymentMethods = listOf(paymentMethodType)
```
jreij, line 102: Minor: this change is unnecessary

sessions-core/src/main/java/com/adyen/checkout/sessions/CheckoutSession.kt
```diff
@@ -8,15 +8,18 @@
...
 import com.adyen.checkout.sessions.model.setup.SessionSetupResponse
+import kotlinx.parcelize.Parcelize
 
 // TODO docs
+@Parcelize
 data class CheckoutSession(
     val sessionSetupResponse: SessionSetupResponse,
     val order: Order?,
-) {
+) : Parcelable {
```
jreij, line 22: Is this still needed?

jreij:
LGTM 👍  We just need to wait until the other PRs are merged then rebase and apply the same changes to the new components

jreij (approved):

PR #1059, opened 2023-02-14
Return session result in drop-in

jreij:
## Description
We were missing the full session result in drop-in.

## Checklist 

COAND-680

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInResult.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInResult.kt
@@ -10,6 +10,7 @@ package com.adyen.checkout.dropin
 
 import com.adyen.checkout.dropin.service.DropInServiceResult
 import com.adyen.checkout.dropin.service.SessionDropInService
+import com.adyen.checkout.sessions.model.SessionPaymentResult
 
 /**
  * A class that contains the final result of Drop-in.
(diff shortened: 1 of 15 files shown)


etman55 (dismissed):

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInServiceResult.kt
```diff
@@ -81,6 +82,11 @@ sealed class DropInServiceResult : BaseDropInServiceResult() {
         override val reason: String? = null,
         override val dismissDropIn: Boolean = false
     ) : DropInServiceResult(), DropInServiceResultError
+
+    /**
+     * For internal use only.
+     */
```
etman55, line 88: If it's for internal use only I don't think we need to add a comment just explaining that.

drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInResult.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * A class that contains the final result of Drop-in.
+ */
+sealed class SessionDropInResult {
+
+    /**
+     * Drop-in was dismissed by the user before it has completed.
+     */
+    class CancelledByUser : SessionDropInResult()
```
etman55, line 23: I guess this should be object

jreij: It's left like that on purpose, if we ever need to change it to a class to add some parameters it will be a breaking change for merchants

drop-in/src/main/java/com/adyen/checkout/dropin/service/DropInServiceResult.kt
jreij: hmm yeah fair point

etman55 (approved):

PR #1070, opened 2023-02-16
Restructure packaging for action component

etman55:
image

COAND-497


action/src/main/java/com/adyen/checkout/action/internal/ui/GenericActionDelegate.kt
```diff
@@ -6,7 +6,7 @@
  * Created by josephj on 19/9/2022.
  */
 
-package com.adyen.checkout.action
+package com.adyen.checkout.action.internal.ui
```
jreij, line 9: Class should be restricted

jreij (approved):

OscarSpruit:
I wonder if we should make these 3 classes internal, cause if we change them it impacts the interface of some other public classes. WDYT @jreij @etman55 ?

image

etman55:
In my understanding for `ActionHandlingConfigurationBuilder` is already internal and for other they're already used in other modules and they're annotated with `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)`

jreij (approved):

PR #1071, opened 2023-02-16
Restructure packaging for await component

etman55:
image

COAND-497

### DIFF
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitDelegate.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitDelegate.kt
@@ -6,14 +6,17 @@
  * Created by oscars on 16/8/2022.
  */
 
-package com.adyen.checkout.await
+package com.adyen.checkout.await.internal.ui
 
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.await.internal.ui.model.AwaitOutputData
 import com.adyen.checkout.components.base.ActionDelegate
 import com.adyen.checkout.components.base.DetailsEmittingDelegate
 import com.adyen.checkout.components.base.StatusPollingDelegate
 import com.adyen.checkout.components.base.ViewableDelegate
 import com.adyen.checkout.components.ui.ViewProvidingDelegate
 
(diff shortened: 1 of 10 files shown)


await/src/main/java/com/adyen/checkout/await/internal/ui/model/AwaitOutputData.kt
```diff
@@ -5,10 +5,12 @@
  *
  * Created by caiof on 25/8/2020.
  */
-package com.adyen.checkout.await
+package com.adyen.checkout.await.internal.ui.model
 
+import androidx.annotation.RestrictTo
 import com.adyen.checkout.components.base.OutputData
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jreij, line 13: Can't this be internal?

await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitDelegate.kt
```diff
@@ -6,14 +6,17 @@
...
 
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.await.internal.ui.model.AwaitOutputData
 import com.adyen.checkout.components.base.ActionDelegate
 import com.adyen.checkout.components.base.DetailsEmittingDelegate
 import com.adyen.checkout.components.base.StatusPollingDelegate
 import com.adyen.checkout.components.base.ViewableDelegate
 import com.adyen.checkout.components.ui.ViewProvidingDelegate
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jreij, line 19: Can this be internal?

etman55: No as we pass it through the component constructor.

await/src/main/java/com/adyen/checkout/await/internal/ui/model/AwaitOutputData.kt
etman55: This is used inside `AwaitDelegate` and I need to make `AwaitDelegate` internal to make it work and I can't make the delegate internal.

jreij (approved):

PR #1075, opened 2023-02-16
Restructure packaging for PayByBank

onurkaragunlu:
Restructure packaging for PayByBank

Screenshot 2023-02-16 at 11 59 40

COAND-497

### DIFF
--- a/paybybank/src/main/java/com/adyen/checkout/paybybank/internal/ui/view/PayByBankView.kt
+++ b/paybybank/src/main/java/com/adyen/checkout/paybybank/internal/ui/view/PayByBankView.kt
@@ -6,7 +6,7 @@
  * Created by ozgur on 28/9/2022.
  */
 
-package com.adyen.checkout.paybybank
+package com.adyen.checkout.paybybank.internal.ui.view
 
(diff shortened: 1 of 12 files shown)


paybybank/src/main/java/com/adyen/checkout/paybybank/internal/ui/view/PayByBankView.kt
```diff
@@ -6,7 +6,7 @@
  * Created by ozgur on 28/9/2022.
  */
 
-package com.adyen.checkout.paybybank
+package com.adyen.checkout.paybybank.internal.ui.view
```
jreij, line 9: Can this class be internal?

onurkaragunlu: I am not sure. Are there any cases in which we directly use the view instead of Adyen component view?

jreij: I don't think so

jreij (approved):

PR #1076, opened 2023-02-16
Restructure packaging for instant

OscarSpruit:
## Description
Restructure packaging for instant

image

COAND-497


instant/src/main/java/com/adyen/checkout/instant/internal/ui/InstantPaymentDelegate.kt
```diff
@@ -6,7 +6,7 @@
  * Created by ozgur on 4/11/2022.
  */
 
-package com.adyen.checkout.instant
+package com.adyen.checkout.instant.internal.ui
```
jreij, line 9: Can this be internal?

jreij (approved):

PR #1081, opened 2023-02-17
Restructure packaging for wechat

etman55:
image

COAND-497


wechatpay/src/main/java/com/adyen/checkout/wechatpay/internal/util/WeChatPayUtils.kt
```diff
@@ -5,10 +5,12 @@
...
  * Created by caiof on 17/9/2019.
  */
-package com.adyen.checkout.wechatpay
+package com.adyen.checkout.wechatpay.internal.util
 
 import android.content.Intent
+import androidx.annotation.RestrictTo
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 object WeChatPayUtils {
```
OscarSpruit, line 14: I think this class should be public. Merchants should be able to call `isResultIntent` to verify the result

OscarSpruit (approved):

PR #1086, opened 2023-02-22
Restructure packaging for components-core

OscarSpruit:
## Description
Good luck reviewing this one 😬 

image

COAND-497


components-core/src/main/java/com/adyen/checkout/components/core/internal/provider/ActionComponentProvider.kt
```diff
@@ -15,12 +15,15 @@ import androidx.lifecycle.LifecycleOwner
...
-import com.adyen.checkout.components.model.payments.response.Action
+import com.adyen.checkout.components.core.action.Action
+import com.adyen.checkout.components.core.internal.ActionComponent
+import com.adyen.checkout.components.core.internal.ActionComponentCallback
+import com.adyen.checkout.components.core.internal.Component
+import com.adyen.checkout.components.core.internal.Configuration
+import com.adyen.checkout.components.core.internal.ui.ActionDelegate
 
 // TODO SESSIONS: DOCS
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
etman55, line 26: I guess we don't need `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` on `getDelegate()` method anymore after adding this to the whole interface

components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/ActionDelegate.kt
```diff
@@ -6,16 +6,18 @@
  * Created by josephj on 24/8/2022.
  */
 
-package com.adyen.checkout.components.base
+package com.adyen.checkout.components.core.internal.ui
```
etman55, line 9: hmmm I am thinking this should be inside action module WDYT?

components-core/src/main/java/com/adyen/checkout/components/core/internal/provider/ActionComponentProvider.kt
```diff
@@ -5,7 +5,7 @@
  *
  * Created by caiof on 13/5/2019.
  */
-package com.adyen.checkout.components
+package com.adyen.checkout.components.core.internal.provider
```
etman55, line 8: same here, should this be inside action module?

components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/ViewableDelegate.kt
```diff
@@ -6,10 +6,13 @@
  * Created by josephj on 24/8/2022.
  */
 
-package com.adyen.checkout.components.base
+package com.adyen.checkout.components.core.internal.ui
```
etman55, line 9: I think this should be inside ui-core module WDYT?

components-core/src/main/java/com/adyen/checkout/components/core/internal/provider/ActionComponentProvider.kt
OscarSpruit: I'd like to keep it there to make extra clear that it shouldn't be used

components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/ViewableDelegate.kt
OscarSpruit: I think we should leave it here. `ui-core` is more "real" UI related, so custom views and styling. There are some not so UI related classes in there, but let's try and keep that to a minimum

etman55 (approved):

PR #1092, opened 2023-02-24
Restructure packaging for drop in module

etman55:
image
image

COAND-497


drop-in/src/main/java/com/adyen/checkout/dropin/DropInResultContract.kt
```diff
@@ -13,7 +13,8 @@ import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
 import androidx.activity.result.contract.ActivityResultContract
-import com.adyen.checkout.dropin.ui.DropInActivity
+import com.adyen.checkout.dropin.internal.ui.DropInActivity
+import com.adyen.checkout.dropin.internal.ui.model.DropInResultContractParams
 
 internal class DropInResultContract : ActivityResultContract<DropInResultContractParams, DropInResult?>() {
```
OscarSpruit, line 19: Since it's internal it should be safe to move to the internal package

drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInResultContract.kt
```diff
@@ -13,7 +13,7 @@ import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
 import androidx.activity.result.contract.ActivityResultContract
-import com.adyen.checkout.dropin.ui.DropInActivity
+import com.adyen.checkout.dropin.internal.ui.DropInActivity
 import com.adyen.checkout.sessions.core.SessionPaymentResult
 
 internal class SessionDropInResultContract :
```
OscarSpruit, line 19: Since it's internal it should be safe to move to the internal package

drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInResultContractParams.kt
```diff
@@ -9,7 +9,6 @@
 package com.adyen.checkout.dropin
 
 import androidx.annotation.RestrictTo
-import com.adyen.checkout.dropin.service.SessionDropInService
 import com.adyen.checkout.sessions.core.CheckoutSession
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
OscarSpruit, line 14: Should also be moved to internal package

OscarSpruit: And I think we can make it an `internal class`

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInResultContractParams.kt
```diff
@@ -6,11 +6,12 @@
...
-package com.adyen.checkout.dropin
+package com.adyen.checkout.dropin.internal.ui.model
 
 import androidx.annotation.RestrictTo
 import com.adyen.checkout.components.core.PaymentMethodsApiResponse
-import com.adyen.checkout.dropin.service.DropInService
+import com.adyen.checkout.dropin.DropInConfiguration
+import com.adyen.checkout.dropin.DropInService
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
OscarSpruit, line 16: This can also be an `internal class` I think

etman55: This class is used in `DropIn.registerForDropInResult()` as a result so I guess it can be `internal` also use as a parameter for `DropIn.startPayment()`

drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInResultContractParams.kt
etman55: This class is used in `DropIn.registerForDropInResult()` as a result so I guess it can be `internal` also use as a parameter for `DropIn.startPayment()`

OscarSpruit: Oh right, it's the generic type of the `ActivityResultLauncher`

OscarSpruit (approved):

PR #1093, opened 2023-02-27
Display installments for /sessions integrations

etman55:
## Description
- Add Installment options params for cards to be used internally instead of card configuration
-  Add installment options mapper

## Checklist 
- [x] Code is unit tested

COAND-613

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
+++ b/card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
@@ -10,11 +10,11 @@ package com.adyen.checkout.card
 import android.content.Context
 import com.adyen.checkout.action.GenericActionConfiguration
 import com.adyen.checkout.action.internal.ActionHandlingPaymentMethodConfigurationBuilder
+import com.adyen.checkout.components.core.Amount
+import com.adyen.checkout.components.core.PaymentMethod
 import com.adyen.checkout.components.core.internal.ButtonConfiguration
 import com.adyen.checkout.components.core.internal.ButtonConfigurationBuilder
 import com.adyen.checkout.components.core.internal.Configuration
-import com.adyen.checkout.components.core.PaymentMethod
-import com.adyen.checkout.components.core.Amount
 import com.adyen.checkout.core.Environment
 import kotlinx.parcelize.Parcelize
 import java.util.Locale
@@ -135,6 +135,9 @@ class CardConfiguration private constructor(
          *
          * Default is true.
          *
+         * When using `sessions` show store payment field will be ignored and replaced with the value
+         * sent to `/sessions` call.
+         *
          * @param showStorePaymentField [Boolean]
          * @return [CardConfiguration.Builder]
          */
@@ -215,6 +218,9 @@ class CardConfiguration private constructor(
 
         /**
          * Configures the installment options to be provided to the shopper.
+         *
+         * When using `sessions` installment configuration will be ignored and replaced with the value
+         * sent to `/sessions` call.
          *
          * @param installmentConfiguration The configuration object for installment options.
          * @return [CardConfiguration.Builder]
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
@@ -13,8 +13,8 @@ import com.adyen.checkout.card.CardBrand
 import com.adyen.checkout.card.CardConfiguration
 import com.adyen.checkout.card.KCPAuthVisibility
 import com.adyen.checkout.card.SocialSecurityNumberVisibility
-import com.adyen.checkout.components.core.internal.ui.model.ComponentParams
 import com.adyen.checkout.components.core.PaymentMethod
+import com.adyen.checkout.components.core.internal.ui.model.ComponentParams
 import com.adyen.checkout.core.internal.util.LogUtil
 import com.adyen.checkout.core.internal.util.Logger
 import com.adyen.checkout.sessions.core.SessionSetupConfiguration
@@ -66,7 +66,7 @@ internal class CardComponentParamsMapper {
             isHideCvcStoredCard = isHideCvcStoredCard ?: false,
             socialSecurityNumberVisibility = socialSecurityNumberVisibility ?: SocialSecurityNumberVisibility.HIDE,
             kcpAuthVisibility = kcpAuthVisibility ?: KCPAuthVisibility.HIDE,
-            installmentConfiguration = installmentConfiguration,
+            installmentParams = null,
             addressParams = addressConfiguration?.mapToAddressParam() ?: AddressParams.None
         )
     }
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
@@ -21,7 +21,7 @@ import com.adyen.checkout.sessions.core.SessionSetupConfiguration
 import com.adyen.checkout.ui.core.internal.ui.model.AddressFieldPolicy
 import com.adyen.checkout.ui.core.internal.ui.model.AddressParams
 
-internal class CardComponentParamsMapper {
+internal class CardComponentParamsMapper(private val installmentsParamsMapper: InstallmentsParamsMapper) {
 
     fun mapToParamsDefault(
         cardConfiguration: CardConfiguration,
@@ -66,7 +66,10 @@ internal class CardComponentParamsMapper {
             isHideCvcStoredCard = isHideCvcStoredCard ?: false,
             socialSecurityNumberVisibility = socialSecurityNumberVisibility ?: SocialSecurityNumberVisibility.HIDE,
             kcpAuthVisibility = kcpAuthVisibility ?: KCPAuthVisibility.HIDE,
-            installmentParams = null,
+            installmentParams = installmentsParamsMapper?.mapToInstallmentParams(
+                sessionSetupConfiguration?.installmentOptions,
+                installmentConfiguration
+            ),
             addressParams = addressConfiguration?.mapToAddressParam() ?: AddressParams.None
         )
     }
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOptionParams.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOptionParams.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2023 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 24/2/2023.
+ */
+
+package com.adyen.checkout.card.internal.ui.model
+
+import android.os.Parcelable
+import com.adyen.checkout.card.CardBrand
+import kotlinx.parcelize.Parcelize
+
+/**
+ * InstallmentOptionParams is used for defining the details of installment options.
+ *
+ * Note: All values specified in [values] must be greater than 1.
+ */
+internal sealed class InstallmentOptionParams : Parcelable {
+    abstract val values: List<Int>
+    abstract val includeRevolving: Boolean
+
+    /**
+     * @param values see [InstallmentOptionParams.values]
+     * @param includeRevolving see [InstallmentOptionParams.includeRevolving]
+     * @param cardBrand a [CardBrand] to apply the given options
+     */
+    @Parcelize
+    data class CardBasedInstallmentOptions(
+        override val values: List<Int>,
+        override val includeRevolving: Boolean,
+        val cardBrand: CardBrand
+    ) : InstallmentOptionParams()
+
+    /**
+     * @param values see [InstallmentOptionParams.values]
+     * @param includeRevolving see [InstallmentOptionParams.includeRevolving]
+     */
+    @Parcelize
+    data class DefaultInstallmentOptions(
+        override val values: List<Int>,
+        override val includeRevolving: Boolean
+    ) : InstallmentOptionParams()
+}
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2023 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by atef on 22/2/2023.
+ */
+
+package com.adyen.checkout.card.internal.ui.model
+
+import com.adyen.checkout.card.CardBrand
+import com.adyen.checkout.card.InstallmentConfiguration
+import com.adyen.checkout.card.InstallmentOptions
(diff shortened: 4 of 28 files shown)


jreij:
One of the commits is missing the ticket number

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/JsonUtils.kt
```diff
@@ -95,4 +132,41 @@ object JsonUtils {
...
+     * @return A [List] of integers, or null if the jsonArray was null.
+     */
+    @JvmStatic
+    fun parseOptIntegerList(jsonArray: JSONArray?): List<Int>? {
+        if (jsonArray == null) {
+            return null
+        }
+        val list: MutableList<Int> = ArrayList()
+        for (i in 0 until jsonArray.length()) {
+            val item = jsonArray.optInt(i, -1)
```
jreij, line 149: Why are we using -1 here? If a value in this array is not an integer we should throw an exception, -1 could hide an integration error we have

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/JsonUtils.kt
```diff
@@ -95,4 +132,41 @@ object JsonUtils {
...
+     * @param intList The [List] of integers to be serialized.
+     * @return The populated [JSONArray]. Could be null.
+     */
+    @JvmStatic
+    fun serializeOptIntegerList(intList: List<Int?>?): JSONArray? {
+        if (intList == null) {
+            return null
+        }
+        return JSONArray().apply {
+            intList.filter { it != -1 }.forEach {
```
jreij, line 167: Here if a value is -1 we skip it and if it's null we add it as null? What if we actually need to serialize -1 on purpose?

example-app/src/main/java/com/adyen/checkout/example/service/RequestUtils.kt
```diff
@@ -86,10 +93,13 @@ fun getSessionRequest(
         shopperEmail = shopperEmail,
         allowedPaymentMethods = allowedPaymentMethods,
         storePaymentMethodMode = storePaymentMethodMode,
-        recurringProcessingModel = recurringProcessingModel
+        recurringProcessingModel = recurringProcessingModel,
+        installmentOptions = installmentOptions
     )
 }
 
+private const val DEFAULT_INSTALLMENT_OPTION = "card"
```
jreij, line 101: This should be at the bottom with the rest of the constants

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/JsonUtils.kt
```diff
@@ -54,6 +59,38 @@ fun JSONArray.toStringPretty(): String {
     }
 }
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Throws(JSONException::class)
+inline fun <reified T : ModelObject> JSONObject.jsonToMap(
```
jreij, line 64: What's the difference between these 2 methods?

sessions-core/src/main/java/com/adyen/checkout/sessions/core/SessionSetupInstallmentOptions.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.adyen.checkout.core.internal.data.model.optIntList
+import com.adyen.checkout.core.internal.data.model.optStringList
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+data class SessionSetupInstallmentOptions(
+    val plans: List<String> = emptyList(),
+    val preselectedValue: Int? = 2,
```
jreij, line 23: Why do we have `2` as a default value?

sessions-core/src/main/java/com/adyen/checkout/sessions/core/SessionSetupInstallmentOptions.kt
```diff
@@ -0,0 +1,60 @@
...
+    val values: List<Int>? = listOf(2)
```
jreij, line 24: Why do we have `listOf(2)` as a default value?

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOption.kt
```diff
@@ -8,7 +8,10 @@
 
 package com.adyen.checkout.card.internal.ui.model
 
-internal enum class InstallmentOption(val type: String?) {
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jreij, line 13: Let's not expose this class just to use it in the example app, we can hardcode the value we want to use there

jreij:
Let's put a comment on `CardConfiguration.Builder.setInstallmentConfigurations` to mention that this configuration is ignored in a sessions flow and that the value provided to the `/sessions` API call will always be used instead.

Same for `CardConfiguration.Builder.setShowStorePaymentField` from the previous PR I guess

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOptionParams.kt
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * InstallmentOptionParams is used for defining the details of installment options.
+ *
+ * Note: All values specified in [values] must be greater than 1.
+ */
+internal sealed class InstallmentOptionParams : Parcelable {
+    abstract val values: List<Int>
+    abstract val includeRevolving: Boolean
+
+    companion object {
```
jreij, line 25: We usually put the companion at the bottom

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOptionParams.kt
```diff
@@ -0,0 +1,68 @@
...
+        private const val STARTING_INSTALLMENT_VALUE = 2
```
jreij, line 26: Same question as before, do we need this default value on the SDK?

sessions-core/src/main/java/com/adyen/checkout/sessions/core/SessionSetupInstallmentOptions.kt
etman55: we don't need it here actually, it was suppose to be inside example app

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/JsonUtils.kt
etman55: hmm just wanted to separate the logic, but we can combine them.

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOptionParams.kt
etman55: This was done on `InstallmentConfiguration` I just copied it. I guess we need it when using a constructor with a max value.

etman55: we probably won't need it will remove it

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentOptionParams.kt
etman55: we probably won't need it, it was just copied from `InstallmentConfiguration` to be used with max value constructors

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/JsonUtils.kt
jreij: We can separate them as long as we give them clear different names to explain what each one does, and make one of them private since it's not supposed to be used directly

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/JsonUtils.kt
```diff
@@ -95,4 +123,47 @@ object JsonUtils {
...
+     * @param intList The [List] of integers to be serialized.
+     * @return The populated [JSONArray]. Could be null.
+     */
+    @JvmStatic
+    fun serializeOptIntegerList(intList: List<Int?>?): JSONArray? {
+        if (intList == null) {
+            return null
+        }
+        return JSONArray().apply {
+            intList.filterNotNull().forEach {
```
jreij, line 164: Is there a reason to allow null values in the list? And if we do why should we filter them out? If i am trying to serialize a list of `1, 2, null, 3` i probably want the null to be there

example-app/src/main/java/com/adyen/checkout/example/data/storage/KeyValueStorage.kt
```diff
@@ -28,6 +28,8 @@ interface KeyValueStorage {
     fun isSplitCardFundingSources(): Boolean
     fun isAddressFormEnabled(): Int
     fun getInstantPaymentMethodType(): String
+    fun isInstallmentOptionsEnabled(): Int
```
jreij, line 31: Can't we have one configuration for both sessions and non sessions?

etman55: it's going to be a long list 
- no installment configuration & no session installments
- default installment configuration & no session installments
- default installment configuration & session installments
- default revolving installment configuration & no session installments
- default revolving installment configuration & session installments
- card based installment configuration  & no session installments
- card based installment configuration  & session installments

jreij: How about the same 4 options that work for sessions and non sessions at the same time?

example-app/src/main/res/values/arrays.xml
```diff
@@ -20,6 +20,20 @@
         <item>"2"</item>
     </array>
 
+    <array name="enable_card_installment_options_entries" >
+        <item>"None"</item>
+        <item>"default installment options"</item>
```
jreij, line 25: Please capitalize the first letter

example-app/src/main/res/values/arrays.xml
```diff
@@ -20,6 +20,20 @@
...
+        <item>"default installment options with revolving"</item>
+        <item>"card base installment options"</item>
```
jreij, line 27: `based` not `base`, also can you mention that it's VISA so we know how to test?

card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
```diff
@@ -216,6 +216,9 @@ class CardConfiguration private constructor(
         /**
          * Configures the installment options to be provided to the shopper.
          *
+         * When using `sessions` installment configuration will be ignored and replaced with the value
```
jreij, line 219: Can you add a similar comment to `CardConfiguration.Builder.setShowStorePaymentField`?

card/src/main/java/com/adyen/checkout/card/internal/util/InstallmentUtils.kt
```diff
@@ -25,30 +27,30 @@ internal object InstallmentUtils {
      * Create a list of installment options from [InstallmentConfiguration].
```
jreij, line 27: This doc should be updated, there's no more `InstallmentConfiguration` here

example-app/src/main/res/values/strings.xml
```diff
@@ -44,6 +44,8 @@
     <string name="split_card_funding_sources_key">split_card_funding_sources</string>
     <string name="split_card_funding_sources_title">Split Card Funding Sources</string>
     <string name="enable_card_address_form_key">enable_card_address_form</string>
+    <string name="enable_card_installment_options_title">Card installment options type</string>
+    <string name="enable_card_installment_options_key">enable_card_installment_options_key</string>
```
jreij, line 48: This reads like a boolean now but it actually can have 4 values right?

etman55: True and the reason for that because it's enable the installment option with the 3 options or disable it so I didn't know what else to call it WDYT?

jreij: Maybe something like `mode` instead of `enable` to make it look like an enum not boolean?

card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt
jreij: Same for bcmc, we missed this one

example-app/src/main/res/values/strings.xml
jreij: Make sure you use the same naming everywhere, there's still stuff that start with `enable_card_installment_options_...` or `settingsInstallmentOptionType` that are not consistent with the rest

sessions-core/src/main/java/com/adyen/checkout/sessions/core/SessionSetupInstallmentOptions.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.adyen.checkout.core.internal.data.model.ModelObject
+import com.adyen.checkout.core.internal.data.model.optIntList
+import com.adyen.checkout.core.internal.data.model.optStringList
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+data class SessionSetupInstallmentOptions(
+    val plans: List<String> = emptyList(),
```
jreij, line 22: Does this need to be non-nullable?

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
```diff
@@ -13,15 +13,15 @@ import com.adyen.checkout.card.CardBrand
...
 import com.adyen.checkout.components.core.PaymentMethod
+import com.adyen.checkout.components.core.internal.ui.model.ComponentParams
 import com.adyen.checkout.core.internal.util.LogUtil
 import com.adyen.checkout.core.internal.util.Logger
 import com.adyen.checkout.sessions.core.SessionSetupConfiguration
 import com.adyen.checkout.ui.core.internal.ui.model.AddressFieldPolicy
 import com.adyen.checkout.ui.core.internal.ui.model.AddressParams
 
-internal class CardComponentParamsMapper {
+internal class CardComponentParamsMapper(private val installmentsParamsMapper: InstallmentsParamsMapper? = null) {
```
jreij, line 24: Do we need to allow `installmentsParamsMapper` to be null?

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.adyen.checkout.sessions.core.SessionSetupInstallmentOptions
+
+internal class InstallmentsParamsMapper {
+
+    internal fun mapToInstallmentParams(
+        sessionInstallmentOptions: Map<String, SessionSetupInstallmentOptions?>?,
+        installmentConfiguration: InstallmentConfiguration?
+    ): InstallmentParams? {
+        return if (sessionInstallmentOptions != null) {
+            var defaultOptions: InstallmentOptionParams.DefaultInstallmentOptions? = null
```
jreij, line 23: Minor: let's extract this logic to a method as well? This way it reads a bit nicer:
```kotlin
return sessionInstallmentOptions?.mapToInstallmentParams()
? installmentConfiguration?.mapToInstallmentParams()
```

WDYT?

jreij (approved):

PR #1095, opened 2023-02-27
Make PaymentComponentState an interface 

ozgur00:
## Description
Have an implementation of PaymentComponentState for every payment method.

COAND-727

### DIFF
--- a/bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentState.kt
+++ b/bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentState.kt
@@ -12,12 +12,12 @@ import com.adyen.checkout.components.core.PaymentComponentState
 import com.adyen.checkout.components.core.PaymentComponentData
 import com.adyen.checkout.components.core.paymentmethod.BacsDirectDebitPaymentMethod
 
-class BacsDirectDebitComponentState(
-    paymentComponentData: PaymentComponentData<BacsDirectDebitPaymentMethod>,
-    isInputValid: Boolean,
-    isReady: Boolean,
+data class BacsDirectDebitComponentState(
+    override val data: PaymentComponentData<BacsDirectDebitPaymentMethod>,
+    override val isInputValid: Boolean,
+    override val isReady: Boolean,
     val mode: BacsDirectDebitMode = BacsDirectDebitMode.INPUT,
-) : PaymentComponentState<BacsDirectDebitPaymentMethod>(paymentComponentData, isInputValid, isReady) {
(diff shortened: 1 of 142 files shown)


bacs/src/main/java/com/adyen/checkout/bacs/BacsDirectDebitComponentState.kt
```diff
@@ -13,11 +13,11 @@ import com.adyen.checkout.components.core.PaymentComponentData
 import com.adyen.checkout.components.core.paymentmethod.BacsDirectDebitPaymentMethod
 
 class BacsDirectDebitComponentState(
```
OscarSpruit, line 15: Should we make this a `data class` as well?

giftcard/src/main/java/com/adyen/checkout/giftcard/GiftCardComponentState.kt
```diff
@@ -18,8 +18,8 @@ import kotlinx.parcelize.Parcelize
  */
 @Parcelize
 class GiftCardComponentState(
```
OscarSpruit, line 20: Should we also make this one a `data class`?

OscarSpruit: And do you know why this one has `@Parcelize` and others don't? Should we make it consistent?

googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayComponentState.kt
```diff
@@ -13,8 +13,8 @@ import com.adyen.checkout.components.core.paymentmethod.GooglePayPaymentMethod
 import com.google.android.gms.wallet.PaymentData
 
 class GooglePayComponentState(
```
OscarSpruit, line 15: Also `data class`?

giftcard/src/main/java/com/adyen/checkout/giftcard/GiftCardComponentState.kt
ozgur00: I think it's because this is the only one we keep in the saved state. I say let's leave them like this until we need to be parcelized.

OscarSpruit (approved):

OscarSpruit (dismissed):

jreij:
One minor thing to fix, in `ExampleSessionsDropInService.onSubmit` we check if `state.data.paymentMethod is BlikPaymentMethod` while now we can do `state is BlikComponentState`

jreij (approved):

PR #1100, opened 2023-03-03
Add stored payment method for ACH

onurkaragunlu:
ACH Direct Debit payment method has been added. Stored payment method logic has been changed. 
When the user clicks one of the stored payment methods while the component state is valid, a popup is shown to get payment.

| Stored Card  | ACH|
| ------------- | ------------- |
|  |   |

## Checklist 
- [x] Code is unit tested

COAND-619

### DIFF
--- a/ach/src/main/java/com/adyen/checkout/ach/internal/ui/StoredACHDirectDebitDelegate.kt
+++ b/ach/src/main/java/com/adyen/checkout/ach/internal/ui/StoredACHDirectDebitDelegate.kt
@@ -1,0 +1,190 @@
+/*
+ * Copyright (c) 2023 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by onurk on 24/2/2023.
+ */
+
+package com.adyen.checkout.ach.internal.ui
+
+import androidx.lifecycle.LifecycleOwner
+import com.adyen.checkout.ach.ACHDirectDebitComponentState
+import com.adyen.checkout.ach.internal.ui.model.ACHDirectDebitComponentParams
+import com.adyen.checkout.ach.internal.ui.model.ACHDirectDebitInputData
+import com.adyen.checkout.ach.internal.ui.model.ACHDirectDebitOutputData
+import com.adyen.checkout.components.core.OrderRequest
+import com.adyen.checkout.components.core.PaymentComponentData
+import com.adyen.checkout.components.core.StoredPaymentMethod
+import com.adyen.checkout.components.core.internal.PaymentComponentEvent
+import com.adyen.checkout.components.core.internal.PaymentObserverRepository
+import com.adyen.checkout.components.core.internal.data.api.AnalyticsRepository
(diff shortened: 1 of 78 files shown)


ach/src/main/java/com/adyen/checkout/ach/internal/ui/model/ACHDirectDebitOutputData.kt
```diff
@@ -18,7 +18,9 @@ internal data class ACHDirectDebitOutputData(
     var bankLocationId: FieldState<String>,
     var ownerName: FieldState<String>,
     val addressState: AddressOutputData,
-    val addressUIState: AddressFormUIState
+    val addressUIState: AddressFormUIState,
+    val isStoredPaymentMethodEnable: Boolean,
```
jreij, line 22: Typo, should be `Enabled` not `Enable`

ach/src/main/java/com/adyen/checkout/ach/internal/ui/StoredACHDirectDebitDelegate.kt
```diff
@@ -0,0 +1,190 @@
...
+    private val _outputDataFlow = MutableStateFlow(createOutputData())
+    override val outputDataFlow: Flow<ACHDirectDebitOutputData> = _outputDataFlow
+
+    override val outputData: ACHDirectDebitOutputData
+        get() = _outputDataFlow.value
+
+    private val _componentStateFlow = MutableStateFlow(createComponentState())
+    override val componentStateFlow: Flow<ACHDirectDebitComponentState> = _componentStateFlow
+
+    private val _viewFlow: MutableStateFlow<ComponentViewType?> = MutableStateFlow(ACHDirectDebitComponentViewType)
```
jreij, line 71: Is the `ACHDirectDebitComponentViewType` needed here or can we just set `MutableStateFlow(null)`?

jreij: If merchants use the standalone stored component it will show an empty view now while it actually shouldn't, right?

jreij (approved):

ach/src/test/java/com/adyen/checkout/ach/internal/ui/model/ACHDirectDebitComponentParamsMapperTest.kt
```diff
@@ -117,6 +118,81 @@ internal class ACHDirectDebitComponentParamsMapperTest {
         assertEquals(expectedAddressParams, params.addressParams)
     }
 
+    @Test
+    fun `when the isStorePaymentFieldVisible  in configuration is false, isStorePaymentFieldVisible in component params should be false`() {
+        val achConfiguration = getAchConfigurationBuilder().apply {
```
jreij, line 123: @onurkaragunlu just noticed this while working on something else
FYI you can use the builder pattern here without the apply:
```kotlin
val achConfiguration = getAchConfigurationBuilder()
    .setShowStorePaymentField(false)
    .build()
```

PR #1105, opened 2023-03-06
Add Duitnow payment method.

onurkaragunlu:
Duitnow has been added to the supported payment methods

## Checklist 

COAND-621

### DIFF
--- a/qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
+++ b/qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
@@ -36,6 +36,7 @@ import com.adyen.checkout.core.internal.util.LogUtil
 import com.adyen.checkout.core.internal.util.Logger
 import com.adyen.checkout.qrcode.internal.QRCodeCountDownTimer
 import com.adyen.checkout.qrcode.internal.ui.model.QRCodeOutputData
+import com.adyen.checkout.qrcode.internal.ui.model.QRCodePaymentMethodConfig
 import com.adyen.checkout.qrcode.internal.ui.model.QrCodeUIEvent
 import com.adyen.checkout.ui.core.internal.RedirectHandler
 import com.adyen.checkout.ui.core.internal.ui.ComponentViewType
@@ -152,19 +153,12 @@ internal class DefaultQRCodeDelegate(
             return
         }
 
-        val viewType = when (action.paymentMethodType) {
-            PaymentMethodTypes.PAY_NOW -> {
-                maxPollingDurationMillis = PAY_NOW_MAX_POLLING_DURATION
-                QrCodeComponentViewType.FULL_QR_CODE
-            }
-            PaymentMethodTypes.UPI_QR -> {
-                maxPollingDurationMillis = UPI_MAX_POLLING_DURATION
-                QrCodeComponentViewType.FULL_QR_CODE
-            }
-            else -> {
-                maxPollingDurationMillis = DEFAULT_MAX_POLLING_DURATION
-                QrCodeComponentViewType.SIMPLE_QR_CODE
-            }
+        var viewType = QrCodeComponentViewType.SIMPLE_QR_CODE
+
+        action.paymentMethodType?.let {
+            val qrConfig = QRCodePaymentMethodConfig.getByPaymentMethodType(it)
+            viewType = qrConfig.viewType
+            maxPollingDurationMillis = qrConfig.maxPollingDurationMillis
         }
         _viewFlow.tryEmit(viewType)
 
@@ -218,11 +212,18 @@ internal class DefaultQRCodeDelegate(
             qrImageUrl = String.format(QR_IMAGE_BASE_PATH, componentParams.environment.baseUrl, encodedQrCodeData)
         }
 
+        var messageTextResource: Int? = null
+        action.paymentMethodType?.let {
+            val qrConfig = QRCodePaymentMethodConfig.getByPaymentMethodType(it)
+            messageTextResource = qrConfig.messageTextResource
+        }
+
         val outputData = QRCodeOutputData(
             isValid = isValid,
             paymentMethodType = action.paymentMethodType,
             qrCodeData = action.qrCodeData,
-            qrImageUrl = qrImageUrl
+            qrImageUrl = qrImageUrl,
+            messageTextResource = messageTextResource
         )
         _outputDataFlow.tryEmit(outputData)
     }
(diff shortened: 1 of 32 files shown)


qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
```diff
@@ -313,13 +317,15 @@ internal class DefaultQRCodeDelegate(
...
         private val VIEWABLE_PAYMENT_METHODS = listOf(
+            PaymentMethodTypes.DUIT_NOW,
             PaymentMethodTypes.PIX,
             PaymentMethodTypes.PAY_NOW,
             PaymentMethodTypes.UPI_QR,
         )
 
         @VisibleForTesting
         internal const val PAYLOAD_DETAILS_KEY = "payload"
+        private val DOIT_NOW_MAX_POLLING_DURATION = 90.seconds.inWholeMilliseconds
```
ozgur00, line 328: Typo. :)

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
```diff
@@ -153,6 +153,10 @@ internal class DefaultQRCodeDelegate(
         }
 
         val viewType = when (action.paymentMethodType) {
+            PaymentMethodTypes.DUIT_NOW -> {
```
ozgur00, line 156: This when block will get even more polluted when PromptPay is added. I suggest creating an enum that holds `maxPollingDuration`, `componentViewType`, and also `paymentMethodType` with a helper function that maps from `paymentMethodType` to this enum and we'll read these constants directly from there. It'll be less work when we add new payment methods that fall into this flow.

ozgur00:
Commit message shouldn't have a dot at the end of it

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
```diff
@@ -152,18 +153,14 @@ internal class DefaultQRCodeDelegate(
...
-                QrCodeComponentViewType.FULL_QR_CODE
-            }
-            PaymentMethodTypes.UPI_QR -> {
-                maxPollingDurationMillis = UPI_MAX_POLLING_DURATION
-                QrCodeComponentViewType.FULL_QR_CODE
-            }
-            else -> {
-                maxPollingDurationMillis = DEFAULT_MAX_POLLING_DURATION
-                QrCodeComponentViewType.SIMPLE_QR_CODE
+        var viewType = QrCodeComponentViewType.SIMPLE_QR_CODE
```
ozgur00, line 156: It's better to have a `DEFAULT` in the enum with these values to keep everything in the same place. We can return `DEFAULT` if we can't find any matching enum entry.

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/model/QRCodePaymentMethodConfig.kt
```diff
@@ -0,0 +1,52 @@
...
+    );
+
+    companion object {
+        fun getByPaymentMethodTye(paymentMethodType: String): QRCodePaymentMethodConfig? {
+            return values().firstOrNull { it.paymentMethodType == paymentMethodType }
+        }
+    }
+}
+
+private val DUIT_NOW_MAX_POLLING_DURATION = 90.seconds.inWholeMilliseconds
```
ozgur00, line 50: These should be in the companion object for consistency.

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/view/FullQRCodeView.kt
```diff
@@ -152,12 +153,8 @@ internal class FullQRCodeView @JvmOverloads constructor(
     }
 
     @StringRes
-    private fun getMessageTextResource(paymentMethodType: String?): Int? {
-        return when (paymentMethodType) {
-            PaymentMethodTypes.PAY_NOW -> R.string.checkout_qr_code_pay_now
-            PaymentMethodTypes.UPI_QR -> R.string.checkout_qr_code_upi
-            else -> null
-        }
+    private fun getMessageTextResource(paymentMethodType: String): Int? {
```
ozgur00, line 156: I think this should be done in delegate and we should have a field in `QRCodeOutputData` for the resourceId. Here we should just set it to keep view as simple as possible.

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/model/QRCodePaymentMethodConfig.kt
onurkaragunlu: I can't do this because of following explanation
https://stackoverflow.com/questions/57560580/kotlin-enum-constructor-argument-must-be-initialized-with-companion-object-con

jreij: Do we need these variables anyway? If we only use them to initialize the enum how about we directly use `maxPollingDurationMillis = 90.seconds.inWholeMilliseconds,`?

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
onurkaragunlu: I thought it but I didn't do it because we might face the same problem like UNKNOWN in CardType and we don't have a payment method type for the default value.

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/model/QRCodePaymentMethodConfig.kt
onurkaragunlu: I thought it but I was not sure about whether they are magic number.

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegate.kt
ozgur00: Yeah but the use case is a bit different here. For instance in `AddressSpecification` we have `DEFAULT` to represent a ui state. I believe it's better to have `DEFAULT` as an entry.

qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/model/QRCodePaymentMethodConfig.kt
ozgur00: What @jreij suggested sounds good to me too. I think it's fine to use it directly in the definition.

ozgur00 (approved):

PR #1108, opened 2023-03-09
Use amount set in sessions

jreij:
## Description
For integrations using `/sessions`, the amount should come from the API response and not from the client side configuration.

## Checklist 
- [x] Code is unit tested

COAND-729


components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Created by josephj on 7/3/2023.
+ */
+
+package com.adyen.checkout.components.core.internal.ui.model
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.components.core.Amount
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface SessionParams {
```
OscarSpruit, line 15: Do we really need this to be an interface? Can't we use the same data class for testing and other things?

jreij: Hmm yeah looking at it now it doesn't make much sense anymore

OscarSpruit (approved):

Issue #1103, opened 2023-03-06
Fatal Exception: com.adyen.checkout.core.exception.ComponentException: Stored Blik is not yet supported in this flow.


michal-chojnacki:
**Describe the bug**
In current version of drop-in app is crashing when I try to pay with saved blik payment method

```
Fatal Exception: com.adyen.checkout.core.exception.ComponentException: Stored Blik is not yet supported in this flow.
       at com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodListDialogFragment.onStoredPaymentMethodSelected(PaymentMethodListDialogFragment.kt:134)
       at com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodAdapter.onStoredPaymentMethodClick(PaymentMethodAdapter.kt:228)
       at com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodAdapter.bindStoredPaymentMethod$lambda-5$lambda-4(PaymentMethodAdapter.kt:122)
       at com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodAdapter.$r8$lambda$72BzwUPHHO5rlFQ6A1wtgvxi84s()
       at com.adyen.checkout.dropin.ui.paymentmethods.PaymentMethodAdapter$$ExternalSyntheticLambda5.onClick(:4)
       at com.adyen.checkout.components.ui.view.AdyenSwipeToRevealLayout$gestureListener$1.onSingleTapConfirmed(AdyenSwipeToRevealLayout.kt:212)
       at android.view.GestureDetector$GestureHandler.handleMessage(GestureDetector.java:329)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:201)
       at android.os.Looper.loop(Looper.java:288)
       at android.app.ActivityThread.main(ActivityThread.java:7842)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```

**To Reproduce**
- Make payment with blik and save it as a payment method.
- Try to have second payment. On screen with all payment methods select saved blik method.

**Expected behavior**
App is not crashing and making payment with saved payment blik method is possible

OscarSpruit:
[...]

PR #1112, opened 2023-03-14
Add support for stored Blik payment method (v4)

OscarSpruit:
## Description
Stored Blik payment methods are now fully supported. Before when trying to open a stored Blik payment method from the payment methods list we threw an exception. Now we show a confirmation dialog and when the user confirms we make the payments call.

https://user-images.githubusercontent.com/17701279/225013423-92bb3262-f4b4-4577-89c7-912eeb57d34d.mp4

## Checklist 
- [ ] Code is unit tested
- [x] Link to related issues: #1103 

COAND-734

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
@@ -27,11 +27,12 @@ import com.adyen.checkout.components.model.payments.request.PaymentMethodDetails
 import com.adyen.checkout.components.ui.view.AdyenSwipeToRevealLayout
 import com.adyen.checkout.components.util.PaymentMethodTypes
 import com.adyen.checkout.core.exception.CheckoutException
-import com.adyen.checkout.core.exception.ComponentException
 import com.adyen.checkout.core.log.LogUtil
 import com.adyen.checkout.core.log.Logger
 import com.adyen.checkout.dropin.R
 import com.adyen.checkout.dropin.databinding.FragmentPaymentMethodsListBinding
+import com.adyen.checkout.dropin.getComponentFor
+import com.adyen.checkout.dropin.ui.PayButtonFormatter
 import com.adyen.checkout.dropin.ui.base.DropInBottomSheetDialogFragment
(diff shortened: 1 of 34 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -199,4 +202,48 @@ class PaymentMethodListDialogFragment :
...
+                )
+            ) { dialog, _ ->
+                dialog.dismiss()
+                onConfirmationButtonClicked(storedPaymentMethod)
+            }
+            .show()
+    }
+
+    private fun onConfirmationButtonClicked(storedPaymentMethod: StoredPaymentMethod) {
+        val component = getComponentFor(
```
jreij, line 231: We need to pass a key here to create a new component per stored payment methods (you can use `storedPaymentMethod.id`).
Otherwise suppose we have 2 stored blik, if we click on the first one then cancel then click on the 2nd one we will still get the same component as the 1st.

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
```diff
@@ -199,4 +202,48 @@ class PaymentMethodListDialogFragment :
...
+            fragment = this,
+            storedPaymentMethod = storedPaymentMethod,
+            dropInConfiguration = dropInViewModel.dropInConfiguration,
+            amount = dropInViewModel.amount,
+        )
+
+        component.observe(viewLifecycleOwner) { componentState ->
+            if (componentState.isValid) {
+                protocol.requestPaymentsCall(componentState)
```
jreij, line 240: Could this fire multiple times? On config changes or something?

drop-in/src/main/java/com/adyen/checkout/dropin/ui/paymentmethods/PaymentMethodListDialogFragment.kt
OscarSpruit: Good point 👍

jreij (approved):

PR #1114, opened 2023-03-15
Update kdocs for 3ds2

OscarSpruit:
COAND-724


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Authentication3DS2Exception.kt
```diff
@@ -11,7 +11,7 @@ import com.adyen.checkout.core.exception.ComponentException
 
 /**
  * This exception is just an indication that the 3DS2 Authentication did not finish as expected.
- * Can be caused by an actual error or by user cancellation.
+ * Can be caused by an actual error, like a timeout or a runtime exception.
```
jreij, line 14: Why are we removing `user cancellation`?

OscarSpruit: `Cancelled3DS2Exception` is thrown when the user cancels

jreij (approved):

PR #1116, opened 2023-03-17
Create phone number input view

OscarSpruit:
## Description
Before we had phone number input in separate components, but with this the logic and views are extracted into one place. Makes it easily reusable and prevents duplication.

## Checklist 
- [x] Code is unit tested

COAND-735


mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/model/MBWayInputData.kt
```diff
@@ -10,6 +10,7 @@ package com.adyen.checkout.mbway.internal.ui.model
 import com.adyen.checkout.components.core.internal.ui.model.InputData
 
 internal data class MBWayInputData(
-    var countryCode: String = "",
-    var localPhoneNumber: String = ""
+    // The input data is taken care of by PhoneNumberDelegate.
+    // We have a dummy field here just in case there will be extra input fields added later.
+    var dummy: String = "",
```
jreij, line 15: I'm not sure about this field 🤔 Can we include `PhoneNumberInputData` inside `MBWayInputData` just like we do with the address data?

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/ui/model/PhoneNumberInputData.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package com.adyen.checkout.ui.core.internal.ui.model
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.components.core.internal.ui.model.InputData
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class PhoneNumberInputData(
+    var countryCode: String = "",
+    var everythingAfterCountryCode: String = "",
```
jreij, line 17: Nitpick: can we rename this? 😅 I don't have a much better suggestion, maybe something like `numberWithoutCountryCode`?

OscarSpruit: Haha I still needed to change that indeed. There is not an official name for it since it consists out of multiple parts, but some sources just call it phone number. Should we call it `phoneNumber` as well?

onurkaragunlu: I think `phoneNumber` is clear. We can use it

OscarSpruit:
We need to think about the structure of all custom views, so I will close this PR for now

PR #1121, opened 2023-03-21
Add kdocs for all "simple" payment methods

OscarSpruit:
## Description
Add kdocs for all payment method modules that basically only have a component, component state and configuration.

COAND-724

### DIFF
--- a/ach/src/main/java/com/adyen/checkout/ach/ACHDirectDebitComponent.kt
+++ b/ach/src/main/java/com/adyen/checkout/ach/ACHDirectDebitComponent.kt
@@ -32,7 +32,7 @@ import com.adyen.checkout.ui.core.internal.util.mergeViewFlows
 import kotlinx.coroutines.flow.Flow
 
 /**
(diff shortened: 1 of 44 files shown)


jreij (approved):

ach/src/main/java/com/adyen/checkout/ach/ACHDirectDebitComponent.kt
```diff
@@ -32,7 +32,7 @@ import com.adyen.checkout.ui.core.internal.util.mergeViewFlows
 import kotlinx.coroutines.flow.Flow
 
 /**
- * An [PaymentComponent] that supports the 'ach' payment method.
+ * A [PaymentComponent] that supports the 'ach' payment method.
```
ozgur00, line 35: I think it looks better to link the `txVariant` to `PaymentMethodTypes` using `[com.adyen.checkout.components.core.PaymentMethodTypes.ACH]`. Then people can click through the docs.

PR #1138, opened 2023-03-28
Add missing kdocs for sessions-core

jreij:
COAND-724

### DIFF
--- a/sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSession.kt
+++ b/sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSession.kt
@@ -11,13 +11,16 @@ package com.adyen.checkout.sessions.core
 import com.adyen.checkout.components.core.Order
 import com.adyen.checkout.components.core.PaymentMethod
 
-// TODO docs
+/**
+ * A class holding the data required to launch Drop-in or a component with the sessions flow.
+ * Use [CheckoutSessionProvider.createSession] to create this class.
+ */
 data class CheckoutSession(
     val sessionSetupResponse: SessionSetupResponse,
     val order: Order?,
 ) {
     fun getPaymentMethod(paymentMethodType: String): PaymentMethod? {
-        return sessionSetupResponse.paymentMethods?.paymentMethods.orEmpty().firstOrNull {
+        return sessionSetupResponse.paymentMethodsApiResponse?.paymentMethods.orEmpty().firstOrNull {
             it.type == paymentMethodType
         }
     }
--- a/sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSessionProvider.kt
+++ b/sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSessionProvider.kt
@@ -8,31 +8,29 @@
 
 package com.adyen.checkout.sessions.core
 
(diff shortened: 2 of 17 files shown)


sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSession.kt
```diff
@@ -11,13 +11,16 @@ package com.adyen.checkout.sessions.core
 import com.adyen.checkout.components.core.Order
 import com.adyen.checkout.components.core.PaymentMethod
 
-// TODO docs
+/**
+ * A class holding the data required to launch Drop-in or a component with teh sessions flow.
```
OscarSpruit, line 15: Teh

sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSessionProvider.kt
```diff
@@ -8,31 +8,29 @@
...
 import com.adyen.checkout.sessions.core.internal.CheckoutSessionInitializer
-import kotlinx.coroutines.CoroutineScope
 
-// TODO docs
 object CheckoutSessionProvider {
-    // TODO docs
+
+    /**
+     * Allows creating a [CheckoutSession] from the response of the /sessions endpoint.
+     * This is a suspend function that executes a network call. Make sure you run this in a background thread.
```
OscarSpruit, line 19: Should we use `withContext` to prevent mistakes?

sessions-core/src/main/java/com/adyen/checkout/sessions/core/SessionComponentCallback.kt
```diff
@@ -9,27 +9,175 @@
...
+     * to be thrown.
+     *
+     * See https://docs.adyen.com/api-explorer/ for more information on the API documentation.
+     *
+     * @return [true] if you took over the sessions flow, [false] otherwise.
+     */
     fun onOrderRequest(): Boolean = false
-    fun onOrderCancel(order: Order): Boolean = false
 
+    fun onOrderCancel(order: Order): Boolean = false // TODO remove this method?
```
OscarSpruit, line 176: Why should we remove this method?

jreij: It won't be applicable to standalone components

sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSession.kt
jreij: 🙃

sessions-core/src/main/java/com/adyen/checkout/sessions/core/CheckoutSessionProvider.kt
jreij: Can you check again?

sessions-core/src/main/java/com/adyen/checkout/sessions/core/SessionComponentCallback.kt
OscarSpruit: Ahh right, we'll probably remove it then when working more on partial payments

OscarSpruit (approved):

PR #1139, opened 2023-03-29
Fixes on class visibility

jreij:
## Description
- Make `CheckoutCurrency` public
- Create public `AdyenLogger` to change the log level
- Remove explicit companion imports

## Checklist 

COAND-497


3ds2/src/test/java/com/adyen/checkout/adyen3ds2/Adyen3DS2ComponentTest.kt
```diff
@@ -47,7 +48,7 @@ internal class Adyen3DS2ComponentTest(
 
     @BeforeEach
     fun before() {
-        Logger.setLogcatLevel(Logger.NONE)
+        AdyenLogger.setLogLevel(Logger.NONE)
```
OscarSpruit, line 51: Just an idea that doesn't have to do with this PR: we should create our own test extension which does this and possibly other generic setup things

OscarSpruit (approved):

PR #1147, opened 2023-04-04
Separate CardView into default and stored views

onurkaragunlu:
Card view has been separated into StoreCardView and DefaultCardview.

## Checklist 
- [x] Code is unit tested

COAND-740


card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    private fun outputDataChanged(cardOutputData: CardOutputData) {
+        binding.editTextCardNumber.setText(
+            localizedContext.getString(
+                R.string.card_number_4digit,
+                cardOutputData.cardNumberState.value
+            )
+        )
+        binding.editTextExpiryDate.date = cardOutputData.expiryDateState.value
+        binding.textInputLayoutSecurityCode.takeIf { isVisible }?.requestFocus()
```
OscarSpruit, line 139: Why are we requesting focus when the output data changes?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        setDetectedCardBrand(cardOutputData)
+    }
+
+    private fun setDetectedCardBrand(cardOutputData: CardOutputData) {
+        val detectedCardTypes = cardOutputData.detectedCardTypes
+        binding.cardBrandLogoImageViewPrimary.strokeWidth = RoundCornerImageView.DEFAULT_STROKE_WIDTH
+        binding.cardBrandLogoImageViewPrimary.loadLogo(
+            environment = cardDelegate.componentParams.environment,
+            txVariant = detectedCardTypes[0].cardBrand.txVariant,
```
OscarSpruit, line 149: What if the list is empty?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
```diff
@@ -0,0 +1,173 @@
...
+            placeholder = R.drawable.ic_card,
+            errorFallback = R.drawable.ic_card,
+        )
+    }
+
+    override fun highlightValidationErrors() {
+        cardDelegate.outputData.let {
+            val securityCodeValidation = it.securityCodeState.validation
+            if (securityCodeValidation is Validation.Invalid) {
```
OscarSpruit, line 158: We also have to show the actual error with `binding.textInputLayoutSecurityCode.showError(...)`

card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
onurkaragunlu: I should move it to `initSecurityCodeInput`.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
onurkaragunlu: Since this is for stored card, it should not be empty ,right?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
onurkaragunlu: nice catch! I will update here.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardView.kt
OscarSpruit: We can never be fully sure. Better to be safe and only load the logo if there is a detected card type

OscarSpruit (approved):

PR #1150, opened 2023-04-05
Cleanup build output

OscarSpruit:
## Description
- Replace manifest package name with namespace in build.gradle
- Ignore formatting warning for online banking string res
- Update Detekt ruleset and apply  new rules
- Remove unnecessary logs of dependency checksum check
- Fix some warnings

## Checklist 


config/gradle/dependenciesCheck.gradle
```diff
@@ -49,8 +49,6 @@ afterEvaluate {
 
                     assert checksum == calculatedChecksum: "Checksum ($algorithm) does not match: $dependency.name"
                 }
-
-                logger.lifecycle("\u2705 $configuration: $dependency.name")
```
jreij, line 53: Why do we need to remove this?

paybybank/src/main/res/values-cs-rCZ/strings.xml
```diff
@@ -8,6 +8,6 @@
   -->
 
 <resources>
-    <string name="checkout_pay_by_bank_search">Hledat...</string>
+    <string name="checkout_pay_by_bank_search">Hledat…</string>
```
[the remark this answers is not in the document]
OscarSpruit: Don't worry, I already did that a while back. That's why this got replaced with `…`

config/gradle/dependenciesCheck.gradle
OscarSpruit: Yes, imo it doesn't really add anything. It's only interesting to know when it fails

jreij (approved):

PR #1156, opened 2023-04-12
Do not send storePaymentMethod if shopper cannot explicitly set it

jreij:
## Description
Currently we are sending `storePaymentMethod = false` by default even if no switch is displayed, which can override some merchant configurations on the API and cause issues.

We need to send this parameter as null in this case.

## Checklist 
- [x] Code is unit tested

COAND-745


ach/src/main/java/com/adyen/checkout/ach/internal/ui/StoredACHDirectDebitDelegate.kt
```diff
@@ -136,7 +136,7 @@ internal class StoredACHDirectDebitDelegate(
             ownerName = FieldState(ownerName, Validation.Valid),
             addressState = AddressValidationUtils.makeValidEmptyAddressOutput(inputData.address),
             addressUIState = AddressFormUIState.NONE,
-            isStoredPaymentMethodEnabled = isStorePaymentSelected,
+            shouldStorePaymentMethod = isStorePaymentMethodSwitchChecked,
```
OscarSpruit, line 139: Why do we have to set this when the payment method is already stored?

OscarSpruit (approved):

PR #1161, opened 2023-04-14
Don't show border on stored GooglePay

OscarSpruit:
## Description
| Pre-selected | Payment method list |
|-|-|
| image | image |

## Checklist 

COAND-747


drop-in/src/main/java/com/adyen/checkout/dropin/internal/util/StoredUtils.kt
```diff
@@ -29,6 +29,7 @@ internal fun StoredPaymentMethod.mapStoredModel(
                 lastFour = lastFour.orEmpty(),
                 expiryMonth = expiryMonth.orEmpty(),
                 expiryYear = expiryYear.orEmpty(),
+                drawIconBorder = !brand.orEmpty().contains(PaymentMethodTypes.GOOGLE_PAY),
```
ozgur00, line 32: shall we have a list of payment method types that doesn't require a border on the icon? in the future if we need to do it for another one we would just add the payment method type to that list and we won't have to change anything here. what do you think?

PR #1174, opened 2023-04-25
set amount for cardConfiguration

pinpong:
Updated configuration to show the amount


jreij:
Hi @pinpong, thanks for the contribution! I only have 2 comments:
- Can you also update all the other configurations that are not calling `setAmount`? They are all in the same class and the flow is the same among them anyway.
- Please note that you are on the `develop` branch which contains lots of breaking changes and unreleased features as part of upcoming major release. If you would like to work on the code that mirrors the actual latest released version please checkout the `v4` branch.

pinpong:
removed the setAmount exception handling because build function throws another one that was not catched.

jreij:
> removed the setAmount exception handling because build function throws another one that was not catched.

What kind of exception are you getting? And what value are you passing to the amount?

pinpong:
It was just a hint to the changes. `setAmount()` can throw a CheckoutException as well `build()`.
`build()` was not catched within the the configuration builders.

jreij (approved):
LGTM ✨

PR #1179, opened 2023-04-28
Make obfuscation work correctly

OscarSpruit:
## Description
The example app now works with obfuscation and loading logo's as well. Everything else already seems to work.

## Checklist 

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/data/api/model/AdditionalData.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/data/api/model/AdditionalData.kt
@@ -8,6 +8,9 @@
 
 package com.adyen.checkout.example.data.api.model
 
+import androidx.annotation.Keep
+
+@Keep
 data class AdditionalData(
     val allow3DS2: String = "false",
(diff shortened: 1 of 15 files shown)


example-app/src/main/java/com/adyen/checkout/example/data/api/model/AdditionalData.kt
```diff
@@ -8,6 +8,9 @@
 
 package com.adyen.checkout.example.data.api.model
 
+import androidx.annotation.Keep
+
+@Keep
```
jreij, line 13: Is this better than keeping all classes in `com.adyen.checkout.example.data.api.model`?

OscarSpruit: It's just preference. I like this more because it's explicit and if you move classes you don't have to manually update the proguard rules. For `LogoSize` I did add a proguard rule though, because I am not sure if `@Keep` works well in libraries

jreij (approved):

PR #1206, opened 2023-05-09
Send brand in paymentComponentData for single branded cards

ozgur00:
## Checklist 
- [x] Code is unit tested

COAND-761

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
@@ -647,11 +647,7 @@ internal class DefaultCardDelegate(
                 taxNumber = stateOutputData.kcpBirthDateOrTaxNumberState.value
             }
 
-            if (isDualBrandedFlow(stateOutputData.detectedCardTypes)) {
-                brand = DetectedCardTypesUtils.getSelectedCardType(
-                    detectedCardTypes = stateOutputData.detectedCardTypes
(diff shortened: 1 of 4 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -733,6 +730,19 @@ internal class DefaultCardDelegate(
         }
     }
 
+    private fun getCardBrand(stateOutputData: CardOutputData): String? {
```
OscarSpruit, line 733: It looks like only `detectedCardTypes` is used from the output data. Only passing that would look a bit cleaner

jreij:
@ozgur00 can we make sure to update BCMC as well in case it has the same issue?

OscarSpruit (approved):

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -733,6 +730,18 @@ internal class DefaultCardDelegate(
         }
     }
 
+    private fun getCardBrand(detectedCardTypes: List<DetectedCardType>): String? {
+        return if (isDualBrandedFlow(detectedCardTypes)) {
```
jreij, line 734: Minor: instead of calling `?.cardBrand?.txVariant` 3 times you can do it only once after the whole if statement

jreij (approved):

PR #1218, opened 2023-05-23
Cash App Pay component

jreij:
## Description
Adding a new component to support Cash App Pay in both the normal and stored flows.

## Checklist 

COAND-634


cashapppay/src/main/java/com/adyen/checkout/cashapppay/DefaultCashAppPayDelegate.kt
```diff
@@ -0,0 +1,231 @@
...
+            getOnFileAction(cashAppParams, outputData),
+        )
+
+        if (actions.isEmpty()) {
+            throw ComponentException("Cannot launch Cash App Pay, you need to either pass an amount or store the shopper account")
+        }
+
+        withContext(Dispatchers.IO) {
+            // must be called from a background thread according to Cash App Pay docs
+            cashAppPay.createCustomerRequest(actions, cashAppParams.returnUrl)
```
OscarSpruit, line 85: Does this actually do IO work? If not, the default dispatcher would be better

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -253,16 +261,21 @@ internal fun getComponentFor(
     dropInConfiguration: DropInConfiguration,
     amount: Amount
 ): PaymentComponent<PaymentComponentState<in PaymentMethodDetails>, Configuration> {
-
+    val context = fragment.requireContext()
```
OscarSpruit, line 264: Is it possible that we call this before the fragment is actually created or after it's been destroyed?

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
```diff
@@ -285,87 +298,98 @@ internal fun getComponentFor(
...
```
OscarSpruit, line 301: Same question here

drop-in/src/main/res/layout/fragment_cash_app_pay_component.xml
```diff
@@ -0,0 +1,96 @@
...
+
+    <TextView
+        android:id="@+id/header"
+        style="@style/AdyenCheckout.TextAppearance.HeaderTitle"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:textIsSelectable="false"
+        tools:text="Header" />
+
+    <LinearLayout
```
OscarSpruit, line 29: This `LinearLayout` is unnecessary I think

cashapppay/src/main/java/com/adyen/checkout/cashapppay/DefaultCashAppPayDelegate.kt
jreij: This is from the Cash App SDK, I think we should explicitly specify `Dispatchers.IO` right?
```kotlin
  /**
   * Create customer request given list of [CashAppPayPaymentAction].
   *
   * Must be called from a background thread.
   * ...
   */
  @WorkerThread
  fun createCustomerRequest(paymentActions: List<CashAppPayPaymentAction>, redirectUri: String?)
```

drop-in/src/main/java/com/adyen/checkout/dropin/ComponentParsingProvider.kt
jreij: Seeing where we call it from now, it's mostly in fragments inside `onCreate`

cashapppay/src/main/java/com/adyen/checkout/cashapppay/DefaultCashAppPayDelegate.kt
OscarSpruit: Checked the code and it does a network call, so `IO` is perfect 👍

OscarSpruit (approved):

PR #1269, opened 2023-07-19
Telemetry: setup

jreij:
## Description
First set of features required for the new telemetry integration.

## Checklist 
- [x] Code is unit tested

COAND-481

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/DefaultAnalyticsRepository.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/DefaultAnalyticsRepository.kt
@@ -25,13 +25,13 @@ class DefaultAnalyticsRepository(
     private val clientKey: String,
 ) : AnalyticsRepository {
 
-    override suspend fun sendAnalyticsEvent() {
+    override suspend fun setupAnalytics() {
         runSuspendCatching {
             val analyticsSetupRequest = analyticsMapper.getAnalyticsSetupRequest(packageName, locale, source)
             analyticsService.setupAnalytics(analyticsSetupRequest, clientKey)
-            Logger.v(TAG, "Analytics event sent")
+            Logger.v(TAG, "Analytics setup call successful")
         }
-            .onFailure { e -> Logger.e(TAG, "Failed to send analytics event", e) }
+            .onFailure { e -> Logger.e(TAG, "Failed to send analytics setup call", e) }
     }
 
     companion object {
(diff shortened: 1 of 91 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/internal/data/model/AnalyticsSetupRequest.kt
```diff
@@ -0,0 +1,111 @@
...
+    val channel: String?,
+    val platform: String?,
+    val locale: String?,
+    val component: String?,
+    val flavor: String?,
+    val deviceBrand: String?,
+    val deviceModel: String?,
+    val referrer: String?,
+    val systemVersion: String?,
+    val containerWidth: Long?,
```
OscarSpruit, line 34: Do you think this will ever be bigger than `Int.MAX_VALUE`?

jreij: Definitely not, but I was just mirroring the types on the backend

OscarSpruit: Ohh that makes sense 👍

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/DefaultAnalyticsRepository.kt
```diff
@@ -22,18 +23,38 @@ class DefaultAnalyticsRepository(
...
         }
-            .onFailure { e -> Logger.e(TAG, "Failed to send analytics event", e) }
     }
 
     companion object {
         private val TAG = LogUtil.getTag()
     }
+
+    @VisibleForTesting
+    internal sealed class State {
```
ozgur00, line 54: Names of subclasses are too similar to each other and that kinda makes it hard to read. I think we can rename the parent class to `InitializationState` and sub classes to
`Initializing` -> `InProgress`
`Initialized` -> `Ready`
`InitializationFailed` -> `Failed`

Couldn't find a replacement for `Uninitialized`. We can leave it as is or may be you or @OscarSpruit have an idea?

OscarSpruit: I like your suggestion

jreij: Fixed 👍

ozgur00 (approved):

PR #1283, opened 2023-08-02
Allow CVC to be empty when bin lookup returns "hidden"

jreij:
## Description
CVC field is still being validated as required when bin lookup returns that the CVC should be "hidden". 
We should treat hidden as optional and allow this field to be empty when bin lookup returns "hidden".

## Checklist 
- [x] Code is unit tested

COAND-783

### DIFF
--- a/card/src/test/java/com/adyen/checkout/card/internal/util/CardValidationUtilsTest.kt
+++ b/card/src/test/java/com/adyen/checkout/card/internal/util/CardValidationUtilsTest.kt
@@ -8,168 +8,520 @@
 
 package com.adyen.checkout.card.internal.util
 
+import com.adyen.checkout.card.CardBrand
+import com.adyen.checkout.card.CardType
+import com.adyen.checkout.card.R
+import com.adyen.checkout.card.internal.data.model.Brand
+import com.adyen.checkout.card.internal.data.model.DetectedCardType
 import com.adyen.checkout.card.internal.ui.model.ExpiryDate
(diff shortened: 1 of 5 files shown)


card/src/test/java/com/adyen/checkout/card/internal/util/CardValidationUtilsTest.kt
```diff
@@ -8,168 +8,520 @@
...
+            val actual = CardValidationUtils.validateExpiryDate(
+                expiryDate = expiryDate,
+                fieldPolicy = Brand.FieldPolicy.REQUIRED,
+                calendar = mockCalendarInstance
+            )
+            assertEquals(FieldState(expiryDate, Validation.Valid), actual)
+        }
+
+        @Test
+        fun `date is more than 30 years in the future then result should be invalid as not within max range`() {
```
OscarSpruit, line 163: A parameterised test would have saved you some work and is also easier to maintain

jreij: Considered it multiple times but ended up not doing it, do you think it would be valuable to change the current tests to parameterised tests?

OscarSpruit: No not necessarily. I do prefer them though as it saves me from writing extra code and it's easier to add/change cases later on

OscarSpruit (approved):

PR #1287, opened 2023-08-07
On bin changed callback

OscarSpruit:
## Description
Add on bin value callback to CardComponent. This callback is called when the shopper enters their card number and the bin value changes.

## Checklist 
- [x] Code is unit tested

COAND-732


drop-in/src/main/java/com/adyen/checkout/dropin/BaseDropInServiceContract.kt
```diff
@@ -94,4 +94,11 @@ interface BaseDropInServiceContract {
      * Set a callback that will be called when a redirect is made.
      */
     fun onRedirect() = Unit
+
+    /**
+     * Set a callback that will be called when the bin value changes.
```
jreij, line 99: Should we clarify that this is only applicable to the card component?

jreij (approved):

PR #1289, opened 2023-08-08
Feature/on bin lookup callback

OscarSpruit:
## Description
Add on bin lookup callback to CardComponent. This callback is called when a bin lookup is performed.

## Checklist 
- [x] Code is unit tested

COAND-732


card/src/main/java/com/adyen/checkout/card/BinLookupData.kt
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Class that contains data about the entered card number.
+ *
+ * @param brand The detected brand.
+ * @param paymentMethodVariant The type of payment method. See the
+ * [docs](https://docs.adyen.com/development-resources/paymentmethodvariant/) for more information.
+ * @param isReliable Indicates whether the data is reliable. If true, the card number is checked with Adyen systems.
+ * If false, the data is provided by a local detection algorithm.
+ */
+data class BinLookupData(
```
jreij, line 12: I wonder if we should call this something else, bin lookup is an internal naming and might not be obvious. What about "card details"? Can we sync with other platforms and product?

card/src/main/java/com/adyen/checkout/card/BinLookupData.kt
```diff
@@ -0,0 +1,16 @@
...
+    val brand: String,
+    val paymentMethodVariant: String?,
```
jreij, line 14: Maybe we should call this `paymentMethodType`, since it's called `type` on the API

card/src/main/java/com/adyen/checkout/card/BinLookupData.kt
OscarSpruit: I do like "card details" more as well. The callback is already called onBinLookup though, because it's name they use for web. You think we should change it?

card/src/main/java/com/adyen/checkout/card/BinLookupData.kt
OscarSpruit: There is a [docs page](https://docs.adyen.com/development-resources/paymentmethodvariant/) on this field and web also uses this name. Are you sure we should change it?

jreij: Nevermind then 👍

card/src/main/java/com/adyen/checkout/card/BinLookupData.kt
jreij: Yeah I would say card details is clearer but if iOS and web use bin lookup then we could go for either 🤷

OscarSpruit: On iOS it's called `didChangeCardBrand`. So we can choose for consistency with one of the platforms or we do our own naming

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -233,6 +236,9 @@ internal class DefaultCardDelegate(
                     "New detected card types emitted - detectedCardTypes: ${detectedCardTypes.map { it.cardBrand }} " +
                         "- isReliable: ${detectedCardTypes.firstOrNull()?.isReliable}"
                 )
+                if (outputData.detectedCardTypes != detectedCardTypes) {
```
ozgur00, line 239: Can we do 
```
detectCardTypeRepository.detectedCardTypesFlow
            .distinctUntilChanged()
            .onEach { detectedCardTypes ->
```
instead of having this if? I think it wouldn't change the behaviour. It would get triggered fewer times but should be fine. What do you think?

OscarSpruit: It does actually change the behaviour, because currently we update the output data even when the detected card types are the same as before. I like your suggestion though, because it doesn't make sense to update the output data if nothing changed

ozgur00 (approved):

PR #1290, opened 2023-08-08
Suppress ProGuard issues coming from Cash App SDK

OscarSpruit:
## Description
Suppress the errors that ProGuard gives. The issue stems from the Cash App SDK.


jreij (approved):

jreij:
@OscarSpruit can we revert this?

OscarSpruit:
We probably can now we updated the Cash App SDK, but we should test it to be sure

PR #1292, opened 2023-08-08
Propagate details on error, cancel and timeout from 3ds2 sdk

ozgur00:
## Description
Update 3ds2 sdk to 2.2.14 and start using the new onCompletion callback to get challenge results. Before this change details were emitted only for completion event, however starting from now they will be emitted for all results received from 3ds2 sdk for merchants to be able to get more information about the errors they encounter during the flow.

## Checklist 
- [x] Code is unit tested

COAND-748


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
```diff
@@ -404,32 +400,59 @@ internal class DefaultAdyen3DS2Delegate(
...
         Logger.d(TAG, "challenge timed out")
+        try {
+            val details = makeDetails(result.transactionStatus, result.additionalDetails)
+            emitDetails(details)
+        } catch (e: CheckoutException) {
+            exceptionChannel.trySend(e)
+        } finally {
+            closeTransaction()
+        }
         exceptionChannel.trySend(Authentication3DS2Exception("Challenge timed out."))
```
jreij, line 428: Wondering if we should still emit exceptions since we're emitting details now 🤔

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
```diff
@@ -404,32 +400,59 @@ internal class DefaultAdyen3DS2Delegate(
...
+        try {
+            val details = makeDetails(result.transactionStatus, result.additionalDetails)
+            emitDetails(details)
+        } catch (e: CheckoutException) {
+            exceptionChannel.trySend(e)
+        } finally {
+            closeTransaction()
         }
+        exceptionChannel.trySend(Authentication3DS2Exception("An error occured during challenge."))
         closeTransaction()
```
jreij, line 443: `closeTransaction()` will be called twice in this flow right?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
```diff
@@ -404,32 +400,59 @@ internal class DefaultAdyen3DS2Delegate(
...
+    override fun onCompletion(result: ChallengeResult) {
+        when (result) {
+            is ChallengeResult.Cancelled -> onCancelled(result)
+            is ChallengeResult.Completed -> onCompleted(result.transactionStatus)
+            is ChallengeResult.Error -> onError(result)
+            is ChallengeResult.Timeout -> onTimeout(result)
+        }
     }
 
+    // no need to call this, it is called before onCompletion
```
jreij, line 455: Can you explain this a bit more?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
ozgur00: Removed the extra call.

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
ozgur00: Apparently now `close()` is called internally in 3ds2 sdk. I left a comment like this to discuss if we should call close anyway ourselves because when I see the example implementation in 3ds2 sdk they also do what we do. I say let's keep our implementation for close.

jreij (approved):

PR #1297, opened 2023-08-15
Allow more flexibility with dialog in drop-in

OscarSpruit:
## Description
Make showing a dialog optional when `DropInService` gets an error. Also, give the option to show a dialog when the payment flow is finished.

I didn't want to change the Error class too much and for Finished I wanted to restrict nullability, so they are inconsistent. What do you prefer: consistency or backwards compatibility?

Screenshot_20230815_163258

## Checklist 

COAND-779


drop-in/src/main/java/com/adyen/checkout/dropin/DropInServiceResult.kt
```diff
@@ -179,8 +189,20 @@ internal sealed class SessionDropInServiceResult : BaseDropInServiceResult() {
...
     class Finished(val result: SessionPaymentResult) : SessionDropInServiceResult()
 }
+
+/**
+ * Represents the data shown in the dialog when the payment flow is finished.
+ *
+ * @param title The title displayed in the dialog.
+ * @param message The message displayed in the dialog.
+ */
+data class FinishedDialog(
```
jreij, line 205: What if we use the same class for the finished and error states?

OscarSpruit: That is one of the points I meant to discuss. The error state already had message stuff, so I left it like that to prevent API changes. For the finished state I wanted to force nullability, so I used this class. Oh and naming suggestions are welcome for this class, I don't really like this name :D

jreij: I think the `FinishedDialog` class you added is much better than what we currently have in the `Error` case:
- `errorMessage` is used to set the dialog message, but if not set then there will be a dialog with a generic message anyway, very counter intuitive
- The title cannot be set by the merchant
- `reason` is only returned back in the final result, also not clear that it won't be shown to the user
- Even though we do specify all these points in the kdocs, it should be a bit more intuitive
- And with `showDialog` now we will have 2 separate fields for that dialog (3 if we add title), better group them in a class

So renaming `FinishedDialog` to something more generic and using it in both classes seems to me worth the breaking change. WDYT?

Also, I didn't really get this point, can you explain a bit more?
> For the finished state I wanted to force nullability, so I used this class

OscarSpruit: > For the finished state I wanted to force nullability, so I used this class

What I mean is that for the finished case we don't have default values, so by making the `title` and `message` strings non-nullable we force merchants to give us these. For errors we do have default values, so title and message are nullable to make use of the defaults. If we use the same class for error and finished we need to figure out how to handle this

jreij: Ahh ok I understand 👍 It could be 2 similar classes then?

ozgur00 (approved):

PR #1298, opened 2023-08-15
Add event to navigate to the payment methods list

OscarSpruit:
## Description
Add event to navigate to the payment methods list

## Checklist 

COAND-736

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/DropInServiceResult.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/DropInServiceResult.kt
@@ -82,7 +82,15 @@ sealed class DropInServiceResult : BaseDropInServiceResult() {
         override val dismissDropIn: Boolean = false
     ) : DropInServiceResult(), DropInServiceResultError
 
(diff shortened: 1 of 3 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/DropInServiceResult.kt
```diff
@@ -81,6 +81,8 @@ sealed class DropInServiceResult : BaseDropInServiceResult() {
         override val reason: String? = null,
         override val dismissDropIn: Boolean = false
     ) : DropInServiceResult(), DropInServiceResultError
+
+    class ToPaymentMethodsList : DropInServiceResult()
```
jreij, line 85: Wondering if we should add the payment methods list as an optional parameter here, to allow reloading drop-in with new payment methods

OscarSpruit: I will check if I can get that to work

jreij (approved):

PR #1304, opened 2023-08-22
Fix QR code actions for InstantPaymentComponent

OscarSpruit:
## Description
Using a payment method like Duitnow would crash. This was because the delegate received by the QR code view was not the correct delegate type yet. 

## Checklist 

COAND-786

### DIFF
--- a/instant/src/main/java/com/adyen/checkout/instant/InstantPaymentComponent.kt
+++ b/instant/src/main/java/com/adyen/checkout/instant/InstantPaymentComponent.kt
@@ -32,7 +32,10 @@ class InstantPaymentComponent internal constructor(
     ViewableComponent,
     ActionHandlingComponent by actionHandlingComponent {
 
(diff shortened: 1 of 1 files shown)


instant/src/main/java/com/adyen/checkout/instant/InstantPaymentComponent.kt
```diff
@@ -32,7 +32,9 @@ class InstantPaymentComponent internal constructor(
     ViewableComponent,
     ActionHandlingComponent by actionHandlingComponent {
 
-    override val delegate: ComponentDelegate get() = actionHandlingComponent.activeDelegate
+    // Using actionHandlingComponent.activeDelegate will crash for QR code actions. This is a workaround for the actual
```
ozgur00, line 35: Since this is a workaround maybe put a `FIXME` or `TODO` as a reminder?

ozgur00 (approved):

PR #1318, opened 2023-08-28
Add session ID to telemetry data

jreij:
## Description
Include the sessionId when making the initial telemetry data call.

## Checklist 
- [x] Code is unit tested

COAND-793

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsRepositoryData.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsRepositoryData.kt
@@ -31,26 +31,31 @@ data class AnalyticsRepositoryData(
     val paymentMethods: List<String>,
     val sessionId: String?,
 ) {
+
     constructor(
         application: Application,
         componentParams: ComponentParams,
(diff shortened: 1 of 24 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsRepositoryData.kt
```diff
@@ -39,24 +40,53 @@ data class AnalyticsRepositoryData(
...
         source = AnalyticsSource.PaymentComponent(componentParams.isCreatedByDropIn, paymentMethod),
         paymentMethodType = paymentMethod.type,
+        sessionId = null,
+    )
+
+    constructor(
+        application: Application,
+        componentParams: ComponentParams,
+        paymentMethod: PaymentMethod,
+        sessionId: String,
```
OscarSpruit, line 50: Looks like f you make this nullable and give it null as default value, the other 2 constructors are no longer needed

jreij: Of course 🤦

OscarSpruit (approved):

PR #1350, opened 2023-09-27
Implement JWE encryption

OscarSpruit:
## Description
- Replace old encryption with JWE encryption
- Rename every occurrence of `encrypter` to `encryptor`

## Checklist 
- [x] Code is unit tested

COAND-655


cse/src/main/java/com/adyen/checkout/cse/CardEncryptor.kt
```diff
@@ -8,19 +8,15 @@
...
-import com.adyen.checkout.cse.internal.DefaultCardEncrypter
-import com.adyen.checkout.cse.internal.DefaultGenericEncrypter
+import com.adyen.checkout.cse.internal.CardEncryptorFactory
 
 /**
  * Allows the encryption of card data to be sent to Adyen's APIs.
  * Use this class with the custom card integration.
  */
-object CardEncrypter {
+object CardEncryptor {
```
jreij, line 17: Is this a breaking change?

cse/src/main/java/com/adyen/checkout/cse/GenericEncryptor.kt
```diff
@@ -8,18 +8,15 @@
...
-import com.adyen.checkout.cse.internal.DateGenerator
-import com.adyen.checkout.cse.internal.DefaultGenericEncrypter
+import com.adyen.checkout.cse.internal.GenericEncryptorFactory
 
 /**
  * Allows the encryption of any type of data to be sent to Adyen's APIs.
  * Use this class with custom component integrations.
  */
-object GenericEncrypter {
+object GenericEncryptor {
```
jreij, line 17: Is this a breaking change?

cse/src/main/java/com/adyen/checkout/cse/internal/CardEncryptorFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+
+package com.adyen.checkout.cse.internal
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object CardEncryptorFactory {
+
+    fun provide(): BaseCardEncryptor {
+        val dateGenerator = DateGenerator()
```
jreij, line 17: Is there a way to not instantiate these dependencies here? I think previously we were doing that in the provider and using provide only in the public facing classes

cse/src/main/java/com/adyen/checkout/cse/internal/JSONWebEncryptor.kt
```diff
@@ -0,0 +1,209 @@
...
+    fun encrypt(publicKey: String, payload: String): String {
+        val pubKey = generatePublicKey(publicKey)
+        val cek = generateCEK()
+        val encryptedKey = Base64String(encryptCEK(pubKey, cek))
+        val jweObject = encrypt(payload, cek, encryptedKey)
+        return serialize(jweObject)
+    }
+
+    private fun generatePublicKey(publicKey: String): PublicKey {
+        val keyComponents = publicKey.split("|")
```
jreij, line 52: We were using `ValidationUtils.isPublicKeyValid` in the previous implementation, could we use it again?

cse/src/main/java/com/adyen/checkout/cse/CardEncryptor.kt
OscarSpruit: Yes, back then I hoped to merge it before v4. Let's revert it

cse/src/main/java/com/adyen/checkout/cse/internal/CardEncryptorFactory.kt
OscarSpruit: The intention here was to make `DateGenerator` and `JSONWebEncryptor` `internal` and as a side effect it cleans up the component providers a bit

jreij (approved):

PR #1373, opened 2023-10-26
Fix wrong restricted API warnings

jreij:
## Description
Remove `@RestrictTo` annotations from interfaces containing public functions.

## Checklist 
- [x] Link to related issues

COAND-803

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/ComponentAvailableCallback.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/ComponentAvailableCallback.kt
@@ -7,9 +7,6 @@
  */
 package com.adyen.checkout.components.core
 
-import androidx.annotation.RestrictTo
-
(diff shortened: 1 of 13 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/ComponentAvailableCallback.kt
```diff
@@ -7,9 +7,6 @@
  */
 package com.adyen.checkout.components.core
 
-import androidx.annotation.RestrictTo
-
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 interface ComponentAvailableCallback {
```
OscarSpruit, line 10: If this is no longer internal we should probably also make `WeChatPayProvider` public and `PaymentMethodAvailabilityCheck` as well

OscarSpruit (approved):

OscarSpruit (approved):

PR #1376, opened 2023-11-01
Override payment method names - customisable approach

araratthehero:
## Description
This change allows merchants to add custom display information to payment methods, which will be shown instead of the values received in the payment methods api response.

- This PR adds the necessary customization option for merchants without introducing a breaking change. It will also be easy to remove it in the future if necessary.

- `IgnoredCustomizedField` annotation has been introduced, to show a warning when using fields which are marked with the `IgnoredCustomizedField` annotation. The annotation hints that there is a customized field which should be used instead.

- Additionally now we show the card name (Visa, Mastercard...), instead of `Credit Card` for stored payment methods.

Here is an example on how merchants can add `PaymentMethodCustomDisplayInformation` to `PaymentMethod` or `StoredPaymentMethod` before starting drop-in:

```
val paymentMethodsApiResponse = paymentMethodsApiResponse.apply {
    addPaymentMethodCustomDisplayInformation(
        type = "scheme",
        PaymentMethodCustomDisplayInformation("CUSTOM CARD NAME"),
    ) {
        // Optional predicate
        it.name == "Credit Card" && it.brand == "Some brand"
    }
    ...
}

DropIn.startPayment(
    context,
    dropInLauncher,
    paymentMethodsApiResponse,
    dropInConfiguration,
    DropInService::class.java,
)
```

Similarly when using sessions:
```
val checkoutSession = checkoutSession.apply {
    addPaymentMethodCustomDisplayInformation(
        type = "scheme",
        PaymentMethodCustomDisplayInformation("CUSTOM CARD NAME"),
    ) {
        // Optional predicate
        it.name == "Credit Card" && it.brand == "Some brand"
    }
    ...
}

DropIn.startPayment(
    context,
    sessionDropInLauncher,
    checkoutSession,
    dropInConfiguration
)
```

## Checklist 
- [x] Link to related issues

COAND-804

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
@@ -17,13 +17,12 @@ import com.adyen.checkout.core.internal.data.model.ModelUtils.deserializeOptList
 import com.adyen.checkout.core.internal.data.model.ModelUtils.serializeOpt
 import com.adyen.checkout.core.internal.data.model.ModelUtils.serializeOptList
 import com.adyen.checkout.core.internal.data.model.getStringOrNull
-import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
 
-@OptIn(IgnoredCustomizedField::class)
 @Parcelize
+@OptIn(IgnoredCustomizedField::class)
 data class PaymentMethod(
     val type: String? = null,
     @IgnoredCustomizedField
@@ -34,14 +33,11 @@ data class PaymentMethod(
     val issuers: List<Issuer>? = null,
     val configuration: Configuration? = null,
     val details: List<InputDetail>? = null,
-    // Gives the option to set a customizable fields for the payment method
+    // This property is used to allow setting a customizable fields for the payment method
     var customDisplayInformation: PaymentMethodCustomDisplayInformation? = null
 ) : ModelObject() {
 
     fun getMerchantCustomizableName() = customDisplayInformation?.name ?: name
-
-    @IgnoredOnParcel
-    var test: String? = null
 
     companion object {
         private const val TYPE = "type"
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
@@ -7,6 +7,7 @@
  */
 package com.adyen.checkout.components.core
 
+import com.adyen.checkout.components.core.internal.util.IgnoredCustomizedField
 import com.adyen.checkout.core.exception.ModelSerializationException
 import com.adyen.checkout.core.internal.data.model.JsonUtils.parseOptStringList
 import com.adyen.checkout.core.internal.data.model.JsonUtils.serializeOptStringList
@@ -16,21 +17,31 @@ import com.adyen.checkout.core.internal.data.model.ModelUtils.deserializeOptList
 import com.adyen.checkout.core.internal.data.model.ModelUtils.serializeOpt
 import com.adyen.checkout.core.internal.data.model.ModelUtils.serializeOptList
 import com.adyen.checkout.core.internal.data.model.getStringOrNull
+import kotlinx.parcelize.IgnoredOnParcel
 import kotlinx.parcelize.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
 
+@OptIn(IgnoredCustomizedField::class)
 @Parcelize
 data class PaymentMethod(
-    var type: String? = null,
-    var name: String? = null,
-    var brands: List<String>? = null,
-    var brand: String? = null,
-    var fundingSource: String? = null,
-    var issuers: List<Issuer>? = null,
-    var configuration: Configuration? = null,
-    var details: List<InputDetail>? = null,
+    val type: String? = null,
+    @IgnoredCustomizedField
+    val name: String? = null,
+    val brands: List<String>? = null,
+    val brand: String? = null,
+    val fundingSource: String? = null,
+    val issuers: List<Issuer>? = null,
+    val configuration: Configuration? = null,
+    val details: List<InputDetail>? = null,
+    // Gives the option to set a customizable fields for the payment method
+    var customDisplayInformation: PaymentMethodCustomDisplayInformation? = null
 ) : ModelObject() {
+
+    fun getMerchantCustomizableName() = customDisplayInformation?.name ?: name
+
+    @IgnoredOnParcel
+    var test: String? = null
 
     companion object {
         private const val TYPE = "type"
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
@@ -21,8 +21,8 @@ import kotlinx.parcelize.Parcelize
 import org.json.JSONException
 import org.json.JSONObject
 
-@Parcelize
 @OptIn(IgnoredCustomizedField::class)
+@Parcelize
 data class PaymentMethod(
     val type: String? = null,
     @IgnoredCustomizedField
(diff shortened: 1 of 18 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
```diff
@@ -35,9 +34,11 @@ internal class CardComponentDialogFragment : BaseComponentDialogFragment() {
...
         Logger.d(TAG, "onViewCreated")
 
-        // try to get the name from the payment methods response
-        binding.header.text = dropInViewModel.getPaymentMethods()
-            .find { it.type == PaymentMethodTypes.SCHEME }?.name
+        binding.header.text = if (isStoredPayment) {
+            storedPaymentMethod.merchantCustomizableName
+        } else {
+            paymentMethod.merchantCustomizableName
+        }
```
araratthehero, line 41: This part is aligned with design team. Figma will be updated accordingly.
https://youtrack.is.adyen.com/issue/COAND-804/Allow-merchants-to-override-the-payment-method-names-that-show-up-in-Drop-in#focus=Comments-102-905829.0-0

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/BaseComponentDialogFragment.kt
```diff
@@ -43,7 +43,7 @@ internal abstract class BaseComponentDialogFragment :
     var paymentMethod: PaymentMethod = PaymentMethod()
     var storedPaymentMethod: StoredPaymentMethod = StoredPaymentMethod()
     lateinit var component: PaymentComponent
-    private var isStoredPayment = false
+    var isStoredPayment = false
```
OscarSpruit, line 46: Maybe make it `protected`, so only this class and it's children can access it

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
OscarSpruit: So it is okay that the specific brand name is shown for stored cards instead of the the generic "Credit card"?

components-core/src/main/java/com/adyen/checkout/components/core/extension/PaymentMethodsApiResponseExtension.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Calling this function multiple times will cause the custom display information to be overridden.
+ * This might be useful when [customDisplayInformation] should be updated when localization settings change.
+ *
+ * @param type Updates payment methods matching the given type.
+ * @param customDisplayInformation Customizable information object to override the default display values.
+ * @param predicate Updates payment methods not only matching the type but also matching the given predicate.
+ */
+fun PaymentMethodsApiResponse.addPaymentMethodCustomDisplayInformation(
+    type: String,
```
OscarSpruit, line 27: Wouldn't it be easier if the merchant just passes the `PaymentMethod`? It would be less flexible, but also less error prone 🤔

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
```diff
@@ -20,18 +21,25 @@ import kotlinx.parcelize.Parcelize
...
 data class PaymentMethod(
-    var type: String? = null,
-    var name: String? = null,
-    var brands: List<String>? = null,
-    var brand: String? = null,
-    var fundingSource: String? = null,
-    var issuers: List<Issuer>? = null,
-    var configuration: Configuration? = null,
-    var details: List<InputDetail>? = null,
+    val type: String? = null,
```
OscarSpruit, line 27: Changing `var` to `val` is sort of a breaking change 🤔 I'm all for it, but I'm not sure if we should do it now

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
```diff
@@ -20,18 +21,25 @@ import kotlinx.parcelize.Parcelize
...
+    @IgnoredCustomizedField
+    val name: String? = null,
+    val brands: List<String>? = null,
+    val brand: String? = null,
+    val fundingSource: String? = null,
+    val issuers: List<Issuer>? = null,
+    val configuration: Configuration? = null,
+    val details: List<InputDetail>? = null,
+    // This property is used to allow setting a customizable fields for the payment method
+    var customDisplayInformation: PaymentMethodCustomDisplayInformation? = null
```
OscarSpruit, line 37: What about we just use the `name` field for all of this? We could override the name in `addPaymentMethodCustomDisplayInformation` and still have the flexibility of `PaymentMethodCustomDisplayInformation`, but avoid adding any logic to this class

components-core/src/main/java/com/adyen/checkout/components/core/extension/PaymentMethodsApiResponseExtension.kt
```diff
@@ -0,0 +1,50 @@
...
+    customDisplayInformation: PaymentMethodCustomDisplayInformation,
+    predicate: (PaymentMethod) -> Boolean = { true }
```
OscarSpruit, line 29: Nice! I like this idea

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
araratthehero: Yes exactly. I discussed it with Leo and Arjen. They think it makes sense to show the stored card name and that it is more descriptive.
They also want to include the card name in the drop in list for stored payment methods, but that will be part of bigger design changes.

I also asked Arjen to include the stored card payment component in Figma, since it was not mapped yet. He will also look into making the stored card component UI more accessible (color wise).

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
araratthehero: I see, perhaps merchants manipulate these fields to get custom behavior in their implementations?

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
araratthehero: Why I chose to go with this approach is because `name` field is what we get from backend and when serializing, we use that field.
I think it is best to keep `name` field unchanged, for some custom use-cases, perhaps merchants want to give a custom display information by name (through creating a custom predicate). Or they have some other logic around the `name` field?

components-core/src/main/java/com/adyen/checkout/components/core/extension/PaymentMethodsApiResponseExtension.kt
araratthehero: It is possible indeed, but in that case they need to filter the payment methods themselves, which might add some extra code when using the `addPaymentMethodCustomDisplayInformation` method. In that case we might also need to support passing payment methods list to the `addPaymentMethodCustomDisplayInformation` method, if their filter result is a list.
Perhaps what we could do is modify the `PaymentMethodTypes` object, by making it an `enum` (by also allowing custom types) and asking merchants to pass an enum value instead of a String for `type`?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
OscarSpruit: Perfect 👍

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
OscarSpruit: Yes indeed, let's discuss it with the rest of the team

components-core/src/main/java/com/adyen/checkout/components/core/extension/PaymentMethodsApiResponseExtension.kt
OscarSpruit: Ah yes didn't take about the filtering yet. An enum would probably be best indeed as it covers both sides. Let's discuss if and how we could do that

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethod.kt
OscarSpruit: I see and that makes sense, but the `name` field is only ment for display purposes. `name` is localised in the backend, so it would be difficult to built logic around it (`type` should be used for that). And `name` is never send back to the backend, so it shouldn't impact anything. Let's maybe discuss with the team as well

PR #1377, opened 2023-11-01
Fix frozen screen when dismissing gift card fragment

OscarSpruit:
## Description
When a user dismisses the gift card fragment drop-in would not be terminated. This blocks the user with an invisible activity and makes the merchant app seem unresponsive. By simply overriding onCancel this was fixed.

## Checklist 

COAND-817

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInBottomSheetDialogFragment.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInBottomSheetDialogFragment.kt
@@ -10,6 +10,7 @@ package com.adyen.checkout.dropin.internal.ui
 
 import android.app.Dialog
 import android.content.Context
+import android.content.DialogInterface
 import android.os.Bundle
 import android.view.KeyEvent
 import android.widget.FrameLayout
@@ -79,6 +80,11 @@ internal abstract class DropInBottomSheetDialogFragment : BottomSheetDialogFragm
 
     open fun onBackPressed(): Boolean {
         return false
+    }
+
+    override fun onCancel(dialog: DialogInterface) {
+        Logger.d(TAG, "onCancel")
+        protocol.terminateDropIn()
     }
 
     companion object {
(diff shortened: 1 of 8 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/GiftCardComponentDialogFragment.kt
```diff
@@ -129,6 +130,12 @@ internal class GiftCardComponentDialogFragment : DropInBottomSheetDialogFragment
         protocol.showError(null, getString(R.string.component_error), componentError.errorMessage, true)
     }
 
+    override fun onCancel(dialog: DialogInterface) {
```
araratthehero, line 133: I can see that every subclass of `DropInBottomSheetDialogFragment` has this method overridden. It looks like all bottom sheet dialog fragments should have this behavior. What do you think of moving this function to the `DropInBottomSheetDialogFragment` and removing them from all of its subclasses?

`ActionComponentDialogFragment` has a different implementation though. This one could stay.

araratthehero (commented):
Looks good!

I just have one suggestion.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/GiftCardComponentDialogFragment.kt
OscarSpruit: Good point, let me look into it

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInBottomSheetDialogFragment.kt
```diff
@@ -81,6 +82,11 @@ internal abstract class DropInBottomSheetDialogFragment : BottomSheetDialogFragm
         return false
     }
 
+    override fun onCancel(dialog: DialogInterface) {
+        Logger.d(TAG, "onCancel")
```
araratthehero, line 86: What do you think of keeping `super.onCancel(dialog)` call here, in case the logic of `onCancel()` changes in `DialogFragment`?
Though I am not sure if it would add any value.

araratthehero (approved):
Looks good to me!

Just wanted to know what you think of keeping the `super.onCancel()` call.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInBottomSheetDialogFragment.kt
OscarSpruit: If they would change the super implementation it would most likely only have unwanted side effects, so I think it's better to leave it out

PR #1378, opened 2023-11-07
Make it possible to override version and platform for analytics

OscarSpruit:
## Description
Make it possible to override version and platform for analytics

## Checklist 
- [x] Code is unit tested

COAND-815

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
@@ -30,9 +30,9 @@ class AnalyticsMapper {
         sessionId: String?,
     ): AnalyticsSetupRequest {
         return AnalyticsSetupRequest(
-            version = BuildConfig.CHECKOUT_VERSION,
+            version = VERSION,
             channel = ANDROID_CHANNEL,
-            platform = ANDROID_PLATFORM,
+            platform = PLATFORM,
             locale = locale.toString(),
             component = getComponentQueryParameter(source),
             flavor = getFlavorQueryParameter(source),
@@ -75,5 +75,23 @@ class AnalyticsMapper {
         private const val DROP_IN_COMPONENT = "dropin"
         private const val ANDROID_PLATFORM = "android"
         private const val ANDROID_CHANNEL = "android"
+
+        private var PLATFORM = ANDROID_PLATFORM
+        private var VERSION = BuildConfig.CHECKOUT_VERSION
+
+        @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+        fun overrideForCrossPlatform(
+            platform: String,
+            version: String,
+        ) {
+            PLATFORM = platform
+            VERSION = version
+        }
+
+        @VisibleForTesting
+        internal fun resetToDefaults() {
+            PLATFORM = ANDROID_PLATFORM
+            VERSION = BuildConfig.CHECKOUT_VERSION
+        }
     }
 }
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
@@ -73,24 +73,23 @@ class AnalyticsMapper {
 
     companion object {
         private const val DROP_IN_COMPONENT = "dropin"
-        private const val ANDROID_PLATFORM = "android"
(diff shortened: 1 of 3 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
```diff
@@ -75,5 +75,24 @@ class AnalyticsMapper {
         private const val DROP_IN_COMPONENT = "dropin"
         private const val ANDROID_PLATFORM = "android"
         private const val ANDROID_CHANNEL = "android"
+
+        private var VERSION = BuildConfig.CHECKOUT_VERSION
+        private var PLATFORM = ANDROID_PLATFORM
+
+        @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+        fun overrideVersion(version: String) {
```
jreij, line 83: Sneaky 😅 I like this solution more than mine

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
```diff
@@ -75,5 +75,24 @@ class AnalyticsMapper {
...
+            VERSION = version
+        }
+
+        @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+        fun overridePlatform(platform: String) {
```
jreij, line 88: I suggest making the argument an enum, with only 2 options: `REACT_NATIVE` and `FLUTTER`, wdyt?

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
```diff
@@ -75,5 +75,24 @@ class AnalyticsMapper {
...
+            PLATFORM = platform
+        }
+
+        @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+        fun resetToDefaults() {
```
jreij, line 93: Could this be needed anywhere other than for testing?

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
OscarSpruit: Yeah maybe that's better

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
OscarSpruit: Can't think of a use case, but I thought it could be nice to have. Shall I make it `internal`, so only the tests can use it?

jreij: Yeah let's do this for now, we'll make it public if we need it later

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
jreij: Not sure if this is better, but what if we merge these 2 methods and change the name to something like `overrideForCrossPlatform`, to make it even more obvious that this method shouldn't be used by anyone, and since our flutter/RN libs will always call both methods at the same time 🤔

OscarSpruit: Good one, I like that idea

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsPlatform.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package com.adyen.checkout.components.core.internal.data.api
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+enum class AnalyticsPlatform(val value: String) {
+    ANDROID("android"),
+    FLUTTER("flutter"),
+    REACT_NATIVE("react-native"),
```
jreij, line 17: Did we confirm these values with backend/data?

OscarSpruit: Yes ✅

jreij (dismissed):

PR #1380, opened 2023-11-09
Override payment method names

araratthehero:
## Description
This change allows merchants to add custom information to payment methods, which will be shown instead of the values received in the payment methods api response.

- For now we decided to add this functionality only for payment methods. If there is demand, we can support the same functionality for stored payment methods.
- Additionally now we show the card name (`Visa`, `Mastercard`...), instead of `Credit Card` for stored payment methods.

Here is an example on how merchants can override payment method name when building `DropInConfiguration`:

```
DropInConfiguration.Builder(
    shopperLocale,
    environment,
    clientKey,
)
    .addCardConfiguration(getCardConfiguration())
    .addCashAppPayConfiguration(getCashAppPayConfiguration())
    ...
    .overridePaymentMethodName("scheme", "Custom card name")
    .build()
```

## Checklist 
- [x] Code is unit tested
- [x] Link to related issues

COAND-804


drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -376,6 +379,11 @@ class DropInConfiguration private constructor(
             return this
         }
 
+        fun overridePaymentMethodName(paymentMethodType: String, name: String): Builder {
```
araratthehero, line 382: This makes it easy for merchants to override payment method name. With design changes, we might introduce a description field for payment methods.
I decided to keep it as simple as possible for now, but later we can also offer a method, which will require `DropInPaymentMethodInformation`, instead of `name` string.

jreij: Good idea 👍 Let's make sure we document the method and mention that it can be called multiple times, for different payment methods

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInPaymentMethodInformation.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package com.adyen.checkout.dropin.internal.ui.model
+
+import android.os.Parcelable
+import com.adyen.checkout.components.core.PaymentMethod
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+class DropInPaymentMethodInformation(
```
jreij, line 16: Can this class be internal?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModelFactory.kt
```diff
@@ -63,4 +68,17 @@ internal class DropInViewModelFactory(
         @Suppress("UNCHECKED_CAST")
         return DropInViewModel(bundleHandler, orderStatusRepository, analyticsRepository) as T
     }
+
+    internal fun PaymentMethodsApiResponse.overridePaymentMethodInformation(
```
jreij, line 72: This method probably doesn't need to be `internal` if it's inside `DropInViewModelFactory`

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInPaymentMethodInformation.kt
araratthehero: Yes, it will be fixed with the next commit

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModelFactory.kt
araratthehero: This will also be fixed with the next commit. I moved it out from the `DropInViewModelFactory`.

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
jreij: A few things on the documentation, all minor comments/nitpicking:
- We generally refer to it as "Drop-in" instead of "drop in" in other code docs as well as the online docs.
- The wording of "filtering" might be confusing since we're not filtering the list itself, but not sure about this one.
- The "Only the latest value will be shown..." line can be removed since it describes a scenario caused by a wrong integration from merchants, it's good that we handle it internally but there's no need to document it if they shouldn't use it this way.
- With most docs you already explain all the logic at the start, so personally I prefer to keep the `@param`s section very short and only describe what the parameter is. For example, paymentMethodType: the type of the `PaymentMethod` and name: the name to be displayed.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModelFactory.kt
```diff
@@ -35,7 +37,9 @@ internal class DropInViewModelFactory(
     override fun <T : ViewModel> create(key: String, modelClass: Class<T>, handle: SavedStateHandle): T {
         val bundleHandler = DropInSavedStateHandleContainer(handle)
 
-        val dropInConfiguration: DropInConfiguration = requireNotNull(bundleHandler.dropInConfiguration)
+        val dropInConfiguration: DropInConfiguration = requireNotNull(bundleHandler.dropInConfiguration).apply {
+            bundleHandler.overridePaymentMethodInformation(overriddenPaymentMethodInformation)
```
jreij, line 41: Also nitpicking: I like to use `apply` to modify or execute something in the context object itself (`dropInConfiguration` in this case). While here we're using it to read `dropInConfiguration.overriddenPaymentMethodInformation` and pass it on to modify the `paymentMethodsApiResponse` inside the `bundleHandler`. So first thing I assumed while reading this is that `bundleHandler` is a property inside `dropInConfiguration` and we're modifying it.

If you agree then I would suggest using `bundleHandler.dropInConfiguration?.let` or even just explicitly calling `dropInConfiguration.overriddenPaymentMethodInformation`, it does add more code but it also adds more clarity. But feel free to disagree, this is really just a personal preference that I thought I'd share.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInPaymentMethodInformation.kt
```diff
@@ -0,0 +1,22 @@
...
+import android.os.Parcelable
+import com.adyen.checkout.components.core.PaymentMethod
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+internal data class DropInPaymentMethodInformation(
+    val name: String
+) : Parcelable
+
+internal fun PaymentMethod.overrideInformation(information: DropInPaymentMethodInformation) {
```
jreij, line 20: Minor: do you think that moving this closer to `overridePaymentMethodInformation` makes more sense? I also see the benefit of keeping it here since you have visibility over what's inside `DropInPaymentMethodInformation`, so not quite sure about this one.

drop-in/src/test/java/com/adyen/checkout/dropin/internal/ui/DropInViewModelFactoryTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.adyen.checkout.components.core.PaymentMethod
+import com.adyen.checkout.components.core.PaymentMethodsApiResponse
+import com.adyen.checkout.components.core.StoredPaymentMethod
+import com.adyen.checkout.dropin.internal.ui.model.DropInPaymentMethodInformation
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Test
+import org.mockito.kotlin.mock
+
+class DropInViewModelFactoryTest {
```
jreij, line 20: Class could be internal

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
```diff
@@ -35,9 +34,11 @@ internal class CardComponentDialogFragment : BaseComponentDialogFragment() {
         super.onViewCreated(view, savedInstanceState)
         Logger.d(TAG, "onViewCreated")
 
-        // try to get the name from the payment methods response
-        binding.header.text = dropInViewModel.getPaymentMethods()
-            .find { it.type == PaymentMethodTypes.SCHEME }?.name
+        binding.header.text = if (isStoredPayment) {
```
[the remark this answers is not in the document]
jreij: I just saw in your commit message and PR description that you did this change on purpose, sorry for my clumsiness 😅 So if you already double checked with product/web/design feel free to ignore my comment

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModelFactory.kt
araratthehero: I agree that it might be confusing to why `.apply` was used here. I will modify it.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInPaymentMethodInformation.kt
araratthehero: That was exactly how I thought about it, to keep the `overrideInformation` next to `DropInPaymentMethodInformation` for visibility and it would make it easy to modify them if a new field is added.

jreij (approved):
🚀

PR #1389, opened 2023-11-14
Display amounts for installments

araratthehero:
## Description
As part of this PR, we allow merchants to show amount next to installments for card component.

## Checklist 
- [x] Code is unit tested
- [x] Link to related issues

COAND-802

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
@@ -16,7 +16,6 @@ import android.widget.BaseAdapter
 import android.widget.Filter
 import android.widget.Filterable
 import androidx.annotation.RestrictTo
-import androidx.annotation.StringRes
 import androidx.recyclerview.widget.RecyclerView
 import com.adyen.checkout.card.databinding.InstallmentViewBinding
 import com.adyen.checkout.card.internal.ui.model.InstallmentOption
@@ -70,7 +69,6 @@ internal class InstallmentListAdapter(
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class InstallmentModel(
-    @StringRes val textResId: Int,
     val numberOfInstallments: Int?,
     val option: InstallmentOption,
     val amount: Amount?,
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
@@ -71,10 +71,10 @@ internal class InstallmentListAdapter(
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class InstallmentModel(
     @StringRes val textResId: Int,
-    val monthValue: Int?,
+    val numberOfInstallments: Int?,
     val option: InstallmentOption,
     val amount: Amount?,
-    val shopperLocale: Locale?,
+    val shopperLocale: Locale,
     val showAmount: Boolean
 )
 
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
@@ -21,6 +21,8 @@ import androidx.recyclerview.widget.RecyclerView
 import com.adyen.checkout.card.databinding.InstallmentViewBinding
 import com.adyen.checkout.card.internal.ui.model.InstallmentOption
 import com.adyen.checkout.card.internal.util.InstallmentUtils
+import com.adyen.checkout.components.core.Amount
+import java.util.Locale
 
 // We need context to inflate the views and localizedContext to fetch the strings
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
@@ -69,8 +71,11 @@ internal class InstallmentListAdapter(
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class InstallmentModel(
     @StringRes val textResId: Int,
-    val value: Int?,
-    val option: InstallmentOption
+    val monthValue: Int?,
+    val option: InstallmentOption,
+    val amount: Amount?,
+    val shopperLocale: Locale?,
+    val showAmount: Boolean
 )
 
 internal class InstallmentFilter(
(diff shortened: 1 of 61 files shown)


card/src/main/res/values/strings.xml
```diff
@@ -32,6 +32,7 @@
 
     <string name="checkout_card_installments_title">Number of installments</string>
     <string name="checkout_card_installments_option_regular">%s months</string>
+    <string name="checkout_card_installments_option_regular_with_price">%1$sx %2$s</string>
```
jreij, line 35: We should move this to the bottom of the file and add `translatable="false"` to it, then do the exact same change in `strings.xml.tt`. This is necessary for our translations process.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentListAdapter.kt
```diff
@@ -66,8 +68,11 @@ internal class InstallmentListAdapter(
 
 internal data class InstallmentModel(
     @StringRes val textResId: Int,
-    val value: Int?,
-    val option: InstallmentOption
+    val monthValue: Int?,
```
jreij, line 71: Good idea renaming this, makes it clearer 👍 However I still had to look it up to understand exactly what it means. The [documentation in the API explorer](https://docs.adyen.com/api-explorer/Checkout/71/post/payments#request-installments-value) says this:
> Defines the number of installments.

Maybe it can help us pick a better name?

card/src/main/res/values/strings.xml
araratthehero: I left it here since I was not sure if in some locales they use the multiplies differently. E.g. `%2$s x%1$s`.
I will check how other teams handle this. If this does not make sense to localize, I will add `translatable="false"` and move it to the bottom of the file.

Thanks for pointing this out!

card/src/main/res/values-cs-rCZ/strings.xml
```diff
@@ -32,6 +32,7 @@
 
     <string name="checkout_card_installments_title">Počet splátek</string>
     <string name="checkout_card_installments_option_regular">%s měsíců</string>
+    <string name="checkout_card_installments_option_regular_with_price">%s× %s</string>
```
OscarSpruit, line 35: This looks like the Arabic `x`, should we verify if this is formatted correctly?

components-core/src/main/java/com/adyen/checkout/components/core/internal/util/NumberExtension.kt
```diff
@@ -0,0 +1,22 @@
...
+import java.util.Locale
+
+/**
+ * Format the [Int] to be displayed to the user based on the Locale.
+ *
+ * @param locale The locale the number will be formatted with.
+ * @return A formatted string displaying value.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun Int.format(locale: Locale): String = NumberFormat.getInstance(locale).format(this)
```
jreij, line 22: Can we rename this to something more descriptive, to indicate that it formats the integer to a localized string maybe?

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/CurrencyUtilsTest.kt
```diff
@@ -0,0 +1,67 @@
...
+package com.adyen.checkout.components.core.internal.util
+
+import com.adyen.checkout.components.core.Amount
+import com.adyen.checkout.core.exception.CheckoutException
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertThrows
+import org.junit.Test
+import java.util.Locale
+
+internal class CurrencyUtilsTest {
```
jreij, line 18: 👏

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/CurrencyUtilsTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        val formattedAmount = CurrencyUtils.formatAmount(amount, locale)
+
+        assertEquals("$100.50", formattedAmount)
+    }
+
+    @Test
+    fun `assert currency does nothing, if currency code is supported`() {
+        val currencyCode = "EUR"
+
+        val formattedAmount = CurrencyUtils.assertCurrency(currencyCode)
```
jreij, line 54: If you want to make sure this doesn't throw anything, maybe check out `assertDoesNotThrow`? And then the name could be `does not throw an exception` instead of `does nothing`?

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/CurrencyUtilsTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        val locale = Locale.forLanguageTag("ar-LB")
+
+        val formattedAmount = CurrencyUtils.formatAmount(amount, locale)
+
+        assertEquals("ل.ل.\u200F ١٠٠٫٥٠", formattedAmount)
+    }
+
+    @Test
+    fun `format amount with en-US locale`() {
+        val amount = Amount("USD", 10050L)
```
jreij, line 42: Minor one but maybe you can use a different number in each test?

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/NumberExtensionTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Test
+import java.util.Locale
+
+internal class NumberExtensionTest {
+
+    @Test
+    fun `integer format formats with en-US locale`() {
+        val locale = Locale.forLanguageTag("en-US")
+
+        assertEquals("1", 1.format(locale))
```
jreij, line 21: This is unnecessary in this case specifically but check out `@ParameterizedTest` if you didn't know about it. It might be nice for these types of tests and we have a few examples of it in the codebase.

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/NumberExtensionTest.kt
```diff
@@ -0,0 +1,36 @@
...
+        assertEquals("5", 5.format(locale))
+        assertEquals("10", 10.format(locale))
+        assertEquals("15", 15.format(locale))
+    }
+
+    @Test
+    fun `integer format formats with ar-LB locale`() {
```
jreij, line 28: I see you're getting on my good side already 🇱🇧 😆

card/src/main/res/values-cs-rCZ/strings.xml
araratthehero: I checked and it is formatted correctly

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/NumberExtensionTest.kt
jreij: Let's keep it like this then 👍

jreij (approved):

PR #1391, opened 2023-11-15
Rename telemetry to analytics

jreij:
## Description
Rename telemetry (only used in the example app) to analytics to preserve consistency with the SDK and because analytics is a broader term that fits our use case more than telemetry.

## Checklist 
- [ ] Code is unit tested


araratthehero (approved):
Looks good to me! 📊

PR #1393, opened 2023-11-16
Update ktlint to 1.0.1

jreij:
## Description
Update ktlint to version 1.0.1

## Checklist 


action-core/src/main/java/com/adyen/checkout/action/core/internal/ActionHandlingPaymentMethodConfigurationBuilder.kt
```diff
@@ -22,7 +22,11 @@ import com.adyen.checkout.voucher.VoucherConfiguration
 import com.adyen.checkout.wechatpay.WeChatPayActionConfiguration
 import java.util.Locale
 
-@Suppress("UNCHECKED_CAST")
+@Suppress(
+    "UNCHECKED_CAST",
+    "ktlint:standard:discouraged-comment-location",
+    "ktlint:standard:type-parameter-list-spacing",
```
araratthehero, line 28: What do you think of modifying the constructor declaration and moving the comment outside of the class? E.g.
```
/**
 * Initialize a configuration builder with the required fields.
 *
 * @param shopperLocale The [Locale] of the shopper.
 * @param environment The [Environment] to be used for internal network calls from the SDK to Adyen.
 * @param clientKey Your Client Key used for internal network calls from the SDK to Adyen.
 */
abstract class ActionHandlingPaymentMethodConfigurationBuilder<
    ConfigurationT : Configuration,
    BuilderT : BaseConfigurationBuilder<ConfigurationT, BuilderT>
    >(
    shopperLocale: Locale,
    environment: Environment,
    clientKey: String
) : BaseConfigurationBuilder<ConfigurationT, BuilderT>(shopperLocale, environment, clientKey),
    ActionHandlingConfigurationBuilder<BuilderT> {
...
```
Though I am not sure which one is better, since the way I sent above, the code might be less readable, but that way we can remove both of the Suppressed warnings. 😶‍🌫️

Same applies to the rest of the Suppressed `ktlint` warnings in this PR.

araratthehero (dismissed):
Looks good to me! ✨

I just have one minor discussion point, which is not a blocker.

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
```diff
@@ -75,22 +76,22 @@ class AnalyticsMapper {
         private const val DROP_IN_COMPONENT = "dropin"
         private const val ANDROID_CHANNEL = "android"
 
-        private var PLATFORM = AnalyticsPlatform.ANDROID.value
-        private var VERSION = BuildConfig.CHECKOUT_VERSION
+        private var platform = AnalyticsPlatform.ANDROID.value
```
jreij, line 79: @OscarSpruit let me know if you suppress the warning on these variable or maybe rename them to indicate that they are technically constants.

OscarSpruit: > Names of top-level or object properties which hold objects with behavior or mutable data should use camel case names:

Taken from [the official Kotlin conventions guide](https://kotlinlang.org/docs/coding-conventions.html#property-names). So maybe we just accept the lowercase naming? Or we rename them?

action-core/src/main/java/com/adyen/checkout/action/core/internal/ActionHandlingPaymentMethodConfigurationBuilder.kt
OscarSpruit: The main issue with that approach is that the comment appears on the class and not on the constructor. So if you hover over this class it tells you about something that is not directly relevant

components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/AnalyticsMapper.kt
jreij: What if we keep them camel case but name them something more descriptive? Like `actualPlatform` and `actualVersion`? I can't think of a good name :/

OscarSpruit: `actual...` sounds good to me. It makes it a bit more clear imo

OscarSpruit (approved):

PR #1395, opened 2023-11-21
Add component for Twint payment method

OscarSpruit:
## Description
Twint payments can now be handled through the `InstantPaymentComponent`. The payments call will return a Twint action which will then be handled with the new `TwintActionComponent`. It is now also configurable if you want the `InstantPaymentComponent` to handle a payment with a SDK (if there is one, if not will fall back to web flow) or with a web flow.

## Checklist 
- [x] Code is unit tested

COAND-806


boleto/src/main/java/com/adyen/checkout/boleto/internal/ui/DefaultBoletoDelegate.kt
```diff
@@ -235,6 +235,7 @@ internal class DefaultBoletoDelegate(
             paymentMethod = GenericPaymentMethod(
```
OscarSpruit, line 235: Should we make a specific payment method class per pm? I think `GenericPaymentMethod` is also barely used anywhere else

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -71,9 +75,17 @@ class DropInConfiguration private constructor(
 ) : Configuration {
 
     internal fun <T : Configuration> getConfigurationForPaymentMethod(paymentMethod: String): T? {
-        if (availablePaymentConfigs.containsKey(paymentMethod)) {
+        // For instant payment we cannot use the payment method as key as we don't have a definitive list of supported
+        // instant payment methods.
+        val key = if (InstantPaymentComponent.PROVIDER.isPaymentMethodSupported(paymentMethod)) {
```
OscarSpruit, line 80: Let me know if you have a better idea on how to do this

twint-sdk/build.gradle
```diff
@@ -0,0 +1,143 @@
+configurations.maybeCreate("default")
+artifacts.add("default", file('TwintSdk-android-7.0.0.aar'))
```
OscarSpruit, line 2: This looks a bit scary, but this is the only way I could make publishing work.

At first I included the AAR in the twint module, but this causes the AAR to be nested inside of the AAR created when publishing the twint module. Nested (aka fat) AARs are not supported. Instead, we now publish the twint AAR as a separate dependency

wechatpay/src/main/java/com/adyen/checkout/wechatpay/internal/ui/DefaultWeChatDelegate.kt
```diff
@@ -110,8 +110,7 @@ internal class DefaultWeChatDelegate(
 
     @SuppressWarnings("ReturnCount")
     override fun handleAction(action: Action, activity: Activity) {
-        @Suppress("UNCHECKED_CAST")
-        val sdkAction = (action as? SdkAction<WeChatPaySdkData>)
```
OscarSpruit, line 114: Only doing check would crash if you pass `SdkAction<TwintSdkData>`, so I changed it

action-core/build.gradle
```diff
@@ -39,8 +39,9 @@ dependencies {
     api project(':await')
     api project(':qr-code')
     api project(':redirect')
-    compileOnly project(':wechatpay')
+    api project(':twint')
```
jreij, line 42: Should we make this `compileOnly`? It's only going to be used in some markets and it's not even the default Twint flow (looking at the `shouldUseSdk` configuration)

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -47,6 +49,8 @@ import kotlinx.parcelize.Parcelize
 import java.util.Locale
 import kotlin.collections.set
 
+private const val INSTANT_PAYMENT = "INSTANT_PAYMENT"
```
jreij, line 52: Can we move this to the companion?

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
jreij: Yep makes sense

twint/src/main/java/com/adyen/checkout/twint/Twint.kt
```diff
@@ -0,0 +1,52 @@
...
+package com.adyen.checkout.twint
+
+import androidx.activity.ComponentActivity
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import ch.twint.payment.sdk.Twint
+import ch.twint.payment.sdk.TwintPayResult
+import com.adyen.checkout.core.exception.CheckoutException
+
+object Twint {
```
jreij, line 18: Let's rename this to avoid merchants confusing it with `ch.twint.payment.sdk.Twint`?

wechatpay/src/main/java/com/adyen/checkout/wechatpay/internal/ui/DefaultWeChatDelegate.kt
araratthehero: If you pass `SdkAction<TwintSdkData>`, doesn't `(action as? SdkAction<WeChatPaySdkData>)` return `null`?

OscarSpruit: Nope, because of type erasure. `action` would be cast to `SdkAction<*>` and it doesn't care about the generic

action-core/build.gradle
OscarSpruit: Good point 👍 `shouldUseSdk` is true by default though, so it would be the default flow. If `:twint` is not included by the merchant and they forget to set `shouldUseSdk` to false, then the whole flow wouldn't work as it can't fall back to web in this case 🤔

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
OscarSpruit: Yes and no. Both the `Builder` and the actual configuration use this field. If we put it in an `companion object` it has to be `public/internal` meaning other classes could access it. Now it's at least file private

instant/src/main/java/com/adyen/checkout/instant/InstantPaymentConfiguration.kt
```diff
@@ -36,6 +39,8 @@ class InstantPaymentConfiguration private constructor(
      */
     class Builder : ActionHandlingPaymentMethodConfigurationBuilder<InstantPaymentConfiguration, Builder> {
 
+        private var shouldUseSdk: Boolean? = null
```
ozgur00, line 42: Should we have an enum instead of a boolean? I feel that suits better for prioritization and makes it clearer for merchants to understand that it's not enabling or disabling, it's rather a priority choice. Also prevents breaking changes if at some point we have to add another way to resolve these.

boleto/src/main/java/com/adyen/checkout/boleto/internal/ui/DefaultBoletoDelegate.kt
jreij: I agree but let's not change them for existing components if they cause a breaking change

twint-sdk/build.gradle
jreij: Let's discuss this more before merging

action-core/build.gradle
jreij: > shouldUseSdk is true by default though, so it would be the default flow

Good point, I missed that.

> the whole flow wouldn't work as it can't fall back to web in this case

I think it would be a wrong integration from the merchant. So it's better to fail rather than use the SDK and risk them getting unwanted results. Added to the fact that we avoid shipping this `aar` to everyone using components 🙃

twint/src/main/java/com/adyen/checkout/twint/Twint.kt
jreij: No then it makes sense to keep it like this 👍

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -99,6 +103,7 @@ internal class CheckoutConfigurationProvider @Inject constructor(
         InstantPaymentConfiguration.Builder(shopperLocale, environment, clientKey)
             .setAmount(amount)
             .setAnalyticsConfiguration(getAnalyticsConfiguration())
+            .setUseSdk(true)
```
jreij, line 106: That's the default value right?

twint/consumer-rules.pro
jreij, line 1: Do the Twint docs mention anything about manually adding any proguard rules?

twint/src/main/java/com/adyen/checkout/twint/Twint.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    private var twintObject: Twint? = null
+
+    private var onResultListener: ((TwintPayResult) -> Unit)? = null
+
+    fun initialize(activity: ComponentActivity) {
```
jreij, line 24: Let's add some kdocs to this and the class 🙏

action-core/build.gradle
OscarSpruit: Alright, I'll change it to `compileOnly`

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
OscarSpruit: Indeed, but I added it for showcase purposes

twint/src/main/java/com/adyen/checkout/twint/Twint.kt
OscarSpruit: Ahhh yes of course

twint/consumer-rules.pro
OscarSpruit: They don't, so I assume they have the right rules in place. I'll test it to make sure though

instant/src/main/java/com/adyen/checkout/instant/InstantPaymentConfiguration.kt
OscarSpruit: Good suggestion, it could make it more clear indeed

twint/src/main/java/com/adyen/checkout/twint/Twint.kt
OscarSpruit: Added it, let me know if I could improve the docs

twint/src/main/java/com/adyen/checkout/twint/Twint.kt
```diff
@@ -0,0 +1,64 @@
...
+ * with [TwintSdkData] can be handled.
+ */
+object Twint {
+
+    private var twintObject: Twint? = null
+
+    private var onResultListener: ((TwintPayResult) -> Unit)? = null
+
+    /**
+     * Initializes the [Twint] object. This method **must** be called in [ComponentActivity.onCreate] or before, because
```
jreij, line 33: Minor: I understand the reason behind mentioning `ComponentActivity` but I'm wondering if saying "your activity's onCreate" could be clearer, WDYT?

OscarSpruit: Yeah it sounds more clear and it should already be clear that a `ComponentActivity` should be used

PR #1398, opened 2023-11-22
Add pinRequired configuration for gift card component

OscarSpruit:
## Description
Add configuration to hide the PIN input field for gift cards.

## Checklist 
- [x] Code is unit tested

COAND-741

### DIFF
--- a/giftcard/src/main/java/com/adyen/checkout/giftcard/internal/ui/GiftCardDelegate.kt
+++ b/giftcard/src/main/java/com/adyen/checkout/giftcard/internal/ui/GiftCardDelegate.kt
@@ -41,4 +41,6 @@ internal interface GiftCardDelegate :
     fun resolveBalanceResult(balanceResult: BalanceResult)
 
     fun resolveOrderResponse(orderResponse: OrderResponse)
+
+    fun isPinRequired(): Boolean
 }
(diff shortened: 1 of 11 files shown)


giftcard/src/main/java/com/adyen/checkout/giftcard/internal/ui/GiftCardDelegate.kt
```diff
@@ -41,4 +41,6 @@ internal interface GiftCardDelegate :
     fun resolveBalanceResult(balanceResult: BalanceResult)
 
     fun resolveOrderResponse(orderResponse: OrderResponse)
+
+    fun isPinRequired(): Boolean
```
jreij, line 45: Would this be better as part of the `GiftCardOutputData`?

OscarSpruit: Hmm maybe, but then we would potentially do unnecessary initialisation when creating `GiftCardView`. I don't really have a strong preference, do you?

jreij: Not really, up to you 👍

OscarSpruit: Okay then I'll keep it this way

jreij (approved):

PR #1404, opened 2023-11-30
Convert Card Component with sessions example integration to Compose

OscarSpruit:
## Description

https://github.com/Adyen/adyen-android/assets/17701279/03647498-b617-4eca-be52-3467b7ead0f5

## Checklist 

COAND-826


example-app/src/main/java/com/adyen/checkout/example/ui/theme/ExampleTheme.kt
```diff
@@ -0,0 +1,120 @@
+package com.adyen.checkout.example.ui.theme
```
jreij, line 1: Some of these new files are missing copyrights

OscarSpruit: Oh no not again

OscarSpruit: Fixed it!

example-app/src/main/java/com/adyen/checkout/example/ui/card/compose/SessionsCardScreen.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+@Composable
+private fun CardComponent(
+    configuration: CardConfiguration,
+    componentData: SessionsCardComponentData,
+    action: Action?,
+    onActionConsumed: () -> Unit,
+    modifier: Modifier = Modifier,
```
araratthehero, line 117: Nits: Usually modifier is the first argument in Composable functions. I think it would be good to keep it consistent.

example-app/src/main/java/com/adyen/checkout/example/ui/compose/ResultContent.kt
```diff
@@ -0,0 +1,58 @@
...
+        val tint = when (resultState) {
+            ResultState.SUCCESS -> LocalCustomColorScheme.current.success
+            ResultState.PENDING -> LocalCustomColorScheme.current.warning
+            ResultState.FAILURE -> MaterialTheme.colorScheme.error
+        }
+        Icon(
+            painter = painterResource(id = resultState.drawable),
+            contentDescription = null,
+            tint = tint,
+            modifier = modifier.size(100.dp),
```
araratthehero, line 47: Here it is best to use `Modifier`, instead of `modifier`. Otherwise it will use the modifier passed to the `ResultContent`, which might introduce buggy behavior in some cases.

example-app/src/main/java/com/adyen/checkout/example/ui/card/compose/SessionsCardScreen.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        if (isLoading) {
+            CircularProgressIndicator()
+        }
+
+        if (toastMessage != null) {
+            val context = LocalContext.current
+            LaunchedEffect(toastMessage) {
+                Toast.makeText(context, toastMessage, Toast.LENGTH_SHORT).show()
+                viewModel.toastMessageConsumed()
```
araratthehero, line 93: I would suggest creating a function argument, instead of directly calling a method on viewMode here.
E.g. onTastMessageConsumed: () -> Unit as an argument for MainContent and in SessionsCardScreen I would just pass viewModel:: toastMessageConsumed to the MainContent.

araratthehero: Same could work for viewModel::actionConsumed

example-app/src/main/java/com/adyen/checkout/example/ui/compose/ResultContent.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.ui.unit.dp
+import com.adyen.checkout.example.ui.theme.LocalCustomColorScheme
+
+@Composable
+internal fun ResultContent(
+    resultState: ResultState,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier.padding(16.dp),
```
araratthehero, line 34: This does not need to be done now, but it would be nice to have a `Dimen` class, similar to the `Color` class, where we could define all our constant dimen values.

araratthehero (commented):
Great work! 🚀

I have a couple of comment, most of which are minor suggestions.

example-app/src/main/java/com/adyen/checkout/example/ui/card/SessionsCardUiState.kt
```diff
@@ -0,0 +1,24 @@
...
+import androidx.compose.runtime.Immutable
+import com.adyen.checkout.card.CardConfiguration
+import com.adyen.checkout.components.core.action.Action
+import com.adyen.checkout.example.ui.compose.ResultState
+
+@Immutable
+internal data class SessionsCardUiState(
+    val cardConfiguration: CardConfiguration,
+    val isLoading: Boolean = false,
+    val toastMessage: String? = null,
```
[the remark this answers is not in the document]
OscarSpruit: `toastMessage` is maybe not the right name, because `toast` is an implementation detail. I want to show a one time message here, so it's not really a state. It's actually an event. Also it's not only for errors, but for any sort of message event.

Maybe just renaming the field and using a type alias for String would make this better?

araratthehero (approved):
Looks good to me! 🚀

PR #1418, opened 2023-12-14
Add Google Pay + sessions flow + compose to example app

jreij:
## Description
Add Google Pay to the example app, including the availability check and the Google Pay compose dynamic button.

## Checklist 

COAND-788


example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayActivity.kt
```diff
@@ -0,0 +1,82 @@
...
+        // Insert return url in extras, so we can access it in the ViewModel through SavedStateHandle
+        val returnUrl = RedirectComponent.getReturnUrl(applicationContext) + "/sessions/googlepay"
+        intent = (intent ?: Intent()).putExtra(RETURN_URL_EXTRA, returnUrl)
+
+        setContent {
+            val useDarkTheme = when (nightThemeRepository.theme) {
+                NightTheme.DAY -> false
+                NightTheme.NIGHT -> true
+                NightTheme.SYSTEM -> isSystemInDarkTheme()
+            }
```
araratthehero, line 49: What do you think of making an extension property or a function out of this?
This pattern will be used in every activity, which supports compose. E.g. `nightThemeRepository.shouldUseDarkTheme()` or `nightThemeRepository.useDarkTheme`

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
```diff
@@ -0,0 +1,205 @@
...
+@Composable
+private fun handleNewIntent(
+    newIntent: Intent?,
+    componentData: SessionsGooglePayComponentData?,
+    onNewIntentHandled: () -> Unit
+): Boolean {
+    if (newIntent == null) return false
+    val googlePayComponent = getGooglePayComponent(componentData = componentData)
+    googlePayComponent.handleIntent(newIntent)
+    onNewIntentHandled()
```
araratthehero, line 189: I would use `LaunchedEffect` here, the way it is done in `handleAction()`. That way if the state gets updated frequently, the body of the `LaunchedEffect(newIntent)` will be executed only when the value of `newIntent` changes and not on every recomposition.

araratthehero: For extra safety, I would do the same for `handleActivityResult()`

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayViewModel.kt
```diff
@@ -0,0 +1,209 @@
...
+
+    fun availabilityChecked() {
+        updateState { it.copy(checkAvailability = null) }
+    }
+
+    fun buttonClicked() {
+        updateState { it.copy(uiState = SessionsGooglePayUIState.ShowComponent) }
+    }
+
+    fun actionConsumed() {
```
araratthehero, line 186: What do you think of renaming this method to `onActionConsumed()` to keep the naming consistent. Similarly `availabilityChecked `, `buttonClicked `, `activityResultHandled`, `newIntentHandled ` could be renamed too.

araratthehero (dismissed):
Looks good to me! 🚢

araratthehero (approved):

PR #1434, opened 2024-01-09
Voucher refactor and Multibanco support

araratthehero:
## Description

In this PR:
- Added support for `Multibanco` voucher.
- Refactored the `FullVoucherView` and made it more generic.
- Created two new vouchers called `BACS_VOUCHER` and `BOLETO_VOUCHER` for concrete voucher implementations, to not introduce breaking changes, which work with deprecated styles.
- Added a new option to save voucher as an image, if vouchers do not have `url` or `downloadUrl`.
- Second action button on voucher will move to the bottom of the first button if there is not enough space.
- Introduced a new approach to handle runtime permissions.
- Refactored QR code action to match the voucher.

## Checklist 
- [x] Code is unit tested
- [x] Aligned with design
- [x] Link to related issues

COAND-372


components-core/src/main/java/com/adyen/checkout/components/core/internal/PermissionRequestParams.kt
```diff
@@ -0,0 +1,21 @@
...
+package com.adyen.checkout.components.core.internal
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.internal.ui.PermissionHandlerCallback
+
+/**
+ * Runtime permission request data
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class PermissionRequestParams(
```
OscarSpruit, line 18: We use `...Params` for the internal configuration objects, so this name might be a bit confusing. Maybe `PermissionRequestData` would solve this confusion? What do you think?

components-core/src/main/java/com/adyen/checkout/components/core/internal/util/DateUtils.kt
```diff
@@ -78,4 +79,20 @@ object DateUtils {
...
+
+    /**
+     * Format server date pattern to regular date pattern (30/03/2023).
+     *
+     * @param date Date to be formatted
+     * @param shopperLocale
+     * @param dateFormat Date pattern
+     */
+    fun formatDateToString(
+        date: Date,
```
OscarSpruit, line 91: Is `Date` working well with different time zones?

voucher/src/main/res/layout/full_voucher_view.xml
```diff
@@ -91,51 +93,54 @@
...
+        android:id="@+id/space_buttons"
+        android:layout_width="match_parent"
+        android:layout_height="@dimen/standard_margin"
         app:layout_constraintTop_toBottomOf="@id/textView_reference_code" />
 
-    <com.google.android.material.button.MaterialButton
-        android:id="@+id/button_downloadPdf"
-        style="@style/AdyenCheckout.Voucher.ButtonDownloadPdf"
-        android:layout_width="wrap_content"
+    <com.google.android.flexbox.FlexboxLayout
```
OscarSpruit, line 102: What do we need this for exactly?

araratthehero: When button labels are big, we need to move the second button below the first button. `FlexboxLayout` solves that issue, but it might not be exactly what we want.
I am thinking of removing it and handling that case with a custom code.

What we want is, when the buttons are on one line, they need to be aligned to the start and the end of the view. But when they are above one another, they need to be centered.

action-core/src/main/java/com/adyen/checkout/action/core/internal/ui/DefaultGenericActionDelegate.kt
```diff
@@ -129,18 +136,26 @@ internal class DefaultGenericActionDelegate(
...
         delegate.exceptionFlow
             .onEach { exceptionChannel.trySend(it) }
             .launchIn(coroutineScope)
     }
 
-    private fun observeDetails(delegate: ActionDelegate) {
-        if (delegate !is DetailsEmittingDelegate) return
+    private fun observePermissionRequests(delegate: ActionDelegate) {
+        if (delegate !is PermissionRequestingDelegate) return
         Logger.d(TAG, "Observing details")
```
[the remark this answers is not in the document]
araratthehero: Good catch!

RELEASE_NOTES.md
```diff
@@ -11,3 +11,16 @@
...
 - You can now override payment method names in drop in by using `DropInConfiguration.Builder.overridePaymentMethodName(type, name)`
 - For stored cards, Drop-in will show the card name ("Visa", "Mastercard"...), instead of "Credit Card"
+- Add support for Multibanco voucher.
+- Permission request is being delegated to the `ActionComponentCallback`, `SessionComponentCallback` or `ComponentCallback` to handle it and return result through callback,
+- All vouchers will start in expanded mode
+- For voucher actions which have no `url` or `downloadUrl`, "Save as image" option will be offered to save the Voucher in `Downloads` folder.
+- Vouchers will save an image to user's phone with the following name format "Payment method type" + "Formatted data and time" (e.g. multibanco-2024-01-09T16_41_10).
+
+## Deprecated
+- The `AdyenCheckout.Voucher.Description.Bacs` style is deprecated. Use `AdyenCheckout.SimpleVoucher.Description` instead.
```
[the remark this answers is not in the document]
araratthehero: In that case every time we add a new voucher type, we need to add a new style for it. Which I think is okay. I can create one base style which will affect all vouchers and concrete styles for specific vouchers.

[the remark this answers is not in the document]
OscarSpruit: Let's discuss it with the team. I don't really have a preference, but I can imagine some merchants would like to style different vouchers differently

components-core/src/main/java/com/adyen/checkout/components/core/internal/util/DateUtils.kt
araratthehero: Creating a `Date()` instance, returns the current date and time set on the device. Are there other points I should pay attention to?

voucher/src/main/res/layout/full_voucher_view.xml
OscarSpruit: This is actually possible with just a `ConstraintLayout` by using the `Flow` tag. For example something like this:
```XML
<androidx.constraintlayout.helper.widget.Flow
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        app:constraint_referenced_ids="button1,button2"
        app:flow_horizontalStyle="spread_inside"
        app:flow_wrapMode="aligned"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toTopOf="parent" />
```

components-core/src/test/java/com/adyen/checkout/components/core/internal/util/DateUtilsTest.kt
```diff
@@ -0,0 +1,144 @@
...
+        val formattedDate = DateUtils.formatDateToString(calendar, pattern)
+
+        assertEquals(expectedFormattedDate, formattedDate)
+    }
+
+    companion object {
+        @JvmStatic
+        fun parseDateToView() = listOf(
+            // month, year, expectedFormattedDate
+            Arguments.arguments("1", "2021", "1/21"),
```
OscarSpruit, line 72: Nit: if you import `arguments()` directly it's possible to omit the verbose `Arguments.` part

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ActionComponentDialogFragment.kt
```diff
@@ -70,6 +79,7 @@ internal class ActionComponentDialogFragment :
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         _binding = FragmentGenericActionComponentBinding.inflate(inflater)
+        setInitViewState(BottomSheetBehavior.STATE_EXPANDED)
```
OscarSpruit, line 82: How does this impact other actions?

qr-code/src/test/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegateTest.kt
```diff
@@ -65,31 +76,91 @@ internal class DefaultQRCodeDelegateTest(
     private lateinit var statusRepository: TestStatusRepository
     private lateinit var paymentDataRepository: PaymentDataRepository
     private lateinit var delegate: DefaultQRCodeDelegate
-    private lateinit var fileDownloader: FileDownloader
+    private lateinit var imageSaver: ImageSaver
 
     @BeforeEach
     fun beforeEach() {
-        fileDownloader = spy(FileDownloader(context))
+        imageSaver = mock<ImageSaver>()
```
OscarSpruit, line 83: I'd suggest to put this as `@Mock private val imageSaver: ImageSaver` in the constructor with the other mocks

qr-code/src/test/java/com/adyen/checkout/qrcode/internal/ui/DefaultQRCodeDelegateTest.kt
```diff
@@ -385,38 +456,154 @@ internal class DefaultQRCodeDelegateTest(
     }
 
     @Test
-    fun `test download qr image with successful result`() = runTest {
-        delegate.initialize(CoroutineScope(UnconfinedTestDispatcher()))
+    fun `test refreshStatus`() = runTest {
```
OscarSpruit, line 459: Is it possible to give these test another name that follows the `when..., then...` format maybe? 😬

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/ImageSaver.kt
```diff
@@ -0,0 +1,198 @@
...
+        } catch (e: SecurityException) {
+            Result.failure(CheckoutException("Security violation: ", e))
+        }
+    }
+
+    private fun String.toURL(): URL? {
+        return try {
+            URL(this)
+        } catch (e: MalformedURLException) {
+            Logger.e(TAG, "toURL: $e")
```
OscarSpruit, line 187: I know this code comes from the old implementation, but could you maybe improve this message?

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/PermissionHandlerExtension.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import android.content.Context
+import android.content.pm.PackageManager
+import androidx.core.content.ContextCompat
+import com.adyen.checkout.core.PermissionHandlerCallback
+import com.adyen.checkout.core.internal.ui.PermissionHandler
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.resume
+
+suspend fun PermissionHandler.checkPermission(context: Context, requiredPermission: String): Boolean? =
```
OscarSpruit, line 19: This can be `internal`

voucher/src/main/java/com/adyen/checkout/voucher/internal/ui/view/FullVoucherView.kt
```diff
@@ -120,39 +120,68 @@ class FullVoucherView @JvmOverloads constructor(
...
-        binding.textViewExpirationLabel.isVisible = isVisible
-        binding.textViewExpirationDate.isVisible = isVisible
-        binding.expiryDateSeparator.isVisible = isVisible
+    private fun onDownloadPdfClicked() {
+        delegate.downloadVoucher(context)
+    }
+
+    private fun onSaveAsImageClicked() {
+        binding.layoutButtons.isVisible = false
+        doOnNextLayout {
```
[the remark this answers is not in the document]
araratthehero: When hiding the view and saving it as an image, it does not lay out the view yet, so it saves the empty space of the buttons as well. But if we wait for the view to get laid out again, it also updates the measurements.

Here is the saved image when saving the view before it gets laid out again after hiding the views.

Screenshot 2024-01-24 at 16 01 17

voucher/src/main/java/com/adyen/checkout/voucher/internal/ui/VoucherViewProvider.kt
```diff
@@ -22,15 +28,19 @@ internal object VoucherViewProvider : ViewProvider {
         context: Context,
     ): ComponentView {
         return when (viewType) {
-            VoucherComponentViewType.SIMPLE_VOUCHER -> VoucherView(context)
-            VoucherComponentViewType.FULL_VOUCHER -> FullVoucherView(context)
+            BACS_VOUCHER -> BacsVoucherView(context)
+            SIMPLE_VOUCHER -> SimpleVoucherView(context)
+            BOLETO_VOUCHER -> BoletoVoucherView(context)
+            FULL_VOUCHER -> FullVoucherView(context)
```
jreij, line 34: Let's simplify this and group similar views together

voucher/src/main/res/values/styles.xml
```diff
@@ -10,25 +10,38 @@
 
     <style name="AdyenCheckout.Voucher" />
 
-    <style name="AdyenCheckout.Voucher.Description.Bacs" parent="AdyenCheckout.TextAppearance.Primary">
+    <!-- This style is deprecated and will be removed in future releases. Please override
+         AdyenCheckout.SimpleVoucher.Description instead. -->
+    <style name="AdyenCheckout.Voucher.Description.Bacs" parent="AdyenCheckout.SimpleVoucher.Description">
```
jreij, line 15: Let's not use these styles anymore: `AdyenCheckout.Voucher.Description.Bacs` and `AdyenCheckout.Voucher.Description.Boleto` and make new ones that are more generic.
We shouldn't delete them to avoid breaking changes but maybe let's just make them empty, add a TODO comment to delete them in the next major release and mention in the release notes that they won't work anymore

voucher/src/main/res/values/styles.xml
```diff
@@ -10,25 +10,38 @@
...
 
-    <style name="AdyenCheckout.Voucher.Description.Boleto" parent="AdyenCheckout.TextAppearance">
+    <!-- This style is deprecated and will be removed in future releases. Please override
+         AdyenCheckout.FullVoucher.Description instead. -->
+    <style name="AdyenCheckout.Voucher.Description.Boleto" parent="AdyenCheckout.FullVoucher.Description">
+        <item name="android:textSize">12sp</item>
+        <item name="android:text">@string/checkout_voucher_introduction</item>
+    </style>
+
+    <style name="AdyenCheckout.FullVoucher.Description" parent="AdyenCheckout.TextAppearance">
```
jreij, line 33: I like the new separation between simple voucher styles and full voucher styles, but we should fix the naming.
If you go now to the example app and try to override the style from there, you will get an error:
Screenshot 2024-01-24 at 17 33 43

The reason for this is that the style `AdyenCheckout.FullVoucher` is not declared anywhere. That's why you can see an empty style `<style name="AdyenCheckout.Voucher" />` at the top of this file (and similarly in all other styles files).

So we can declare 2 new styles: `AdyenCheckout.SimpleVoucher` and `AdyenCheckout.FullVoucher` but since we already have `AdyenCheckout.Voucher` I will suggest using `AdyenCheckout.Voucher.Simple` and `AdyenCheckout.Voucher.Full`.

WDYT?

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/PermissionHandlerExtension.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        requestPermission(
+            context = context,
+            requiredPermission = requiredPermission,
+            callback = object : PermissionHandlerCallback {
+                override fun onPermissionGranted(requestedPermission: String) {
+                    if (requestedPermission == requiredPermission) {
+                        continuation.resume(true)
+                    } else {
+                        continuation.resume(null)
```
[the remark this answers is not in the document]
OscarSpruit: Thinking as a merchant `null` would be a bit unclear. Maybe we could use an enum to make it more clear?

voucher/src/main/java/com/adyen/checkout/voucher/internal/ui/model/VoucherInformationField.kt
```diff
@@ -0,0 +1,16 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ararat on 22/12/2023.
+ */
+
+package com.adyen.checkout.voucher.internal.ui.model
+
+import androidx.annotation.StringRes
+
+data class VoucherInformationField(
```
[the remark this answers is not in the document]
OscarSpruit: Oh I see, in that case a `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` will do

voucher/src/main/res/values/styles.xml
araratthehero: Good catch, I agree with your suggested changes.

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/PermissionHandlerExtension.kt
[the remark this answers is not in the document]
araratthehero: This is for now an internal class which will only be used by us. But I agree that enum would make it more clear. 👍

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ActionComponentDialogFragment.kt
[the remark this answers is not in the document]
araratthehero: When testing on smaller device, this is what I get.
Screenshot_20240125_113004

On bigger devices it works well too. Could you verify with the latest changes?

voucher/src/main/java/com/adyen/checkout/voucher/internal/ui/view/FullVoucherView.kt
```diff
@@ -67,47 +77,37 @@ class FullVoucherView @JvmOverloads constructor(
...
-            localizedContext
-        )
-        binding.buttonDownloadPdf.setLocalizedTextFromStyle(
-            R.style.AdyenCheckout_Voucher_ButtonDownloadPdf,
-            localizedContext
-        )
-        binding.textViewExpirationLabel.setLocalizedTextFromStyle(
-            R.style.AdyenCheckout_Voucher_ExpirationDateLabel,
-            localizedContext
+            localizedContext,
```
jreij, line 88: Localization with standalone components does not fully work with the new labels.
The best way to test this is to set the locale in the test app settings to a different value than the device locale, then use the instant payment component to test multibanco, you should see some strings are still in English.

You can check the `initLocalizedStrings` method and the `localizedContext` variable to see how we support this currently (you can find them in every view) but I will explain more in person.

voucher/src/main/res/layout/full_voucher_view.xml
```diff
@@ -91,51 +93,55 @@
...
         android:id="@+id/button_copyCode"
-        style="@style/AdyenCheckout.Voucher.ButtonCopyCode"
+        style="@style/AdyenCheckout.Voucher.ActionButton"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        app:layout_constraintEnd_toStartOf="@+id/button_downloadPdf"
-        app:layout_constraintHorizontal_chainStyle="packed"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toBottomOf="@id/textView_reference_code" />
+        android:text="@string/checkout_voucher_copy_code" />
```
jreij, line 121: Generally, with components we don't set the text attrbute in the layout xml files, we set them in `styles.xml` instead. That is to allow merchants to override the resource there if needed.
So we use `tools:text` instead of `android:text` in this file.
One other advantage of this (more like a side effect) is that we can spot localization issues like the one I mentioned in the previous comment, because the text will simply not appear if you forgot to localize it and you used `tools:text`. I can also explain more later.

jreij (approved):

jreij (approved):

PR #1435, opened 2024-01-09
Simplified configuration

OscarSpruit:
## Description
With these changes it is now easier to configure the SDK. `CheckoutConfiguration` eliminates the need to pass all common configuration values to every configuration object and by using Kotlin DSL you can create configurations in a more declarative way.

The PR is a little big, sorry for that.

## Checklist 
- [x] Code is unit tested

COAND-811


ozgur00:
We can get rid of the `configuration` part at the end of extension functions. Looks better imo

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.kt
```diff
@@ -106,3 +107,33 @@ class Adyen3DS2Configuration private constructor(
         }
     }
 }
+
+fun CheckoutConfiguration.adyen3DS2Configuration(
```
ozgur00, line 111: for example here. and all other places that it applies to

ozgur00:
We can consider making the `private var`s in `Builder`s to make the code more Kotlin DSL friendly. For now in order to not have any breaking changes we could keep the functions as well.

ozgur00:
After discussing it with @OscarSpruit and @jreij, we should leave considering nested component configurations in drop-in for v6. I think it's more intuitive for current users. I understand this way of using it covers merchants using both component and drop-in integrations. But we should keep the other alternative in mind and re-evaluate this before v6 depending on whether we're getting complaints or not.

ozgur00:
Since the common fields such as `shopperLocale`, `amount` etc. are not expected to be set within the drop-in or component configurations eventually we should remove them. We can already mark them as deprecated. I think they should be immutable fields since they're passed in the constructor.

ach/src/main/java/com/adyen/checkout/ach/ACHDirectDebitConfiguration.kt
```diff
@@ -122,8 +124,37 @@ class ACHDirectDebitConfiguration private constructor(
...
                 addressConfiguration = addressConfiguration,
-                isStorePaymentFieldVisible = isStorePaymentFieldVisible
+                isStorePaymentFieldVisible = isStorePaymentFieldVisible,
             )
         }
     }
 }
+
+@Suppress("FunctionName")
+fun CheckoutConfiguration.ACHDirectDebitConfiguration(
```
jreij, line 134: Can we use `achDirectDebitConfiguration` for consistency?

eps/src/main/java/com/adyen/checkout/eps/EPSConfiguration.kt
```diff
@@ -91,3 +93,38 @@ class EPSConfiguration private constructor(
         }
     }
 }
+
+@Suppress("FunctionName")
+fun CheckoutConfiguration.EPSConfiguration(
```
jreij, line 98: Same question as ACH

mbway/src/main/java/com/adyen/checkout/mbway/MBWayConfiguration.kt
```diff
@@ -94,3 +96,38 @@ class MBWayConfiguration private constructor(
         }
     }
 }
+
+@Suppress("FunctionName")
+fun CheckoutConfiguration.MBWayConfiguration(
```
jreij, line 101: Same question

qr-code/src/main/java/com/adyen/checkout/qrcode/QRCodeConfiguration.kt
```diff
@@ -70,3 +71,34 @@ class QRCodeConfiguration private constructor(
         }
     }
 }
+
+@Suppress("FunctionName")
+fun CheckoutConfiguration.QRCodeConfiguration(
```
jreij, line 76: Same question

upi/src/main/java/com/adyen/checkout/upi/UPIConfiguration.kt
```diff
@@ -93,3 +94,46 @@ class UPIConfiguration(
         )
     }
 }
+
+@Suppress("FunctionName")
+fun CheckoutConfiguration.UPIConfiguration(
```
jreij, line 99: You know the drill 😆

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -49,65 +50,59 @@ internal class CheckoutConfigurationProvider @Inject constructor(
...
+            amount = amount,
+            analyticsConfiguration = getAnalyticsConfiguration(),
+        ) {
+            // Drop-in
+            dropIn {
+                setEnableRemovingStoredPaymentMethods(true)
+            }
+
+            // Payment methods
+            bacsDirectDebit()
```
jreij, line 67: The configurations with empty blocks could give the impression that we need to declare them even if we don't need to configure anything. WDYT about removing them?

components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        repeat(size) {
+            val key = parcel.readString()!!
+            val configClass = parcel.readSerializable() as Class<Configuration>
+            val config = parcel.readParcelable<Configuration>(configClass.classLoader)!!
+            availableConfigurations[key] = config
+        }
+    }
+
+    fun addConfiguration(key: String, configuration: Configuration) {
```
jreij, line 57: Can this and `addActionConfiguration` be restricted?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/provider/ComponentParsingProvider.kt
```diff
@@ -1,948 +0,0 @@
...
-    fragment: Fragment,
-    paymentMethod: PaymentMethod,
-    dropInConfiguration: DropInConfiguration,
-    amount: Amount?,
-    componentCallback: ComponentCallback<*>,
-    sessionDetails: SessionDetails?,
-    analyticsRepository: AnalyticsRepository,
-    onRedirect: () -> Unit,
-): PaymentComponent {
-    val dropInParams = dropInConfiguration.mapToParams(amount)
```
jreij, line 594: This logic (passing the amount) was used to ensure the amount is updated in the partial payments flow. I cannot find the same logic in the new implementation so I think partial payments with the advanced flow might not be showing the right amount

ach/src/main/java/com/adyen/checkout/ach/internal/ui/model/ACHDirectDebitComponentParamsMapper.kt
```diff
@@ -53,29 +53,16 @@ internal class ACHDirectDebitComponentParamsMapper(
...
                 AddressParams.FullAddress(
                     supportedCountryCodes = supportedCountryCodes,
-                    addressFieldPolicy = AddressFieldPolicyParams.Required
+                    addressFieldPolicy = AddressFieldPolicyParams.Required,
                 )
             }
         }
     }
 
-    private fun ACHDirectDebitComponentParams.override(
```
jreij, line 65: If someone manually set a configuration for a certain component inside drop-in but with conflicting fields (different amount or shopper locale or client key etc), will the component still take the drop-in configuration as the source of truth?

OscarSpruit: The values from `CheckoutConfiguration` will always be used no matter what you configure in drop-in or for a component. Only sessions is able to override these values

drop-in/src/main/java/com/adyen/checkout/dropin/internal/provider/ComponentParsingProvider.kt
OscarSpruit: Good catch! Let me check this

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.kt
```diff
@@ -106,3 +107,33 @@ class Adyen3DS2Configuration private constructor(
...
+            amount?.let { setAmount(it) }
+            analyticsConfiguration?.let { setAnalyticsConfiguration(it) }
+        }
+        .apply(configuration)
+        .build()
+    addActionConfiguration(config)
+    return this
+}
+
+fun CheckoutConfiguration.getAdyen3DS2Configuration(): Adyen3DS2Configuration? {
```
jreij, line 125: Do these need to be public?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInComponentParamsMapper.kt
```diff
@@ -9,28 +9,28 @@
...
-        }
+        val dropInConfiguration = checkoutConfiguration.getDropInConfiguration()
+        return DropInComponentParams(
+            shopperLocale = checkoutConfiguration.shopperLocale,
+            environment = checkoutConfiguration.environment,
+            clientKey = checkoutConfiguration.clientKey,
+            analyticsParams = AnalyticsParams(checkoutConfiguration.analyticsConfiguration),
+            isCreatedByDropIn = true,
+            amount = overrideAmount,
+            showPreselectedStoredPaymentMethod = dropInConfiguration?.showPreselectedStoredPaymentMethod ?: true,
```
jreij, line 30: Since we have the default values here now, shall we make them nullable in the configuration builder and add their default values to the kdocs, just like the rest of the builders?

drop-in/src/test/java/com/adyen/checkout/dropin/DropInConfigurationTest.kt
```diff
@@ -0,0 +1,143 @@
...
+import com.adyen.checkout.redirect.RedirectConfiguration
+import com.adyen.checkout.redirect.getRedirectConfiguration
+import com.adyen.checkout.wechatpay.WeChatPayActionConfiguration
+import com.adyen.checkout.wechatpay.getWeChatPayActionConfiguration
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNotNull
+import org.junit.jupiter.api.Test
+import java.util.Locale
+
+internal class DropInConfigurationTest {
```
jreij, line 25: I have a suggestion for 1 more scenario to test:

```kotlin
val checkoutConfiguration = CheckoutConfiguration() {
    dropIn {
        ...
    }
    card {
        ...
    }
    someOtherPM {
        ...
    }
}
// assert drop-in and component configs are added
```

Maybe it can be combined with the first test?
I tried to think of more use cases to test but I'm not sure how practical they are so I'll leave it up to you to decide if there are other valid use cases to test.

drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -91,14 +92,31 @@ object DropIn {
...
+            context = context,
+            dropInLauncher = dropInLauncher,
+            checkoutSession = checkoutSession,
+            checkoutConfiguration = dropInConfiguration.toCheckoutConfiguration(),
+            serviceClass = serviceClass,
+        )
+    }
+
+    @JvmStatic
+    fun startPayment(
```
jreij, line 106: The new methods are missing kdocs

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.kt
OscarSpruit: They don't, but I thought it would maybe nice to have for merchants

drop-in/src/test/java/com/adyen/checkout/dropin/DropInConfigurationTest.kt
OscarSpruit: Hmm I don't think that's in scope for this test. Payment method configs are not added to the drop-in config when using `CheckoutConfiguration`. So the tests for the respective payment method configs should cover these cases

jreij: Ah good one 👍

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -49,65 +46,53 @@ internal class CheckoutConfigurationProvider @Inject constructor(
...
+            }
+
+            googlePay {
+                setCountryCode(keyValueStorage.getCountry())
+            }
+
+            // Actions
+            adyen3DS2 {
+                setThreeDSRequestorAppURL("https://www.adyen.com")
+            }
```
araratthehero, line 89: Currently I am able to call methods of every available implicit receiver inside a lambda, like:
```
adyen3DS2 {
    adyen3DS2 {
        adyen3DS2 {
            ...
        }
    }
}
```

It can be limitted by using a `@DslMarker` annotation.

You can create a new annotation:
```
@DslMarker
@Target(AnnotationTarget.CLASS, AnnotationTarget.TYPE)
annotation class CheckoutConfigurationMarker
```

and apply the `@CheckoutConfigurationMarker` annotation to `CheckoutConfiguration` class and to all extension function receivers like:

```
fun CheckoutConfiguration.adyen3DS2(
    configuration: (@CheckoutConfigurationMarker Adyen3DS2Configuration.Builder).() -> Unit = {}
): CheckoutConfiguration {
    ...
}
```

After that this should be the result:
Screenshot 2024-01-26 at 10 50 27

components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Locale
+
+class CheckoutConfiguration(
+    override val shopperLocale: Locale,
+    override val environment: Environment,
+    override val clientKey: String,
+    override val amount: Amount? = null,
+    override val analyticsConfiguration: AnalyticsConfiguration? = null,
+    @IgnoredOnParcel
+    private val config: CheckoutConfiguration.() -> Unit = {},
```
jreij, line 28: Can we use a better name for this?

components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
```diff
@@ -0,0 +1,105 @@
...
+import android.os.Parcelable
+import android.os.Parcelable.CONTENTS_FILE_DESCRIPTOR
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.components.core.internal.Configuration
+import com.adyen.checkout.core.Environment
+import kotlinx.parcelize.IgnoredOnParcel
+import java.util.Locale
+
+class CheckoutConfiguration(
+    override val shopperLocale: Locale,
```
jreij, line 22: Can we add support for context instead of locale?

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
OscarSpruit: That's really nice! I was wondering before if something like this was possible

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.kt
```diff
@@ -106,3 +107,33 @@ class Adyen3DS2Configuration private constructor(
...
+    return CheckoutConfiguration(
+        shopperLocale = shopperLocale,
+        environment = environment,
+        clientKey = clientKey,
+        amount = amount,
+        analyticsConfiguration = analyticsConfiguration,
+    ) {
+        addActionConfiguration(this@toCheckoutConfiguration)
+    }
+}
```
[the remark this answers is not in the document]
OscarSpruit: I played around with this, but to do it nicely I have to refactor some things. I propose to do that in a new PR, because this one is already too big. Is that okay?

components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
OscarSpruit: Changed it to `configurationBlock`. Is that okay?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/Adyen3DS2Configuration.kt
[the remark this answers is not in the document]
araratthehero: Sure, sounds good!

jreij (approved):

jreij (approved):

PR #1443, opened 2024-01-22
Exclude files from test coverage

OscarSpruit:
## Description
Exclude project specific files from test coverage since they can't be unit tested.

### DIFF
--- a/config/gradle/jacoco.gradle
+++ b/config/gradle/jacoco.gradle
@@ -10,7 +10,7 @@ if (project.hasProperty('android')) {
 
 project.afterEvaluate {
     jacoco {
-        toolVersion = "0.8.9"
+        toolVersion = "$jacoco_version"
     }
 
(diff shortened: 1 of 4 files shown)


araratthehero (dismissed):

config/gradle/jacoco.gradle
```diff
@@ -106,10 +115,14 @@ ext.coverageExclusions = [
     '**/SecurityCodeInput.*',
     '**/SocialSecurityNumberInput.*',
 
-    // Test classes
-    '**/*Test.*',
-    '**/Test*.*',
+    // Project specific files that can't be unit tested
```
jreij, line 118: What about excluding `View`s as well for now until we have proper testing?

OscarSpruit: We already did, but I just pushed something to make it work better

jreij (approved):

PR #1447, opened 2024-01-23
Add support for native redirects

OscarSpruit:
## Description
Add support for the native redirect action.

## Checklist 
- [x] Code is unit tested

COAND-805

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
@@ -62,9 +62,10 @@ data class PaymentComponentData<PaymentMethodDetailsT : PaymentMethodDetails>(
                 return try {
                     JSONObject().apply {
                         putOpt(PAYMENT_METHOD, serializeOpt(modelObject.paymentMethod, PaymentMethodDetails.SERIALIZER))
+                        putOpt(ORDER, serializeOpt(modelObject.order, OrderRequest.SERIALIZER))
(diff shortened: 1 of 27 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
```diff
@@ -37,6 +37,7 @@ data class PaymentComponentData<PaymentMethodDetailsT : PaymentMethodDetails>(
     var dateOfBirth: String? = null,
     var socialSecurityNumber: String? = null,
     var installments: Installments? = null,
+    var supportNativeRedirect: Boolean? = true,
```
jreij, line 40: Leaving a comment so we don't forget to check if this should be changed to `false`

RELEASE_NOTES.md
```diff
@@ -59,6 +59,7 @@ CheckoutConfiguration(
   - Convenience Stores Japan. Payment method type: **econtext_stores**
   - Online Banking Japan. Payment method type: **econtext_online**.
   - Seven-Eleven: Payment method type: **econtext_seven_eleven**
+- For redirect actions, the `RedirectComponent` is now able to deal with native app redirects.
```
jreij, line 62: Can you double check whether this should be in the release notes?

jreij (changes requested):
Just leaving this so we don't forget, we shouldn't have release notes for this feature now

jreij (approved):

PR #1449, opened 2024-01-24
Add missing parent styles in XML

jreij:
## Description
Some styles are not overridable and will cause a build error, because their parent style is not declared in `styles.xml`.

## Checklist 

COAND-841

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,7 @@
 
 ## Fixed
 - For drop-in with sessions, error dialogs will no longer display user unfriendly messages.
+- Overriding some of the XML styles without specifying a parent style no longer causes a build error.
 
 ## Changed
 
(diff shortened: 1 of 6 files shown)


RELEASE_NOTES.md
```diff
@@ -13,6 +13,7 @@
 
 ## Fixed
 - For drop-in with sessions, error dialogs will no longer display user unfriendly messages.
+- Overriding some XML styles no longer causes a build error.
```
OscarSpruit, line 16: It only happened for these styles if you'd omit the `parent`. Should we express more that it's kind of an edge case?

OscarSpruit (approved):

RELEASE_NOTES.md
jreij: I see your point but I can't find a way to clarify it, any suggestions?

OscarSpruit: Maybe something like:
Overriding XML styles in some cases could cause a build error. This is now fixed.

PR #1466, opened 2024-02-05
Apply shopper locale if provided to the sessions API call

jreij:
## Description
The sessions setup call returns a shopper locale if the merchant set this value in the `/sessions` call. We will now read this value and apply it on our end.

## Checklist 
- [x] Code is unit tested

COAND-847

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/model/Adyen3DS2ComponentParamsMapper.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/model/Adyen3DS2ComponentParamsMapper.kt
@@ -62,6 +62,7 @@ internal class Adyen3DS2ComponentParamsMapper(
         if (sessionParams == null) return this
         return copy(
             amount = sessionParams.amount ?: amount,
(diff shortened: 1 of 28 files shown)


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/model/Adyen3DS2ComponentParamsMapper.kt
```diff
@@ -62,6 +62,7 @@ internal class Adyen3DS2ComponentParamsMapper(
         if (sessionParams == null) return this
         return copy(
             amount = sessionParams.amount ?: amount,
+            shopperLocale = sessionParams.shopperLocale ?: shopperLocale,
```
araratthehero, line 65: I would write a test for this mapper too.

jreij: Good catch 👍 I added it

araratthehero (approved):

jreij:
We will need to change the behavior of this feature so I'm closing this for now.

Issue #530, opened 2021-11-09
Ability to provide Logger instances


rahulsainani:
Sometimes while analysing payment issues in qa/release builds it might be useful to know what's happening inside the SDK. The logs are visible while debugging and developing but generally qa/release builds have logcat(`android.util.Log`) and Adyen logging disabled. Since Adyen uses it's own Logger instance it is hard to achieve this. 

An idea would be to use patterns used in libraries like [Timber](https://github.com/JakeWharton/timber) and [Beagle](https://github.com/pandulapeter/beagle) where there is a Logger [interface/abstract class](https://github.com/JakeWharton/timber/blob/trunk/timber/src/main/java/timber/log/Timber.kt#L20) that defines the contract and prepares the logs/ handle common functionality for all log levels. Clients could provide their own instance of Logger which might log to their own backend or services like Firebase or Sentry on release builds so issues can be solved better. Adyen SDK could include the default logger which might log to logcat for debugging builds and can still make sure all sensitive logs are handled by the default logger and other loggers won't get access to it. 

An example of how Timber is configured with a CrashReportingTree

```
class CrashReportingTree : Timber.Tree() {

    override fun log(priority: Int, tag: String?, message: String, t: Throwable?) {
        if (priority != Log.VERBOSE) {
            Firebase.crashlytics.log("${getPriorityLabel(priority)}/ $message")
        }
        t?.let {
            Firebase.crashlytics.recordException(it)
        }
    }

    private fun getPriorityLabel(priority: Int): String =
        when (priority) {
            Log.ASSERT -> "A"
            Log.DEBUG -> "D"
            Log.INFO -> "I"
            Log.WARN -> "W"
            Log.ERROR -> "E"
            else -> "X"
        }
}

// Setting up
    if (BuildConfig.DEBUG) {
            Timber.plant(Timber.DebugTree())
    } else {
            Timber.plant(CrashReportingTree())
    }

```

Adyen logger could also be theoretically setup similarly. This was originally discussed in #514 in a [comment](https://github.com/Adyen/adyen-android/issues/514#issuecomment-957441707).

jreij:
Hi, thanks for the suggested! We discussed this internally following your comment on the other issue and we added this feature to the pipeline 🙂 

We don't have an ETA for it unfortunately but we will make sure we update this ticket when we have a PR.

PR #1467, opened 2024-02-05
Improved logging

OscarSpruit:
## Description
Logging is now improved:
- A custom `AdyenLogger` instance can be set with `AdyenLogger.setLogger`. This gives the ability to handle logs in your own custom way.
- Use `adyenLog(AdyenLogLevel, throwable) { "your log" }` to log something
    - The tag is automatically generated (no more need for a static tag). The way we generate tags is more performant than we did before.
    - The log message is given with a lambda. This is better for performance when logging is turned off.
- Use `LoggingExtension::class` in your tests to change the default Android logging to println logging (prevents crashes).

## Checklist 
- [x] Code is unit tested
- [x] Fixes: #530 

COAND-486


checkout-core/src/main/java/com/adyen/checkout/core/AdyenLogger.kt
```diff
@@ -8,17 +8,75 @@
...
+        tag: String,
+        message: String,
+        throwable: Throwable?,
+    )
+
+    companion object {
+
+        @PublishedApi
+        @Volatile
+        internal var logger: AdyenLogger = LogcatLogger()
```
araratthehero, line 35: When setting the logger, we delegate logging completely to that specific logger. I wonder if it would be beneficial to keep a list of loggers instead, to allow `LogcatLogger` and a custom logger(s) to handle logging simultaneously.
What do you think about this?

OscarSpruit: I did take that into consideration, but decided having one logger would be better. Having multiple loggers could mean that the log will be logged multiple times, bloating the logs. Next to that, some merchants might not want to log to logcat for a good reason. Finally, logcat is not always working (take unit tests as example). Having a list with `LogcatLogger` always in there would mean it would crash if unavailable.

araratthehero (approved):

araratthehero (approved):

araratthehero (approved):

drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -243,6 +243,11 @@ object DropIn {
     }
 
     private fun updateDefaultLogLevel(context: Context) {
-        Logger.updateDefaultLogLevel(BuildUtils.isDebugBuild(context))
+        val logLevel = if (BuildUtils.isDebugBuild(context)) {
+            AdyenLogLevel.DEBUG
+        } else {
+            AdyenLogLevel.NONE
+        }
+        AdyenLogger.setLogLevel(logLevel)
```
jreij, line 251: @OscarSpruit this is now causing an issue where drop-in overrides the log level explicitly set by the merchant. Check out `Logger.isLogLevelInitialized` in the old code

PR #1469, opened 2024-02-08
Draft proposal to update the Google Pay section of the sample application

JlUgia:
## Description
Proposal to update sample application to simplify the integration.
*Note*: This changes don't enforce API correctness, and are only meant to be used as guiding proposals. The code hasn't been run or tested. This review does not include thoughts from our Android counterparts, which we are still seeking.

The changes include:
- Hoist state and view model to the parent activity, and free the composable screen from the dependency
- [Tentative] Updated action related states to use `sealed`/`abstract` class. This should do without the need to `null`ify optional parameters after having taken action.
- Proposed removing the round trip to the viewModel on payment start and completed.
- Separate one off events from the UI state for increased clarity.
- Proposed to instantiate the Google Pay component only once (if possible – ie.: the `componentData` does not change). In addition, the component could be created once at the activity level and send as a param to the composable screen. That would allow us to handle the intent right away and free the screen and viewModel from that additional logic and extra UDF cycle.

Feel free to comment and modify.


example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayEvents.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Created by josephj on 13/12/2023.
+ */
+
+package com.adyen.checkout.example.ui.googlepay.compose
+
+import android.content.Intent
+import com.adyen.checkout.components.core.action.Action
+
+internal abstract class SessionsGooglePayEvents internal constructor() {
+    object None : SessionsGooglePayEvents()
```
jreij, line 15: Neat! We probably can't have simultaneous events in this flow so this should work

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
```diff
@@ -50,12 +46,17 @@ import com.google.pay.button.PayButton
 @Composable
 internal fun SessionsGooglePayScreen(
     useDarkTheme: Boolean,
+    googlePayState: SessionsGooglePayState,
+    eventsState: SessionsGooglePayEvents,
+    onGooglePayLauncherResult: (apiTaskResult: ApiTaskResult<PaymentData>) -> Unit,
     onBackPressed: () -> Unit,
-    viewModel: SessionsGooglePayViewModel,
 ) {
+    lateinit var googlePayComponent: GooglePayComponent
```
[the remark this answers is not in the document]
jreij: Another question, is there any case where this field might not get initialized and causes a problem? Asking because you definitely know more than me about compose 😅

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayViewModel.kt
```diff
@@ -162,69 +161,34 @@ internal class SessionsGooglePayViewModel @Inject constructor(
...
-        }
-    }
-
-    fun onGooglePayStarted() {
-        updateState { it.copy(startGooglePay = null) }
-    }
-
-    fun onActionConsumed() {
-        updateState { it.copy(actionToHandle = null) }
+        // TODO Call Adyen API
```
jreij, line 187: Do you plan on completing this part or do we do it ourselves after merge?

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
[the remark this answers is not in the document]
JlUgia: That's a question for us, since the component is only accessed in response to events and state changes under our control. In other words, if we can guarantee that the `fetchSession` action (triggering the `SessionsGooglePayEvents.ComponentData` event) completes before any of the following happens:
- A `SessionsGooglePayEvents.Action` or `SessionsGooglePayEvents.Intent` event is issued.
- The pay button is clicked
- The state becomes `SessionsGooglePayState.ShowButton` or `SessionsGooglePayState.ShowComponent`.

...then the logic must be safe.
My understanding is that this is the case at least for all Google Pay related events, since Google Pay availability is only requested one the session call returns. Is that also true for `Action`, or `Intent` events sent by the component?

On that note, I see that we are handling the `SessionsGooglePayState.ShowComponent` state in the compose UI, but that value is never set elsewhere.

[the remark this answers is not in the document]
JlUgia: The var is function local, so access modifiers are not necessary.

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayViewModel.kt
JlUgia: How is that done today?
What I've seen so far is that today, this is sent back to the compose flow, which calls the delegate's `handlePaymentResult` through the Google Pay component. I see that this creates a component state and emits it through `_componentStateFlow`. Where is this then handled?
I was hoping we could find a way to start this operation directly from the viewModel, since, after all, initiating a backend call should not need to update the UI flow.

jreij: Yes what you said is correct. After `_componentStateFlow` emits, the component has all the data it needs to send the payment to the backend. After that you will receive one of the 3 callbacks implemented in `SessionsGooglePayViewModel`: `onAction`, `onFinished` or `onError`.

To simplify this you can remove the `onGooglePayLauncherResult` method altogether and directly call the component inside the launcher:
```kotlin
    lateinit var googlePayComponent: GooglePayComponent

    val googlePayLauncher = rememberLauncherForActivityResult(
        contract = TaskResultContracts.GetPaymentDataResult(),
        onResult = googlePayComponent::handlePaymentResult,
    )
```
WDYT?

To clarify why I didn't do this initially, I was worried about keeping an instance of the component inside the screen, to be used later. Which is why every time I needed the component I had to go back to the view model and send the component data back to fetch the component instance again. But after seeing your `lateinit` solution which we discussed earlier I think this could work.

JlUgia: This works for me

jreij (approved):
🚀

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayEvents.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2023 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 13/12/2023.
+ */
+
+package com.adyen.checkout.example.ui.googlepay.compose
+
+internal abstract class SessionsGooglePayEvents internal constructor() {
```
jreij, line 11: @JlUgia I just noticed this line, is there any benefit from using this over `internal sealed class SessionsGooglePayEvents {`?

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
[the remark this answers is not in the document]
jreij: The tricky part here is with `Intent`. This is usually triggered after a redirect (custom tabs, external browser, app switch) has completed and we returned to the app through a deep link. With Google Pay that's mostly after the Google Pay component is finished and a 3DS challenge was launched in the browser (which is rare).

Right now we have this code in the `SessionsGooglePayActivity`:

```kotlin 
    override fun onNewIntent(intent: Intent) {
        ..
        sessionsGooglePayViewModel.onNewIntent(intent) // this triggers a `SessionsGooglePayEvents.Intent`
    }
```

The only case I can think of where this code executes before the `SessionsGooglePayEvents.ComponentData` event has been called is if the activity got destroyed in the background.

But then even without a redirect, I don't think this sample covers that scenario (app destroyed in the background and recreated), because we are fetching a new session when the view model is created so we lose the previous state of the component completely. I am not sure if any of our samples covers that use case, it's a tricky one to deal with.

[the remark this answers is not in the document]
JlUgia: For compose, note that the view model can outlive the activity (and hence the data there can be reused in subsequent activity recreations).

If the 3DS flow is added to the app, as a developer, I'd like to see one of the following (in order of preference):
- A good practice sample that I can use in my app (including the session renewal)
- A marked commented warning letting developers know about lifecycle challenges

If this is not deemed relevant for a simple sample application, I'd consider removing it otherwise.

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayEvents.kt
JlUgia: The main difference is that `sealed class`es enforce exhaustive checking (eg.: a `switch` statement will be forced to include a `case` statement for each member in the class), which has backwards-compatibility implications (especially for lists that are not meant to be exhaustive).

My rule of thumb is to use `sealed class` for exhaustive lists that are always meant to be handled in full. For example view states in a layout (loading, complete, error), and `abstract class` for non-exhaustive lists with unrelated items like `SesssionsGooglePayEvents` (even more so if they are part of an API that is expected to change).

Let me know if that makes sense.

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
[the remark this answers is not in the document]
jreij: Thanks for the input! I think we definitely would like to provide an example that fully works with all edge case scenarios in mind.

> and hence the data there can be reused in subsequent activity recreations

Yes and in this case our samples would work. However (correct me if I'm wrong) if the system decides to kill the app in the background to free up memory (just like when you switch on "Don't Keep Activities" in the developer options) then everything in memory will be destroyed including the view model. Which means the only thing we can use to recover the data is the `SaveStateHandle`. We already use it on the SDK side but not for all scenarios, so we need to investigate this more.

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayEvents.kt
jreij: This is a really good solution 💯  Especially in library development where adding something to a sealed class can technically cause a breaking change

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
```diff
@@ -71,26 +69,39 @@ internal fun SessionsGooglePayScreen(
...
+            }
+            is SessionsGooglePayEvents.Intent -> {
+                LaunchedEffect(eventsState.intent) {
+                    googlePayComponent.handleIntent(eventsState.intent)
+                }
+            }
+        }
+
         SessionsGooglePayContent(
+            googlePayComponent = googlePayComponent,
```
jreij, line 90: Btw I just tested the branch locally and this `lateinit` field is causing a crash here (and a few other places) when the screen is being created:
```
kotlin.UninitializedPropertyAccessException: lateinit property googlePayComponent has not been initialized
```

I'm not sure anymore if this solution will work. I'll try to find a solution but let me know if you have any suggestions.

JlUgia: I didn't have a client key implementation readily available, and haven't tested it. Let me take a look tomorrow. What's the quickest way to have a client key added?

jreij: Yeah sorry my bad, I should have tested it myself before merging but I didn't want to keep stalling this PR so I thought I'll fix any errors after merging 😅 

If you have an Adyen test account already then you can just follow [this step in the integration guide](https://docs.adyen.com/development-resources/client-side-authentication/#get-your-client-key). Otherwise don't worry about it, I'll try to figure it out.

PR #1470, opened 2024-02-08
Inject coroutine dispatchers

OscarSpruit:
## Description
Inject coroutines dispatchers instead of using them directly. This makes the code better testable.

Bonus:
- Reduced complexity of `DefaultAdyen3DS2Delegate` methods
- Removed dispatcher injection from `TestDispatcherExtension`
  - It was only used in one test and there were warnings about it

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/data/api/SubmitFingerprintService.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/data/api/SubmitFingerprintService.kt
@@ -12,23 +12,25 @@ import com.adyen.checkout.adyen3ds2.internal.data.model.SubmitFingerprintRequest
 import com.adyen.checkout.adyen3ds2.internal.data.model.SubmitFingerprintResponse
 import com.adyen.checkout.core.internal.data.api.HttpClient
 import com.adyen.checkout.core.internal.data.api.post
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
 internal class SubmitFingerprintService(
     private val httpClient: HttpClient,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) {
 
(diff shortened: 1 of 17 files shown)


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/data/api/SubmitFingerprintService.kt
```diff
@@ -12,23 +12,25 @@ import com.adyen.checkout.adyen3ds2.internal.data.model.SubmitFingerprintRequest
 import com.adyen.checkout.adyen3ds2.internal.data.model.SubmitFingerprintResponse
 import com.adyen.checkout.core.internal.data.api.HttpClient
 import com.adyen.checkout.core.internal.data.api.post
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
 internal class SubmitFingerprintService(
     private val httpClient: HttpClient,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
```
araratthehero, line 21: What do you think of calling this `dispatcher` or `coroutineDispatcher` to be more generic?
It could also be changed in other files.

OscarSpruit: Hmm yeah I was thinking about that. `dispatcher` is a bit too generic imo, so then I went with `coroutineDispatcher`. But then I thought "what if you have 2 or more dispatchers in a class?", then you need to make some sort of distinction.

WDYT? I'm fine with just `coroutineDispatcher` as well, since this case isn't happening anywhere

araratthehero: I think it is good to have a generic name, since you inject a different type of dispatcher from tests for example. Or in some cases you might want to inject a different dispatcher based on the call sight implementation.

araratthehero (approved):

PR #1485, opened 2024-02-21
Make configurations optional with the sessions flow

jreij:
## Description
- With the sessions flow, when starting drop-in or creating a component, the configuration parameter can be made optional.
- When `CheckoutSessionProvider.createSession` to create a `CheckoutSession`, we can pass the `environment` and `clientKey` instead of the whole configuration.

## Checklist 
- [x] Code is unit tested

COAND-853

### DIFF
--- a/checkout-core/src/main/java/com/adyen/checkout/core/internal/util/LocaleUtil.kt
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/internal/util/LocaleUtil.kt
@@ -8,6 +8,7 @@
 package com.adyen.checkout.core.internal.util
 
 import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.exception.CheckoutException
 import java.util.IllformedLocaleException
 import java.util.Locale
 
@@ -25,6 +26,9 @@ object LocaleUtil {
      */
     @JvmStatic
     fun toLanguageTag(locale: Locale): String {
+        if (!isValidLocale(locale)) {
+            throw CheckoutException("Invalid shopper locale: $locale.")
+        }
         return locale.toLanguageTag()
     }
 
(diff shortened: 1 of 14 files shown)


checkout-core/src/main/java/com/adyen/checkout/core/internal/util/LocaleUtil.kt
```diff
@@ -36,7 +40,11 @@ object LocaleUtil {
      */
     @JvmStatic
     fun fromLanguageTag(tag: String): Locale {
-        return Locale.forLanguageTag(tag)
+        val locale = Locale.forLanguageTag(tag)
+        if (!isValidLocale(locale)) {
```
OscarSpruit, line 44: Isn't this redundant? `forLanguageTag` already safely parses the tag, so I think it won't return an invalid locale

jreij: Fair point. But I'm not sure if it's obvious that `forLanguageTag` only returns valid locales. And I'm wondering if our validation and `forLanguageTag` work the exact same way internally. Or if we'll need to change our validation in the future and forget to re-add it here.
WDYT?

jreij: Yeah we've had this validation before and it caused issues, I think it considered some locales invalid that were actually valid.

In all cases the validation is now very basic and maybe only prevents obviously invalid use cases. But my point is that if we change it later on to something more robust and we don't call it inside `fromLanguageTag` it might cause issues.

Anyway happy to remove it if you think it's bad to have this redundancy.

OscarSpruit: Hmm yeah that is the problem with hardcoded lists, if countries or languages change they don't reflect the truth anymore.

I don't think we will ever create a more robust validation with all this in mind, so I think it's safe to remove

jreij: 👍

OscarSpruit (dismissed):

OscarSpruit (approved):

PR #1487, opened 2024-02-23
Allow shoppers remove stored cards for session integration

araratthehero:
## Description
- Now merchants do not have to override `onRemoveStoredPaymentMethod` to support removing stored payment methods for session integration.
- A new toggle has been added to the settings page of the example app to allow removing stored payment methods.
- `SessionConfiguration` object has been created to optimize `SessionParams` and keep configuration related params separate.

To be done:
- Align error messages with other platforms.

## Checklist 
- [x] Code is unit tested

COAND-835

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -83,3 +83,4 @@ CheckoutConfiguration(
 - When `CheckoutSessionProvider.createSession` to create a `CheckoutSession`, you can pass the `environment` and `clientKey` instead of the whole configuration.
 - In drop-in all actions will start in expanded mode
 - When using the Google Pay component, it is no longer necessary to manually import the `3ds2` module to handle transactions that require a native 3DS2 challenge. 
+- Removing stored payment methods are being handled internally if you are using Sessions integration. You do not need to override `onRemoveStoredPaymentMethod` function anymore.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -84,3 +84,4 @@ CheckoutConfiguration(
 - In drop-in all actions will start in expanded mode
 - When using the Google Pay component, it is no longer necessary to manually import the `3ds2` module to handle transactions that require a native 3DS2 challenge. 
 - Removing stored payment methods are being handled internally if you are using Sessions integration. You do not need to override `onRemoveStoredPaymentMethod` function anymore.
+- If you are using `DropInServiceResult.Error` without specifying an error message, the default has changed from `Error sending payment. Please try again.` to `An unknown error occurred`.
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
@@ -16,6 +16,7 @@ import java.util.Locale
 data class SessionParams(
     val enableStoreDetails: Boolean?,
     val installmentConfiguration: SessionInstallmentConfiguration?,
+    val showRemovePaymentMethodButton: Boolean?,
     val amount: Amount?,
     val returnUrl: String?,
     val shopperLocale: Locale?,
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
@@ -14,8 +14,11 @@ import java.util.Locale
 
 /**
  * Object that holds values set during sessions setup call.
- * [SessionParams] values should always have higher priority than values set in client side configurations. Otherwise
- * it can cause server error, since specific configuration is not enabled, but it is being used.
+ * [SessionParams] values by default should have higher priority than values set in client side configurations.
+ * Otherwise it can cause server error, since specific configuration is not enabled, but it is being used.
+ *
+ * Only for some specific cases, if they do not cause server error, client side configurations can have higher priority
+ * than [SessionParams] values.
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class SessionParams(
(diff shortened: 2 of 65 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInParamsMapper.kt
```diff
@@ -39,4 +43,10 @@ internal class DropInParamsMapper {
     fun getShopperLocale(checkoutConfiguration: CheckoutConfiguration, sessionParams: SessionParams?): Locale? {
         return checkoutConfiguration.shopperLocale ?: sessionParams?.shopperLocale
     }
+
+    private fun getIsRemovingStoredPaymentMethodsEnabled(
+        dropInConfiguration: DropInConfiguration?,
+        sessionParams: SessionParams?
+    ) = dropInConfiguration?.isRemovingStoredPaymentMethodsEnabled
```
jreij, line 50: This code allows the client side configuration to override the sessions configuration, is that intended?

sessions-core/src/test/java/com/adyen/checkout/sessions/core/internal/SessionInteractorTest.kt
```diff
@@ -707,6 +708,39 @@ internal class SessionInteractorTest(
         }
     }
 
+    @Nested
+    @DisplayName("when an update payment methods call is requested and")
```
jreij, line 712: I think the test names need to be fixed

sessions-core/src/main/java/com/adyen/checkout/sessions/core/internal/ui/model/SessionParamsFactory.kt
```diff
@@ -34,16 +35,19 @@ object SessionParamsFactory {
...
-            installmentConfiguration = SessionInstallmentConfiguration(
-                installmentOptions = sessionSetupConfiguration?.installmentOptions?.map {
-                    it.key to SessionInstallmentOptionsParams(
-                        plans = it.value?.plans,
-                        preselectedValue = it.value?.preselectedValue,
-                        values = it.value?.values,
-                    )
-                }?.toMap(),
-                showInstallmentAmount = sessionSetupConfiguration?.showInstallmentAmount,
+            configuration = SessionConfiguration(
```
jreij, line 38: The field is marked as nullable but it doesn't look like it can be null

araratthehero: The way we have it implemented it is not null, but I kept it null in the `SessionConfiguration` to keep it consistent with other fields in `SessionParams` and have them all nullable. Which also allows us to pass null in tests, where we do not need to create an `SessionConfiguration` object and pass null for all of it's arguments.

Though I do not mind making it non-nullable. Let me know what you think.

sessions-core/src/test/java/com/adyen/checkout/sessions/core/internal/SessionInteractorTest.kt
araratthehero: Good catch!

jreij: Yes but these tests are not about "when an update payment methods call is requested and" but about removing a stored payment methods

araratthehero: Ahh I see. Totally missed that and confidently copied that part. 😅

jreij:
@araratthehero I will review this again after you remove the draft status and rebase 🙏

components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
```diff
@@ -12,10 +12,16 @@ import androidx.annotation.RestrictTo
 import com.adyen.checkout.components.core.Amount
 import java.util.Locale
 
+/**
+ * Object that holds values set during sessions setup call.
+ * [SessionParams] values should always have higher priority than values set in client side configurations. Otherwise
```
jreij, line 17: This is correct for most attributes but there are exceptions. For example `shopperLocale` can be overridden by a client side configuration

drop-in/src/main/java/com/adyen/checkout/dropin/SessionDropInService.kt
```diff
@@ -222,6 +223,27 @@ open class SessionDropInService : BaseDropInService(), SessionDropInServiceInter
...
+            val result = sessionInteractor.removeStoredPaymentMethod(storedPaymentMethodId)
+
+            val serviceResult = when (result) {
+                SessionCallResult.RemoveStoredPaymentMethod.Successful ->
+                    RecurringDropInServiceResult.PaymentMethodRemoved(
+                        storedPaymentMethodId,
+                    )
+
+                is SessionCallResult.RemoveStoredPaymentMethod.Error -> RecurringDropInServiceResult.Error(
+                    errorDialog = ErrorDialog(),
```
jreij, line 238: What if we explicitly set the error message here even if it will match the default one? I'm thinking it might be tough for someone to read this line and figure out why aren't we setting any error message before finding out it will fall back to the default one anyway

components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/SessionParams.kt
araratthehero: Adjusted it a bit to give more context

RELEASE_NOTES.md
```diff
@@ -82,3 +82,5 @@ CheckoutConfiguration(
 - When `CheckoutSessionProvider.createSession` to create a `CheckoutSession`, you can pass the `environment` and `clientKey` instead of the whole configuration.
 - In drop-in all actions will start in expanded mode
 - When using the Google Pay component, it is no longer necessary to manually import the `3ds2` module to handle transactions that require a native 3DS2 challenge. 
+- Removing stored payment methods are being handled internally if you are using Sessions integration. You do not need to override `onRemoveStoredPaymentMethod` function anymore.
+- Default api error message for drop-in changed from `Error sending payment. Please try again.` to `An unknown error occurred`.
```
jreij, line 86: This can be any error that the merchant faced, we can't assume it's API. So maybe we can sat if you use `DropInServiceResult.Error` without specifying an error message, the default has changed?

araratthehero: That is better indeed

jreij (approved):

PR #1488, opened 2024-02-26
Continuation of #1469: Draft proposal to update the Google Pay section of the sample application

JlUgia:
Moved all Google Pay related logic from inside the composable, to confine the lifecycle of the `lateinit var` holding the Google Pay component, and make it easier to use.

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
@@ -12,6 +12,7 @@ package com.adyen.checkout.example.ui.googlepay.compose
 
 import android.app.Activity
 import androidx.activity.compose.rememberLauncherForActivityResult
+import androidx.activity.result.ActivityResultLauncher
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
@@ -36,6 +37,8 @@ import com.adyen.checkout.components.compose.AdyenComponent
 import com.adyen.checkout.components.compose.get
 import com.adyen.checkout.example.ui.compose.ResultContent
 import com.adyen.checkout.googlepay.GooglePayComponent
+import com.google.android.gms.tasks.Task
+import com.google.android.gms.wallet.PaymentData
 import com.google.android.gms.wallet.contract.TaskResultContracts
 import com.google.pay.button.ButtonTheme
 import com.google.pay.button.ButtonType
@@ -48,14 +51,6 @@ internal fun SessionsGooglePayScreen(
     eventsState: SessionsGooglePayEvents,
     onBackPressed: () -> Unit,
 ) {
-    lateinit var googlePayComponent: GooglePayComponent
-
-    val activity = LocalContext.current as Activity
-    val googlePayLauncher = rememberLauncherForActivityResult(
-        contract = TaskResultContracts.GetPaymentDataResult(),
-        onResult = googlePayComponent::handlePaymentResult,
-    )
-
     Scaffold(
         modifier = Modifier.windowInsetsPadding(WindowInsets.ime),
         topBar = {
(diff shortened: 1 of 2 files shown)


example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
```diff
@@ -118,7 +117,9 @@ private fun SessionsGooglePayContent(
 
             is SessionsGooglePayState.ShowButton -> {
                 PayButton(
-                    onClick = onButtonClicked,
+                    onClick = {
+                        googlePayComponent.startGooglePayScreen(googlePayLauncher)
```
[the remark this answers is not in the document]
JlUgia: The screenshot seems to indicate the leak as stemming from inside `PayButton`, and not `googlePayComponent`, is that accurate?
Do you have the full report? Is this `LeakCanary`?

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
```diff
@@ -118,7 +117,9 @@ private fun SessionsGooglePayContent(
...
+                    },
                     allowedPaymentMethods = googlePayComponent.getGooglePayButtonParameters().allowedPaymentMethods,
```
OscarSpruit, line 123: Currently, this line is crashing when there is an action after the payment (for example 3DS2 was triggered), because `googlePayComponent` is null. When an action is triggered I think this method will recompose, which will cause `googlePayComponent` to be null again. 

One way I can think of to mitigate this issue is to use `var googlePayComponent: GooglePayComponent? by remember { mutableStateOf(null) }`. I'm not very experienced with compose though, so I'm not sure if this is a proper solution. How do you think we can tackle this issue?

JlUgia: That's right, we need to `remember` this value. Let me propose something.

JlUgia: I'm worried about using an optional type (and needing to null check every call to `googlePayComponent`) when we know the lifecycle of the variable well (it gets initialized at the beginning, and can be relied upon from that point on).
I feel that some of that is due to the composable views and the google pay component lifecycles not being in tune (as in, the component does things that don't necessarily concern the view, like initializing itself when the component data is received).
Two initial ideas I see to alleviate the impact from the component in the view are:
- Keeping the component and some of its operation in a viewModel, that surfaces UI-related state changes to the UI when necessary.
- Decoupling some initialization parameters in the component that are not needed right away. For example some of the first actions issued by the google pay component is `startGooglePayScreen`. This method does not required any of the parameters fetched at init time.

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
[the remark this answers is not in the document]
dmengelt: @jreij just to rule out an older memory leak we had: what is the Play Services version of your device/emulator? 

```
Settings -> Apps -> "See all X apps" -> Click the search icon top right
-> Search for "Google Play Services" -> scroll to the bottom of the screen
```

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
OscarSpruit: > Keeping the component and some of its operation in a viewModel, that surfaces UI-related state changes to the UI when necessary.

Not sure if this is really possible. Our components extend `ViewModel`. Having a reference from a `ViewModel` to, in this case, `GooglePayComponent` sounds like bad practice and could potentially cause memory leaks.

> Decoupling some initialization parameters in the component that are not needed right away. For example some of the first actions issued by the google pay component is startGooglePayScreen. This method does not required any of the parameters fetched at init time.

Yes, I think this the way to go. Moving the initializing of the component up in the tree could probably resolve the nullability issue of my previous proposal as well. I was thinking to even move the component to the activity, but maybe that goes against any compose best practices:

image

example-app/src/main/java/com/adyen/checkout/example/ui/googlepay/compose/SessionsGooglePayScreen.kt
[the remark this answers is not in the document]
dmengelt: thanks @OscarSpruit

jreij:
We are redesigning the whole Google Pay component flow, closing this PR for now.

PR #1489, opened 2024-02-29
Restore action handling state

OscarSpruit:
## Description
When launching a redirect in the browser, the merchant app could be killed by the system. Finishing the redirect in that case would fail, because the state was lost. With these changes we can restore the state and finish the payment.

## Checklist 
- [x] Code is unit tested

COAND-857


action-core/src/main/java/com/adyen/checkout/action/core/internal/ui/DefaultGenericActionDelegate.kt
```diff
@@ -101,35 +103,41 @@ internal class DefaultGenericActionDelegate(
...
 
-            delegate.initialize(coroutineScope)
+        delegate.handleAction(action, activity)
+    }
 
-            observeDetails(delegate)
-            observeExceptions(delegate)
-            observePermissionRequests(delegate)
-            observeViewFlow(delegate)
+    private fun createDelegateAndObserve(action: Action, application: Application) {
```
ozgur00, line 121: we can separate it into two functions for creating and observing.

OscarSpruit: Let's say we have those 2 methods, would you then use it like this:
```Kotlin
fun someFun() {
    createDelegate(action, activity.application)
    observeDelegate(_delegate)
}
```
or call it from the `createDelegate` method?:
```Kotlin
fun createDelegate(action: Action, application: Application) {
    val delegate = ...
    observeDelegate(delegate)
}
```

ozgur00: not a strong preference but second one seems clearer to me. the idea behind it was the function does 2 separate things and could be more granular.

ozgur00 (dismissed):

OscarSpruit:
Let's wait a bit with merging this one

action-core/src/main/java/com/adyen/checkout/action/core/internal/ui/DefaultGenericActionDelegate.kt
```diff
@@ -76,6 +78,15 @@ internal class DefaultGenericActionDelegate(
     override fun initialize(coroutineScope: CoroutineScope) {
         adyenLog(AdyenLogLevel.DEBUG) { "initialize" }
         _coroutineScope = coroutineScope
+        restoreState()
+    }
+
+    private fun restoreState() {
+        adyenLog(AdyenLogLevel.DEBUG) { "Restoring state" }
+        val action: Action? = savedStateHandle[ACTION_KEY]
```
jreij, line 86: Can we use `SavedStateHandleProperty` for this?

OscarSpruit: Good one!

RELEASE_NOTES.md
```diff
@@ -14,6 +14,7 @@
 - Address Lookup not displaying validation error on Card Component when no address has been selected.
 - Actions no longer crash when your app uses obfuscation.
 - Drop-in no longer throws an error while handling a 3DS2 challenge on API 66 and below.
+- When the app process is killed after calling `handleAction`, then the state will now be restored and `handleIntent` can be called to continue the payment process.
```
jreij, line 17: I think we can make this a bit more broad, since we fixed many issues related to activity state restoration

OscarSpruit: Something like this?
```
When the app process is dies during action handling, then the state will now be restored and the payment process can be continued.
```

jreij: 👍

ozgur00 (approved):

PR #1490, opened 2024-02-29
Update 3DS2 parameters in example app

jreij:
## Description

Right now we have 2 switches in the example app. Their value is reflected in the `additionalData` in the `/payments` and `/sessions` request body.

These fields are now deprecated and need to be replaced with [authenticationData](https://docs.adyen.com/api-explorer/Checkout/71/post/payments#request-authenticationData).
Additionally, it is not obvious how the 2 switches work currently because we can get 3 states in fact and not 4.

This PR replaces the 2 switches with one `3D Secure` list selector having 3 options: `Prefer native - Redirect - Disabled`.

## Checklist 

COAND-861


example-app/src/main/java/com/adyen/checkout/example/data/api/model/PaymentsRequest.kt
```diff
@@ -30,13 +30,12 @@ data class PaymentsRequestData(
...
     val returnUrl: String,
-    val additionalData: AdditionalData,
+    val authenticationData: AuthenticationData,
     val threeDSAuthenticationOnly: Boolean,
     val shopperIP: String,
     val reference: String,
     val channel: String,
     val lineItems: List<Item>,
     val shopperEmail: String? = null,
-    val threeDS2RequestData: ThreeDS2RequestDataRequest?,
```
araratthehero, line 40: I see that [this field](https://docs.adyen.com/api-explorer/Checkout/71/post/payments#request-threeDS2RequestData) is not deprecated here. Since it is replaced with `authenticationData.threeDSRequestData`, what do you think of asking backend to deprecated it?
Though they do not look like the same objects to me.

example-app/src/main/res/values/strings.xml
```diff
@@ -42,10 +42,8 @@
     <string name="shopper_information">Shopper Information</string>
     <string name="shopper_reference_key">shopper_reference</string>
     <string name="shopper_reference_title">Shopper Reference</string>
-    <string name="threeds2_key">threeds2</string>
-    <string name="threeds2_title">3DS2</string>
-    <string name="execute3D_key">execute_3d</string>
-    <string name="execute3D_key_title">Execute 3D</string>
+    <string name="threeds_mode_key">threeds_mode</string>
+    <string name="threeds2_mode_title">3D Secure</string>
```
araratthehero, line 46: I would change this to:

```suggestion
    <string name="threeds_mode_title">3D Secure</string>
```

example-app/src/main/java/com/adyen/checkout/example/data/api/model/PaymentsRequest.kt
jreij: The object is not deprecated but apparently not applicable for checkout, so we should not be using it.

example-app/src/main/res/values/strings.xml
jreij: Fixed 👍

araratthehero (approved):

PR #1501, opened 2024-03-05
Analytics

araratthehero:
## Description
- Architecture for analytics system
- Network calls for analytics service and events
- `AnalyticsManager` implementation in MBWay delegate

## Checklist 
- [ ] Code is unit tested (we will add tests in a separate PR)
- [ ] Link to related issues

COAND-844


components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/remote/AnalyticsTrackRequestProvider.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import com.adyen.checkout.components.core.internal.analytics.AnalyticsEvent
+import com.adyen.checkout.components.core.internal.analytics.AnalyticsPlatformParams
+import com.adyen.checkout.components.core.internal.data.model.AnalyticsTrackInfo
+import com.adyen.checkout.components.core.internal.data.model.AnalyticsTrackLog
+import com.adyen.checkout.components.core.internal.data.model.AnalyticsTrackRequest
+
+internal class AnalyticsTrackRequestProvider {
+
+    operator fun invoke(
```
ozgur00, line 19: is there a specific reason this being an operator function?

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/AnalyticsManager.kt
```diff
@@ -0,0 +1,24 @@
...
+
+package com.adyen.checkout.components.core.internal.analytics
+
+import androidx.annotation.RestrictTo
+import kotlinx.coroutines.CoroutineScope
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+interface AnalyticsManager {
+
+    fun initialize(owner: Any, coroutineScope: CoroutineScope)
```
ozgur00, line 17: instead of having the owner to be the type of `Any` we can introduce an interface like `AnalyticsOwner` that would be implemented in `DropInViewModel` and `delegates` so we would be more precise what we need here. that could also prevent undesired calls for this function (not sure if that would ever happen, but if we try to call this in a place we would immediately know it should be an `AnalyticsOwner`

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManager.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    private var _coroutineScope: CoroutineScope? = null
+    private val coroutineScope: CoroutineScope get() = requireNotNull(_coroutineScope)
+
+    private var timerJob: Job? = null
+
+    private var ownerReference: String? = null
+
+    override fun initialize(owner: Any, coroutineScope: CoroutineScope) {
+        if (isInitialized) return
```
ozgur00, line 44: for cases like this where we do nothing logging what happened is a good idea to debug issues in the future

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManager.kt
```diff
@@ -0,0 +1,137 @@
...
+                onSuccess = { attemptId ->
+                    checkoutAttemptId = attemptId?.also { startTimer() }
+                },
+                onFailure = { adyenLog(AdyenLogLevel.WARN, it) { "Failed to fetch checkoutAttemptId." } },
+            )
+        }
+    }
+
+    override fun trackEvent(event: AnalyticsEvent) {
+        if (cannotSendEvents()) return
```
ozgur00, line 69: same thing for logging cases where we don't do anything

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/GenericEvents.kt
```diff
@@ -0,0 +1,130 @@
...
+    fun focus(
+        component: String,
+        target: String,
+    ) = AnalyticsEvent.Info(
+        component = component,
+        type = AnalyticsEvent.Info.Type.FOCUS,
+        target = target,
+    )
+
+    fun unFocus(
```
ozgur00, line 72: `unfocus` should be without uppercase `F`

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/remote/AnalyticsTrackRequestProvider.kt
OscarSpruit: Not necessarily, but this class is supposed to have 1 responsibility and this sort of reflects that

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/api/HttpClientExt.kt
```diff
@@ -66,7 +66,7 @@ suspend fun <T : ModelObject, R : ModelObject> HttpClient.post(
     adyenLog(AdyenLogLevel.VERBOSE) { "request - ${requestJson.toStringPretty()}" }
 
     val result = runAndLogHttpException { post(path, requestJson.toString(), queryParameters) }
-    val resultJson = JSONObject(String(result, Charsets.UTF_8))
+    val resultJson = JSONObject(String(result, Charsets.UTF_8).ifBlank { "{}" })
```
jreij, line 69: Nit: `.ifBlank { "{}" }` seems to me a bit hacky (not sure this is the right word).
What if we create an empty `JSONObject()` when the string is blank?

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/AnalyticsPlatformParams.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.annotation.RestrictTo
+import androidx.annotation.VisibleForTesting
+import com.adyen.checkout.components.core.BuildConfig
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object AnalyticsPlatformParams {
+
+    const val CHANNEL = "android"
+
+    var platform = AnalyticsPlatform.ANDROID.value
```
jreij, line 20: In the previous code we named these `actualPlatform` and `actualVersion` (check `AnalyticsMapper`). Is it intended to change them to `platform` and `version`?

OscarSpruit: Yes, before we used both forms with and without 'actual' prefixed, so it would be confusing if didn't have it. Now this class is just providing them, so it would be confusing when you're consuming this class

jreij: Ah good point. But the code now looks a bit confusing because `CHANNEL` is different from the rest:
```kotlin
version = AnalyticsPlatformParams.version,
channel = AnalyticsPlatformParams.CHANNEL,
platform = AnalyticsPlatformParams.platform,
```
Can we make all 3 values look the same for an external consumer?

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/AnalyticsManager.kt
ozgur00: Fair point. I'm thinking any class that implements that interface gets to define owner the way it wants to. I think we should avoid that somehow. I can't think of a way other than an interface for that. Also not sure if I'm being overly careful with this. @araratthehero I'm curious what you think about this?

araratthehero: Having an interface for this case will add extra work every time you want to use analytics manager. Analytics usage should be as simple as possible without extra implementations. 
Since we need the `owner` only for identification purposes, it is best to not introduce a limitation for the owner to be of a specific type and keep it as simple as possible.
`owner::class.qualifiedName` is generic and gives a unique value, which solves the issue for our use case, to identify if `clear()` function is called from the same owner. If we add extra usage for the owner other than identification, I would suggest reviewing what we have.

But if you have any idea how we can identify the caller without extra implementations or getting the `qualifiedName` of the caller class would be good to discuss.

jreij (approved):

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/DefaultNewAnalyticsRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+package com.adyen.checkout.components.core.internal.analytics.data
+
+import com.adyen.checkout.components.core.internal.analytics.AnalyticsEvent
+import com.adyen.checkout.components.core.internal.analytics.data.local.AnalyticsLocalDataStore
+import com.adyen.checkout.components.core.internal.analytics.data.remote.AnalyticsRemoteDataStore
+import com.adyen.checkout.components.core.internal.analytics.data.remote.AnalyticsSetupProvider
+import com.adyen.checkout.components.core.internal.analytics.data.remote.AnalyticsTrackRequestProvider
+
+internal class DefaultNewAnalyticsRepository(
+    private val localInfoDataStore: AnalyticsLocalDataStore<AnalyticsEvent.Info>,
```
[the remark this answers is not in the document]
araratthehero: We discussed this with @OscarSpruit and it would not add much value for us, since the repository implementation is quite simple already. Also we can not really remove the dependency on analytics types if we add a new abstraction.
I would only do that if the logic gets more complicated or the analytics behavior changes. Wdyt?

ozgur00 (approved):

PR #1517, opened 2024-03-15
Update address lookup docs

ozgur00:
## Description

COAND-870

### DIFF
--- a/docs/ADDRESS_LOOKUP.md
+++ b/docs/ADDRESS_LOOKUP.md
@@ -9,9 +9,52 @@
   ```
 ## Integrating with Address Lookup Functionality
 - If you're integrating with Drop-in:
-    - Implement the mandatory `onAddressLookupQueryChanged(query: String)` callback and optional `onAddressLookupCompletion(lookupAddress: LookupAddress)` callback.
+    - Implement the mandatory `onAddressLookupQueryChanged(query: String)` callback and optional `onAddressLookupCompletion(lookupAddress: LookupAddress)` callback in your implementation of DropInService.
     - Pass the result of these actions by using `AddressLookupDropInServiceResult` class.
+      ```kotlin
+      override fun onAddressLookupQueryChanged(query: String) {
+          // Use the query parameter to make a call to your desired maps endpoint
+          // Map the address objects LookupAddress object and 
+          // pass it back to Drop-in as follows:
+          sendAddressLookupResult(AddressLookupDropInServiceResult.LookupResult(options))  
+      }
+
+      // Optionally use this callback to get the complete details for a LookupAddress option selected by the shopper
+      override fun onAddressLookupCompletion(lookupAddress: LookupAddress): Boolean {
+          // Make a call to your desired maps endpoint and pass the complete 
+          // LookupAddress object back to Drop-in as follows:
+          sendAddressLookupResult(AddressLookupDropInServiceResult.LookupComplete(completeLookupAddress))
+          return true // Return false if you do not need to make an api call
+         
+      } 
+      ```
 - If you're integrating with standalone `CardComponent`:
     - Set `AddressLookupCallback` via `CardComponent.setAddressLookupCallback(AddressLookupCallback)` function to receive the related events.
     - Pass the result of these actions by calling `CardComponent.setAddressLookupResult(addressLookupResult: AddressLookupResult)`.
+    ```kotlin
+    cardComponent.setAddressLookupCallback(object : AddressLookupCallback {
+        override fun onQueryChanged(query: String) {
+            // Use the query parameter to make a call to your desired maps endpoint
+            // Map the address objects LookupAddress object and 
+            // pass it back to card component as follows:
+            cardComponent.updateAddressLookupOptions(options)
+        }
+
+  // Optionally use this callback to get the complete details for a LookupAddress option selected by the shopper
+        override fun onLookupCompletion(lookupAddress: LookupAddress): Boolean {
+            // Make a call to your desired maps endpoint and pass the complete 
+            // LookupAddress object back to card component as follows:
+            cardComponent.setAddressLookupResult(AddressLookupResult.Completed(completeLookupAddress))
+            return true // Return false if you do not need to make an api call
+        }
+    })
+    ```
     - Delegate back pressed event to `CardComponent` by calling `CardComponent.handleBackPress()` which returns true if the back press is handled by Adyen SDK and false otherwise.
+    ```kotlin
+    override fun onBackPressed() {
+        if (cardComponent?.handleBackPress() == true) return
+        super.onBackPressed()
+    }
+    ```
+## Screenshots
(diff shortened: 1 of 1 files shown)


docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
 - If you're integrating with Drop-in:
     - Implement the mandatory `onAddressLookupQueryChanged(query: String)` callback and optional `onAddressLookupCompletion(lookupAddress: LookupAddress)` callback.
```
jreij, line 12: Suggestion as discussed earlier: consider removing these line and keeping the code sample only to make the page less crowded

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
...
+      } 
+      ```
 - If you're integrating with standalone `CardComponent`:
     - Set `AddressLookupCallback` via `CardComponent.setAddressLookupCallback(AddressLookupCallback)` function to receive the related events.
     - Pass the result of these actions by calling `CardComponent.setAddressLookupResult(addressLookupResult: AddressLookupResult)`.
+    ```kotlin
+    cardComponent.setAddressLookupCallback(object : AddressLookupCallback {
+        override fun onQueryChanged(query: String) {
+            // Use the query parameter to make a call to your desired maps endpoint
+            // Map the address objects LookupAddress object and pass it back to Drop-in as follows:
```
jreij, line 44: `Drop-in` should be fixed

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
...
+            val options: List<LookupAddress> = // mapped list of address objects
+            cardComponent.updateAddressLookupOptions(options)
+        }
+
+        override fun onLookupCompletion(lookupAddress: LookupAddress): Boolean {
+            // Optionally use this callback to get the complete details for a LookupAddress option selected by the shopper
+            // If you need to use this callback to make a call to get complete details for the selected address option you must return true.
+            // You can also use this function conditionally. In that case, return true if you perform an api call, false otherwise.
+            // Make a call to your desired maps endpoint and pass the complete LookupAddress object back to Drop-in as follows:
```
jreij, line 53: `Drop-in` should be fixed

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
...
+      ```kotlin
+      override fun onAddressLookupQueryChanged(query: String) {
+          // Use the query parameter to make a call to your desired maps endpoint 
+          // Map the address objects LookupAddress object and pass it back to Drop-in as follows:
+          val options: List<LookupAddress> = // mapped list of address objects
+          sendAddressLookupResult(AddressLookupDropInServiceResult.LookupResult(options))  
+      }
+
+      override fun onAddressLookupCompletion(lookupAddress: LookupAddress): Boolean {
+          // Optionally use this callback to get the complete details for a LookupAddress option selected by the shopper
```
jreij, line 23: Maybe you can reduce the line size to make the code section readable without scrolling horizontally?

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
...
+          // If you need to use this callback to make a call to get complete details for the selected address option you must return true.
+          // You can also use this function conditionally. In that case, return true if you perform an api call, false otherwise.
```
jreij, line 25: can you use `API` instead of `api`?

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
...
+    ```kotlin
+    cardComponent.setAddressLookupCallback(object : AddressLookupCallback {
+        override fun onQueryChanged(query: String) {
+            // Use the query parameter to make a call to your desired maps endpoint
+            // Map the address objects LookupAddress object and pass it back to Drop-in as follows:
+            val options: List<LookupAddress> = // mapped list of address objects
+            cardComponent.updateAddressLookupOptions(options)
+        }
+
+        override fun onLookupCompletion(lookupAddress: LookupAddress): Boolean {
```
jreij, line 49: I feel like this code section and comment is too complex. I have a few suggestions, feel free to take any that you like:
- Add a small comment on top of the method, something like `// Optional: use this callback if you need to make an extra API call after the shopper selects an address option`. Makes it easy to decide whether you should continue reading the whole section or skip it.
- Write the whole sample assuming they will always make an API call if they implemented this method, and only add a comment next to `return true` that they can return false if they do not need to make the API call in certain cases. I find it easier when the docs focus on the most standard use case but mentions the extra options in case someone needs them.
- Make the comments a bit shorter 😅

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
...
+    ```
     - Delegate back pressed event to `CardComponent` by calling `CardComponent.handleBackPress()` which returns true if the back press is handled by Adyen SDK and false otherwise.
+    ```kotlin
+    override fun onBackPressed() {
+        if (cardComponent?.handleBackPress() == true) return
+        super.onBackPressed()
+    }
+    ```
+## Screenshots
+![address-lookup](https://github.com/Adyen/adyen-android/assets/6615094/ee671606-fbf4-4674-96ec-99a804e53d2e)
```
jreij, line 73: Minor: it could be nice if all screenshots included the device frame as well. If you take the screenshot from inside Android Studio you can then [add a frame to it afterwards](https://developer.android.com/studio/debug/am-screenshot). Or even just drag and drop screenshots into this [device art generator](https://developer.android.com/distribute/marketing-tools/device-art-generator).

docs/ADDRESS_LOOKUP.md
```diff
@@ -11,7 +11,63 @@
 - If you're integrating with Drop-in:
     - Implement the mandatory `onAddressLookupQueryChanged(query: String)` callback and optional `onAddressLookupCompletion(lookupAddress: LookupAddress)` callback.
     - Pass the result of these actions by using `AddressLookupDropInServiceResult` class.
+      ```kotlin
+      override fun onAddressLookupQueryChanged(query: String) {
```
jreij, line 15: Can we mention that these code samples are inside their implementation of `DropInService`? People using sessions might not have seen this class at all

docs/ADDRESS_LOOKUP.md
ozgur00: what needs to be fixed there? we always use it like this.
Screenshot 2024-03-18 at 13 25 07
Screenshot 2024-03-18 at 13 25 22
Screenshot 2024-03-18 at 13 25 36

docs/ADDRESS_LOOKUP.md
ozgur00: I think we have the necessary information displayed through these images and that doesn't add much imo.

jreij (approved):

PR #1519, opened 2024-03-18
Analytics testing

OscarSpruit:
## Description
A continuation of #1501, adding the unit tests that were missing there.

COAND-844

### DIFF
--- a/components-core/src/test/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManagerTest.kt
+++ b/components-core/src/test/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManagerTest.kt
@@ -1,0 +1,172 @@
+package com.adyen.checkout.components.core.internal.analytics
+
+import com.adyen.checkout.components.core.internal.analytics.data.AnalyticsRepository
+import com.adyen.checkout.components.core.internal.ui.model.AnalyticsParams
+import com.adyen.checkout.components.core.internal.ui.model.AnalyticsParamsLevel
+import com.adyen.checkout.test.LoggingExtension
+import com.adyen.checkout.test.TestDispatcherExtension
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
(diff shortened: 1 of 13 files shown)


components-core/src/test/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManagerTest.kt
```diff
@@ -0,0 +1,160 @@
...
+@OptIn(ExperimentalCoroutinesApi::class)
+@ExtendWith(MockitoExtension::class, LoggingExtension::class, TestDispatcherExtension::class)
+internal class DefaultAnalyticsManagerTest(
+    @Mock private val analyticsRepository: NewAnalyticsRepository,
+) {
+
+    private lateinit var analyticsManager: DefaultAnalyticsManager
+
+    @BeforeEach
+    fun setup() = runTest {
```
araratthehero, line 39: This could be:

```suggestion
    fun setup() {
```

components-core/src/test/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManagerTest.kt
```diff
@@ -0,0 +1,160 @@
...
+
+            analyticsManager.trackEvent(GenericEvents.rendered("dropin", false))
+
+            verify(analyticsRepository, never()).storeEvent(any())
+        }
+
+        @Test
+        fun `sending events is enabled, then events should be stored`() = runTest {
+            analyticsManager.initialize(this@TrackEventTest, this)
+            val event = GenericEvents.rendered("dropin", false)
```
araratthehero, line 107: What do you think of creating a new `AnalyticsEvent` internally for the usage of this test?

This way we can explicitly set the `shouldForceSend` flag and the tests will not fail if the implementation of `Info` or `Log` changes or any specific event changes its type.

components-core/src/test/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManagerTest.kt
```diff
@@ -0,0 +1,160 @@
...
+
+            verify(analyticsRepository).storeEvent(event)
+        }
+
+        @Test
+        fun `the event should force sending, then the event is sent right away`() = runTest {
+            whenever(analyticsRepository.fetchCheckoutAttemptId()) doReturn "test value"
+            analyticsManager.initialize(this@TrackEventTest, this)
+
+            analyticsManager.trackEvent(GenericEvents.submit("dropin"))
```
araratthehero, line 119: Same applies here for `shouldForceSend`.

araratthehero (approved):

PR #1529, opened 2024-03-22
Analytics - Add events - Card

araratthehero:
## Description
Track analytics for Card component

## Checklist 
- [x] Code is unit tested
- [x] Link to related issues

COAND-845


card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -174,6 +175,9 @@ class DefaultCardDelegate(
     private fun initializeAnalytics(coroutineScope: CoroutineScope) {
         adyenLog(AdyenLogLevel.VERBOSE) { "initializeAnalytics" }
         analyticsManager.initialize(this, coroutineScope)
+
+        val event = GenericEvents.rendered(paymentMethod.type.orEmpty())
```
ozgur00, line 179: it is unlikely but is it okay to send an empty string for the component param?

araratthehero: Indeed it is quite unlikely, but `paymentMethod.type` is nullable and to not add the check we decided to go with `.orEmpty()` approach.
We will monitor the events on data side and check later to see if there are any empty `component`.

ozgur00 (approved):

PR #1530, opened 2024-03-22
Analytics - Add events - Issuers list

araratthehero:
## Description
Track analytics for Issuers list

## Checklist 
- [x] Code is unit tested
- [x] Link to related issues

COAND-845

### DIFF
--- a/issuer-list/src/main/java/com/adyen/checkout/issuerlist/internal/ui/DefaultIssuerListDelegate.kt
+++ b/issuer-list/src/main/java/com/adyen/checkout/issuerlist/internal/ui/DefaultIssuerListDelegate.kt
@@ -18,6 +18,7 @@ import com.adyen.checkout.components.core.PaymentMethodTypes
 import com.adyen.checkout.components.core.internal.PaymentComponentEvent
 import com.adyen.checkout.components.core.internal.PaymentObserverRepository
 import com.adyen.checkout.components.core.internal.analytics.AnalyticsManager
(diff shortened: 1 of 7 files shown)


issuer-list/src/main/java/com/adyen/checkout/issuerlist/internal/ui/DefaultIssuerListDelegate.kt
```diff
@@ -128,6 +135,13 @@ internal class DefaultIssuerListDelegate<
         _outputDataFlow.tryEmit(outputData)
 
         updateComponentState(outputData)
+
+        val event = GenericEvents.selected(
```
OscarSpruit, line 139: Could it be that this method is triggered when something else happens then selecting an issuer?

araratthehero: `onInputDataChanged()` gets called only from `updateInputData()` and it gets called only when an issuer is selected.

I know it is not safe to do this here, since the naming is quite confusing, but after we remove `inputData` and change the way we communicate between the delegate and the view, this will be fixed.

OscarSpruit: Yes, I had the same thought. Just wanted to make sure it will not be called at the wrong time 👍

OscarSpruit (approved):

PR #1533, opened 2024-03-26
Analytics - Add events - Action handled

araratthehero:
## Description
Track action event, when action is handled

## Checklist 
- [x] Code is unit tested
- [x] Link to related issues

COAND-845

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
@@ -34,6 +34,8 @@ import com.adyen.checkout.components.core.internal.ActionObserverRepository
 import com.adyen.checkout.components.core.internal.PaymentDataRepository
 import com.adyen.checkout.components.core.internal.SavedStateHandleContainer
 import com.adyen.checkout.components.core.internal.SavedStateHandleProperty
+import com.adyen.checkout.components.core.internal.analytics.AnalyticsManager
+import com.adyen.checkout.components.core.internal.analytics.GenericEvents
 import com.adyen.checkout.components.core.internal.util.Base64Encoder
 import com.adyen.checkout.components.core.internal.util.bufferedChannel
 import com.adyen.checkout.core.AdyenLogLevel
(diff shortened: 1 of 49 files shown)


action-core/src/main/java/com/adyen/checkout/action/core/internal/ui/DefaultGenericActionDelegate.kt
```diff
@@ -129,6 +132,12 @@ internal class DefaultGenericActionDelegate(
             observeViewFlow(delegate)
         }
 
+        val event = GenericEvents.action(
```
OscarSpruit, line 135: One concern: if merchants use an action component directly to handle the action, then we would not send any event. Maybe we should move this event to each specific action delegate? WDYT?

araratthehero: Hmm, indeed it makes sense to move them to concrete action delegates. Will do it, thanks for the catch!

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
```diff
@@ -78,6 +80,7 @@ internal class DefaultAdyen3DS2Delegate(
     private val coroutineDispatcher: CoroutineDispatcher,
     private val base64Encoder: Base64Encoder,
     private val application: Application,
+    private val analyticsManager: AnalyticsManager?,
```
OscarSpruit, line 83: Just double checking: it's okay for this to be nullable, right? Because standalone action components should not do the initial analytics call?

araratthehero: I thought that we would only create and initialize analytics from payment components and pass them down to the action components. Also since we can not really create an analytics manager in the action component providers, since we would need the payment method type (for source) which we get only when `handleAction()` is being called, I made them nullable. Unless I missed something and it is possible.

What do you think about this? Probably we should check with product?

OscarSpruit: Oh yes you are right. I didn't think about the technical limitations. 

> I thought that we would only create and initialize analytics from payment components

This is what I also thought, so I think it's actually okay to have them nullable. Let's still double check though

OscarSpruit (approved):

PR #1550, opened 2024-04-04
Limit accessibility of CheckoutConfiguration extension methods

OscarSpruit:
## Description
Make methods to get specific configs from CheckoutConfiguration internal

COAND-874

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,9 @@
 - Localization issues in address lookup functionality.
 - Overriding some of the XML styles without specifying a parent style no longer causes a build error.
 
+## Removed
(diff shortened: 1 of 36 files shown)


araratthehero (dismissed):

jreij (changes requested):
Can you add release notes? The `Warning` section in the `5.3.0`  release notes should help

RELEASE_NOTES.md
```diff
@@ -12,6 +12,9 @@
 - Localization issues in address lookup functionality.
 - Overriding some of the XML styles without specifying a parent style no longer causes a build error.
 
+## Deprecated
```
jreij, line 15: I think this should be removed instead of deprecated?

OscarSpruit: Fair enough, forgot we had that tag

PR #1559, opened 2024-04-17
Address memory leaks

OscarSpruit:
## Description
Fix memory leaks:
- Observing flows with wrong coroutine scope
- Creating ViewModel in a custom way
- Unnecessary reference to component
- Reference to `Protocol` in `DropInBottomSheetDialogFragment`
- Bind BaseDropInService to DropInActivity lifecycle

Bonus:
- Fix build warning about unchecked casting

## Checklist 
- [x] Link to related issues

COAND-869


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
```diff
@@ -132,6 +132,7 @@ internal class DropInActivity :
 
         override fun onServiceDisconnected(className: ComponentName) {
             adyenLog(AdyenLogLevel.DEBUG) { "onServiceDisconnected" }
+            serviceBound = false
```
araratthehero, line 135: Is this necessary here?

I see that we already set `serviceBound` to `false` in `DropInActivity:245`

OscarSpruit: `onServiceDisconnected` is only triggered when the service is destroyed by external factors, like the system having not enough memory. In that case it would be nice for us to know if the service is still bound or not

araratthehero: I see, makes sense.

araratthehero (approved):

PR #1560, opened 2024-04-18
Feature/top bar navigation

araratthehero:
## Description
This PR includes the following changes:
- Swipe to dismiss Drop-in has been removed
- Top navigation has been added to all Drop-in screens
- Going back from actions now dismisses Drop-in, instead of opening it again
- Gift card confirmation screen now has a title, which shows the payment method name

## Checklist 
- [x] Link to related issues

COAND-868


RELEASE_NOTES.md
```diff
@@ -19,6 +19,8 @@
 - You can no longer use functions like `CheckoutConfiguration.getCardConfiguration()` or `CheckoutConfiguration.getDropInConfiguration()` to get configurations from the  `CheckoutConfiguration` object. When starting Drop-in or Components, pass the full `CheckoutConfiguration` object.
 
 ## Changed
+- Top navigation has been added to Drop-in and dragging gesture has been disabled which caused Drop-in to dismiss.
```
jreij, line 22: The existing release notes are from the latest release, we need to clear the file

jreij, line 22: Can you add a label for the PR?

drop-in/src/main/res/layout/bottom_sheet_toolbar.xml
```diff
@@ -0,0 +1,51 @@
...
+    style="@style/AdyenCheckout.BottomSheetToolbar"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content">
+
+    <androidx.appcompat.widget.AppCompatImageView
+        android:id="@+id/imageView_back"
+        style="@style/AdyenCheckout.Toolbar.Navigation"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:src="@drawable/ic_back"
```
jreij, line 21: Consider moving this to styles.xml to allow it to be overridden (same for the other icon)

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
```diff
@@ -65,6 +59,20 @@ internal class CardComponentDialogFragment : BaseComponentDialogFragment(), Addr
             .launchIn(viewLifecycleOwner.lifecycleScope)
     }
 
+    private fun initToolbar() = with(binding.bottomSheetToolbar) {
```
jreij, line 62: I found an edge case scenario that displays the wrong icon:
- Load drop-in with only one payment method that can display multiple screens (card + address lookup or BACS)
- Configure drop-in with `setSkipListWhenSinglePaymentMethod(true)` 
- Open drop-in and open the second screen
- A back icon is expected to show but a close icon is actually shown
- Tapping the icon does go back to the first screen so the tap action works correctly

I don't think we can do this now without adding extra functionality, let's discuss whether it's worth it to handle this

araratthehero: Good point. I did not like leaving it like that, so I fixed it. Could you please test it again?

bacs/src/main/java/com/adyen/checkout/bacs/internal/ui/DefaultBacsDirectDebitDelegate.kt
```diff
@@ -139,22 +139,24 @@ internal class DefaultBacsDirectDebitDelegate(
         }
     }
 
+    override fun canHandleBackPress() = isModeConfirmation()
+
     override fun handleBackPress(): Boolean {
-        val isConfirmationMode = _componentStateFlow.value.mode == BacsDirectDebitMode.CONFIRMATION
-        return if (isConfirmationMode) {
+        return if (isModeConfirmation()) {
```
ozgur00, line 145: a small one: i think we should call `canHandleBackPress()` here to check instead of `isModeConfirmation()` directly then we know we will handle back press if we can, then if we want to know how we decide that we can always go see the `canHandleBackPress()` details.

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -485,15 +485,19 @@ class DefaultCardDelegate(
         addressLookupDelegate.initialize(coroutineScope, inputData.address)
     }
 
+    override fun canHandleBackPress() = isViewTypeAddressLookup()
+
     override fun handleBackPress(): Boolean {
-        return if (_viewFlow.value == CardComponentViewType.AddressLookup) {
+        return if (isViewTypeAddressLookup()) {
```
ozgur00, line 491: i'd use the same approach here too

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/CardComponentDialogFragment.kt
jreij: Great effort with this! I like the general idea but my main issue with the implementation is that it only works for us internally (drop-in). Merchants using standalone components won't be able to benefit from it (as we're observing `component.viewFlow` which is internal). It would be great if this became part of some public functionality. 

Personally I wouldn't mind if we keep this part out of the scope of this PR as it's getting quite big and all of the solutions are going to be complicated enough to deserve their own PR.

araratthehero: After discussing with design and product, we decided to leave it as it is for now.

RELEASE_NOTES.md
```diff
@@ -8,45 +8,6 @@
...
-  | [AndroidX Compose BoM](https://developer.android.com/develop/ui/compose/bom/bom-mapping)                     | **2024.04.01**                |
-  | [AndroidX Compose Compiler](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.12) | **1.5.12**                    |
-  | [AndroidX Lifecycle](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.7.0)                | **2.7.0**                     |
-  | [Google Pay](https://developers.google.com/pay/api/android/support/release-notes#feb-24)                     | **19.3.0**                    |
-  | [Google Pay Compose Button](https://github.com/google-pay/compose-pay-button/releases/tag/v1.0.0)            | **1.0.0**                     |
-  | [Kotlin](https://github.com/JetBrains/kotlin/releases/tag/v1.9.24)                                           | **1.9.24**                    |
-  | [Kotlin coroutines](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.8.0)                         | **1.8.0**                     |
+- Top navigation has been added to Drop-in and dragging gesture has been disabled which caused Drop-in to dismiss.
```
jreij, line 12: Consider grouping these release notes under one point. Up to you of course, this is very subjective.
For example:
- drop-in navigation improvements:
  - top nav bar
  - drag gesture
  - navigation back from action

araratthehero: Grouping looks fine to me

jreij (approved):

PR #1584, opened 2024-04-25
Fix initial log level

OscarSpruit:
## Description
This fixes a bug where drop-in would override the log level that a merchant has already set before.

## Checklist 
- [x] Code is unit tested

COAND-889

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
@@ -256,6 +256,6 @@ object DropIn {
         } else {
             AdyenLogLevel.NONE
         }
-        AdyenLogger.setLogLevel(logLevel)
+        AdyenLogger.setInitialLogLevel(logLevel)
     }
 }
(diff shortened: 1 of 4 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/DropIn.kt
```diff
@@ -256,6 +256,6 @@ object DropIn {
         } else {
             AdyenLogLevel.NONE
         }
-        AdyenLogger.setLogLevel(logLevel)
+        AdyenLogger.setInitialLogLevel(logLevel)
```
jreij, line 259: Not in the scope of this PR but do you think we can do this for components as well?

jreij: What if we use the app startup library for this? It could be a nice place to introduce this library 👀

jreij: Will do that in a separate task

jreij (approved):

PR #1585, opened 2024-04-26
Update 3ds2 sdk

OscarSpruit:
## Description
Update 3ds2 sdk to 2.2.17

## Checklist 
- [x] Code is unit tested
- [x] Link to related issues

COAND-886


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/data/model/ChallengeResult.kt
```diff
@@ -38,7 +40,7 @@ internal class ChallengeResult private constructor(val isAuthenticated: Boolean,
             jsonObject.put(KEY_TRANSACTION_STATUS, transactionStatus)
             jsonObject.putOpt(KEY_AUTHORISATION_TOKEN, authorisationToken)
             jsonObject.putOpt(KEY_SDK_ERROR, errorDetails)
-            val payload = AndroidBase64Encoder().encode(jsonObject.toString())
+            val payload = Base64.encode(jsonObject.toString().toByteArray())
```
jreij, line 43: Is it safe to change this now and use an experimental API?

OscarSpruit: We are already using this in other places as well, so imo it's safe to use. Only risk is that the api will change in the future, but then we just change accordingly

jreij: Then let's look into replacing the remaining usages of `AndroidBase64Encoder`

jreij (approved):

PR #1586, opened 2024-04-29
Remove opened triggers from the release notes check action

jreij:
## Description
Having both the `opened` and `labeled` triggers as well as the `cancel-in-progress` flag on the `release-notes-check` action causes an issue where some PRs get stuck in a pending check state if they were labeled right before or right after being opened. This turned out to be a [known limitation](https://docs.github.com/en/actions/using-jobs/using-concurrency#using-concurrency-in-different-scenarios).

For now we will remove the `opened` trigger which solves this but it also means that every PR has to be labeled. Which is why we'll introduce the `Chore` label for PRs which we now keep unlabeled.

## Checklist 
- [x] PR is labelled 


araratthehero (dismissed):

araratthehero (approved):

.github/workflows/check_labels.yml
```diff
@@ -0,0 +1,47 @@
+name: Check Labels
+
+# Every PR should have a label and some labels should include an update to the release notes
+on:
+  pull_request:
+    branches-ignore:
+      - 'main'
+    types: [ synchronize, labeled, unlabeled ]
```
OscarSpruit, line 8: Is `synchronize` not going to cause similar issues as before?

jreij: Only if we push and label at the same time, but in that case unlabeled could cause the same issue. It's a tough one to fully fix as long as github does not support this themselves but I guess it's the closest we can get.

araratthehero (approved):

araratthehero:
@jreij Is it expected that the `release-notes-check` is pending for this PR? I would expect it to not be visible at all?

.github/workflows/check_labels.yml
OscarSpruit: Ah okay that should be fine. I didn't know when `synchronize` is triggered

OscarSpruit (approved):

jreij:
@araratthehero yeah, the rules for the required checks are set in the repo's settings not in the code, so it's still waiting for the `release-notes-check` job to be completed. But since I renamed it to `labels-check` it will not run here, we'll have to fix the settings first.

PR #1590, opened 2024-04-30
Update 3DS2 SDK [v4]

OscarSpruit:
## Description
Updated the 3DS2 SDK to 2.2.17 on v4

COAND-900


jreij (changes requested):
Can we add release notes?

PR #1611, opened 2024-05-02
Improve client key validation

jreij:
## Description
- Validate client key and environment together not separately.
- Fix inaccurate regex for client key.
- Add more tests.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-905

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/util/ValidationUtils.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/util/ValidationUtils.kt
@@ -17,19 +17,8 @@ object ValidationUtils {
     private const val PHONE_REGEX = "^\\D*(\\d\\D*){9,14}$"
     private val PHONE_PATTERN = Pattern.compile(PHONE_REGEX)
 
-    private const val CLIENT_KEY_REGEX = "([a-z]){4}_([a-zA-Z0-9]){32}"
-    private val CLIENT_KEY_PATTERN = Pattern.compile(CLIENT_KEY_REGEX)
-
-    private const val CLIENT_KEY_TEST_PREFIX = "test_"
-    private const val CLIENT_KEY_LIVE_PREFIX = "live_"
-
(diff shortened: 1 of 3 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/internal/util/ValidationUtils.kt
```diff
@@ -55,25 +45,20 @@ object ValidationUtils {
...
+            Environment.APSE,
+            Environment.AUSTRALIA,
+            Environment.EUROPE,
+            Environment.INDIA,
+            Environment.UNITED_STATES -> LIVE_CLIENT_KEY_PATTERN.matcher(clientKey).matches()
 
-        return (isLiveEnvironment && clientKey.startsWith(CLIENT_KEY_LIVE_PREFIX)) ||
-            (isTestEnvironment && clientKey.startsWith(CLIENT_KEY_TEST_PREFIX)) ||
-            (!isLiveEnvironment && !isTestEnvironment)
+            else -> GENERIC_CLIENT_KEY_PATTERN.matcher(clientKey).matches()
```
OscarSpruit, line 61: Should we even check this case? Can't think of a situation where a client key wouldn't be live or test 🤔

jreij: Yeah I was in between with this one. Internally we have other environments as well, and although they are not accessible publicly we might want to use them ourselves, so instead of always returning valid or invalid I thought of adding this extra validation. WDYT?

OscarSpruit: Hmm I think it's better to not build in a potential backdoor. When we need this for testing we can disable validation locally

jreij: Agreed, fixed now

OscarSpruit (approved):

PR #1613, opened 2024-05-02
Migrate to KSP

OscarSpruit:
## Description
Migrate from KAPT to KSP. KAPT was only used in the example app for Hilt, so this does not impact our SDK in any way.

`./gradlew clean assDeb --no-build-cache` went from 1m 2s to 38s locally 🚀


build.gradle
```diff
@@ -6,6 +6,7 @@ plugins {
     id 'com.android.application' version "$android_gradle_plugin_version" apply false
     id 'com.android.library' version "$android_gradle_plugin_version" apply false
     id 'org.jetbrains.kotlin.android' version "$kotlin_version" apply false
+    id 'com.google.devtools.ksp' version '1.9.23-1.0.20' apply false
```
araratthehero, line 9: Could you move this version to a variable as we do for other plugins?

araratthehero (dismissed):

PR #1614, opened 2024-05-02
Migrate to non-transitive R classes

OscarSpruit:
## Description
Migrate to non-transitive R classes. R classes must now by fully referenced and modules need an explicit dependency on a library of which the R class is from.

`./gradlew clean assDeb --no-build-cache` went from 1m 2s to 45s locally. Combined with KSP it will be even better


jreij (commented):
Could this have any impact on merchants?

OscarSpruit:
> Could this have any impact on merchants?

I thought it doesn't, but let's test it out

jreij (approved):

PR #1628, opened 2024-05-14
Support changing the custom tabs toolbar and navigation colors

jreij:
## Description
Currently we directly use `colorPrimary` as the toolbar color for our custom tabs where we launch external redirects. This PR allows changing this color as well as 3 other customizable colors (`toolbarColor`, `secondaryToolbarColor`, `navigationBarColor` and `navigationBarDividerColor`).

To preserve consistency with custom styling across the library, we added an XML style that can be overridden and will be applied internally to the custom tabs. This also allows easily setting different light and dark theme colors.

## Checklist 
- [x] PR is labelled 

COAND-907


ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/CustomTabsLauncher.kt
```diff
@@ -0,0 +1,67 @@
...
+    fun launchCustomTab(context: Context, uri: Uri): Boolean {
+        @Suppress("SwallowedException")
+        return try {
+            CustomTabsIntent.Builder()
+                .setShowTitle(true)
+                .setDefaultColorSchemeParams(getDefaultColorSchemeParams(context))
+                .build()
+                .launchUrl(context, uri)
+            true
+        } catch (e: ActivityNotFoundException) {
```
OscarSpruit, line 32: Should we log this exception?

ui-core/src/main/res/values/attrs.xml
```diff
@@ -9,8 +8,15 @@
...
     <declare-styleable name="RoundCornerImageView">
-        <attr name="adyenRadius" format="dimension"/>
-        <attr name="adyenStrokeWidth" format="dimension"/>
-        <attr name="adyenStrokeColor" format="color"/>
+        <attr name="adyenRadius" format="dimension" />
+        <attr name="adyenStrokeWidth" format="dimension" />
+        <attr name="adyenStrokeColor" format="color" />
+    </declare-styleable>
+
+    <declare-styleable name="AdyenCheckoutCustomTabs">
```
OscarSpruit, line 16: Isn't this name supposed to be the same as a custom view? And since we don't have a view, is it necessary to declare a styleable?

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/CustomTabsLauncher.kt
jreij: Right now its callers are logging it based on their own behavior, some of them fall back to another solution (so log level is debug) while other fail without it (log level error). Do you think we should add a verbose log here as well?

ui-core/src/main/res/values/attrs.xml
jreij: You're right I missed that, can you check the latest commit?

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/util/CustomTabsLauncher.kt
OscarSpruit: I was thinking that the exception maybe has some useful information, but on second thought probably not. When this exception is thrown it basically can only mean there is no compatible browser. So I think it's fine like this

PR #1629, opened 2024-05-14
Fix drop-in unable to handle 3DS2 on API v66 and below

jreij:
## Description
With Checkout API v66 and earlier, drop-in throws an error while trying to handle two separate 3DS2 actions returned by the API (fingerprint and challenge).

## Checklist 
- [x] PR is labelled 

COAND-906


OscarSpruit (approved):
Nice! This fixes an issue I was running into in the state PR

PR #1632, opened 2024-05-21
Restore 3DS2 state after process death

OscarSpruit:
## Description
Restore 3DS2 state after process death

Note: labeled this PR as chore as there are already release notes about this from another PR

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-857

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
@@ -97,6 +97,8 @@ internal class DefaultAdyen3DS2Delegate(
 
     private var authorizationToken: String? by SavedStateHandleProperty(AUTHORIZATION_TOKEN_KEY)
 
+    private var action: BaseThreeds2Action? by SavedStateHandleProperty(ACTION_KEY)
+
     override fun initialize(coroutineScope: CoroutineScope) {
         _coroutineScope = coroutineScope
     }
@@ -122,9 +124,11 @@ internal class DefaultAdyen3DS2Delegate(
 
     override fun handleAction(action: Action, activity: Activity) {
         if (action !is BaseThreeds2Action) {
-            exceptionChannel.trySend(ComponentException("Unsupported action"))
+            emitError(ComponentException("Unsupported action"))
             return
         }
+
+        this.action = action
 
         val paymentData = action.paymentData
         paymentDataRepository.paymentData = paymentData
@@ -140,7 +144,7 @@ internal class DefaultAdyen3DS2Delegate(
         activity: Activity,
     ) {
         if (action.token.isNullOrEmpty()) {
-            exceptionChannel.trySend(ComponentException("Fingerprint token not found."))
+            emitError(ComponentException("Fingerprint token not found."))
             return
         }
         identifyShopper(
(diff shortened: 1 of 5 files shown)


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/SharedChallengeStatusHandler.kt
```diff
@@ -0,0 +1,40 @@
...
+
+internal object SharedChallengeStatusHandler : ChallengeStatusHandler {
+
+    var onCompletionListener: ChallengeStatusHandler? = null
+        set(value) {
+            field = value
+            resultQueue?.let { onCompletion(it) }
+        }
+
+    private var resultQueue: ChallengeResult? = null
```
araratthehero, line 23: To me this naming is a little misleading, since it is not a queue of operations.
What do you think of renaming this to `queuedResult` or keep the name the same but change the type to a list (`mutableListOf<() -> Unit>()`)?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
```diff
@@ -97,8 +97,11 @@ internal class DefaultAdyen3DS2Delegate(
 
     private var authorizationToken: String? by SavedStateHandleProperty(AUTHORIZATION_TOKEN_KEY)
```
araratthehero, line 98: Should we get rid of the `authorizationToken` here, if we are storing `action` already?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
```diff
@@ -97,8 +97,11 @@ internal class DefaultAdyen3DS2Delegate(
...
 
+    private var action: BaseThreeds2Action? by SavedStateHandleProperty(ACTION_KEY)
```
araratthehero, line 100: Is storing `action` necessary, since `action` is always passed in `handleAction()`?

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/SharedChallengeStatusHandler.kt
OscarSpruit: `queuedResult` sounds better indeed

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
OscarSpruit: Let me check

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
OscarSpruit: Yes, it seems like we don't do anything with it, but it's the opposite. By storing the action `DefaultGenericActionDelegate` will be able to recreate the 3ds delegate

3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/ui/DefaultAdyen3DS2Delegate.kt
OscarSpruit: Yes just removed it

araratthehero (approved):

PR #1633, opened 2024-05-22
Change phone number country code picker

OscarSpruit:
## Description
- Align country code picker with other platforms
- Sort list by localized country name
- Some design tweaks
- Addition of NEW secondary text color
- Preselected correct country code for MBWay

Screenshot_20240522_150738

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-912

### DIFF
--- a/mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/DefaultMBWayDelegate.kt
+++ b/mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/DefaultMBWayDelegate.kt
@@ -177,7 +177,7 @@ internal class DefaultMBWayDelegate(
     }
 
     companion object {
-        private const val ISO_CODE_PORTUGAL = "PT"
+        internal const val ISO_CODE_PORTUGAL = "PT"
         private const val ISO_CODE_SPAIN = "ES"
 
         private val SUPPORTED_COUNTRIES = listOf(ISO_CODE_PORTUGAL, ISO_CODE_SPAIN)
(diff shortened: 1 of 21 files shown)


mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MbWayView.kt
```diff
@@ -92,10 +91,11 @@ internal class MbWayView @JvmOverloads constructor(
                 onCountrySelected(country)
             }
         }
-        val firstCountry = countries.firstOrNull()
-        if (firstCountry != null) {
-            binding.autoCompleteTextViewCountry.setText(firstCountry.toShortString())
-            onCountrySelected(firstCountry)
+        val initialCountry =
+            countries.firstOrNull { it.isoCode == DefaultMBWayDelegate.ISO_CODE_PORTUGAL } ?: countries.firstOrNull()
```
araratthehero, line 95: Is this necessary here? Since countries should only have the supported countries right.
```suggestion
        val initialCountry = countries.firstOrNull()
```

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/DefaultMBWayDelegate.kt
```diff
@@ -177,7 +178,7 @@ internal class DefaultMBWayDelegate(
     }
 
     companion object {
-        private const val ISO_CODE_PORTUGAL = "PT"
+        internal const val ISO_CODE_PORTUGAL = "PT"
```
araratthehero, line 181: I think this can also change back to private
```suggestion
        private const val ISO_CODE_PORTUGAL = "PT"
```

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MbWayView.kt
OscarSpruit: Yes the list contains Spain and Portugal and the order is not guaranteed. We have to preselect Portugal. Using just `countries.firstOrNull()` could preselect Spain instead

araratthehero: I see. I guess it happens when the names are localized right?

My thinking is to move that logic to the delegate.
What if we create a `getDefaultSupportedCountry()` function in the delegate and return the ISO code or the country itself?

araratthehero (approved):

PR #1636, opened 2024-05-27
Implement Ideal 2.0 flow

OscarSpruit:
## Description
- Decouple `IdealComponent` etc from IssuerList (as much backwards compatible as possible)
- `IdealComponent` now functions like `InstantPaymentComponent` 

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-875

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,11 +8,19 @@
 [//]: # (## Deprecated)
 [//]: # ( - Configurations public constructor are deprecated, please use each Configuration's builder to make a Configuration object)
 
-## Added
+## New
 - UPI now supports `upi_intent` payment apps.
+- The new iDEAL payment flow where the shopper is redirected to the iDEAL payment page to select their bank and authorize the payment.
 
 ## Changed
 - Drop-in navigation improvements:
   - Top navigation has been added
   - Dragging gesture has been disabled which caused Drop-in to dismiss
   - Going back from actions dismisses Drop-in
+
+## Deprecated
+- For `IdealComponent`:
+  - `isConfirmationRequired()` can be removed.
+  - `submit()` can be removed.
+- When configuring iDEAL:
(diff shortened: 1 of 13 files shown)


ozgur00:
Using drop-in, after an initial successful payment when I try to choose ideal we get an error both for sessions and advanced flow. @OscarSpruit did you notice this?

ideal/src/main/java/com/adyen/checkout/ideal/internal/ui/DefaultIdealDelegate.kt
```diff
@@ -0,0 +1,104 @@
...
+        )
+    }
+
+    override fun removeObserver() {
+        observerRepository.removeObservers()
+    }
+
+    private fun createComponentState(): IdealComponentState {
+        val paymentComponentData = PaymentComponentData(
+            paymentMethod = IdealPaymentMethod(
```
ozgur00, line 86: Is `IdealPaymentMethod` class necessary, or is it possible to use `GenericPaymentMethod` here?

OscarSpruit: Technically it is possible, but since these are public classes it would break the API if we change it

RELEASE_NOTES.md
```diff
@@ -8,6 +8,9 @@
 [//]: # (## Deprecated)
 [//]: # ( - Configurations public constructor are deprecated, please use each Configuration's builder to make a Configuration object)
 
+## New
+- The new iDEAL payment flow where the shopper is redirected to the iDEAL payment page to select their bank and authorize the payment.
```
jreij, line 12: Let's add the deprecated methods

jreij (changes requested):

ozgur00 (approved):

PR #1658, opened 2024-06-04
Extend autofill support

OscarSpruit:
## Description
Extend autofill support for:
- Gift cards
- UPI
- Improved address input

## Checklist 
- [x] PR is labelled 

COAND-917


ui-core/src/main/java/com/adyen/checkout/ui/core/internal/ui/view/AddressFormInput.kt
```diff
@@ -324,6 +332,8 @@ class AddressFormInput @JvmOverloads constructor(
     private fun initHouseNumberInput(styleResId: Int?) {
         styleResId?.let { textInputLayoutHouseNumber?.setLocalizedHintFromStyle(it, localizedContext) }
         editTextHouseNumber?.apply {
+            // TODO
+//            setAutofillHints(View.AUTO)
```
OscarSpruit, line 336: There is no autofill hint for house number. Currently, the house number is automatically autofilled into the street field. Should we add logic to extract the number automatically from the street field into the house number field? What do you guys think?

OscarSpruit: We decided to remove autofill hints for this field as there is not much we can do

jreij (approved):

PR #1691, opened 2024-07-09
UPI - Show disclaimer when Continue is pressed without a selection

araratthehero:
## Description
Now a disclaimer is shown when user clicks "Continue" without making a selection for UPI intent.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-946

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,20 +8,5 @@
 [//]: # (## Deprecated)
 [//]: # ( - Configurations public constructor are deprecated, please use each Configuration's builder to make a Configuration object)
 
-## New
-- For Google Pay on Advanced flow, `onSubmit` now returns`threeDS2SdkVersion` in the `paymentMethod` object that you must pass in your [`/payments`](https://docs.adyen.com/api-explorer/Checkout/71/post/payments) request to correctly trigger the 3D Secure 2 flow.
-
-## Fixed
-- On Android API versions 21 to 25, the `NoSuchMethodError` no longer occurs during the 3D Secure 2 challenge flow.
-- When [using R8 to shrink your code](https://developer.android.com/build/shrink-code), `CIRCULAR REFERENCE: com.android.tools.r8.utils.b: Missing class...` errors no longer occur.
-
-## Changed
-- Dependency versions:
-  | Name                                                                                                         | Version                       |
-  |--------------------------------------------------------------------------------------------------------------|-------------------------------|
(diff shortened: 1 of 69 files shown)


RELEASE_NOTES.md
```diff
@@ -9,19 +9,4 @@
...
-  |--------------------------------------------------------------------------------------------------------------|-------------------------------|
-  | [Adyen 3DS2](https://github.com/Adyen/adyen-3ds2-android/releases/tag/2.2.19)                                | **2.2.19**                    |
-  | [Kotlin](https://github.com/JetBrains/kotlin/releases/tag/v1.9.24)                                           | **1.9.24**                    |
-  | [Android Gradle plugin](https://developer.android.com/build/releases/gradle-plugin)                          | **8.4.1**                     |
-  | [AndroidX Compose Compiler](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.14) | **1.5.14**                    |
-  | [Kotlin coroutines](https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.8.1)                         | **1.8.1**                     |
-  | [AndroidX Fragment](https://developer.android.com/jetpack/androidx/releases/fragment#1.7.1)                  | **1.7.1**                     |
+- For UPI Intent a disclaimer will be shown when "Continue" button is pressed without selecting any UPI option.
```
jreij, line 12: "disclaimer" seems a bit confusing to me, it's more of a validation error message, what do you think?

jreij, line 12: Not sure if this should be under "new", feels more like a fix/improvement rather than a new feature (same for the PR label)

araratthehero: We do not really have a good label for this. The ones that I am thinking of are `Enhancement` and `Fix`. But `Enhancement` indicates a new feature request and the `Fix` indicates a fix. This is more of an improvement of a previously implemented feature.

image

jreij: Right, `Enhancement` is for issues not PRs. I think `Fix` applies here even though it's not exactly a bug fix. Think about `Feature` as anything that includes new functionality (mostly adds new things to the public API but not always) and `Fix` as anything that doesn't directly affect integrations but improves the flow for merchants.

jreij: I agree that the naming might be confusing but we mostly use these 2 labels to know if we should make a minor release or a patch, they are not meant to be completely exhaustive to every use case.

jreij (approved):

PR #1697, opened 2024-07-09
Create twint-action module

OscarSpruit:
## Description
Create twint-action module and move existing classes there. In the twint module type aliases are added to make sure it's backwards compatible.

COAND-927


twint/src/main/java/com/adyen/checkout/twint/TwintActionComponent.kt
```diff
@@ -8,60 +8,10 @@
...
-
-        @JvmField
-        val PROVIDER = TwintActionComponentProvider()
-    }
-}
+@Deprecated(
+    "This class has been moved to a new package",
+    ReplaceWith("TwintActionComponent", "com.adyen.checkout.twint.action.TwintActionComponent"),
+)
+typealias TwintActionComponent = TwintActionComponent
```
araratthehero, line 17: Could you later add in release notes that the component and the configuration are deprecated?

PR #1700, opened 2024-07-10
Create Twint payment component

OscarSpruit:
## Description
Create a Twint payment component.
Next step is to populate the view, so shoppers can use a switch to store their Twint.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-927

### DIFF
--- a/twint/src/main/java/com/adyen/checkout/twint/TwintComponent.kt
+++ b/twint/src/main/java/com/adyen/checkout/twint/TwintComponent.kt
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2024 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 10/7/2024.
+ */
+
+package com.adyen.checkout.twint
+
+import androidx.lifecycle.LifecycleOwner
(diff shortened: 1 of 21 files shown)


twint/src/main/java/com/adyen/checkout/twint/TwintComponent.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.adyen.checkout.ui.core.internal.ui.ButtonDelegate
+import com.adyen.checkout.ui.core.internal.ui.ComponentViewType
+import com.adyen.checkout.ui.core.internal.ui.ViewableComponent
+import com.adyen.checkout.ui.core.internal.util.mergeViewFlows
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * A [PaymentComponent] that supports the [PaymentMethodTypes.TWINT] payment method.
+ */
+class TwintComponent internal constructor(
```
jreij, line 38: This one should also not allow skipping the list of payment methods right? We need to rethink about the implementation of that logic

OscarSpruit: Twint is already included in this logic, because it's in `SUPPORTED_ACTION_ONLY_PAYMENT_METHODS`

PR #1704, opened 2024-07-12
Clean up and update translations

araratthehero:
## Description
- The translations script is adjusted to always add a new line at the end of files, which would conflict with IDE settings before and some files would have a new line added by the IDE and some of them would not.
- The new line has been removed before untranslated texts, which would cause to have an empty line generated right before the `</resources>` closing tag.
- Translations are now in sync with other platforms

## Additionally
- Translations script has been updated to add an empty line at the end of each generated translation file. PR is opened in the translations repository.

## Checklist 
- [x] PR is labelled 

COAND-961


card/src/main/res/values/strings.xml
```diff
@@ -11,16 +11,16 @@
     <string name="checkout_card_holder_hint">Name on card</string>
     <string name="checkout_card_expiry_date_hint">Expiry date</string>
     <string name="checkout_card_expiry_date_optional_hint">Expiry date (optional)</string>
-    <string name="checkout_card_security_code_hint">CVC / CVV</string>
-    <string name="checkout_card_security_code_optional_hint">CVC / CVV (optional)</string>
+    <string name="checkout_card_security_code_hint">Security code</string>
```
jreij, line 14: There's quite a few changes in this PR, i think we need release notes especially for the more prominent changes like this one

araratthehero: I just added the list of updated string resources

OscarSpruit (approved):

PR #1715, opened 2024-07-19
Meal voucher - Create MealVoucherView

araratthehero:
## Description
Create meal voucher view

## Checklist 
- [x] PR is labelled 

COAND-683

### DIFF
--- a/ui-core/src/main/java/com/adyen/checkout/ui/core/internal/ui/view/SecurityCodeInput.kt
+++ b/ui-core/src/main/java/com/adyen/checkout/ui/core/internal/ui/view/SecurityCodeInput.kt
@@ -1,11 +1,11 @@
 /*
- * Copyright (c) 2022 Adyen N.V.
+ * Copyright (c) 2024 Adyen N.V.
  *
  * This file is open source and available under the MIT license. See the LICENSE file for more info.
  *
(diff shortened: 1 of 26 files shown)


ui-core/src/main/java/com/adyen/checkout/ui/core/internal/ui/view/SecurityCodeInput.kt
```diff
@@ -15,7 +15,7 @@ class SecurityCodeInput @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
     defStyleAttr: Int = 0
-) : CardNumberInput(context, attrs, defStyleAttr) {
+) : AdyenTextInputEditText(context, attrs, defStyleAttr) {
```
araratthehero, line 18: We figured that it was by mistake that `SecurityCodeInput` was extending `CardNumberInput`?

OscarSpruit: Yeah looks like it indeed. Could you remove `open` from `CardNumberInput` as well? So we can't accidentally make the same mistake

OscarSpruit (approved):

PR #1721, opened 2024-07-26
Fix skip to payment method list

OscarSpruit:
## Description
Separated `SUPPORTED_ACTION_ONLY_PAYMENT_METHODS` into two, one for drop-in skip to logic and one for `InstantPayComponent` supported payment methods.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-962

### DIFF
--- a/instant/src/main/java/com/adyen/checkout/instant/InstantPaymentComponent.kt
+++ b/instant/src/main/java/com/adyen/checkout/instant/InstantPaymentComponent.kt
@@ -6,6 +6,7 @@ import androidx.lifecycle.viewModelScope
 import com.adyen.checkout.action.core.internal.ActionHandlingComponent
 import com.adyen.checkout.action.core.internal.DefaultActionHandlingComponent
 import com.adyen.checkout.action.core.internal.ui.GenericActionDelegate
+import com.adyen.checkout.components.core.PaymentMethodTypes
 import com.adyen.checkout.components.core.internal.ComponentEventHandler
 import com.adyen.checkout.components.core.internal.PaymentComponent
 import com.adyen.checkout.components.core.internal.PaymentComponentEvent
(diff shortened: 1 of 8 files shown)


araratthehero (dismissed):

instant/src/main/java/com/adyen/checkout/instant/InstantPaymentComponent.kt
```diff
@@ -74,5 +75,21 @@ class InstantPaymentComponent internal constructor(
...
         @JvmField
         val PROVIDER = InstantPaymentComponentProvider()
+
+        /**
+         * These are the payment method types known to be supported by the [InstantPaymentComponent]. There are
+         * payment method types that are not listed here that are actually supported. See
+         * [InstantPaymentComponentProvider.isPaymentMethodSupported] for more details.
+         */
+        @JvmField
+        val PAYMENT_METHOD_TYPES = listOf(
```
jreij, line 85: Even with the comment explaining it I'm not sure about this list being public. From an external perspective I don't see a reason why it should be part of the public API since it has no use for someone else using the library

OscarSpruit: I do agree with you, but other components also have this field public and imo it could be private for those as well. I did this out of consistency

jreij: What if we put this list somewhere completely different? it doesn't have to have the same name or even be public

OscarSpruit: Sure, let me find a nice place

OscarSpruit: I like the idea of extracting this logic. Shall we merge this PR first, so we can work on it in your PR? Or you think it's better if I do it here?

araratthehero: Sure, let's merge this first. We can discuss the other approach later

araratthehero (approved):

PR #1722, opened 2024-07-29
UPI - Improve payment methods list

araratthehero:
## Description
- Removed UPI payment methods from unsupported list
- Move the Drop-in payment method filtering to `DropInViewModel`
- Filtering is done in `getPaymentMethods()` method, to apply it for features like `skipListWhenSinglePaymentMethod`

I changed the initial approach of creating a `PaymentMethodManager`, since it was redundant with the current Drop-in structure. It could be a better fit after we refactor Drop-in.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-965

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
@@ -133,6 +133,9 @@ object PaymentMethodTypes {
         SEPA,
         TWINT,
         UPI,
+        UPI_INTENT,
+        UPI_COLLECT,
+        UPI_QR,
         WECHAT_PAY_SDK,
     )
 
@@ -178,8 +181,5 @@ object PaymentMethodTypes {
         WECHAT_PAY_MINI_PROGRAM,
         WECHAT_PAY_QR,
         WECHAT_PAY_WEB,
-        UPI_INTENT,
-        UPI_COLLECT,
-        UPI_QR,
     )
 }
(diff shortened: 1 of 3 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/model/DropInPaymentMethodManager.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package com.adyen.checkout.dropin.internal.ui.model
+
+import com.adyen.checkout.components.core.PaymentMethodTypes
+import com.adyen.checkout.components.core.PaymentMethodTypes.UPI_COLLECT
+import com.adyen.checkout.components.core.PaymentMethodTypes.UPI_INTENT
+import com.adyen.checkout.components.core.PaymentMethodTypes.UPI_QR
+
+// TODO: Does it really make sense to have this here?
+internal object DropInPaymentMethodManager : PaymentMethodManager {
```
OscarSpruit, line 17: I think this should be a `class`

drop-in/src/test/java/com/adyen/checkout/dropin/internal/ui/model/TestPaymentMethodManager.kt
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright (c) 2024 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ararat on 25/7/2024.
+ */
+
+package com.adyen.checkout.dropin.internal.ui.model
+
+internal object TestPaymentMethodManager : PaymentMethodManager {
```
OscarSpruit, line 11: And this one should be a `class` as well

OscarSpruit (approved):

PR #1725, opened 2024-07-31
Meal Voucher - Create gift card protocol

ozgur00:
## Description
GiftCardProtocol is responsible for providing details for the differences in Gift Card and Meal Voucher flows such as ComponentViewType or the payment method object.

## Checklist 
- [x] PR is labelled 

COAND-693

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
@@ -39,6 +39,7 @@ object PaymentMethodTypes {
     const val MEAL_VOUCHER_FR_GROUPEUP = "mealVoucher_FR_groupeup"
     const val MEAL_VOUCHER_FR_NATIXIS = "mealVoucher_FR_natixis"
     const val MEAL_VOUCHER_FR_SODEXO = "mealVoucher_FR_sodexo"
+    const val MEAL_VOUCHER_FR = "mealVoucher_FR"
     const val MOLPAY_MALAYSIA = "molpay_ebanking_fpx_MY"
     const val MOLPAY_THAILAND = "molpay_ebanking_TH"
     const val MOLPAY_VIETNAM = "molpay_ebanking_VN"
--- a/components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/PaymentMethodDetails.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/PaymentMethodDetails.kt
@@ -65,7 +65,12 @@ abstract class PaymentMethodDetails : ModelObject() {
                 DotpayPaymentMethod.PAYMENT_METHOD_TYPE -> DotpayPaymentMethod.SERIALIZER
                 EPSPaymentMethod.PAYMENT_METHOD_TYPE -> EPSPaymentMethod.SERIALIZER
                 EntercashPaymentMethod.PAYMENT_METHOD_TYPE -> EntercashPaymentMethod.SERIALIZER
-                GiftCardPaymentMethod.PAYMENT_METHOD_TYPE -> GiftCardPaymentMethod.SERIALIZER
+                GiftCardPaymentMethod.PAYMENT_METHOD_TYPE,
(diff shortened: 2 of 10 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
```diff
@@ -36,6 +36,7 @@ object PaymentMethodTypes {
     const val GOOGLE_PAY_LEGACY = "paywithgoogle"
     const val IDEAL = "ideal"
     const val MB_WAY = "mbway"
+    const val MEAL_VOUCHER_FR = "mealVoucher_FR" // not a txVariant, only to be used for /payments call
```
jreij, line 39: In that case in there any benefit for merchants to have this as a public value?

ozgur00: Good point. Restricted the access to library group.

components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/PaymentMethodDetails.kt
```diff
@@ -65,7 +65,12 @@ abstract class PaymentMethodDetails : ModelObject() {
                 DotpayPaymentMethod.PAYMENT_METHOD_TYPE -> DotpayPaymentMethod.SERIALIZER
                 EPSPaymentMethod.PAYMENT_METHOD_TYPE -> EPSPaymentMethod.SERIALIZER
                 EntercashPaymentMethod.PAYMENT_METHOD_TYPE -> EntercashPaymentMethod.SERIALIZER
-                GiftCardPaymentMethod.PAYMENT_METHOD_TYPE -> GiftCardPaymentMethod.SERIALIZER
+                GiftCardPaymentMethod.PAYMENT_METHOD_TYPE,
+                PaymentMethodTypes.MEAL_VOUCHER_FR_GROUPEUP,
+                PaymentMethodTypes.MEAL_VOUCHER_FR_NATIXIS,
+                PaymentMethodTypes.MEAL_VOUCHER_FR_SODEXO,
+                PaymentMethodTypes.MEAL_VOUCHER_FR -> GiftCardPaymentMethod.SERIALIZER
```
jreij, line 72: Should we create a separate PaymentMethod object for meal vouchers, in case we need to change it from the `GiftCardPaymentMethod` later? (which would cause a breaking change)

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
```diff
@@ -94,6 +96,10 @@ object PaymentMethodTypes {
     const val WECHAT_PAY_QR = "wechatpayQR"
     const val WECHAT_PAY_WEB = "wechatpayWeb"
 
+    // Non-public payment method types only for internal use
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jreij, line 100: Not sure if this should be internal to the package itself since it doesn't fit in with this file but I'll leave it for you to decide

components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/PaymentMethodDetails.kt
ozgur00: For now we will go on with the current solution, then consider refactoring the flow.

components-core/src/main/java/com/adyen/checkout/components/core/PaymentMethodTypes.kt
ozgur00: Since this file is `public` it doesn't fit here, however it acts as a `txVariant` and used both while building the payment method object and getting the serializer. The `paymentMethod.type` field we get from `/paymentMethods` call is passed as the `brand` parameter, so we cannot use that. That's why among all the possible options this is the most meaningful place for this to be in.

jreij: 👍

jreij (approved):

PR #1760, opened 2024-08-21
Track rendered event on drop-in

OscarSpruit:
## Description
Track rendered event on drop-in with configData.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-967


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInConfigDataGenerator.kt
```diff
@@ -0,0 +1,21 @@
...
+package com.adyen.checkout.dropin.internal.ui
+
+import com.adyen.checkout.dropin.internal.ui.model.DropInParams
+
+internal class DropInConfigDataGenerator {
+
+    fun generate(configuration: DropInParams): Map<String, String> {
+        return mapOf(
+            "skipPaymentMethodList" to configuration.skipListWhenSinglePaymentMethod.toString(),
+            "openFirstStoredPaymentMethod" to configuration.showPreselectedStoredPaymentMethod.toString(),
```
araratthehero, line 18: I think we can also send `configuration.isRemovingStoredPaymentMethodsEnabled`?

OscarSpruit: The other platforms don't send this. Do you think we should add it anyway and align later?

araratthehero: Yes, I think it is best to send what we can at this stage. Later others can send that too when they add the functionality.

araratthehero (approved):

PR #1770, opened 2024-08-30
Update payment method headers

ozgur00:
## Description
Introduce a header on top bar. Change the translations for sections in the payment methods list.

## Checklist 
- [x] PR is labelled 

COAND-975


drop-in/src/main/res/values/strings.xml
```diff
@@ -12,15 +12,16 @@
     <string name="action_failed">There was an error while processing your payment. Please try again later.</string>
     <string name="time_out_error">Request timed out. Please try again.</string>
     <string name="unknown_error">An unknown error occurred</string>
-    <string name="store_payment_methods_header">Your payment methods</string>
-    <string name="other_payment_methods">Select other method</string>
+    <string name="store_payment_methods_header">Stored</string>
```
jreij, line 15: Let's name these strings something more specific to avoid conflicting with the merchants' own strings. Maybe prefix them with `checkout_` to follow the pattern?

drop-in/src/main/res/values/strings.xml
```diff
@@ -12,15 +12,16 @@
...
+    <string name="other_payment_methods">others</string>
```
jreij, line 16: Should this be `Others` with capital first letter?

drop-in/src/main/res/values/styles.xml
```diff
@@ -51,8 +51,9 @@
         <item name="android:paddingRight">@dimen/standard_margin</item>
         <item name="android:textAllCaps">true</item>
         <item name="android:textIsSelectable">false</item>
-        <item name="android:textSize">14sp</item>
+        <item name="android:textSize">12sp</item>
```
jreij, line 54: WDYT about deprecating this style and creating a new one? If anyone is currently overriding this style they will not get a nice result since these labels are now subtitles instead of titles

drop-in/src/main/res/values/strings.xml
ozgur00: It probably should, but it comes from translations like this.

drop-in/src/main/res/values/strings.xml
ozgur00: For this PR I only updated the relevant ones, for the remaining ones we can do them in a separate PR to keep this one focused on headers.

jreij: We can't update any string that's already released anyway or else it's a breaking change, but maybe we should make sure all the strings that meal vouchers (or related PRs) recently introduced are prefixed

jreij (approved):

jreij (changes requested):
The release notes file has some weird extra blank lines

ozgur00:
> The release notes file has some weird extra blank lines

Weird. Gotta be fixed now.

jreij (approved):

PR #1783, opened 2024-09-03
Settings screen improvements

jreij:
## Description
- Refactored settings screen to compose.
- Merged country and currency settings into one country selector setting.
- Added placeholders, hints and validations when editing settings.
- Rearranged settings and categories.

## Checklist 
- [x] PR is labelled 

COAND-860


example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
```diff
@@ -0,0 +1,267 @@
...
+            .padding(
+                start = ExampleTheme.dimensions.grid_2,
+                end = ExampleTheme.dimensions.grid_2,
+                top = ExampleTheme.dimensions.grid_2,
+            )
+            .fillMaxWidth(),
+    ) {
+        Text(
+            title,
+            style = MaterialTheme.typography.labelLarge,
```
OscarSpruit, line 166: Can we use a larger type here? To make the header and items more distinguishable

example-app/src/main/java/com/adyen/checkout/example/di/MapperModule.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.adyen.checkout.example.ui.settings.viewmodel.SettingsUIMapper
+import com.adyen.checkout.example.ui.theme.DefaultUIThemeRepository
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+object MapperModule {
```
OscarSpruit, line 23: You can remove this module. Dagger is smart enough to provide this classes automatically

example-app/src/main/java/com/adyen/checkout/example/di/ProviderModule.kt
```diff
@@ -0,0 +1,26 @@
...
+import android.app.Application
+import com.adyen.checkout.example.provider.LocaleProvider
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+object ProviderModule {
```
OscarSpruit, line 20: Same here, but you have to annotate the context in `LocaleProvider` with `@ApplicationContext`

example-app/src/main/java/com/adyen/checkout/example/di/StorageModule.kt
```diff
@@ -28,9 +31,19 @@ object StorageModule {
         PreferenceManager.getDefaultSharedPreferences(appContext)
 
     @Provides
-    fun provideKeyValueStorage(appContext: Application, sharedPreferences: SharedPreferences): KeyValueStorage =
-        DefaultKeyValueStorage(appContext, sharedPreferences)
+    fun provideSharedPreferencesManager(sharedPreferences: SharedPreferences): SharedPreferencesManager =
```
OscarSpruit, line 34: This class can be simplified to:
```Kotlin
@Module
@InstallIn(SingletonComponent::class)
internal abstract class StorageModule {

    @Binds
    abstract fun bindKeyValueStorage(defaultKeyValueStorage: DefaultKeyValueStorage): KeyValueStorage

    companion object {

        @Provides
        fun provideSharedPreferences(appContext: Application): SharedPreferences =
            PreferenceManager.getDefaultSharedPreferences(appContext)

        @Provides
        fun provideAssetManager(appContext: Application): AssetManager = appContext.assets
    }
}
```
You have to add `@Inject constructor` to `DefaultKeyValueStorage` though

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
```diff
@@ -0,0 +1,267 @@
...
+import com.adyen.checkout.example.ui.settings.viewmodel.SettingsViewModel
+import com.adyen.checkout.example.ui.theme.ExampleTheme
+
+@Composable
+internal fun SettingsScreen(
+    onBackPressed: () -> Unit,
+    viewModel: SettingsViewModel = hiltViewModel(),
+) {
+    Scaffold(
+        modifier = Modifier.windowInsetsPadding(WindowInsets.ime),
```
araratthehero, line 55: Is `windowInsetsPadding` necessary here or using `innerPadding` is already enough?

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
```diff
@@ -0,0 +1,267 @@
...
+import com.adyen.checkout.example.ui.settings.model.SettingsUIState
+import com.adyen.checkout.example.ui.settings.viewmodel.SettingsViewModel
+import com.adyen.checkout.example.ui.theme.ExampleTheme
+
+@Composable
+internal fun SettingsScreen(
+    onBackPressed: () -> Unit,
+    viewModel: SettingsViewModel = hiltViewModel(),
+) {
+    Scaffold(
```
araratthehero, line 54: I would move this to the second `SettingsScreen` composable and keep only viewModel related logic in this Composable, which acts like a bridge, between the viewModel and the main Composable functions.

It would look like this:
```
@Composable
internal fun SettingsScreen(
    onBackPressed: () -> Unit,
    viewModel: SettingsViewModel = hiltViewModel(),
) {
    val uiState by viewModel.uiState.collectAsState()
    
    SettingsScreen(
        uiState = uiState,
        onBackPressed = onBackPressed,
        onItemClicked = viewModel::onItemClicked,
        onEditSettingConsumed = viewModel::onEditSettingConsumed,
        onTextSettingChanged = viewModel::onTextSettingChanged,
        onListSettingChanged = viewModel::onListSettingChanged,
        onSwitchSettingChanged = viewModel::onSwitchSettingChanged,
    )
}
```

This way the `SettingsScreen()` will be a fully defined composable, which does not depend on the viewModel. Otherwise to reuse the composable for the entire screen, you have to use the `SettingsScreen()` which requires the `viewModel`.

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
```diff
@@ -0,0 +1,267 @@
...
+                onEditSettingConsumed()
+            },
+        )
+    }
+}
+
+@Composable
+private fun SettingsItemsList(
+    settingsCategories: List<SettingsCategory>,
+    onItemClicked: (SettingsItem) -> Unit,
```
araratthehero, line 119: `onItemClicked` suggests that in every item click this callback will be fired, but in the code it only happens for Text setting items. What do you think of changing the name to something like `onTextSettingClicked`?

example-app/src/main/java/com/adyen/checkout/example/ui/settings/viewmodel/SettingsViewModel.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    fun onListSettingChanged(identifier: SettingsIdentifier, selectedItem: EditSettingsData.SingleSelectList.Item) {
+        settingsEditor.editSetting(identifier, selectedItem)
+        fetchSettings()
+    }
+
+    fun onSwitchSettingChanged(identifier: SettingsIdentifier, newValue: Boolean) {
+        settingsEditor.editSetting(identifier, newValue)
+        fetchSettings()
+    }
```
araratthehero, line 68: What do you think of only leaving one `onSettingChanged()` function here, which would accept an `identifier` and `value` (with type `EditSettingsData`), which could be passed to the `settingsEditor`, which in turn would distinguish the differences between setting items.

This way if we add or remove new type of settings, we do not need to modify the `SettingsViewModel`.

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
```diff
@@ -0,0 +1,267 @@
...
+                onTextSettingChanged(uiState.settingToEdit.identifier, it)
+                onEditSettingConsumed()
+            },
+            onListSettingChanged = {
+                onListSettingChanged(uiState.settingToEdit.identifier, it)
+                onEditSettingConsumed()
+            },
+            onDismiss = {
+                onEditSettingConsumed()
+            },
```
araratthehero, line 111:
```suggestion
            onTextSettingChanged = {
                onTextSettingChanged(uiState.settingToEdit.identifier, it)
            },
            onListSettingChanged = {
                onListSettingChanged(uiState.settingToEdit.identifier, it)
            },
            onDismiss = {
                onEditSettingDismissed()
            },
```

What do you think of changing the `onEditSettingConsumed()` to have a more descriptive name, like `onEditSettingConsumed()` and only call it when the dialog is dismissed? Since calling `onTextSettingChanged()` or `onListSettingChanged()` means that the settings is changed and the dialog should hide right?

example-app/src/main/java/com/adyen/checkout/example/ui/compose/GenericDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+    onDismiss: () -> Unit,
+) {
+    Dialog(onDismiss) {
+        Surface(
+            modifier = modifier.fillMaxWidth(),
+            shape = MaterialTheme.shapes.medium,
+            shadowElevation = 1.dp,
+        ) {
+            Column {
+                Column(Modifier.padding(ExampleTheme.dimensions.grid_2)) {
```
araratthehero, line 48: I would avoid adding paddings in `GenericDialog`, since it does not create content, but gets it through arguments and each composables passed to the `GenericDialog` can define their own paddings.

It will also solve this pressed state background color not being fully visible.
Screenshot 2024-09-11 at 14 31 29 (2)

example-app/src/main/java/com/adyen/checkout/example/data/storage/SharedPreferencesManager.kt
```diff
@@ -0,0 +1,77 @@
...
+ * Created by josephj on 29/8/2024.
+ */
+
+package com.adyen.checkout.example.data.storage
+
+import android.content.SharedPreferences
+import androidx.core.content.edit
+
+class SharedPreferencesManager(
+    val sharedPreferences: SharedPreferences
```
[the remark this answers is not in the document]
jreij: Not really because `getEnum` is `inline` and uses it, but I can make both `internal`

example-app/src/main/java/com/adyen/checkout/example/ui/compose/GenericDialog.kt
```diff
@@ -0,0 +1,91 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.Dialog
+import com.adyen.checkout.example.ui.theme.ExampleTheme
+
+@Suppress("LongParameterList")
+@Composable
+fun GenericDialog(
```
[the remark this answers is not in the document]
jreij: We won't change this since it's not really an alert dialog but a regular dialog

example-app/src/main/java/com/adyen/checkout/example/data/storage/KeyValueStorage.kt
```diff
@@ -8,172 +8,189 @@
...
+import com.adyen.checkout.example.data.storage.SharedPreferencesEntry.SPLIT_CARD_FUNDING_SOURCES
+import com.adyen.checkout.example.data.storage.SharedPreferencesEntry.THREEDS_MODE
 
 @Suppress("TooManyFunctions")
 interface KeyValueStorage {
     fun getShopperReference(): String
+    fun setShopperReference(shopperReference: String?)
     fun getAmount(): Amount
+    fun setAmount(amount: Long?)
+    fun setCurrency(currency: String?)
```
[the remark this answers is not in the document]
jreij: Yes, in the settings screen we edit the numeric amount value in one place and the combination of the currency + country in another place. However, inside the app (when calling the SDK and the Checkout API), we fetch the country on its own and the combination of the numeric amount value and the currency inside one object which is `Amount`.

example-app/src/main/java/com/adyen/checkout/example/data/storage/SharedPreferencesEntry.kt
```diff
@@ -0,0 +1,80 @@
...
+ * Copyright (c) 2024 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 29/8/2024.
+ */
+
+package com.adyen.checkout.example.data.storage
+
+enum class SharedPreferencesEntry(val key: String, val defaultValue: Any?) {
```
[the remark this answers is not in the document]
jreij: My intention with separating this class from settings is that it can contain non settings related prefs as well. If you check its usages you can see that the settings related classes do not directly access it. WDYT?

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/EditSettingListFieldDialog.kt
```diff
@@ -0,0 +1,83 @@
...
+        },
+        content = {
+            LazyColumn {
+                items(settingToEdit.items) { item ->
+                    Text(
+                        modifier = Modifier
+                            .clickable { onConfirm(item) }
+                            .padding(vertical = ExampleTheme.dimensions.grid_2)
+                            .fillMaxWidth(),
+                        text = stringFromUIText(uiText = item.text),
```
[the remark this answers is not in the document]
jreij: Yes indeed

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
jreij: I just followed what we do in other screens, but I don't think we need it for this screen 👍

example-app/src/main/java/com/adyen/checkout/example/ui/settings/composable/SettingsScreen.kt
jreij: Nice!

example-app/src/main/java/com/adyen/checkout/example/data/storage/SharedPreferencesEntry.kt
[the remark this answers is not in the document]
araratthehero: I was thinking of keeping preferences separate for each screen/flow. But the list is not that long, so I think it is fine to keep it in one place.

example-app/src/main/java/com/adyen/checkout/example/ui/settings/viewmodel/SettingsViewModel.kt
jreij: Check out the last 2 commits and let me know what do you think

example-app/src/main/java/com/adyen/checkout/example/ui/compose/GenericDialog.kt
```diff
@@ -32,31 +32,28 @@ import com.adyen.checkout.example.ui.theme.ExampleTheme
...
+                if (title != null && content != null) {
+                    Spacer(Modifier.size(ExampleTheme.dimensions.grid_2))
                 }
+                content?.invoke()
                 if (dismissButton != null || confirmButton != null) {
                     Row(
-                        Modifier
-                            .padding(ExampleTheme.dimensions.grid_1)
-                            .fillMaxWidth(),
+                        Modifier.fillMaxWidth(),
```
araratthehero, line 56: The space between the buttons and the content has been removed

Screenshot 2024-09-12 at 09 21 00

example-app/src/main/java/com/adyen/checkout/example/ui/settings/viewmodel/SettingsViewModel.kt
araratthehero: Looks good! 👍

example-app/src/main/java/com/adyen/checkout/example/ui/compose/GenericDialog.kt
jreij: Indeed, fixed it.

araratthehero (approved):

PR #1786, opened 2024-09-06
Fix setInteractionBlocked blocking PaymentInProgressView cancel button

ozgur00:
## Description
When `setInteractionBlocked(true)` is called it used to block touch event for everything displayed in `AdyenComponentView`, including `PaymentInProgressView`. Now it only blocks the submit button when set.

## Checklist 
- [x] PR is labelled 

COAND-978


ozgur00:
Closing since the solution for the problem requires further discussion.

PR #1811, opened 2024-10-08
Expose card input field validation functionality to public

ozgur00:
## Description
Added public `CardNumberValidator`, `CardExpiryDateValidator` and `CardSecurityCodeValidator` classes to make our internal validation available for merchants.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-985


RELEASE_NOTES.md
```diff
@@ -15,6 +15,7 @@
   - Up. Payment method type: **mealVoucher_FR_groupeup**.
   - Natixis. Payment method type: **mealVoucher_FR_natixis**.
   - Sodexo. Payment method type: **mealVoucher_FR_sodexo**.
+- Added `CardNumberValidator`, `CardExpiryDateValidator` and `CardSecurityCodeValidator` classes to make validation functionality for the corresponding fields available to public.
```
jreij, line 18: Maybe let's mention at the start that this is for API only merchants with a custom card component?

card/api/card.api
```diff
@@ -161,34 +161,6 @@ public final class com/adyen/checkout/card/BuildConfig {
 	public fun <init> ()V
 }
 
-public final class com/adyen/checkout/card/CardBrand : android/os/Parcelable {
```
jreij, line 164: This is a breaking change, let's add a type alias and test that all public functionality is not affected?

card/api/card.api
```diff
@@ -297,51 +269,6 @@ public final class com/adyen/checkout/card/CardConfigurationKt {
 	public static synthetic fun card$default (Lcom/adyen/checkout/components/core/CheckoutConfiguration;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lcom/adyen/checkout/components/core/CheckoutConfiguration;
 }
 
-public final class com/adyen/checkout/card/CardType : java/lang/Enum {
```
jreij, line 300: Same here, breaking change

checkout-core/src/main/java/com/adyen/checkout/core/ui/model/ExpiryDate.kt
```diff
@@ -3,13 +3,10 @@
...
- * Created by ararat on 17/7/2024.
+ * Created by ozgur on 4/10/2024.
  */
-package com.adyen.checkout.ui.core.internal.ui.model
+package com.adyen.checkout.core.ui.model
 
-import androidx.annotation.RestrictTo
-
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class ExpiryDate(
```
jreij, line 10: Can we document this class to make it clear what kind of values are we expecting for `expiryMonth` and `expiryYear`?

jreij, line 10: Should `EMPTY_DATE` and `INVALID_DATE` be made internal/restricted? Maybe they could be even moved to an internal class so that merchants don't see/use them. I see they are used inside the validator but I don't think the public validator should depend on them

checkout-core/src/main/java/com/adyen/checkout/core/ui/validation/CardExpiryDateValidator.kt
```diff
@@ -3,23 +3,40 @@
...
+
+    /**
+     * Validate expiry date.
+     *
+     * @param expiryDate Expiry date.
+     * @param calendar Calendar instance to be used to compare given expiry date.
+     *
+     * @return Validation result.
+     */
     fun validateExpiryDate(
```
jreij, line 40: Is there a need for this to be public?

ozgur00: Made it internal.

checkout-core/src/main/java/com/adyen/checkout/core/ui/model/ExpiryDate.kt
ozgur00: Good one, updated the Validator

checkout-core/src/main/java/com/adyen/checkout/core/ui/validation/CardExpiryDateValidationResult.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Possible validation results for expiry date validation. (@see [CardExpiryDateValidator.validateExpiryDate]
+ */
+sealed interface CardExpiryDateValidationResult {
+    class Valid : CardExpiryDateValidationResult
+    interface Invalid : CardExpiryDateValidationResult {
+        class TooFarInTheFuture : Invalid
+        class TooOld : Invalid
+        class OtherReason : Invalid
```
jreij, line 19: This should only happen when the date is not in the correct bounds right? Can we rename it to indicate that?

checkout-core/src/main/java/com/adyen/checkout/core/ui/validation/CardExpiryDateValidator.kt
```diff
@@ -58,7 +68,7 @@ object ExpiryDateValidationUtils {
 
     private fun dateExists(expiryDate: ExpiryDate): Boolean {
         return (
-            expiryDate !== ExpiryDate.EMPTY_DATE &&
+            !expiryDate.isEmptyDate() &&
```
jreij, line 71: This condition can be removed

jreij (approved):

PR #1812, opened 2024-10-08
Analytics - Send different values when checkoutAttemptId is not available or it failed to fetch

araratthehero:
## Description

## Checklist 
- [ ] PR is labelled 
- [ ] Code is unit tested

COAND-982


components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/CheckoutAttemptIdState.kt
```diff
@@ -0,0 +1,18 @@
...
+ * Created by ararat on 8/10/2024.
+ */
+
+package com.adyen.checkout.components.core.internal.analytics
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+sealed class CheckoutAttemptIdState {
+    data class Available(val checkoutAttemptId: String) : CheckoutAttemptIdState()
```
OscarSpruit, line 15: If you put `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)` on all these sub classes, then they shouldn't show up in the .api file

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManager.kt
```diff
@@ -101,31 +105,23 @@ internal class DefaultAnalyticsManager(
...
-            analyticsRepository.sendEvents(checkoutAttemptId)
+            analyticsRepository.sendEvents(checkoutAttemptIdState.checkoutAttemptId)
         }.fold(
             onSuccess = { /* Not necessary */ },
             onFailure = { throwable -> adyenLog(AdyenLogLevel.WARN, throwable) { "Failed sending analytics events" } },
         )
     }
 
-    override fun getCheckoutAttemptId(): String? = checkoutAttemptId
+    override fun getCheckoutAttemptIdState() = checkoutAttemptIdState
```
OscarSpruit, line 122: I think the state should be something private to `DefaultAnalyticsManager` and then this method can stay as it was

components-core/src/main/java/com/adyen/checkout/components/core/internal/util/CheckoutAttemptIdMapper.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.adyen.checkout.components.core.internal.util
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.components.core.internal.analytics.AnalyticsManager
+import com.adyen.checkout.components.core.internal.analytics.CheckoutAttemptIdState
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+const val CHECKOUT_ATTEMPT_ID_NOT_FETCHED = "checkoutAttemptId-not-fetched"
```
OscarSpruit, line 16: All these fields and methods should be internal or even moved to `DefaultAnalyticsManager` IMO. These two `val`s are now used in tests, but I think if you make the change I suggested earlier and hardcode a value in `TestAnalyticsManager` then tests don't have to know about these

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/DefaultAnalyticsManager.kt
araratthehero: That was my initial implementation, but I changed it, since `checkoutAttemptId-not-fetched` and `fetch-checkoutAttemptId-failed` values are not really related to the analytics. They are only necessary to be sent in the `/payments` call, based on if the `checkoutAttemptId` is fetched or not. Those two values are not the actual values of the `checkoutAttemptId`, that is why I separated them, to make sure that the Analytics manager is not aware of them and if they change in the future, they will not change the logic of the analytics manager. They are relevant only for the `/payments` call.

Do you think there is a better way to handle this and keep them separate from the analytics manager?

OscarSpruit: Hmm I see your point, but I think it actually is related to analytics. Currently, the `getCheckoutAttemptId` method is only used by delegates for the `/payments` call. The only other place we use it is to make analytics requests, but we don't make analytics requests if we have no attempt id. Making the not fetched and failed values valid attempt ids imo

araratthehero: I was thinking of exposing a state and whenever necessary we can assign custom values to the `checkoutAttemptId`, since those are not values which are returned from backend. But I guess it is just thinking about cases which we do not have now.
I can move everything to the analytics manager and later we can make changes if necessary.

OscarSpruit: Yes, I think that's a good idea

OscarSpruit (approved):

PR #1815, opened 2024-10-09
Replace JSONObject's opt functions with nullable extensions

jreij:
## Description
- Replace JSONObject's opt functions (`optString`, `optBoolean`, `optInt`, `optLong` and `optDouble`) with nullable extensions. These functions are non nullable and return arbitrary default values which could cause unexpected bugs.
- Add a lint rule to prevent their use.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-1003

### DIFF
--- a/lint/src/main/java/com/adyen/checkout/lint/JSONOptFunctions.kt
+++ b/lint/src/main/java/com/adyen/checkout/lint/JSONOptFunctions.kt
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2024 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 14/8/2024.
+ */
+
+package com.adyen.checkout.lint
+
+import com.android.tools.lint.detector.api.Category
+import com.android.tools.lint.detector.api.Detector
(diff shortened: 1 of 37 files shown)


lint/src/main/java/com/adyen/checkout/lint/JSONOptFunctions.kt
```diff
@@ -0,0 +1,73 @@
...
+internal val JSON_OPT_FUNCTIONS_ISSUE = Issue.create(
+    id = "JSONOptFunctions",
+    briefDescription = "JSONObject \"opt\" functions should not be used directly",
+    explanation = """
+        JSONObject's optString, optBoolean, optInt, optLong and optDouble functions are non nullable and return
+        arbitrary default values which could cause unexpected bugs. Use an internal extension function that returns
+        null instead.
+    """.trimIndent().replace(Regex("(\n*)\n"), "$1"),
+    implementation = Implementation(JSONOptFunctionsDetector::class.java, Scope.JAVA_FILE_SCOPE),
+    category = Category.I18N,
```
OscarSpruit, line 30: `CUSTOM_LINT_CHECKS` is a better category I think

araratthehero (approved):

PR #1819, opened 2024-10-11
Clarify code docs for analytics level none

jreij:
## Description
With the change in #1801 we should clarify what does  `AnalyticsLevel.NONE` mean.

## Checklist 
- [x] PR is labelled 

COAND-980

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/AnalyticsConfiguration.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/AnalyticsConfiguration.kt
@@ -34,7 +34,7 @@ enum class AnalyticsLevel {
     ALL,
 
     /**
-     * No analytics are sent from the library.
+     * Only Drop-in/Components analytics are not sent from the library.
      */
     NONE,
 }


ozgur00 (approved):

components-core/src/main/java/com/adyen/checkout/components/core/AnalyticsConfiguration.kt
```diff
@@ -34,7 +34,7 @@ enum class AnalyticsLevel {
     ALL,
 
     /**
-     * No analytics are sent from the library.
+     * Only Drop-in/Components analytics are not sent from the library.
```
araratthehero, line 37: Wouldn't it be more clear, and match with the comment for `ALL` to explain here what we actually send, instead of what we do not send?
Something like `Crucial SDK information is sent from the library, which does not include any information about user behavior`.

jreij: This was done in collaboration with docs, [this section](https://docs.adyen.com/online-payments/analytics-and-data-tracking/#data-we-are-collecting) explains everything in detail, we will link to it in the release notes but we can't add and maintain all this info in the code docs

araratthehero: I see now what `Drop-in/Components analytics` refers to. Without reading the docs it is unclear.

What do you think of adding the link to the docs in the code docs, to make it easier for merchants to access that page?

PR #1820, opened 2024-10-11
Google Pay button styling

OscarSpruit:
## Description
Make it possible to style the Google Pay button both through code and XML. Where code has priority.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-942

### DIFF
--- a/example-app/src/main/res/values-night/styles.xml
+++ b/example-app/src/main/res/values-night/styles.xml
@@ -8,7 +8,7 @@
 
 <resources>
 
-    <style name="GooglePayButton">
-        <item name="buttonTheme">light</item>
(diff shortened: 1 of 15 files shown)


example-app/src/main/res/values-night/styles.xml
```diff
@@ -8,7 +8,7 @@
 
 <resources>
 
-    <style name="GooglePayButton">
-        <item name="buttonTheme">light</item>
+    <style name="AdyenCheckout.GooglePay.Button">
```
jreij, line 11: Do we/merchants still need to do this manually?

OscarSpruit: We now automatically do this

PR #1830, opened 2024-10-15
Analytics - Add support for error events

araratthehero:
## Description
Add support for analytics error events

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-979


components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/DefaultAnalyticsRepository.kt
```diff
@@ -41,17 +43,20 @@ internal class DefaultAnalyticsRepository(
     ) {
         val infoEvents = localInfoDataStore.fetchEvents(remoteDataStore.infoSize)
         val logEvents = localLogDataStore.fetchEvents(remoteDataStore.logSize)
+        val errorEvents = localErrorDataStore.fetchEvents(remoteDataStore.errorSize)
 
-        if (infoEvents.isEmpty() && logEvents.isEmpty()) return
+        if (infoEvents.isEmpty() && logEvents.isEmpty() && errorEvents.isEmpty()) return
```
ozgur00, line 48: nitpick: we can put this condition either to a private function or a `val` to make the code a bit more readable

araratthehero: I left it here to be right below the declarations of the events, to not forget to update the condition if it changes. When I was implementing the error events I missed to update the condition and noticed it later.

Let me know if you still think we should create a private function for it.

araratthehero: In the case of creating a val for it, it will look like this, which imo is not so different from what we have now?

```
val areEventsEmpty = infoEvents.isEmpty() && logEvents.isEmpty() && errorEvents.isEmpty()
if (areEventsEmpty) return
```

ozgur00: Maybe a function would be better then to be able to hide the implementation details, but in any case three conditions tied together can be simplified and here we would only have
```kotlin
if (!haveEventsToTrack) return //!shouldTrackEvents, noEventsToTrack kind of names
```
and that would make this function slightly simpler. 

I understand your reasoning about not forgetting it to update it too though. I'm okay with both. But our unit tests should catch if we forget to update this part, we can improve that part.

components-core/src/test/java/com/adyen/checkout/components/core/internal/analytics/data/DefaultAnalyticsRepositoryTest.kt
```diff
@@ -128,4 +151,48 @@ internal class DefaultAnalyticsRepositoryTest(
             verify(localErrorDataStore, never()).removeEvents(any())
         }
     }
+
+    companion object {
+
+        @OptIn(DirectAnalyticsEventCreation::class)
+        @JvmStatic
+        fun sendEventsSource() = listOf(
```
ozgur00, line 159: I think we should cover the case where all three of the lists are empty and we don't send any event. What do you think?

araratthehero: We already have a test called `there are no events stored, then sending is canceled` for it at `DefaultAnalyticsRepositoryTest:90`.

ozgur00 (approved):

PR #1842, opened 2024-10-18
Introduce DispatcherProvider

ozgur00:
## Description
`DispatcherProvider` allows us to set the `CoroutineDispatcher` we're going to use hence it is now possible to use a custom dispatcher in our ui tests. 

## Checklist 
- [x] PR is labelled 

### DIFF
--- a/checkout-core/src/main/java/com/adyen/checkout/core/DispatcherProvider.kt
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/DispatcherProvider.kt
@@ -1,0 +1,57 @@
+/*
+ * Copyright (c) 2024 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 16/10/2024.
+ */
+
+package com.adyen.checkout.core
+
+import androidx.annotation.RestrictTo
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.MainCoroutineDispatcher
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object DispatcherProvider {
+
+    private var mainFactory = { Dispatchers.Main }
+    val Main: MainCoroutineDispatcher get() = mainFactory()
+
+    private var defaultFactory = { Dispatchers.Default }
+    val Default: CoroutineDispatcher get() = defaultFactory()
+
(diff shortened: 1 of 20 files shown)


checkout-core/src/main/java/com/adyen/checkout/core/DispatcherProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package com.adyen.checkout.core
+
+import androidx.annotation.RestrictTo
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object DispatcherProvider {
```
OscarSpruit, line 16: Nice, this concept is probably the best we can do right now! 

I think we could improve it a little bit by making these `var`s lazy. This will make sure the dispatchers are not initialised on the app's startup, possibly increasing startup times. Something like this:
```Kotlin
private var mainFactory = { Dispatchers.Main }
val Main: MainCoroutineDispatcher get() = mainFactory()

private var defaultFactory = { Dispatchers.Default }
val Default: CoroutineDispatcher get() = defaultFactory()

private var ioFactory = { Dispatchers.IO }
val IO: CoroutineDispatcher get() = ioFactory()

fun overrideMain(dispatcher: MainCoroutineDispatcher) {
    mainFactory = { dispatcher }
}

fun resetMain() {
    mainFactory = { Dispatchers.Main }
}

...
```

example-app/src/androidTest/java/com/adyen/checkout/test/rule/IdlingDispatcherRule.kt
```diff
@@ -45,6 +46,7 @@ class IdlingDispatcherRule : TestRule {
                     unregister(defaultIdlingResource)
                     unregister(ioIdlingResource)
                 }
+                DispatcherProvider.resetAll()
                 overrideDispatchers(defaultDispatcher, ioDispatcher)
```
OscarSpruit, line 50: This line can be removed as we no longer have to manually reset the dispatchers

checkout-core/src/main/java/com/adyen/checkout/core/DispatcherProvider.kt
ozgur00: I like it. done!

OscarSpruit (approved):

PR #1814, opened 2024-10-09
Fetch version name with gradle task

jreij:
## Description
Fetch version name with gradle task instead of shell script.

## Checklist 
- [x] PR is labelled 
- [ ] Changes are tested manually

COAND-832

PR #1847, opened 2024-10-21
Release - New GitHub workflow for releasing

araratthehero:
## Description
- Allow running the job on a specified branch instead of main
- Use new Github secrets for publishing the app on maven central
- Make the Github release draft

## Things to do in the future
- Add a step to verify if the branch name starts with e.g. `release/`
- Separate each step into a different job

## Checklist 
- [x] PR is labelled 

COAND-1013

### DIFF
--- a/.github/workflows/publish_release_old.yml
+++ b/.github/workflows/publish_release_old.yml
@@ -1,0 +1,81 @@
+name: Publish Release - Old
+
+# Every time we merge to main branch we publish a release.
+on:
+  push:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+
+  publish-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 17
+
+      - name: Gradle cache
+        uses: gradle/actions/setup-gradle@v3
+        with:
+          cache-read-only: true
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
+      - name: Gradle check
+        run: ./gradlew check
+
+      # TODO: add more tests or rely on check_release workflow?
+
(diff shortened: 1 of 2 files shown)


jreij:
PR base should not be `main`, the current release process requires `develop` to be up to date with `main` so we cannot merge anything directly to `main` before we are fully ready to remove the old release process

.github/workflows/publish_release_manual.yml
```diff
@@ -0,0 +1,81 @@
...
+
+jobs:
+  
+  publish-release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: main
```
araratthehero, line 19: This needs to change. It should be the branch on which we triggered the action.

jreij (changes requested):
Once we merge the feature branch we will lose the old request action, can we instead keep it and name it something like `old`?

.github/workflows/publish_release_old.yml
```diff
@@ -0,0 +1,81 @@
...
+          SIGNING_SECRET_KEY_RING_FILE: ${{ secrets.SIGNING_SECRET_KEY_RING_FILE }}
+          SONATYPE_STAGING_PROFILE_ID: ${{ secrets.SONATYPE_STAGING_PROFILE_ID }}
+
+      # Get the version name from a script and save to environment variable.
+      - name: Set PROJECT_VERSION
+        run: |
+          echo "▸ Set run permission."
+          chmod +x scripts/version_name.sh
+          echo "▸ Getting version name"
+          PROJECT_VERSION=$(./scripts/version_name.sh)
```
araratthehero, line 64: This need to be updated when [this PR](https://github.com/Adyen/adyen-android/pull/1814) is merged.

araratthehero: Similarly the `publish_release.yml` as well

araratthehero: We do not have to wait for that PR anymore. The changes made [here](https://github.com/Adyen/adyen-android/pull/1848) are enough.

jreij (approved):

PR #1852, opened 2024-10-29
Release - Break release job apart

araratthehero:
## Description
Break release job apart into multiple jobs.
[Here](https://github.com/Adyen/adyen-android/actions/runs/11572437317) is a workflow run which demonstrated how it looks.

## Checklist 
- [x] PR is labelled 

COAND-1013

### DIFF
--- a/.github/workflows/publish_to_maven_central.yml
+++ b/.github/workflows/publish_to_maven_central.yml
@@ -16,11 +16,6 @@ jobs:
           distribution: 'zulu'
           java-version: 17
 
-      - name: Gradle cache
-        uses: gradle/actions/setup-gradle@v3
-        with:
-          cache-read-only: true
-
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
 
--- a/.github/workflows/publish_to_maven_central.yml
+++ b/.github/workflows/publish_to_maven_central.yml
@@ -1,0 +1,45 @@
+name: Publish to Maven Central
+
+on:
+  workflow_call
+
+jobs:
+  publish_to_maven_central:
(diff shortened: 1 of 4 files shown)


.github/workflows/publish_to_maven_central.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'zulu'
+          java-version: 17
+
+      - name: Gradle cache
```
OscarSpruit, line 19: I know we had it before, but should we remove caching for releases?

araratthehero: I think the cache can help us execute gradle tasks faster. I did a small experiment with it.

I just ran different workflows with and without cache.
[Workflow with cache](https://github.com/Adyen/adyen-android/actions/runs/11591406960): It looks like it saves us 3.5 minutes on average.
[Workflow without cache](https://github.com/Adyen/adyen-android/actions/runs/11590456730): The cache has been removed from `check` and `publish_to_maven_central` jobs. The first thing I see is that it downloads the gradle, which IMO takes the extra time. 
image

OscarSpruit: Yes I understand it takes more time, but if the cache is ever corrupt we might publish a build that is broken

araratthehero: Hmm, I see your point. I will remove it then so we are safe.

OscarSpruit (approved):

PR #1855, opened 2024-10-31
Google Pay availability check

OscarSpruit:
## Description
Check Google Pay availability when initialising the component. If unavailable, the button will be hidden.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-943

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/ComponentAvailableCallback.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/ComponentAvailableCallback.kt
@@ -7,6 +7,6 @@
  */
 package com.adyen.checkout.components.core
 
-interface ComponentAvailableCallback {
+fun interface ComponentAvailableCallback {
     fun onAvailabilityResult(isAvailable: Boolean, paymentMethod: PaymentMethod)
 }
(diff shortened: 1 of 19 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/ComponentAvailableCallback.kt
```diff
@@ -7,6 +7,6 @@
  */
 package com.adyen.checkout.components.core
 
-interface ComponentAvailableCallback {
+fun interface ComponentAvailableCallback {
```
jreij, line 10: Is this needed?

OscarSpruit: Not really, but it's a nice little improvement:
```Kotlin
object : ComponentAvailableCallback {
    override fun onAvailabilityResult(isAvailable: Boolean, paymentMethod: PaymentMethod) {
        updateOutputData(isButtonVisible = isAvailable)

        if (!isAvailable) {
            exceptionChannel.trySend(GooglePayUnavailableException())
        }
    }
}
```
vs
```
{ isAvailable, paymentMethod ->
    updateOutputData(isButtonVisible = isAvailable)

    if (!isAvailable) {
        exceptionChannel.trySend(GooglePayUnavailableException())
    }
}
```
Only downside is if we would ever add a method to the interface then we have to remove the fun and that would break integration if merchants use the lambda way

jreij: As long as it's not a breaking change currently, I think it's a good idea. I don't see a reason to add a new function there in v5.

OscarSpruit: It's not breaking, it's possible to use both ways now

jreij (approved):

PR #1856, opened 2024-10-31
Analytics - Level field

araratthehero:
## Description
Add level field to the initial call to determine if the analytics is disabled or not

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-1029

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/remote/DefaultAnalyticsSetupProvider.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/remote/DefaultAnalyticsSetupProvider.kt
@@ -14,12 +14,15 @@ import com.adyen.checkout.components.core.Amount
 import com.adyen.checkout.components.core.internal.analytics.AnalyticsPlatformParams
 import com.adyen.checkout.components.core.internal.analytics.AnalyticsSource
 import com.adyen.checkout.components.core.internal.data.model.AnalyticsSetupRequest
+import com.adyen.checkout.components.core.internal.ui.model.AnalyticsParamsLevel
 import java.util.Locale
 
+@Suppress("LongParameterList")
 internal class DefaultAnalyticsSetupProvider(
     private val application: Application,
     private val shopperLocale: Locale,
     private val isCreatedByDropIn: Boolean,
+    private val analyticsLevel: AnalyticsParamsLevel,
     private val amount: Amount?,
     private val source: AnalyticsSource,
     private val sessionId: String?,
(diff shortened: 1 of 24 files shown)


OscarSpruit:
Looks good to me! Is the backend ready as well?

araratthehero:
It has been merged and will be available on Test anytime soon. I will wait before merging.

OscarSpruit (approved):

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/remote/DefaultAnalyticsSetupProvider.kt
```diff
@@ -57,8 +61,16 @@ internal class DefaultAnalyticsSetupProvider(
         is AnalyticsSource.PaymentComponent -> source.paymentMethodType
     }
 
+    private fun getLevelQueryParameter(analyticsParamsLevel: AnalyticsParamsLevel) = when (analyticsParamsLevel) {
```
ozgur00, line 64: What do you think about making this a field of `AnalyticsParamLevel` enum? Then we wouldn't have to map it and just access it like `analyticsParamsLevel.queryParam`. Although it would increase coupling of this code and `AnalyticsParamLevel`, do you see a problem with that?

components-core/src/main/java/com/adyen/checkout/components/core/internal/ui/model/AnalyticsParams.kt
```diff
@@ -26,14 +26,14 @@ data class AnalyticsParams(
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 enum class AnalyticsParamsLevel(val priority: Int) {
-    NONE(1),
+    INITIAL(1),
     ALL(2),
 }
 
 private fun getLevel(analyticsConfiguration: AnalyticsConfiguration?): AnalyticsParamsLevel {
```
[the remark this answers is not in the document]
araratthehero: In case the `analyticsConfiguration` is not passed, we want to default the value to `ALL`.
For the when statement, it is already exhaustive,  since `null`  case is different than `else` case. If we add a new value, we will have to implement it as well.

components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/data/remote/DefaultAnalyticsSetupProvider.kt
araratthehero: I added the mapping in `DefaultAnalyticsSetupProvider`, since it is purely part of the analytics setup call, and here we define what needs to be sent as part of the initial request based on the passed level. On the other hand `AnalyticsParamsLevel` is used in Analytics manager to decide what events to send.

I prefer keeping the mapping in the `DefaultAnalyticsSetupProvider` and keeping the `AnalyticsParamsLevel` as is. It will make it easier to adjust or change the initial analytics call in the future without touching `AnalyticsParamsLevel`.

Let me know what you think.

ozgur00 (approved):

PR #1872, opened 2024-11-01
Release - Publish release and delete branch

araratthehero:
## Description
Publish the draft release for a specific tag, create the tag and delete the branch.

## Checklist 
- [x] PR is labelled 

COAND-1015

### DIFF
--- a/.github/workflows/finalize_release.yml
+++ b/.github/workflows/finalize_release.yml
@@ -37,8 +37,10 @@ jobs:
       - name: Publish Github release
         uses: ncipollo/release-action@v1
         env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           VERSION_NAME: ${{ needs.generate_version_name.outputs.version_name }}
         with:
+          token: ${{ env.GITHUB_TOKEN }}
           tag: ${{ env.VERSION_NAME }}
           allowUpdates: true
           updateOnlyUnreleased: true
--- a/.github/workflows/finalize_release.yml
+++ b/.github/workflows/finalize_release.yml
@@ -5,8 +5,30 @@ on:
   workflow_dispatch:
 
 jobs:
+  generate_version_name:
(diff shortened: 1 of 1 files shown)


.github/workflows/finalize_release.yml
```diff
@@ -0,0 +1,63 @@
...
+      - uses: actions/checkout@v4
+
+      # Publish Github release and the tag
+      - name: Publish Github release
+        uses: ncipollo/release-action@v1
+        env:
+          VERSION_NAME: ${{ needs.generate_version_name.outputs.version_name }}
+        with:
+          tag: ${{ env.VERSION_NAME }}
+          allowUpdates: true
```
OscarSpruit, line 43: What does this actually mean? Would publishing the same version name override the previous release?

araratthehero: Indeed. If `allowUpdates` is true, then the release gets updated if it already exists, based on `tag`. I also set the `updateOnlyUnreleased` flag to true, which will only update a release if it is not published yet.

OscarSpruit: Ahh I see, that makes sense

.github/workflows/finalize_release.yml
```diff
@@ -0,0 +1,65 @@
...
+
+  remove_release_branch:
+    runs-on: ubuntu-latest
+    needs: publish_draft_release
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Remove release branch
+        uses: dawidd6/action-delete-branch@v3
```
OscarSpruit, line 59: Not sure what does action all does, but wouldn't it be easier to use plain git?

araratthehero: This action allows deleting multiple branches and has some filtering options, but in our case is indeed not necessary.

OscarSpruit (approved):

PR #1873, opened 2024-11-01
Pay by Bank US - UI

ozgur00:
## Description
Implement View, ViewProvider, layout and relevant ui classes.

## Checklist 
- [x] PR is labelled 

COAND-974

### DIFF
--- a/paybybank-us/src/main/java/com/adyen/checkout/paybybankus/internal/DefaultPayByBankUSDelegate.kt
+++ b/paybybank-us/src/main/java/com/adyen/checkout/paybybankus/internal/DefaultPayByBankUSDelegate.kt
@@ -22,11 +22,17 @@ import com.adyen.checkout.components.core.paymentmethod.PayByBankUSPaymentMethod
 import com.adyen.checkout.core.AdyenLogLevel
 import com.adyen.checkout.core.internal.util.adyenLog
 import com.adyen.checkout.paybybankus.PayByBankUSComponentState
+import com.adyen.checkout.paybybankus.R
+import com.adyen.checkout.paybybankus.internal.ui.model.PayByBankUSBrandLogo
+import com.adyen.checkout.paybybankus.internal.ui.model.PayByBankUSOutputData
 import com.adyen.checkout.ui.core.internal.ui.ButtonComponentViewType
 import com.adyen.checkout.ui.core.internal.ui.ComponentViewType
 import com.adyen.checkout.ui.core.internal.ui.PaymentComponentUIEvent
 import com.adyen.checkout.ui.core.internal.ui.PaymentComponentUIState
 import com.adyen.checkout.ui.core.internal.ui.SubmitHandler
+import com.adyen.checkout.ui.core.internal.ui.model.LogoTextItem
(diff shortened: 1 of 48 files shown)


paybybank-us/src/main/java/com/adyen/checkout/paybybankus/internal/DefaultPayByBankUSDelegate.kt
```diff
@@ -125,4 +129,14 @@ internal class DefaultPayByBankUSDelegate(
...
         analyticsManager.clear(this)
     }
+
+    override fun getBrandList(): List<LogoTextItem> {
+        return listOf(
+            LogoItem(PayByBankUSBrandLogo.US_1.path, componentParams.environment),
+            LogoItem(PayByBankUSBrandLogo.US_2.path, componentParams.environment),
+            LogoItem(PayByBankUSBrandLogo.US_3.path, componentParams.environment),
+            LogoItem(PayByBankUSBrandLogo.US_4.path, componentParams.environment),
+            TextItem(R.string.checkout_pay_by_bank_us_more),
```
jreij, line 139: Consider using an icon for this instead of text, if a merchant is using a custom font it might make this look ugly/different

paybybank-us/src/main/java/com/adyen/checkout/paybybankus/internal/PayByBankUSDelegate.kt
```diff
@@ -24,4 +25,6 @@ internal interface PayByBankUSDelegate :
     val componentStateFlow: Flow<PayByBankUSComponentState>
 
     fun setInteractionBlocked(isInteractionBlocked: Boolean)
+
+    fun getBrandList(): List<LogoTextItem>
```
jreij, line 29: Can we use output data for this instead?

ui-core/src/main/java/com/adyen/checkout/ui/core/internal/ui/view/LogoTextAdapter.kt
```diff
@@ -0,0 +1,100 @@
...
+    internal class TextViewHolder(
+        private val binding: TextViewHolderBinding,
+        private val localizedContext: Context
+    ) : RecyclerView.ViewHolder(binding.root) {
+        fun bind(item: TextItem) {
+            binding.textView.text = localizedContext.getString(item.textResId)
+        }
+    }
+
+    object LogoTextItemDiffCallback : DiffUtil.ItemCallback<LogoTextItem>() {
```
jreij, line 82: This can be annotated with restrict to hide it from the public api file

paybybank-us/src/main/java/com/adyen/checkout/paybybankus/internal/DefaultPayByBankUSDelegate.kt
ozgur00: I am not so sure about this. I get your point but I think leaving it as text allows merchants to use some other text such as `and more` or `+` instead of `+ more`. I prefer it staying this way unless you are strongly against it.

jreij (approved):

PR #1884, opened 2024-11-08
5.7.1 release preview

jreij:
## Description
This PR serves to preview the `5.7.1` changes before release. The base branch is a mirror of the `5.7.0` tag.


ozgur00:
The last commit (with message `Fix address lookup view not retaining correct input state`) is applied twice.

PR #1895, opened 2024-11-19
Fix loading state overlapping error dialog in drop-in

OscarSpruit:
## Description
1. Override this in the drop in service:
```Kotlin
override fun onBalanceCheck(paymentComponentState: PaymentComponentState<*>) {
    val result = handleBalanceResponse(null)
    sendBalanceResult(result)
}
```
2. Open drop-in
3. Do a gift card payment
4. Aaaand you're now stuck on a loading state with an error dialog behind it...

The issue is fixed by using `showNow()` when showing the loading state fragment. An extra improvement is to always dismiss the loading state when we get back a result from the merchant. Before we only dismissed the loading state in some specific places.

## Checklist 
- [x] PR is labelled 

COAND-984

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
@@ -277,8 +277,6 @@ internal class DropInActivity :
     private fun errorDialogDismissed(reason: String, terminateDropIn: Boolean) {
         if (terminateDropIn) {
             terminateWithError(reason)
-        } else {
-            setLoading(false)
         }
     }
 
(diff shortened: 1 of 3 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
```diff
@@ -627,7 +625,7 @@ internal class DropInActivity :
         val loadingDialog = getFragmentByTag(LOADING_FRAGMENT_TAG)
         if (showLoading) {
             if (loadingDialog == null && !supportFragmentManager.isDestroyed) {
-                LoadingDialogFragment.newInstance().show(supportFragmentManager, LOADING_FRAGMENT_TAG)
+                LoadingDialogFragment.newInstance().showNow(supportFragmentManager, LOADING_FRAGMENT_TAG)
```
jreij, line 628: Any potential issues from calling `showNow`?

OscarSpruit: Afaik no. `show()` is asynchronous, waiting for the main thread to be ready before the dialog is created and show. This will not guarantee the order of dialogs shown.

`showNow()` immediately creates and shows the fragment, guaranteeing the order in which dialogs are created.

PR #1903, opened 2024-11-26
Pay by Bank US - Stored delegate

ozgur00:
## Description
Support stored pay by bank us delegate.

## Checklist 
- [x] PR is labelled 

COAND-974


drop-in/src/main/java/com/adyen/checkout/dropin/internal/util/StoredUtils.kt
```diff
@@ -54,6 +54,17 @@ internal fun StoredPaymentMethod.mapStoredModel(
             )
         }
 
+        PaymentMethodTypes.PAY_BY_BANK_US -> {
+            GenericStoredModel(
+                id = id.orEmpty(),
+                imageId = type.orEmpty(),
+                isRemovable = isRemovingEnabled,
+                name = label.orEmpty(),
```
jreij, line 62: Is this the only payment method where we will use the `label`?

ozgur00: At this point yes, I talked to backend and we could move all the display logic in the future to backend and use this for all other payment methods.

jreij (approved):

PR #1904, opened 2024-11-27
Analytics - Track submit event after validation

araratthehero:
## Description
We were tracking the submit event regardless if the component data was valid or not, but this PR fixes it. Here we track events when there are values emitted in the `submitFlow`, which happens when the `onSubmit()` is called and the component is valid.

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-1038

### DIFF
--- a/ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
+++ b/ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
@@ -104,7 +104,7 @@ internal class DefaultACHDirectDebitDelegate(
     private val _viewFlow: MutableStateFlow<ComponentViewType?> = MutableStateFlow(ACHDirectDebitComponentViewType)
     override val viewFlow: Flow<ComponentViewType?> = _viewFlow
 
-    override val submitFlow: Flow<ACHDirectDebitComponentState> = submitHandler.submitFlow
+    override val submitFlow: Flow<ACHDirectDebitComponentState> = getTrackedSubmitFlow()
     override val uiStateFlow: Flow<PaymentComponentUIState> = submitHandler.uiStateFlow
     override val uiEventFlow: Flow<PaymentComponentUIEvent> = submitHandler.uiEventFlow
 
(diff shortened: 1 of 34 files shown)


ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
```diff
@@ -104,7 +104,7 @@ internal class DefaultACHDirectDebitDelegate(
     private val _viewFlow: MutableStateFlow<ComponentViewType?> = MutableStateFlow(ACHDirectDebitComponentViewType)
     override val viewFlow: Flow<ComponentViewType?> = _viewFlow
 
-    override val submitFlow: Flow<ACHDirectDebitComponentState> = submitHandler.submitFlow
+    override val submitFlow: Flow<ACHDirectDebitComponentState> = getTrackedSubmitFlow()
```
jreij, line 107: Do we need to assign `override val submitFlow` to `submitHandler.submitFlow.onEach...`? Can we just keep `override val submitFlow = submitHandler.submitFlow` and call `onEach` in the `init` block or `initialize` function?
```kotlin
init { // or override fun initialize(coroutineScope: CoroutineScope) {
    submitFlow.onEach {
        val event = GenericEvents.submit(paymentMethod.type.orEmpty())
        analyticsManager.trackEvent(event)
    }
}
```

araratthehero: That was my initial idea, but it doesn't work that way, since `onEach` transforms the upstream flow to apply a side effect and the transformed flow (downsteam flow) should be collected for `onEach` to work.

jreij: Oh makes sense. But what if we collect it as well?

```kotlin
override fun initialize(coroutineScope: CoroutineScope) {
    submitFlow.onEach {
        val event = GenericEvents.submit(paymentMethod.type.orEmpty())
        analyticsManager.trackEvent(event)
    }
    .launchIn(coroutineScope)
}
```

araratthehero: It still won't work, since the flow is cold and once the value is collected, it will be emitted and then it will not be stored. So the second collector (`PaymentObserverRepository`) will not receive it the first time, but it will receive when there is a second emission (second click to the submit button). For every emission, only one collection will happen. And since the `PaymentObserverRepository` collects the flow after the delegate, it will receive every second emission.

We could solve this issue if we use a shared flow, but then we need to support buffering and cacheing to make sure the values are not lost, but still it is easier to lose values when collection happens and we can't make sure that every emission will be collected. For our current use case, I think having a cold flow is the right solution, which is easier to maintain, is less error prone and will guarantee the callback invocation.

jreij: Ah yes I completely missed that, let's keep it this way for now 👍

jreij (approved):

PR #1923, opened 2024-12-04
Analytics - Track 3ds2 error events

araratthehero:
## Description
Track error events for the 3ds2 component

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-1010


components-core/src/main/java/com/adyen/checkout/components/core/internal/analytics/GenericEvents.kt
```diff
@@ -129,10 +129,12 @@ object GenericEvents {
         component: String,
         event: ErrorEvent,
         target: String? = null,
+        message: String? = null,
     ) = AnalyticsEvent.Error(
         component = component,
         errorType = event.errorType,
         code = event.errorCode,
         target = target,
+        message = message,
```
ozgur00, line 138: I assume we should update previously added error events that don't have a `message` passed to them. Or does it only apply to 3ds2 error events?

araratthehero: I added a new ticket to go through the previously implemented events and add a message if necessary

ozgur00 (approved):

PR #1924, opened 2024-12-04
Fix crash when opening instant pm in drop-in

OscarSpruit:
## Description
Make sure the loading fragment transaction is completed before trying to dismiss it. This fixes a crash that happened when opening instant payment methods in drop-in.

## Checklist 
- [x] PR is labelled 

COAND-984


ozgur00:
Can we add release notes for this fix?

jreij (approved):

PR #1930, opened 2024-12-16
Release - Add update release notes script

ozgur00:
## Description
Add updating release notes for github release after merging the release notes changes pull request to `release-notes` branch. 

1. It requires release notes file to be placed in project root direct directory with the following name pattern: `<VERSION_NAME>.md // ex. 1.2.3.md`
2. It requires the branch for updating release notes to be named in the following format: `release-notes/<version_name>`
3. It fetches the version name from the head commit, which we expect it to be the merge commit from a branch that is named `release-notes/<version_name>`. (`fetch_version_name` step)
4. It fetches the file path to be used for `update_release_notes` step. (`fetch_updated_release_notes_file` step)
5. It updates the existing github release using the version name and release notes file path provided from the previous steps. (`update_release_notes` step)

## Checklist 
- [x] PR is labelled 

The flow is tested, except the PR merge flow which triggers a push on the base branch. That will be tested later with other workflows.

COAND-1034

### DIFF
--- a/.github/workflows/update_release_notes.yml
+++ b/.github/workflows/update_release_notes.yml
@@ -1,0 +1,74 @@
+name: Update release notes
+
+on:
+  push:
+    branches:
+      - 'release-notes'
+
+jobs:
+  fetch_version_name:
+    name: Fetch Version Name
+    needs: create_github_release
+    runs-on: ubuntu-latest
+    outputs:
+      version_name_release_notes: ${{ steps.fetch_version_name.outputs.version_name_release_notes }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Fetch version name for updating release notes
+        id: fetch_version_name
+        run: |
+          chmod +x scripts/fetch_release_notes_version_name.sh
+          commit_message="${{ github.event.head_commit.message }}"
+          VERSION_NAME=$(./scripts/fetch_release_notes_version_name.sh "${commit_message// /}")
+          echo -e "version_name_release_notes=$VERSION_NAME" >> $GITHUB_OUTPUT
+          echo -e $VERSION_NAME
+
+  fetch_updated_release_notes_file:
+    name: Fetch Updated Release Notes File
+    needs: fetch_version_name
+    runs-on: ubuntu-latest
+    outputs:
+      updated_release_notes_filepath: ${{ steps.fetch_updated_release_notes_file.outputs.updated_release_notes_filepath }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Fetch filepath for updated release notes file
+        id: fetch_updated_release_notes_file
+        env:
+          PROJECT_ROOT: ${{ github.workspace }}
+          VERSION_NAME: ${{ needs.fetch_version_name.outputs.version_name_release_notes }}
+        run: |
+          chmod +x scripts/fetch_updated_release_notes_file.sh
+          UPDATED_RELEASE_NOTES_FILEPATH=$(./scripts/fetch_updated_release_notes_file.sh $PROJECT_ROOT $VERSION_NAME)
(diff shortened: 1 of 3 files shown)


.github/workflows/update_release_notes.yml
```diff
@@ -0,0 +1,68 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION_NAME: ${{ needs.fetch_version_name_release_notes.outputs.version_name_release_notes }}
+        with:
+          token: ${{ env.GITHUB_TOKEN }}
+          tag: ${{ env.VERSION_NAME }}
+          allowUpdates: true
+          updateOnlyUnreleased: true
+          omitNameDuringUpdate: true
+          draft: true
+          bodyFile: ${{ needs.fetch_updated_release_notes_file.outputs.updated_release_notes_filepath }}
```
araratthehero, line 68: I would move this to the `env` list, like it is done for the `VERSION_NAME` and use it here as `${{ env.RELEASE_NOTES_FILE }}`

.github/workflows/update_release_notes.yml
```diff
@@ -0,0 +1,68 @@
+name: Update release notes
+
+on:
+  push:
+    branches:
+      - 'release-notes'
+
+jobs:
+  fetch_version_name_release_notes:
```
araratthehero, line 9: What do you think of changing the name to just `fetch_version_name` and omit `release_notes` to keep it more simple?

9.9.9.md
araratthehero, line 1: This file could be removed

scripts/fetch_release_notes_version_name.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+#
+# Copyright (c) 2024 Adyen N.V.
+#
+# This file is open source and available under the MIT license. See the LICENSE file for more info.
+#
+# Created by ozgur on 10/12/2024.
+#
+function fetch_release_notes_version_name() {
```
araratthehero, line 9: Does this script fetch the version name from the first file found that has a name matching the regex?

I am thinking of two cases:
- What if there are more than one release notes files (since all release notes will be available on the `release-notes` branch)
- What if we want to do a hotfix, and the version name is not the highest.

scripts/fetch_updated_release_notes_file.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+#
+# Copyright (c) 2024 Adyen N.V.
+#
+# This file is open source and available under the MIT license. See the LICENSE file for more info.
+#
+# Created by ozgur on 10/12/2024.
+#
+function fetch_updated_release_notes_file() {
```
araratthehero, line 9: We already fetch the version name in `fetch_release_notes_version_name.sh` file. I would either pass the version name to this script or construct the file name in the action itself.

scripts/fetch_release_notes_version_name.sh
ozgur00: I update the script to get the name from the commit name, however that turned out to be trickier than expected since a commit message has space characters and bash splits all the words into separate arguments when it's passed to a script. As a workaround I trim all the whitespace and then pass it to the script, seems like it's working. What do you think? 🤔

araratthehero: I think it would be best to not alter the commit message before processing, so we could work on the actual commit message in the script. Instead of passing it in the arguments, I would directly read the latest commit message inside the script, or add the commit message in the `env` list and use it in the script.

ozgur00: Problem is when we pass the commit message as an argument to a bash command, it automatically splits the string into words and the script takes them as separate arguments which makes it difficult to parse the commit message as a whole and get the version name. That's why I think for now it's best to leave it as is, since the other alternative to it is looking at the changed files, which is also not 100% reliable.

araratthehero (approved):

PR #1936, opened 2024-12-23
Analytics - Track api errors

araratthehero:
## Description
- Track possible API errors

## Checklist 
- [x] PR is labelled 
- [x] Code is unit tested

COAND-1007

### DIFF
--- a/ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
+++ b/ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
@@ -201,6 +201,10 @@ internal class DefaultACHDirectDebitDelegate(
                 },
                 onFailure = { e ->
                     adyenLog(AdyenLogLevel.ERROR) { "Unable to fetch public key" }
+
+                    val event = GenericEvents.error(paymentMethod.type.orEmpty(), ErrorEvent.API_PUBLIC_KEY)
+                    analyticsManager.trackEvent(event)
+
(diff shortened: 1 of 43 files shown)


ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
```diff
@@ -201,6 +201,10 @@ internal class DefaultACHDirectDebitDelegate(
                 },
                 onFailure = { e ->
                     adyenLog(AdyenLogLevel.ERROR) { "Unable to fetch public key" }
+
+                    val event = GenericEvents.error(paymentMethod.type.orEmpty(), ErrorEvent.API_PUBLIC_KEY)
```
jreij, line 205: This logic could be added to new components in the future, have we thought of a way to make sure we don't forget to track events that need to be tracked?

jreij (approved):
Only left one comment but it doesn't have to be addressed in this PR

ach/src/main/java/com/adyen/checkout/ach/internal/ui/DefaultACHDirectDebitDelegate.kt
araratthehero: We did not think about it yet. The aim was to implement all necessary events now. We could create a guide on creating new components in the future, which could contain a step about analytics events too?

jreij: Yeah that should be enough, let's not overcomplicate it. We'll see if this becomes a problem later.

PR #1937, opened 2024-12-23
Release - Make uploading to maven central run manually

araratthehero:
## Description
Allow the workflow to upload the build to maven central to also be run manually

## Checklist 
- [x] PR is labelled 

COAND-1051

### DIFF
--- a/.github/workflows/publish_to_maven_central.yml
+++ b/.github/workflows/publish_to_maven_central.yml
@@ -1,6 +1,13 @@
 name: Publish to Maven Central
 
 on:
+  workflow_dispatch:
+    inputs:
+      version-name:
+        description: "Version name of the build"
+        required: true
+        type: string
+
   workflow_call:
     inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish_to_maven_central.yml
```diff
@@ -1,6 +1,13 @@
 name: Publish to Maven Central
 
 on:
+  workflow_dispatch:
+    inputs:
+      version-name:
```
jreij, line 6: Out of curiosity, how would you pass this input to the job in a manual run?

jreij (approved):

.github/workflows/publish_to_maven_central.yml
araratthehero: It looks similar to this, with one field.

image

jreij: Nice!

PR #1953, opened 2025-01-03
Release - Clean up and remove develop mentions

araratthehero:
## Description
- Changed the `publish_release` to `start_release` to keep the naming consistent
- Removed redundant workflows
- Change `develop` mentions with `main`

## Checklist 
- [x] PR is labelled 

COAND-1019


.github/workflows/check_develop.yml
jreij, line 1: Shouldn't this be changed to check main instead?

.github/workflows/check_labels.yml
```diff
@@ -4,7 +4,6 @@ name: Check Labels
 on:
   pull_request:
     branches-ignore:
-      - 'main'
       - 'release/**'
```
jreij, line 7: This here ignores all PRs targetting branches named `release/**`, but AFAIK we won't have such PRs right?

araratthehero: Good point, this can be removed

.github/workflows/check_develop.yml
araratthehero: Since `check_pr` already does these checks and we do not permit direct pushes to the `main` branch, I decided to remove. Do you think we should still have it?

jreij: I get the point but we also have the same setup currently, `check_pr` already does these checks and we do not permit direct pushes to the `develop` branch. So i'm thinking that there's another reason why we have this job. Can we investigate a bit further?

.github/workflows/start_release.yml
```diff
@@ -1,4 +1,4 @@
-name: Publish Release
+name: Start Release
```
jreij, line 1: Nit: can we rename this to make it clearer that this doesn't actually release anything? For example `Draft/Prepare/Stage release`

jreij (approved):

PR #1975, opened 2025-01-24
 State management changes - MBway

araratthehero:
## Description
- Move state from view to delegate side for MBWay
- Create a state manager which handles state related operations
- Create registries, which can be implemented for concrete payment methods

## Checklist 
- [x] Code is unit tested

COAND-935


mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/DefaultMBWayDelegate.kt
```diff
@@ -45,36 +57,69 @@ internal class DefaultMBWayDelegate(
...
 ) : MBWayDelegate {
 
-    private val inputData = MBWayInputData()
-
-    private val _outputDataFlow = MutableStateFlow(createOutputData())
-    override val outputDataFlow: Flow<MBWayOutputData> = _outputDataFlow
-
-    private val _componentStateFlow = MutableStateFlow(createComponentState())
-    override val componentStateFlow: Flow<MBWayComponentState> = _componentStateFlow
+    private var state = MutableStateFlow(
```
OscarSpruit, line 64: This should be a `val` I think

OscarSpruit (commented):
It looks good, but a bit complicated. Maybe you can walk us through it soon?

ozgur00 (approved):

PR #1999, opened 2025-02-10
Fix Twint overriding store payment method mode

OscarSpruit:
## Description
This fixes an issue where the Twint component would override the `storePaymentMethodMode = enabled` for sessions making it impossible to store Twint.

## Checklist 
- [x] Code is unit tested

COAND-1067

### Fixed
- For Twint with sessions flow, the payment method will now be stored correctly when setting `storePaymentMethodMode` to `enabled`

### DIFF
--- a/twint/src/main/java/com/adyen/checkout/twint/internal/ui/DefaultTwintDelegate.kt
+++ b/twint/src/main/java/com/adyen/checkout/twint/internal/ui/DefaultTwintDelegate.kt
@@ -130,7 +130,7 @@ internal class DefaultTwintDelegate(
             paymentMethod = paymentMethod,
             order = order,
             amount = componentParams.amount,
-            storePaymentMethod = shouldStorePaymentMethod(),
+            storePaymentMethod = if (componentParams.showStorePaymentField) outputData.isStorePaymentSelected else null,
         )
 
         return TwintComponentState(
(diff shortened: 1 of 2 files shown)


twint/src/main/java/com/adyen/checkout/twint/internal/ui/DefaultTwintDelegate.kt
```diff
@@ -130,7 +130,7 @@ internal class DefaultTwintDelegate(
             paymentMethod = paymentMethod,
             order = order,
             amount = componentParams.amount,
-            storePaymentMethod = shouldStorePaymentMethod(),
+            storePaymentMethod = if (componentParams.showStorePaymentField) outputData.isStorePaymentSelected else null,
```
araratthehero, line 133: Does this also need to be fixed for other payment methods or it only happened for Twint?

twint/src/main/java/com/adyen/checkout/twint/internal/ui/DefaultTwintDelegate.kt
jreij, line 1: Release notes:
- I believe the field being overridden is `storePaymentMethod` not `storePaymentMethodMode` right?
- Can you add a short explanation of the impact of the SDK wrongly overriding this value?

twint/src/main/java/com/adyen/checkout/twint/internal/ui/DefaultTwintDelegate.kt
OscarSpruit: Seems to be only Twint

jreij (dismissed):

PR #2001, opened 2025-02-11
Emit specific exception when Google Pay is cancelled

OscarSpruit:
## Description
Emit specific exception when Google Pay is cancelled by a user. This makes it easier for merchant to distinguish between an actual error and a cancellation.

## Checklist 
- [x] Code is unit tested

COAND-1068

### Changed
- For Google Pay, when a user cancels the flow you will now get a `GooglePayCancellationException` in `onError`

### DIFF
--- a/googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/DefaultGooglePayDelegate.kt
+++ b/googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/DefaultGooglePayDelegate.kt
@@ -263,7 +263,7 @@ internal class DefaultGooglePayDelegate(
             }
 
             Activity.RESULT_CANCELED -> {
-                exceptionChannel.trySend(ComponentException("Payment canceled."))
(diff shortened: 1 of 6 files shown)


googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/DefaultGooglePayDelegate.kt
```diff
@@ -225,7 +226,7 @@ internal class DefaultGooglePayDelegate(
 
             CommonStatusCodes.CANCELED -> {
                 adyenLog(AdyenLogLevel.INFO) { "GooglePay payment canceled" }
-                exceptionChannel.trySend(ComponentException("GooglePay payment canceled"))
+                exceptionChannel.trySend(GooglePayCancellationException("GooglePay payment canceled"))
```
ozgur00, line 229: Should we also send this exception on line 266?

jreij (approved):

ozgur00 (approved):

PR #2004, opened 2025-02-12
Fix address lookup formatting

OscarSpruit:
## Description
In both the card view and search results of address lookup the addresses are now nicely formatted in two lines.

## Checklist 
- [x] Code is unit tested

COAND-1069

### Changed
- For address lookup, addresses are now formatted onto two lines

### DIFF
--- a/ui-core/src/test/java/com/adyen/checkout/ui/core/internal/ui/model/AddressOutputDataTest.kt
+++ b/ui-core/src/test/java/com/adyen/checkout/ui/core/internal/ui/model/AddressOutputDataTest.kt
@@ -16,7 +16,7 @@ import org.junit.jupiter.api.Test
 internal class AddressOutputDataTest {
 
     @Test
-    fun addressOutputDataFormatted() {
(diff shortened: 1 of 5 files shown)


ui-core/src/test/java/com/adyen/checkout/ui/core/internal/ui/model/AddressOutputDataTest.kt
```diff
@@ -13,9 +13,10 @@ import com.adyen.checkout.components.core.internal.ui.model.Validation
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Test
 
-class AddressOutputDataTest {
+internal class AddressOutputDataTest {
+
     @Test
-    fun addressOutputDataToString() {
+    fun addressOutputDataFormatted() {
```
jreij, line 19: Minor: can you use a more descriptive test name?

jreij (approved):

PR #2006, opened 2025-02-13
[Card scanning] Setup scanner

OscarSpruit:
## Description
Create and implement card scanner class. Testing can and will be done in a later stage.

COAND-429

### DIFF
--- a/card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
+++ b/card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
@@ -23,8 +23,8 @@ import com.google.android.gms.wallet.PaymentCardRecognitionResult
 import com.google.android.gms.wallet.PaymentsClient
 import com.google.android.gms.wallet.Wallet
 import com.google.android.gms.wallet.WalletConstants
+import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlin.coroutines.resume
-import kotlin.coroutines.suspendCoroutine
 
 class AdyenCardScanner {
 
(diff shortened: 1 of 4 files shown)


card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
```diff
@@ -8,4 +8,97 @@
...
+                    result.creditCardExpirationDate?.month,
+                    result.creditCardExpirationDate?.year,
+                )
+            }
+    }
+
+    fun terminate() {
+        paymentsClient = null
+        isAvailable = false
+        paymentCardRecognitionPendingIntent = null
```
ozgur00, line 102: Not sure if it's necessary, but we can call
```kotlin
paymentCardRecognitionPendingIntent?.cancel()
```
before nullifying it just to be safer. Wdyt?

card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
```diff
@@ -8,4 +8,97 @@
...
+    suspend fun isAvailable(): Boolean {
+        val paymentsClient = paymentsClient ?: error("initialize must be called before checking availability")
+
+        return suspendCoroutine { continuation ->
+            val request = PaymentCardRecognitionIntentRequest.getDefaultInstance()
+            paymentsClient
+                .getPaymentCardRecognitionIntent(request)
+                .addOnSuccessListener { intentResponse ->
+                    paymentCardRecognitionPendingIntent = intentResponse.paymentCardRecognitionPendingIntent
+                    continuation.resume(true)
```
araratthehero, line 58: By the time the listener is called, the coroutine can be cancelled already and it might cause unexpected behavior.

You could use `suspendCancellableCoroutine` and resume the continuation only when `continuation.isActive` is true.
```
if (continuation.isActive) {
    continuation.resume(true)
}
```

We could also make sure that the task is cancelled when the coroutine got cancelled:
```
continuation.invokeOnCancellation {
    paymentsClientTask.cancel()
}
```

What do you think?

card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
OscarSpruit: Oh nice! I wasn't aware this could be cancelled

ozgur00 (approved):

araratthehero (approved):

PR #2007, opened 2025-02-13
PayTo - Add PayId type adapter

araratthehero:
## Description
Add a picker for the PayId type and hide and show views based on chosen option

## Checklist 

COAND-1058


payto/src/main/java/com/adyen/checkout/payto/internal/ui/DefaultPayToDelegate.kt
```diff
@@ -173,4 +178,13 @@ internal class DefaultPayToDelegate(
         removeObserver()
         analyticsManager.clear(this)
     }
+
+    companion object {
+        private val SUPPORTED_PAY_ID_TYPES = listOf(
+            PayIdTypeModel(PayIdType.MOBILE, R.string.checkout_payto_payid_type_phone_number),
+            PayIdTypeModel(PayIdType.EMAIL, R.string.checkout_payto_payid_type_email),
+            PayIdTypeModel(PayIdType.ABN, R.string.checkout_payto_payid_type_abn),
```
OscarSpruit, line 186: What does ABN mean? If it's an abbreviation, maybe we can use the full form, so that people without knowledge of the payment method can understand better 🤔

araratthehero: I will check with the team again to see if there is an alternative name we can use, but I know that that is the only option that is not gonna get localized. So we will keep using ABN for all languages.

payto/src/main/java/com/adyen/checkout/payto/internal/ui/view/PayIdTypeAdapter.kt
```diff
@@ -0,0 +1,98 @@
...
+import android.view.View
+import android.view.ViewGroup
+import android.widget.BaseAdapter
+import android.widget.Filter
+import android.widget.Filterable
+import androidx.recyclerview.widget.RecyclerView
+import com.adyen.checkout.payto.databinding.PayIdTypeViewBinding
+import com.adyen.checkout.payto.internal.ui.model.PayIdTypeModel
+
+internal class PayIdTypeAdapter(
```
ozgur00, line 22: I'm thinking if we can use `SimpleTextItemAdapter`, since we only show a text here. However we would need to refactor it a bit to support `stringResId`s. What do you think?

araratthehero: Indeed I created a new adapter, since the `SimpleTextListAdapter` did not support localization and resIds.
I will check how to add the resIds and localization to the `SimpleTextItemAdapter`. If it makes the adapter more complex, we could probably create another adapter called `SimpleTextListLocalizedAdapter` and support localization?

araratthehero: As discussed I created a `LocalizedTextListAdapter` adapter which supports resIds and localization.

payto/src/main/java/com/adyen/checkout/payto/internal/ui/DefaultPayToDelegate.kt
araratthehero: I checked and ABN is indeed an abbreviation (Australian Business Number), but we will keep it as ABN, since it is known to people who use PayTo.

ozgur00 (approved):

PR #2009, opened 2025-02-18
[Card scanning] Implementation

OscarSpruit:
## Description
The card component now contains a fragment with a button that can handle the activity result from the scanner.
The scanner is wrapped, so that if the `:card-scanning` module is not present the button won't show and scanning is not possible.

The button is not according to design yet and will be replaced later.

https://github.com/user-attachments/assets/69a5cc07-7abe-47a4-9191-c2ad4cfdbc26

## Checklist 

COAND-429


card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
```diff
@@ -79,6 +77,29 @@ class AdyenCardScanner {
...
+                0,
+                0,
+                null,
+            )
+        }
+    }
+
+    private fun startScanner(startIntentSender: (PendingIntent) -> Unit) {
+        val paymentCardRecognitionPendingIntent =
+            paymentCardRecognitionPendingIntent ?: error("isAvailable must be called before starting the scanner")
```
araratthehero, line 99: Imo the name `isAvailable()` is a bit confusing here, since the name suggests that it returns an availability flag. But in our case it does an availability check and also storing of `paymentCardRecognitionPendingIntent`. Is there a way to separate those two actions or perhaps change the name of the function?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -128,6 +135,7 @@ class CardView @JvmOverloads constructor(
         initPostalCodeInput()
         initAddressFormInput(coroutineScope)
         initAddressLookup()
+        initFragment(delegate)
```
araratthehero, line 138: Instead of calling it a fragment, what do you think of giving it a more descriptive name, such as `OptionalFeatureFragment`, `FeatureContainerFragment` or `DynamicContentFragment`?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -63,18 +63,25 @@ import com.adyen.checkout.ui.core.R as UICoreR
 @Suppress("TooManyFunctions", "LargeClass")
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 class CardView @JvmOverloads constructor(
-    context: Context,
+    layoutInflater: LayoutInflater,
```
araratthehero, line 66: Why is the benefit of providing the `layoutInflater` here?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardFragment.kt
```diff
@@ -0,0 +1,92 @@
...
+        super.onActivityResult(requestCode, resultCode, data)
+        if (requestCode != REQUEST_CODE) return
+
+        val result = cardScanner?.getResult(data)
+
+        result?.let {
+            delegate?.updateInputData {
+                val (pan, expiryMonth, expiryYear) = result
+                cardNumber = pan.orEmpty()
+                if (expiryMonth != null && expiryYear != null) {
```
araratthehero, line 73: This will always be true right?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
OscarSpruit: This is actually required for the `FragmentContainerView` to work. By using `context` it will use the wrong context to create a layout inflater and will crash at runtime

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardFragment.kt
OscarSpruit: The Google SDK could possibly fail to recognise fields and pass `null` to us

araratthehero: But we have the following `let` check and it will never be `null`?
```
result?.let {
    ...
    val (pan, expiryMonth, expiryYear) = result
    if (expiryMonth != null && expiryYear != null) {
        ...
    }
    ...
}
```

card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
OscarSpruit: I just combined the `initialize` and `isAvailable` methods into one. I think it makes more sense now. What do you think?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardFragment.kt
OscarSpruit: The `let` checks if the `result` object is null, but then the fields inside this object can also be `null`. In case a user cancels the flow for example the `result` will be `null`. When the Google SDK fails to recognise the expiry date for example, then this field in the result will be `null`

araratthehero: Ah indeed, I just noticed. Thanks for the explanation!

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardFragment.kt
```diff
@@ -0,0 +1,92 @@
...
+import androidx.core.view.isVisible
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.lifecycleScope
+import com.adyen.checkout.card.databinding.FragmentCardBinding
+import com.adyen.checkout.card.internal.ui.CardDelegate
+import com.adyen.checkout.card.internal.util.CardScannerWrapper
+import com.adyen.checkout.core.ui.model.ExpiryDate
+import kotlinx.coroutines.launch
+
+internal class CardFragment : Fragment() {
```
araratthehero, line 25: What do you think of calling this `CardFeatureFragment` too?

OscarSpruit: Because we need a new design for the button, this will most likely change a bit. Is it okay if I rename it in that PR?

araratthehero: Sure

PR #2018, opened 2025-02-20
Add codeQL script

araratthehero:
## Description
Added codeQL script, to override the default configuration. It will run every time there is a push on main and every Sunday (kept the default trigger).

[Here](https://github.com/Adyen/adyen-android/actions/runs/13431791320/job/37525106947) is an example of the workflow run.

## Checklist 

COAND-1070

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,42 @@
+name: CodeQL Analysis
+
+on:
+  push:
+    branches: [ "main" ]
+  schedule:
+    - cron: '29 6 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze (${{ matrix.language }})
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,42 @@
+name: CodeQL Analysis
+
+on:
+  push:
+    branches: [ "main" ]
+  schedule:
+    - cron: '29 6 * * 0'
```
OscarSpruit, line 7: Is there a specific reason to run it every Sunday at 6:29?

araratthehero: Not really. It was there by default, I did not remove it. Do you think we should remove it?

OscarSpruit (approved):

PR #2019, opened 2025-02-20
PayTo - Field validation

araratthehero:
## Description
Add validation for PayTo fields

## Checklist 
- [x] Code is unit tested

COAND-1058

### DIFF
--- a/payto/src/main/java/com/adyen/checkout/payto/internal/ui/model/PayToOutputData.kt
+++ b/payto/src/main/java/com/adyen/checkout/payto/internal/ui/model/PayToOutputData.kt
@@ -11,6 +11,9 @@ package com.adyen.checkout.payto.internal.ui.model
 import com.adyen.checkout.components.core.internal.ui.model.FieldState
 import com.adyen.checkout.components.core.internal.ui.model.OutputData
 import com.adyen.checkout.components.core.internal.ui.model.Validation
+import com.adyen.checkout.components.core.internal.util.ValidationUtils
+import com.adyen.checkout.payto.R
+import com.adyen.checkout.payto.internal.util.PayToValidationUtils
 
 @Suppress("LongParameterList")
 internal class PayToOutputData(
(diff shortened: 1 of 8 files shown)


payto/src/main/java/com/adyen/checkout/payto/internal/ui/model/PayToOutputData.kt
```diff
@@ -33,45 +35,91 @@ internal class PayToOutputData(
...
+                bsbStateBranch,
+                Validation.Invalid(R.string.checkout_payto_bsb_state_branch_invalid),
+            )
+        }
 
-    private fun validateFirstName(firstName: String): FieldState<String> = FieldState(
-        firstName,
-        Validation.Valid
-    )
+    private fun validateFirstName(firstName: String): FieldState<String> = if (firstName.isNotEmpty()) {
```
OscarSpruit, line 108: Should we check for `isNotBlank()` instead maybe?

payto/src/main/java/com/adyen/checkout/payto/internal/util/PayToValidationUtils.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.regex.Pattern
+
+internal object PayToValidationUtils {
+
+    private const val PHONE_NUMBER_REGEX = "^\\+[0-9]{1,3}-[1-9]{1,1}[0-9]{1,29}\$"
+    private val PHONE_NUMBER_PATTERN = Pattern.compile(PHONE_NUMBER_REGEX)
+
+    @Suppress("ktlint:standard:max-line-length", "MaxLineLength")
+    private const val EMAIL_ADDRESS_REGEX =
```
OscarSpruit, line 19: In `ValidationUtils` we also have regexes for phone numbers and emails. Would it be possible to reuse those?

araratthehero: The email regex has been reused, but we should keep the phone number regex.

OscarSpruit (approved):

PR #2025, opened 2025-02-26
PayTo - Component state

araratthehero:
## Description
Generate component state for PayTo

## Checklist 
- [x] Code is unit tested

COAND-1058


payto/src/main/res/values/styles.xml
```diff
@@ -21,11 +21,13 @@
 
     <style name="AdyenCheckout.PayTo.FirstNameEditText" parent="AdyenCheckout.TextInputEditText">
         <item name="android:hint">@string/checkout_payto_first_name_hint</item>
```
OscarSpruit, line 23: The most important input fields already have an autofill hint, but I think some others could also use it. The first and last name inputs for example. [Here](https://developer.android.com/reference/androidx/autofill/HintConstants) is a extended list of hints. You can either hardcode those values in xml or use this constants class through code

OscarSpruit (approved):

PR #2026, opened 2025-02-27
PayTo - Add translations

araratthehero:
## Description
Add translations for PayTo

## Checklist 

COAND-1058


payto/src/main/res/values/strings.xml
```diff
@@ -3,34 +3,38 @@
...
-  ~ Created by ararat on 6/2/2025.
+  ~ Created by ararat on 21/2/2025.
   -->
 
-<!-- TODO Add translations -->
 <resources>
-    <string name="checkout_payto_mode_selection" translatable="false">How would you like to use Payto?</string>
+    <string name="checkout_payto_mode_selection">How would you like to use Payto?</string>
+
+    <string name="checkout_payto_payid_description">Enter the PayID and account details that are connected to your Payto account.</string>
```
OscarSpruit, line 12: Shouldn't it be "PayTo"?

araratthehero: Thanks for flagging it! In the designs we have Payto, but I see that some of the translations were already requested with PayTo. I will verify what is the correct version and fix it accordingly.

araratthehero: I fixed all copies, they should be `PayTo`.

OscarSpruit (approved):

PR #2030, opened 2025-02-28
PayTo - Decrease the size of the reference for example app

araratthehero:
## Description
Decrease the size of the reference for example app to make PayTo function properly. Meanwhile there is a discussion going on internally to see if it is possible to increase the limit for PayTo.

## Checklist 

COAND-1058


jreij (approved):

example-app/src/main/java/com/adyen/checkout/example/service/RequestUtils.kt
```diff
@@ -162,7 +162,7 @@ private const val CARD_BASED_INSTALLMENT_OPTION = "visa"
 private const val ATTEMPT_AUTHENTICATION_TRUE_VALUE = "always"
 private const val ATTEMPT_AUTHENTICATION_FALSE_VALUE = "never"
 
-private fun getReference() = "android-test-components_${System.currentTimeMillis()}"
+private fun getReference() = "android-test_${System.currentTimeMillis()}"
```
jreij, line 165: Minor: should we use either `_` or `-` in both places?

PR #2036, opened 2025-03-03
Format response body accordingly for logging

OscarSpruit:
## Description
This fixes an issue where the log would crash if the response body couldn't be transformed into a JSONObject.

NOTE: the improved logging was not released yet, so there are no release notes needed.

## Checklist 

COAND-1077

### DIFF
--- a/checkout-core/src/main/java/com/adyen/checkout/core/internal/data/api/HttpClientExt.kt
+++ b/checkout-core/src/main/java/com/adyen/checkout/core/internal/data/api/HttpClientExt.kt
@@ -100,6 +100,12 @@ private fun Any.logResponse(response: AdyenApiResponse) {
     response.headers.forEach { (key, value) ->
         adyenLog(AdyenLogLevel.VERBOSE) { "$key: $value" }
     }
-    adyenLog(AdyenLogLevel.VERBOSE) { response.body.toJSONObject().toStringPretty() }
+    adyenLog(AdyenLogLevel.VERBOSE) { response.body.tryToFormatJson() }
     adyenLog(AdyenLogLevel.VERBOSE) { "response - END" }
+}
+
+private fun String.tryToFormatJson(): String = when {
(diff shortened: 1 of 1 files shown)


checkout-core/src/main/java/com/adyen/checkout/core/internal/data/api/HttpClientExt.kt
```diff
@@ -100,6 +100,11 @@ private fun Any.logResponse(response: AdyenApiResponse) {
     response.headers.forEach { (key, value) ->
         adyenLog(AdyenLogLevel.VERBOSE) { "$key: $value" }
     }
-    adyenLog(AdyenLogLevel.VERBOSE) { response.body.toJSONObject().toStringPretty() }
+    val bodyString = when {
+        response.body.startsWith("{") -> JSONObject(response.body).toStringPretty()
```
jreij, line 104: IMO this is still not a safe check, technically the response could start with `"{"` and still fail to cast to `JSONObject` for other reasons.
We should handle any thrown exceptions and not let the API fail because of logging no matter what.
And ideally we should have a more robust way of checking whether something is a `JSONObject` or a `JSONArray` or just a normal string other than the first character.
Let me know your thoughts.

OscarSpruit: Although I understand your concern, I think if we can't successfully transform a string into a json object or array we have bigger problems... After logging the response we again try to create a JSONObject/JSONArray again and we have no exception handling there. I think that's okay as the upstream is supposed to handle any IO exceptions. So IMO safely creating these objects just for logging is overkill. WDYT?

jreij: Up to you. To me the most important thing is not to have issues caused by logging, I don't trust servers 😅

jreij (approved):

PR #2046, opened 2025-03-14
PayTo - Recover state when view gets recreated

araratthehero:
## Description
- Do not include country code in phone number validation, since it is hardcoded and not necessary for the validation and in the view it acts as a prefix
- Recover state from outputData when view gets recreated

## Checklist 
- [x] Code is unit tested

COAND-1058


payto/src/main/java/com/adyen/checkout/payto/internal/ui/view/PayToView.kt
```diff
@@ -71,6 +72,20 @@ internal class PayToView @JvmOverloads constructor(
         initBsbAccountNumberInput()
         initFirstNameInput()
         initLastNameInput()
+
+        updateInputFields(delegate.outputData)
```
OscarSpruit, line 76: Shouldn't this be called before all the inputs are initialized? So it doesn't trigger listeners and creates this sort of loop

OscarSpruit (approved):

PR #2062, opened 2025-04-01
Card scanning analytics

OscarSpruit:
## Description
Add analytics for card scanning. See the ticket for more details. The API did not add the events yet, so it returns an error currently.

## Checklist 
- [x] Code is unit tested

COAND-1082

### DIFF
--- a/card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
+++ b/card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
@@ -17,7 +17,6 @@ import androidx.core.app.ActivityCompat
 import androidx.fragment.app.Fragment
 import com.adyen.checkout.core.AdyenLogLevel
 import com.adyen.checkout.core.Environment
-import com.adyen.checkout.core.exception.CheckoutException
 import com.adyen.checkout.core.internal.util.adyenLog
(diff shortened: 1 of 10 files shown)


card-scanning/src/main/java/com/adyen/checkout/card/scanning/AdyenCardScanner.kt
```diff
@@ -104,14 +103,16 @@ class AdyenCardScanner {
...
         val paymentCardRecognitionPendingIntent =
             paymentCardRecognitionPendingIntent ?: error("The scanner must be initialized before it can be started")
 
-        try {
+        return try {
             startIntentSender(paymentCardRecognitionPendingIntent)
+            true
         } catch (e: IntentSender.SendIntentException) {
-            throw CheckoutException("Failed to start payment card recognition", e)
+            adyenLog(AdyenLogLevel.ERROR, e) { "Failed to start payment card recognition" }
```
araratthehero, line 114: Do we not want to throw CheckoutException anymore?

OscarSpruit: Before the exception wasn't handled, so it would crash. If we propagate the exception to the merchant, then in drop-in we would show a dialog and after terminate the drop-in. That's a bit extreme for this case. So I figured ignoring the exception is best for now and we can improve in v6

PR #2072, opened 2025-04-11
Setup UI theme

OscarSpruit:
## Description
Create base structure for the UI theme. This allows for customization and forms the base for our design system.

COAND-1084

### DIFF
--- a/ui/src/main/java/com/adyen/checkout/ui/theme/AdyenButtonStyles.kt
+++ b/ui/src/main/java/com/adyen/checkout/ui/theme/AdyenButtonStyles.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 10/4/2025.
+ */
+
+package com.adyen.checkout.ui.theme
+
+data class AdyenButtonStyles(
+    val cornerRadius: Int? = null,
+    val primary: AdyenButtonStyle? = null,
+    val secondary: AdyenButtonStyle? = null,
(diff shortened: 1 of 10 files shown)


ui/src/main/java/com/adyen/checkout/ui/internal/ColorTokens.kt
```diff
@@ -0,0 +1,41 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 11/4/2025.
+ */
+
+package com.adyen.checkout.ui.internal
+
+import com.adyen.checkout.ui.theme.AdyenColor
+
+internal object ColorTokensLight {
```
araratthehero, line 13: What do you think of calling these `DefaultColorsLight` or something similar?

ui/src/main/java/com/adyen/checkout/ui/theme/AdyenButtonStyles.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 10/4/2025.
+ */
+
+package com.adyen.checkout.ui.theme
+
+data class AdyenButtonStyles(
+    val cornerRadius: Int? = null,
```
ozgur00, line 12: Do you think we can use `Dp` here?

OscarSpruit: `Dp` is a compose class and the aim is to keep our system agnostic. I was thinking to make clear with kdocs and/or annotations that this value is in dp

ozgur00 (approved):

PR #2073, opened 2025-04-14
Setup compose for ui module

OscarSpruit:
## Description
- Update versions of compose and kotlin dependencies
- Update .api files as a result
- Fix one unit test as a result
- Apply new way of configuring the compose compiler
- Make all theming classes immutable

COAND-1084


araratthehero (commented):
Looks good to me. 

Only I see that the `release_notes_dependency_list.toml` file needs to be updated.

araratthehero (approved):

PR #2078, opened 2025-04-16
v6 - Add initial state management

araratthehero:
## Description
Add core implementation of the state management

COAND-1073

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/internal/ui/state/model/Validation.kt
+++ b/core/src/main/java/com/adyen/checkout/core/internal/ui/state/model/Validation.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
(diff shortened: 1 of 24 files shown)


core/src/main/java/com/adyen/checkout/core/internal/ui/state/model/Validation.kt
```diff
@@ -0,0 +1,26 @@
...
+import androidx.annotation.RestrictTo
+import androidx.annotation.StringRes
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+sealed class Validation {
+
+    /**
+     * Field is valid and can be accepted.
+     */
+    object Valid : Validation()
```
OscarSpruit, line 20: Unfortunately nested sealed classes need to be annotated with `RestrictTo` too

PR #2080, opened 2025-04-16
v6 - Add state management for MBWayDelegate

araratthehero:
## Description
Add state management for MBWayDelegate

COAND-1073

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/model/MBWayTransformerRegistry.kt
+++ b/core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/model/MBWayTransformerRegistry.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/model/MBWayTransformerRegistry.kt
```diff
@@ -0,0 +1,34 @@
...
+internal class MBWayTransformerRegistry : FieldTransformerRegistry<MBWayFieldId> {
+    private val transformers = MBWayFieldId.entries.associateWith { fieldId ->
+        when (fieldId) {
+            MBWayFieldId.COUNTRY_CODE -> DefaultTransformer()
+            MBWayFieldId.PHONE_NUMBER -> LocalPhoneNumberTransformer()
+        }
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> transform(key: MBWayFieldId, value: T): T {
```
ozgur00, line 25: In the interface the field is called `fieldId`, so it's better to use it here as well.

ozgur00 (approved):

PR #2081, opened 2025-04-17
Button composables

OscarSpruit:
## Description
Create button composables for all the different type of buttons we have.
Also add a util that easily allows to preview composables in light and dark mode.

image

COAND-1084


ui/src/main/java/com/adyen/checkout/ui/internal/ButtonDefaults.kt
```diff
@@ -0,0 +1,60 @@
...
+import androidx.compose.ui.graphics.Color
+import com.adyen.checkout.ui.theme.AdyenButtonStyle
+
+internal object ButtonDefaults {
+
+    @Composable
+    fun primaryButtonStyle(style: AdyenButtonStyle?): InternalButtonStyle {
+        val colors = AdyenCheckoutTheme.colors
+        return InternalButtonStyle(
+            backgroundColor = style?.backgroundColor?.toCompose() ?: colors.primary,
```
ozgur00, line 21: Is there something explicitly mentioned in the design that we can refer about which color to use as fallback? If in the future we see an issue with some of these colors due to using fallback, then it would be good to have them documented somewhere.

ui/src/main/java/com/adyen/checkout/ui/internal/Buttons.kt
```diff
@@ -0,0 +1,270 @@
...
+        text = text,
+        modifier = modifier,
+        icon = icon,
+        isLoading = isLoading,
+        cornerRadius = buttonStyles.cornerRadius ?: AdyenCheckoutTheme.elements.cornerRadius,
+        style = ButtonDefaults.primaryButtonStyle(buttonStyles.primary),
+    )
+}
+
+@Preview(showBackground = true)
```
ozgur00, line 54: It would be nice to have a preview where we pass an icon as well, as an example of how it looks. Doesn't have to be this one.

OscarSpruit: After some discussion we decided to remove icons as we don't have them in the design. We could always add them later if really needed

ozgur00 (approved):

PR #2083, opened 2025-04-17
v6 - Add viewState for MBWayDelegate

araratthehero:
## Description
Add viewState for the new MbWay component

COAND-1073


ozgur00 (approved):

core/src/main/java/com/adyen/checkout/core/internal/AdyenComponent.kt
```diff
@@ -11,14 +11,15 @@ package com.adyen.checkout.core.internal
...
 
-internal class AdyenComponent : ViewModel() {
+internal class AdyenComponent(coroutineScope: CoroutineScope) : ViewModel() {
 
     private val paymentFacilitator: PaymentFacilitator
 
     init {
         // TODO - Initialize Payment Flow
-        paymentFacilitator = PaymentFacilitator()
+        paymentFacilitator = PaymentFacilitator(coroutineScope)
```
OscarSpruit, line 22: Can't we pass the `viewModelSope`?

PR #2086, opened 2025-04-22
Co-badged Cards - Display card brand options

ozgur00:
## Description
Use `CardBrandAdapter` in `CardView` to display brand options.

## Checklist 
- [x] Code is unit tested

COAND-1085

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
@@ -20,6 +20,7 @@ import com.adyen.checkout.card.internal.data.api.DetectCardTypeRepository
 import com.adyen.checkout.card.internal.data.model.Brand
 import com.adyen.checkout.card.internal.data.model.DetectedCardType
 import com.adyen.checkout.card.internal.ui.model.CVCVisibility
+import com.adyen.checkout.card.internal.ui.model.CardBrandItem
 import com.adyen.checkout.card.internal.ui.model.CardComponentParams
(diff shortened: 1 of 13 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -405,35 +403,14 @@ class CardView @JvmOverloads constructor(
 
     private fun initCardBrandLogoViews(selectedIndex: Int) {
```
OscarSpruit, line 404: Looking at the designs, shouldn't we remove this logic to show the selected brand inside of the input field?

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -419,6 +422,7 @@ class DefaultCardDelegate(
 
         val firstCardBrand = DetectedCardTypesUtils.getSelectedOrFirstDetectedCardType(
             detectedCardTypes = outputData.detectedCardTypes,
+            selectedCardBrandItem = outputData.dualBrandCardBrands.firstOrNull { it.isSelected },
```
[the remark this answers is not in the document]
ozgur00: Leaving it as is as we discussed. This way we keep component state creation only dependent on output data.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
ozgur00: There was some confusion due to ticket having old designs, fixed now.

OscarSpruit (approved):

PR #2087, opened 2025-04-22
Co-badged Cards - Display localeBrand from BinLookupResponse

ozgur00:
## Description
Use localized brand name provided in the `BinLookupResponse` as brand name display value.

## Checklist 

COAND-1085


card/src/main/java/com/adyen/checkout/card/internal/data/model/Brand.kt
```diff
@@ -28,6 +28,7 @@ data class Brand(
     val expiryDatePolicy: String? = null,
     val panLength: Int? = null,
     val paymentMethodVariant: String? = null,
+    val localeBrand: String? = null,
```
OscarSpruit, line 31: It's probably too late to change on the backend, but can we change this name in our code to something more clear? Maybe something like `displayName` or `localizedName`?

OscarSpruit (approved):

PR #2098, opened 2025-04-25
Focus on first input

OscarSpruit:
## Description
Focus on the first input fields for most payment methods (except PayTo and a few that don't have a input field initially).

## Checklist 

COSDK-138

### Fixed
- The first input field will now automatically be focussed


ozgur00:
Are these changes made to align with other platforms?

ozgur00:
Looks good. My only question is we should do it for `PayTo` and `UPI` since they also contain text inputs, but with more complex UI, did you take a look at them?

OscarSpruit:
> Looks good. My only other question is we should do it for `PayTo` and `UPI` since they also contain text inputs, but with more complex UI, did you take a look at them?

@ozgur00 UPI doesn't have an input field visible from the start. PayTo is very complicated as it has tabs and a type dropdown. Both change the form completely. When I tried to focus on the first input it lead to some weird cases, so I decided it's better to not do it for PayTo

ozgur00 (approved):

PR #2099, opened 2025-04-28
Co-badged Cards: Use getChangePayload in DiffCallback

ozgur00:
## Description
Using `getChangePayload()` enables us to only update radio button, not the whole item in `RecyclerView`. This solves flickering issue on selected state changes.

## Checklist 

COAND-1085

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardBrandAdapter.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardBrandAdapter.kt
@@ -30,6 +30,14 @@ internal class CardBrandAdapter(
         val cardBrandItem = currentList[position]
         holder.bind(cardBrandItem, onItemClicked)
     }
+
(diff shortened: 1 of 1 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardBrandAdapter.kt
```diff
@@ -30,6 +30,14 @@ internal class CardBrandAdapter(
         val cardBrandItem = currentList[position]
         holder.bind(cardBrandItem, onItemClicked)
     }
+
+    override fun onBindViewHolder(holder: CardBrandItemViewHolder, position: Int, payloads: MutableList<Any>) {
+        if (payloads.isEmpty()) {
+            super.onBindViewHolder(holder, position, payloads)
+        } else if (payloads.first() == true) {
```
OscarSpruit, line 37: For safety I think it's better to use `firstOrNull()` here

OscarSpruit (approved):

PR #2110, opened 2025-05-01
v6: AdyenCheckout - Initialize Sessions 

ozgur00:
## Description
In the initialization block of AdyenCheckout we're initializing sessions with the changes in this PR. I have copied a lot of classes from several core modules (`sessions-core`, `components-core`, `checkout-core`). A lot of the size of the change comes from that.

COSDK-441

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/exception/HttpException.kt
+++ b/core/src/main/java/com/adyen/checkout/core/exception/HttpException.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 30/4/2025.
+ */
+
+package com.adyen.checkout.core.exception
+
+import com.adyen.checkout.core.internal.data.model.ErrorResponseBody
+
+/**
+ * Indicates that an internal API call has failed.
+ */
+class HttpException(
+    val code: Int,
+    message: String,
+    val errorBody: ErrorResponseBody?,
+    // TODO - Errors
+// ) : CheckoutException(message)
+) : RuntimeException(message)
--- a/core/src/main/java/com/adyen/checkout/core/exception/ModelSerializationException.kt
+++ b/core/src/main/java/com/adyen/checkout/core/exception/ModelSerializationException.kt
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 30/4/2025.
+ */
+
+package com.adyen.checkout.core.exception
+
+import com.adyen.checkout.core.internal.data.model.ModelObject
+import org.json.JSONException
+
+/**
+ * Exception thrown when an issue occurs during serialization of a [ModelObject].
+ */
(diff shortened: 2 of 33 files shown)


core/src/main/java/com/adyen/checkout/core/exception/ModelSerializationException.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import com.adyen.checkout.core.internal.data.model.ModelObject
+import org.json.JSONException
+
+/**
+ * Exception thrown when an issue occurs during serialization of a [ModelObject].
+ */
+class ModelSerializationException(modelClass: Class<*>, cause: JSONException?) :
+//    CheckoutException("Unexpected exception while serializing ${modelClass.simpleName}.", cause)
+    Exception("Unexpected exception while serializing ${modelClass.simpleName}.", cause)
```
OscarSpruit, line 19: This should be a `RuntimeException` I think

core/src/main/java/com/adyen/checkout/core/internal/data/model/ModelUtils.kt
```diff
@@ -0,0 +1,88 @@
...
+ * Created by ozgur on 30/4/2025.
+ */
+
+package com.adyen.checkout.core.internal.data.model
+
+import org.json.JSONArray
+import org.json.JSONObject
+import java.util.Collections
+
+object ModelUtils {
```
OscarSpruit, line 15: This can be `internal`

core/src/main/java/com/adyen/checkout/core/model/Amount.kt
```diff
@@ -0,0 +1,53 @@
...
+import com.adyen.checkout.core.internal.data.model.ModelObject
+import com.adyen.checkout.core.internal.data.model.getLongOrNull
+import com.adyen.checkout.core.internal.data.model.getStringOrNull
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+data class Amount(
+    var currency: String? = null,
```
OscarSpruit: It would be a good time to convert all these mutable objects to immutable objects with `val`. WDYT?

core/src/main/java/com/adyen/checkout/core/exception/HttpException.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Indicates that an internal API call has failed.
+ */
+class HttpException(
+    val code: Int,
+    message: String,
+    val errorBody: ErrorResponseBody?,
+    // TODO - Errors
+// ) : CheckoutException(message)
+) : Exception(message)
```
araratthehero, line 22: This should also be `RuntimeException` I guess?

core/src/main/java/com/adyen/checkout/core/exception/ModelSerializationException.kt
```diff
@@ -0,0 +1,19 @@
...
+package com.adyen.checkout.core.exception
+
+import com.adyen.checkout.core.internal.data.model.ModelObject
+import org.json.JSONException
+
+/**
+ * Exception thrown when an issue occurs during serialization of a [ModelObject].
+ */
+class ModelSerializationException(modelClass: Class<*>, cause: JSONException?) :
+//    CheckoutException("Unexpected exception while serializing ${modelClass.simpleName}.", cause)
```
araratthehero, line 18: Could you add a todo here too, like `// TODO - Error propagation`?

PR #2112, opened 2025-05-02
Improve updating verification metadata

OscarSpruit:
## Description
- Add a Gradle task that resolves all dependencies
- Clean up verification metadata file
- Apply new Gradle task to CI job

COSDK-446


build.gradle
```diff
@@ -14,6 +14,7 @@ plugins {
     alias libs.plugins.sonarqube
     alias libs.plugins.jacoco
     alias libs.plugins.binary.compatibility.validator
+    id "resolve-dependencies"
```
ozgur00: Nit: I think if we add
```gradle
resolve-dependencies = { id = "resolve-dependencies" }
```
to the plugin section in the `toml` file we can use `alias libs.plugins.resolve.dependencies` here. Would look a bit more unified with the rest of the plugins we apply.

OscarSpruit: Nice suggestion! I didn't know that was possible

ozgur00 (dismissed):

PR #2117, opened 2025-05-09
Co-badged Cards

ozgur00:
## Description
Complete co-badged cards feature.

## Checklist 
- [x] Code is unit tested

COSDK-30

## Release notes
[//]: # (### New)
[//]: # (### Fixed)
[//]: # (List fixes here)

### Improved
- For co-badged cards, brand selection experience is improved. Shoppers now will be able to see the available brand options in a separate section as part of the Card Component.


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -792,13 +746,23 @@ class CardView @JvmOverloads constructor(
...
+        if (isDualBranded) {
+            if (cardBrandAdapter == null) {
+                cardBrandAdapter = CardBrandAdapter { cardBrandItem ->
+                    cardDelegate.updateInputData {
+                        selectedCardBrand = cardBrandItem.brand
+                    }
+                }
+                binding.recyclerViewCobadgeBrands.adapter = cardBrandAdapter
+            }
+            cardBrandAdapter?.submitList(dualBrandData?.brandOptions)
```
araratthehero, line 763: What do you think of creating the `CardBrandAdapter` once when the view gets created and only update the adapter content here when outputData changes, instead of creating a new `CardBrandAdapter` every time?

ozgur00: We only create the adapter for the first time (when it's `null`), then we just submit the updated list every time output data changes. So I think it works the way you mention, just a bit different code. I followed this approach because we use the same for adapters in other places as well.

araratthehero: Good point, I missed it

PR #2122, opened 2025-05-14
Request focus on first input in drop-in

OscarSpruit:
## Description
Directly requesting focus in the view, like we previously had, didn't work for all use cases (compose for example). It is better to let the merchant decide if they want to focus on the `AdyenComponentView` or their own view. In drop-in we are in control, so we can focus on the first input

NOTE: release notes are already provided by https://github.com/Adyen/adyen-android/pull/2098

## Checklist 

COSDK-138

### DIFF
--- a/ach/src/main/java/com/adyen/checkout/ach/internal/ui/view/ACHDirectDebitView.kt
+++ b/ach/src/main/java/com/adyen/checkout/ach/internal/ui/view/ACHDirectDebitView.kt
@@ -39,7 +39,7 @@ internal class ACHDirectDebitView @JvmOverloads constructor(
 ) : LinearLayout(
     context,
     attrs,
-    defStyleAttr,
+    defStyleAttr
 ),
     ComponentView {
     private val binding = AchDirectDebitViewBinding.inflate(LayoutInflater.from(context), this)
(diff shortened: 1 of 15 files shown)


ach/src/main/java/com/adyen/checkout/ach/internal/ui/view/ACHDirectDebitView.kt
```diff
@@ -114,7 +115,11 @@ internal class ACHDirectDebitView @JvmOverloads constructor(
                 }
             }
 
-            requestFocus()
+            doOnNextLayout { view ->
```
jreij, line 118: Can we export this logic into a separate function used by all views?

jreij (approved):

jreij (commented):
The release notes in the initial PR say `For drop-in, the first input field will now automatically be focussed`, is that still correct after this PR is merged or do we need to edit those release notes?

OscarSpruit:
> The release notes in the initial PR say `For drop-in, the first input field will now automatically be focussed`, is that still correct after this PR is merged or do we need to edit those release notes?

@jreij I edited them there to make sure they reflect the changes in this PR

jreij (approved):

PR #2123, opened 2025-05-14
Fix predictive back navigation for drop-in

OscarSpruit:
## Description
Predictive back navigation doesn't work in drop-in. For now, we will opt-out until we have a proper fix.

To test it out, you have to change the compile and target sdk version to 36.

## Checklist 

COSDK-472

### Fixed
- For the `DropInActivity`, predictive back navigation is now disabled to ensure compatibility with Android 16


jreij (dismissed):

jreij (commented):
Release notes: `DropInActivity` is not a publicly documented class, can we just say `drop-in` instead?

PR #2141, opened 2025-06-03
Accessibility - Expiry Date & Security Code

ozgur00:
## Description
Add content description for expiry date and security code fields.
Also updated security code hint, to align with other platforms. Validation errors are yet to be aligned, that is left out to keep this PR focused on accessibility.

## Checklist 

COSDK-484

## Release notes
[//]: # (### New)
[//]: # (### Fixed)
[//]: # (List fixes here)
### Improved
- For Expiry Date and Security Code input fields in Card Component, content descriptions are more expressive and contain format information.

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
@@ -44,10 +44,12 @@ import com.adyen.checkout.ui.core.internal.ui.ComponentView
 import com.adyen.checkout.ui.core.internal.ui.loadLogo
 import com.adyen.checkout.ui.core.internal.ui.model.AddressOutputData
 import com.adyen.checkout.ui.core.internal.ui.view.AdyenTextInputEditText
+import com.adyen.checkout.ui.core.internal.ui.view.ExpiryDateInput
 import com.adyen.checkout.ui.core.internal.ui.view.RoundCornerImageView
 import com.adyen.checkout.ui.core.internal.ui.view.SecurityCodeInput
 import com.adyen.checkout.ui.core.internal.util.hideError
 import com.adyen.checkout.ui.core.internal.util.isVisible
+import com.adyen.checkout.ui.core.internal.util.setAccessibilityDelegateWith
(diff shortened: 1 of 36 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -425,6 +428,15 @@ class CardView @JvmOverloads constructor(
                 binding.textInputLayoutExpiryDate.showError(localizedContext.getString(expiryDateValidation.reason))
             }
         }
+        binding.editTextExpiryDate.setContentDescription()
+    }
+
+    // TODO - Investigate contentDescription in ViewFieldState [COSDK-490]
+    private fun ExpiryDateInput.setContentDescription() {
```
OscarSpruit: Just tested it with TalkBack and it doesn't seem to work. [Here](https://support.google.com/accessibility/android/answer/6378120?sjid=10008584266951357526-EU) it says that `contentDescription` might be unstable when set on an `EditText`.  I think to make it work we either have to add the format to the `hint` of the `TextInputLayout` or we need to add it as a helper text, like below:

image

Also TalkBack isn't able to pronounce "MM/YY" correctly, instead it says "M-Y-Y". Is this acceptable or should we make it say the format in full?

ozgur00: I'll take a look for alternatives about the first part of your comment. Second part about the pronunciation: I had noticed, I've requested translations for `Month, 2 digits, Year, 2 digits` to make it more expressive. Now waiting on translations for that.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -425,6 +429,15 @@ class CardView @JvmOverloads constructor(
...
```
OscarSpruit: `setContentDescription` doesn't really apply anymore now, maybe we can rename it to something like `setAccessibilityDescription`? Also for the method on line 461

OscarSpruit (approved):

PR #2143, opened 2025-06-03
v6: Payment Facilitator Provider

ozgur00:
## Description
Add `PaymentFacilitatorProvider` and factories for sessions and advanced flows.

## Checklist 

COSDK-142

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/internal/PaymentFacilitatorProvider.kt
+++ b/core/src/main/java/com/adyen/checkout/core/internal/PaymentFacilitatorProvider.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 3/6/2025.
+ */
+
+package com.adyen.checkout.core.internal
+
+import androidx.lifecycle.SavedStateHandle
+import com.adyen.checkout.core.AdyenCheckout
+import kotlinx.coroutines.CoroutineScope
+
+internal class PaymentFacilitatorProvider {
+
+    fun provide(
+        adyenCheckout: AdyenCheckout,
+        coroutineScope: CoroutineScope,
+        savedStateHandle: SavedStateHandle,
+    ): PaymentFacilitator {
+        val paymentFacilitatorFactory = if (adyenCheckout.checkoutSession == null) {
+            AdvancedPaymentFacilitatorFactory(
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/adyen/checkout/core/internal/PaymentFacilitatorProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+        adyenCheckout: AdyenCheckout,
+        coroutineScope: CoroutineScope,
+        savedStateHandle: SavedStateHandle,
+    ): PaymentFacilitator {
+        return if (adyenCheckout.checkoutSession == null) {
+            AdvancedPaymentFacilitatorFactory().create(
+                checkoutConfiguration = adyenCheckout.checkoutConfiguration,
+                checkoutCallback = adyenCheckout.checkoutCallback,
+                coroutineScope = coroutineScope,
+            )
```
araratthehero: Nits: What do you think of moving these params, inside the `AdvancedPaymentFacilitatorFactory` as constructor parameters. Then move the `create` function to the `PaymentFacilitatorFactory` interface.
This way it looks more clear imo, to pass the parameters to the `AdvancedPaymentFacilitatorFactory` or `SessionsPaymentFacilitatorFactory` when creating them and call the `create()` function after. It also forces the `create()` function to return a `PaymentFacilitator`.

It would look like this:
```
internal interface PaymentFacilitatorFactory {
    fun create(): PaymentFacilitator
}
```

```
fun provide(
    adyenCheckout: AdyenCheckout,
    coroutineScope: CoroutineScope,
): PaymentFacilitator {
    val paymentFacilitatorFactory = when (adyenCheckout.checkoutSession) {
        null -> AdvancedPaymentFacilitatorFactory(
            checkoutConfiguration = adyenCheckout.checkoutConfiguration,
            checkoutCallback = adyenCheckout.checkoutCallback,
        )

        else -> SessionsPaymentFacilitatorFactory(
            checkoutSession = adyenCheckout.checkoutSession,
            checkoutConfiguration = adyenCheckout.checkoutConfiguration,
            checkoutCallback = adyenCheckout.checkoutCallback,
            savedStateHandle = savedStateHandle,
        )
    }

    return paymentFacilitatorFactory.create()
}
```

ozgur00: It's a good idea, I have tweaked it slightly to not have `coroutineScope` stored in these factory classes, so they would just pass it on.

PR #2154, opened 2025-06-11
Sonatype - Migrate workflows to Sonatype Central portal

araratthehero:
## Description
The workflows are updated to use [OSSRH Staging API](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/) to publish artifacts to Central Portal.

A workflow run can be found [here](https://github.com/Adyen/adyen-android/actions/runs/15579793582). We still get `104` error for the last step, even though it successfully uploads the repository to Central Portal. Currently waiting for a response from Sonatype.

## Checklist 

COSDK-475

### DIFF
--- a/.github/workflows/publish_to_maven_central.yml
+++ b/.github/workflows/publish_to_maven_central.yml
@@ -59,6 +59,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
       # Moved the repository from OSSRH to Central Portal
       - name: Move repository to Central Portal
         env:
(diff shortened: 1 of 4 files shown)


scripts/move_ossrh_repository_to_central_portal.py
```diff
@@ -0,0 +1,160 @@
...
+def _make_api_request(method, url, auth, params=None, headers=None, expected_success_codes=None):
+    # Makes an HTTP request and handles common errors.
+    # Returns the response object on success, None on failure or request exception.
+    if headers is None:
+        headers = {}
+    headers.setdefault("Accept", "application/json") # Default Accept header
+
+    if expected_success_codes is None:
+        if method.upper() == "POST":
+            pass
```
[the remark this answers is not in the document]
araratthehero: It is by design and that method just checks if there is an error code, it prints a message and does not return a result.

scripts/move_ossrh_repository_to_central_portal.py
```diff
@@ -0,0 +1,160 @@
...
+    print("-------------------- UPLOAD RESPONSE END ----------------------")
+    print(f"Upload HTTP Status Code: {response.status_code}")
+
+    # Success conditions: 2xx status codes or specifically HTTP 408
+    if 200 <= response.status_code < 300 or response.status_code == 408:
+        print(f"Repository publish command for key '{repo_key}' processed: Status {response.status_code}. Considered successful.")
+        return True
+    else:
+        print(f"Error: Failed to publish repository with key '{repo_key}'. Status: {response.status_code}.")
+        return False
```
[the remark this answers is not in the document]
araratthehero: This has been removed already

.github/workflows/publish_to_maven_central.yml
```diff
@@ -15,9 +15,32 @@ on:
         type: string
 
 jobs:
-  publish_to_maven_central:
+  # This step should be removed when gradle plugin supports uploading to Central Portal
+  drop_all_open_ossrh_repositories:
```
OscarSpruit, line 19: Could this potentially drop open repositories from other teams?

.github/workflows/publish_to_maven_central.yml
```diff
@@ -15,9 +15,32 @@ on:
...
     runs-on: ubuntu-latest
 
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
```
ozgur00, line 27: I think it's better to create a `.python-version` file to store the version in a dedicated file and not in the script. That way in the future we know where to look for the version, and it also unifies the python version we use throughout all our workflows. Also if we do that we don't have to specify this parameter, we can omit it. I think we can put it in the root directory, but not 100% sure, we can discuss further offline.

https://github.com/pyenv/pyenv?tab=readme-ov-file#understanding-python-version-selection

scripts/move_ossrh_repository_to_central_portal.py
```diff
@@ -0,0 +1,160 @@
+import os
+import sys
+import requests
+import urllib.parse
+
+# --- Configuration ---
+SONATYPE_API_BASE_URL = "https://ossrh-staging-api.central.sonatype.com/manual"
+SEARCH_REPOSITORIES_URL = f"{SONATYPE_API_BASE_URL}/search/repositories"
+UPLOAD_REPO_BASE_URL = f"{SONATYPE_API_BASE_URL}/upload/repository" # Key will be appended
+TIMEOUT_SECONDS = 600
```
ozgur00: Is this based on sonatype documentation? 🤔 If so, would be nice to mention our reasoning as a comment.

scripts/drop_all_open_ossrh_repositories.py
```diff
@@ -0,0 +1,133 @@
...
+    if headers is None:
+        headers = {}
+    headers.setdefault("Accept", "application/json") # Default Accept header
+
+    if expected_success_codes is None:
+        expected_success_codes = [200]
+
+    print(f"Making {method} request to: {url} with params: {params if params else 'None'}")
+    try:
+        response = requests.request(method, url, auth=auth, params=params, headers=headers, timeout=TIMEOUT_SECONDS)
```
ozgur00: Nit: Would be nice and consistent if we wrote argument names for `method` and `url` too.

scripts/move_ossrh_repository_to_central_portal.py
araratthehero: Good point. Currently we get an error for this request and I am waiting for a response from Sonatype. I decided to increase it too allow the request to run for a long time. It roughly take 4 minutes to get a response. I will write a comment here, so it gets changed in the future.

.github/workflows/publish_to_maven_central.yml
araratthehero: Yes indeed, that is on purpose. The problem with the API is that we cannot identify which repositories are uploaded by us.
This step drops only open repositories. In general when a repository is completely uploaded (through `maven_publish`), we make an API call (`move_repository_to_central_portal` step) which closes the repository automatically. So right after the workflow run there will be no open repositories.

It is an edge case if we already have an open repository from another team, but if it happens, either of the team will know about it, since the workflow will fail, which can be communicated between the teams. And a re-run of the workflow should run from a clean state. Note that this solution is temporary, until gradle plugin works are finished (top priority for Sonatype). Ideally with the plugin we should not have this issue.

This script solves a few issues:
- Sometimes it happens that when uploading the artifacts to Sonatype, it creates two repositories instead of one. In that case our workflow will fail, and running it again will drop all open repositories.
- Since we can not get an identifier of the repository we just uploaded, the only way to get an identifier is to call the API to get the list of open repositories. So if there are two repositories which are open we cannot identify the one we just uploaded.
- It makes sure that we start the release from a clean state, in case of a previous workflow run created an incomplete open repository.

ozgur00 (approved):

PR #2159, opened 2025-06-16
V6 - MBWay - Generate componentState from delegate state

araratthehero:
## Description
Now componentState is generated from delegate state

## Checklist 
- [x] Code is unit tested

COSDK-505


core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/MBWayFactory.kt
```diff
@@ -58,6 +90,31 @@ internal class MBWayFactory :
...
+            validationRegistry = MBWayValidatorRegistry(),
+            stateUpdaterRegistry = MBWayStateUpdaterRegistry(),
+            transformerRegistry = transformerRegistry,
+        )
+
+        return MBWayDelegate(
+            coroutineScope = coroutineScope,
+            componentParams = componentParams,
+            analyticsManager = analyticsManager,
+            order = null,
```
ozgur00, line 71: Can we add the order related todo here as well to make sure we won't miss it?

ozgur00 (approved):

PR #2162, opened 2025-06-17
v6 - Add PaymentMethodProvider

ozgur00:
## Description
Add `PaymentMethodProvider` class to instantiate payment methods by `txVariant`.

COSDK-482

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/internal/BaseComponentState.kt
+++ b/core/src/main/java/com/adyen/checkout/core/internal/BaseComponentState.kt
@@ -1,0 +1,14 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 17/6/2025.
+ */
+
+package com.adyen.checkout.core.internal
+
+import com.adyen.checkout.core.paymentmethod.PaymentComponentState
+import com.adyen.checkout.core.paymentmethod.PaymentMethodDetails
+
+internal typealias BaseComponentState = PaymentComponentState<out PaymentMethodDetails>
--- a/core/src/main/java/com/adyen/checkout/core/internal/PaymentMethodProvider.kt
+++ b/core/src/main/java/com/adyen/checkout/core/internal/PaymentMethodProvider.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 13/6/2025.
+ */
+
+package com.adyen.checkout.core.internal
+
+import com.adyen.checkout.core.CheckoutConfiguration
+import com.adyen.checkout.core.internal.ui.PaymentDelegate
+import com.adyen.checkout.core.sessions.CheckoutSession
+import kotlinx.coroutines.CoroutineScope
+
+internal object PaymentMethodProvider {
+
+    private val factories =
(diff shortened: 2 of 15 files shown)


core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/MBWayFactory.kt
```diff
@@ -0,0 +1,63 @@
...
+        coroutineScope: CoroutineScope,
+        checkoutSession: CheckoutSession,
+        checkoutConfiguration: CheckoutConfiguration
+    ): MBWayDelegate {
+        val componentParams = ButtonComponentParamsMapper(CommonComponentParamsMapper()).mapToParams(
+            checkoutConfiguration = checkoutConfiguration,
+
+            // TODO - Add locale support, For now it's hardcoded to US
+            // deviceLocale = localeProvider.getLocale(application)
+            deviceLocale = Locale.US,
```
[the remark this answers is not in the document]
ozgur00: LocaleProvider will be added later once we add `Initializer`s and pass `Context` to our `PaymentMethodFactory` classes.

core/src/main/java/com/adyen/checkout/core/internal/BaseComponentState.kt
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Created by ozgur on 17/6/2025.
+ */
+
+package com.adyen.checkout.core.internal
+
+import com.adyen.checkout.core.paymentmethod.PaymentComponentState
+import com.adyen.checkout.core.paymentmethod.PaymentMethodDetails
+
+typealias BaseComponentState = PaymentComponentState<out PaymentMethodDetails>
```
ozgur00: internal

core/src/main/java/com/adyen/checkout/core/internal/PaymentMethodProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Advanced
+     */
+    fun get(
+        txVariant: String,
+        coroutineScope: CoroutineScope,
+        checkoutConfiguration: CheckoutConfiguration
+    ): PaymentDelegate<BaseComponentState> {
+        return (factories[txVariant] as? AdvancedPaymentMethodFactory)?.create(
```
araratthehero: Is it possible to have a paymentDelegate which supports either of the advanced or sessions flow, but not both of them together?

araratthehero: I am thinking if it makes sense to have two different `SessionsPaymentMethodFactory` and `AdvancedPaymentMethodFactory` factories, instead of having one `PaymentMethodFactory`. If by mistake you forget to implement one of those interfaces in your delegate, it will crash at runtime for one flow only and it might be easy to miss it?

OscarSpruit: I do agree. We can probably abstract the session specific things away. I'm thinking if we can keep sessions as a whole out of payment method related classes. Only `PaymentFacilitator` or some class on a similar level would know about it. The facilitator level could create the session or advanced implementation of an interface and then the payment method factories don't have to be aware of sessions at all

ozgur00: I removed the separate interfaces. Right now we do have `SessionsPaymentFacilitatorFactory` and `AdvancedPaymentFacilitatorFactory` for creating sessions or advanced dependencies, which is indeed higher than PaymentMethodFactory level so there's less dependency on it.

ozgur00: Resolving this conversation, the follow up to it is in PR #2169.

araratthehero (approved):

PR #2163, opened 2025-06-17
v6 example integration

OscarSpruit:
## Description
To see and test what we are making I added an example integration of the v6 stuff we have so far. In the non sessions list there is now a "v6" entry and when you click it, it will open an activity that renders the v6 composable (with the minimum necessary setup).

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/main/ComponentItem.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/main/ComponentItem.kt
@@ -27,11 +27,13 @@ internal sealed class ComponentItem {
         object PayPal : Entry(R.string.paypal_component_entry)
         data class Instant(private val paymentMethodType: String) :
             Entry(R.string.instant_component_entry, listOf(paymentMethodType))
+
         object CardWithSession : Entry(R.string.card_component_with_session_entry)
         object CardWithSessionTakenOver : Entry(R.string.card_component_with_session_taken_over_entry)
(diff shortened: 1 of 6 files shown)


example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6ViewModel.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.adyen.checkout.core.CheckoutConfiguration
+import com.adyen.checkout.core.CheckoutResult
+import com.adyen.checkout.core.Environment
+import com.adyen.checkout.core.paymentmethod.PaymentComponentState
+import com.adyen.checkout.example.BuildConfig
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+
+@HiltViewModel
+internal class V6ViewModel @Inject constructor() : ViewModel(), CheckoutCallback {
```
araratthehero, line 23: What do you think of leaving TODOs in all files which need to be removed later. Or create a ticket to clean up those temporary files when v6 is done?

OscarSpruit: In general I agree, but for the example app I don't think it's necessary. Once v6 is in it's final stages we will have to refactor the example app anyway

araratthehero (dismissed):

araratthehero (approved):

PR #2166, opened 2025-06-18
Use CountryModel instead of a String for MBWay country selector

araratthehero:
## Description
Use CountryModel instead of String for MBWay country selector

## Checklist 
- [x] Code is unit tested

COSDK-505


core/src/test/java/com/adyen/checkout/com/mbway/internal/ui/MBWayComponentStateTest.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    fun `when toComponentState is called with valid inputs, then MBWayComponentState should be created correctly`() {
+        val updatedState = initialState.copy(
+            countryCodeFieldState = initialState.countryCodeFieldState.copy(validation = Validation.Valid),
+            localPhoneNumberFieldState = initialState.localPhoneNumberFieldState.copy(validation = Validation.Valid),
+        )
+
+        val componentState = updatedState.toComponentState(
+            analyticsManager = analyticsManager,
```
OscarSpruit: Not really related to this PR, but wouldn't it be easier to pass the `checkoutAttemptId` directly? Then we don't need to create the analytics manager for simple tests

araratthehero: It makes sense, since the component state does not need to know about the `analyticsManager`.

OscarSpruit (approved):

PR #2172, opened 2025-06-23
v6 - MBWay country code picker

OscarSpruit:
## Description
Open a dialog after the country code picker is clicked. In the dialog the shopper can choose their country code.

Note: the dialog will be replaced after we implemented proper navigation and after #2166 we can display the values according to the designs.

https://github.com/user-attachments/assets/3050c1f9-f14a-4fb2-8beb-364454b717ec

COSDK-506


core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/view/MbWayComponent.kt
```diff
@@ -73,10 +101,91 @@ internal fun MbWayComponent(
...
+                                onDismissRequest()
+                            }
+                            .fillMaxWidth()
+                            .padding(12.dp),
+                    ) {
+                        BodyEmphasized(country)
+
+                        if (isSelected) {
+                            Icon(
+                                imageVector = ImageVector.vectorResource(com.adyen.checkout.test.R.drawable.ic_checkmark),
```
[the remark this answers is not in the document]
OscarSpruit: We temporarily generate the resources of the UI module in a package named `test` to avoid conflicts. This is not an actual testing source and will be renamed later

core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/view/MbWayComponent.kt
```diff
@@ -70,14 +101,104 @@ internal fun MbWayComponent(
...
+
+        Surface(
+            modifier = Modifier.fillMaxSize(),
+            color = AdyenCheckoutTheme.colors.background,
+        ) {
+            Column(
+                verticalArrangement = Arrangement.spacedBy(4.dp),
+                modifier = Modifier
+                    .systemBarsPadding()
+                    .padding(16.dp),
```
araratthehero: What do you think of having dimensions specified someplace else and reused wherever necessary?

core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/view/MbWayComponent.kt
```diff
@@ -70,14 +101,104 @@ internal fun MbWayComponent(
...
+            onDismissRequest = { showCountryCodeDialog = false },
+            viewState = viewState,
+            fieldChangeListener = fieldChangeListener,
         )
     }
 }
 
+@Suppress("LongMethod", "MagicNumber")
+@Composable
+private fun CountryCodeDialog(
```
araratthehero, line 116: I would create a reusable Dialog composable, which will contain the status bar color logic and accept content to be shown. This way we could reuse this composable.

core/src/main/java/com/adyen/checkout/core/mbway/internal/ui/view/MbWayComponent.kt
OscarSpruit: Maybe that's a good idea. Is it OK if I do that in a follow up PR? As it requires some extra setup and I probably need to replace it in some more composables

araratthehero: Sure, sounds good

PR #2178, opened 2025-06-27
Fix expiry date input

OscarSpruit:
## Description
Fixes an issue where after resuming the card component when coming back from the background the expiry date input field would set `11/03` and it wouldn't be editable.

## Checklist 

COSDK-521

### Fixed
- For the card component, the expiry date field works correctly again after resuming the app from the background

### DIFF
--- a/card/src/test/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegateTest.kt
+++ b/card/src/test/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegateTest.kt
@@ -1598,8 +1598,10 @@ internal class DefaultCardDelegateTest(
         private const val TEST_CLIENT_KEY = "test_qwertyuiopasdfghjklzxcvbnmqwerty"
         private const val TEST_CARD_NUMBER = "5555444433331111"
         private const val TEST_EXPIRY_MONTH = "03"
-        private const val TEST_EXPIRY_YEAR = "30"
-        private const val TEST_EXPIRY_DATE = "$TEST_EXPIRY_MONTH/$TEST_EXPIRY_YEAR"
+        private const val TEST_EXPIRY_YEAR = "2030"
+
+        // Take last 2 digits of year to comply with the MM/yy format
+        private val TEST_EXPIRY_DATE = "$TEST_EXPIRY_MONTH/${TEST_EXPIRY_YEAR.takeLast(2)}"
         private const val TEST_SECURITY_CODE = "737"
         private val TEST_ORDER = OrderRequest("PSP", "ORDER_DATA")
         private const val TEST_CHECKOUT_ATTEMPT_ID = "TEST_CHECKOUT_ATTEMPT_ID"
--- a/checkout-core/src/test/java/com/adyen/checkout/core/ui/model/ExpiryDateTest.kt
+++ b/checkout-core/src/test/java/com/adyen/checkout/core/ui/model/ExpiryDateTest.kt
@@ -36,24 +36,29 @@ internal class ExpiryDateTest {
             arguments(ExpiryDate(1, 1), "01/01"),
             arguments(ExpiryDate(12, 1234), "12/34"),
             arguments(ExpiryDate(9, 2008), "09/08"),
+            arguments(ExpiryDate(10, 1998), "10/98"),
+            arguments(ExpiryDate(10, 2098), "10/98"),
             arguments(EMPTY_DATE, "00/00"),
             arguments(INVALID_DATE, "-1/-1"),
         )
 
         @JvmStatic
         fun createExpiryDateFromStringSource() = listOf(
-            arguments("01/01", ExpiryDate(1, 1)),
-            arguments("3/3", ExpiryDate(3, 3)),
-            arguments("03/3", ExpiryDate(3, 3)),
-            arguments("3/03", ExpiryDate(3, 3)),
-            arguments("12/34", ExpiryDate(12, 34)),
-            arguments("09/08", ExpiryDate(9, 8)),
-            arguments("00/00", ExpiryDate(0, 0)),
-            arguments("-9/-18", ExpiryDate(-9, -18)),
+            arguments("01/01", ExpiryDate(1, 2001)),
+            arguments("3/03", ExpiryDate(3, 2003)),
+            arguments("12/34", ExpiryDate(12, 2034)),
+            arguments("09/08", ExpiryDate(9, 2008)),
+            arguments("01/00", ExpiryDate(1, 2000)),
+            // Our correction adds 100 years when the date parsing gives a smaller century than the current
+            arguments("10/98", ExpiryDate(10, 2098)),
+            arguments("3/3", ExpiryDate(3, 103)),
+            arguments("03/3", ExpiryDate(3, 103)),
             // Invalid input should create invalid expiry date
-            arguments("03", ExpiryDate(3, -1)),
-            arguments("asd/fgh", ExpiryDate(-1, -1)),
-            arguments("asdfgh", ExpiryDate(-1, -1)),
+            arguments("00/00", INVALID_DATE), // Month 0 doesn't exist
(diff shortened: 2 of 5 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/StoredCardDelegate.kt
```diff
@@ -396,15 +396,14 @@ internal class StoredCardDelegate(
...
-            adyenLog(AdyenLogLevel.ERROR, e) { "Failed to parse stored Date" }
-            inputData.expiryDate = EMPTY_DATE
+        val expiryMonth = storedPaymentMethod.expiryMonth
+        val expiryYear = storedPaymentMethod.expiryYear
+
+        if (expiryMonth.isNullOrBlank() || expiryYear.isNullOrBlank()) {
+            adyenLog(AdyenLogLevel.WARN) { "Failed to parse stored expiry date" }
+            inputData.expiryDate = ""
+        } else {
+            inputData.expiryDate = "$expiryMonth/$expiryYear"
```
araratthehero: This formats the expiry date identical to `ExpiryDate.toMMyyString()`. I would use the `toMMyyString()` here too, to format the date.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -417,7 +416,7 @@ class CardView @JvmOverloads constructor(
 
     private fun initExpiryDateInput() {
         binding.editTextExpiryDate.setOnChangeListener {
-            val date = binding.editTextExpiryDate.date
+            val date = binding.editTextExpiryDate.rawValue
```
ozgur00: Nit: We can remove the `date` and directly assign the raw value in input data.
```
cardDelegate.updateInputData { expiryDate = binding.editTextExpiryDate.rawValue }
```

card/src/main/java/com/adyen/checkout/card/internal/ui/StoredCardDelegate.kt
araratthehero: Another idea would be to move the `toMMyyString()` out of the `ExpiryDate`, add two arguments for the month and year (`toMMyyString(expiryMonth, expiryYear)`) and have an extension function in `ExpiryDate`, which would have an implementation like: `fun ExpiryDate.toMMyyString() = toMMyyString(expiryMonth, expiryYear)`.

This way to use the `toMMyyString()` you wouldn't need to create an `ExpiryDate` object.

card/src/test/java/com/adyen/checkout/card/internal/ui/CardValidationMapperTest.kt
```diff
@@ -29,7 +28,7 @@ internal class CardValidationMapperTest {
 
         @Test
         fun `date is valid, then correct fieldState should be returned`() {
-            val expiryDate = ExpiryDate(4, 2025) // 04/2025
+            val expiryDate = "04/25"
```
ozgur00, line 31: I like this. Way more readable. 👍

card/src/test/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegateTest.kt
```diff
@@ -809,8 +809,8 @@ internal class DefaultCardDelegateTest(
 
                 with(requireNotNull(paymentComponentData.paymentMethod)) {
                     assertEquals(TEST_CARD_NUMBER, encryptedCardNumber)
-                    assertEquals(TEST_EXPIRY_DATE.expiryMonth.toString(), encryptedExpiryMonth)
-                    assertEquals(TEST_EXPIRY_DATE.expiryYear.toString(), encryptedExpiryYear)
+                    assertEquals(ExpiryDate.from(TEST_EXPIRY_DATE).expiryMonth.toString(), encryptedExpiryMonth)
```
ozgur00: Extracting the expiry month using `ExpiryDate.from()` makes this assertion test two things, so if it fails it could be either due to component state not being created properly, or `ExpiryDate.from()` function not working correctly. I think it would be easier to understand if it fails only under one condition.  In order to avoid this we can either use `TEST_EXPIRY_DATE.take(2)` and `TEST_EXPIRY_DATE.takeLast(2)`, or maybe create two more constants for year and month and use it like:
```kotlin
private const val TEST_EXPIRY_MONTH = "03"
private const val TEST_EXPIRY_YEAR = "30"
private const val TEST_EXPIRY_DATE = "$TEST_EXPIRY_MONTH/$TEST_EXPIRY_YEAR"

// assertion
assertEquals(TEST_EXPIRY_MONTH, encryptedExpiryMonth)
```
What do you think?

checkout-core/src/test/java/com/adyen/checkout/core/ui/model/ExpiryDateTest.kt
```diff
@@ -0,0 +1,46 @@
...
+            arguments(ExpiryDate(12, 1234), "12/34"),
+            arguments(ExpiryDate(9, 2008), "09/08"),
+        )
+
+        @JvmStatic
+        fun createExpiryDateFromStringSource() = listOf(
+            arguments("01/01", ExpiryDate(1, 1)),
+            arguments("12/34", ExpiryDate(12, 34)),
+            arguments("09/08", ExpiryDate(9, 8)),
+        )
```
araratthehero: I would add some cases when the expiry date is not correct or is partial, to verify if it is generated correctly too.

card/src/test/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegateTest.kt
OscarSpruit: Totally agree with you, but the problem here is that the encryption internally converts the date to integers and back to strings. This causes the leading zeros to drop. For example, if we would use `take()`, then we would assert `"03"` against `"3"`. I couldn't quickly come up with a better solutions, so any ideas are welcome

card/src/main/java/com/adyen/checkout/card/internal/ui/StoredCardDelegate.kt
OscarSpruit: I get the point, but `ExpiryDate` uses integers internally and we get string values from the API. I thought it's a bit useless to convert the string values to integers and then convert them right back to a string. WDYT?

araratthehero: I see, my idea was to do something like this:
```
data class ExpiryDate(
    val expiryMonth: Int,
    val expiryYear: Int,
) {
    
    fun toMMyyString() = toMMyyString(expiryMonth.toString(), expiryYear.toString())

    ...
}


fun toMMyyString(expiryMonth: String, expiryYear: String): String {
    val monthDigits = expiryMonth.padStart(2, '0')
    val yearDigits = expiryYear.takeLast(2).padStart(2, '0')
    return "$monthDigits/$yearDigits"
}
```

So the formatting logic is in one function. Wdyt?

OscarSpruit: Oh like that it makes sense 👍

card/src/test/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegateTest.kt
ozgur00: What about if we use `trimStart` to remove the `0`s?
```kotlin
assertEquals(TEST_EXPIRY_MONTH.trimStart('0'), encryptedExpiryMonth)
assertEquals(TEST_EXPIRY_YEAR.trimStart('0'), encryptedExpiryYear)
```

ozgur00 (approved):

araratthehero (approved):

PR #2193, opened 2025-07-04
v5 - Disable co-badged for not supported brands

ozgur00:
## Description
This change makes sure we only display co-badged selection for cards that contain Bcmc, Carte Bancaire or Dankort. 

## Checklist 
- [ ] Code is unit tested

COSDK-529

## Release notes
[//]: # (### New)
[//]: # (### Fixed)
[//]: # (List fixes here)
### Fixed
- For cards, co-badged cards selection will not be offered for not supported brands.


card/src/main/java/com/adyen/checkout/card/internal/util/DualBrandedCardHandler.kt
```diff
@@ -64,4 +95,12 @@ class DualBrandedCardHandler(private val environment: Environment) {
         isSelected = isSelected,
         environment = environment,
     )
+
+    companion object {
+        private val SUPPORTED_CARD_BRANDS = listOf(
```
OscarSpruit, line 105: Wouldn't it be better to have an exclusion list? So that if any new dual branded cards are added older SDK version work right away and merchants don't need to update

ozgur00: I think it's a good point, however I had a discussion with product, and we thought to be on the safer side and have an inclusion list. Since the current situation is basically a new cobadged card type, which is not supported, so there's an issue. With exclusion list we could have a similar situation in our hands, and that we want to avoid by having an inclusion list.

PR #2209, opened 2025-07-10
v6 - Configuration improvements

OscarSpruit:
## Description
The main purpose of this PR is to simplify payment method specific configurations and their builders, but in addition to that quite some clean up happened:

- Configurations and their builders are separated -> no more complicated inheritance
- Fields that are already defined in `CheckoutConfiguration` are removed from specific configurations
- `isSubmitButtonVisible` moved to `CheckoutConfiguration` eliminating the need of `ButtonComponentParams` etc


core/src/main/java/com/adyen/checkout/core/components/CheckoutConfiguration.kt
```diff
@@ -155,3 +161,19 @@ class CheckoutConfiguration(
         }
     }
 }
+
+private fun Parcel.writeNullableBoolean(value: Boolean?) {
```
ozgur00: I think we can remove `nullable` part from the function name for these two functions to align with other functions such as `parcel.readString()` and `parcel.readSerializable()` which are nullable functions.

OscarSpruit: Replaced it with `readValue` and `writeValue` as Gemini suggested in the other PR. No more need for these functions

ozgur00 (approved):

PR #2210, opened 2025-07-11
v5 - Checkout configuration isSubmitButtonVisible

OscarSpruit:
## Description
Submit button visibility is now configurable on a global level through `CheckoutConfiguration`. The changes are backwards compatible.

Note: deprecation warning will be added in a follow up PR.

## Checklist 
- [x] Code is unit tested

### Deprecated
- The `setSubmitButtonVisible` method for configuration objects is now deprecated. Instead use `CheckoutConfiguration`:
```kotlin
CheckoutConfiguration(
    ...
    isSubmitButtonVisible = false
) {
    ...
}
```

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
@@ -110,7 +110,7 @@ class CheckoutConfiguration(
         clientKey = requireNotNull(parcel.readString()),
         amount = parcel.readParcelable(Amount::class.java.classLoader),
         analyticsConfiguration = parcel.readParcelable(AnalyticsConfiguration::class.java.classLoader),
-        isSubmitButtonVisible = parcel.readSerializable() as? Boolean?,
+        isSubmitButtonVisible = parcel.readValue(null) as? Boolean?,
     ) {
         val size = parcel.readInt()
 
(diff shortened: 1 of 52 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/CheckoutConfiguration.kt
```diff
@@ -51,15 +51,18 @@ import java.util.Locale
...
  */
+@Suppress("LongParameterList")
 @CheckoutConfigurationMarker
 class CheckoutConfiguration(
     override val environment: Environment,
     override val clientKey: String,
     override val shopperLocale: Locale? = null,
     override val amount: Amount? = null,
     override val analyticsConfiguration: AnalyticsConfiguration? = null,
+    val isSubmitButtonVisible: Boolean? = null,
```
jreij, line 65: Could this be a breaking change if anyone is passing the `configurationBlock` in the arguments?

OscarSpruit: For most people probably not, but it could be. I will add a secondary constructor and mark it deprecated

OscarSpruit: Marking it as deprecated is not possible, because with the default values the compiler doesn't seem to understand which constructor you are using.

PR #2215, opened 2025-07-14
v5 - Configuration deprecations

OscarSpruit:
## Description
- Add deprecation warnings to configurations, builder and setter methods
- Make builder `var`s public, so they replace the setter methods

## Checklist 

### Deprecated
- Configuration and configuration builder classes are now deprecated. Use `CheckoutConfiguration` instead.
- Configuration setter methods are replaced by property access syntax. For example:
```kotlin
CheckoutConfiguration(
    environment = environment,
    clientKey = clientKey,
) {
    card {
        // Deprecated
        setHolderNameRequired(true)
        // Replacement
        isHolderNameRequired = true
    }
}
```

### DIFF
--- a/card/api/card.api
+++ b/card/api/card.api
@@ -238,7 +238,6 @@ public final class com/adyen/checkout/card/CardConfiguration$Builder : com/adyen
 	public fun <init> (Ljava/util/Locale;Lcom/adyen/checkout/core/Environment;Ljava/lang/String;)V
 	public synthetic fun buildInternal ()Lcom/adyen/checkout/components/core/internal/Configuration;
 	public final fun getAddressConfiguration ()Lcom/adyen/checkout/card/AddressConfiguration;
-	public final fun getHolderNameRequired ()Ljava/lang/Boolean;
 	public final fun getInstallmentConfiguration ()Lcom/adyen/checkout/card/InstallmentConfiguration;
 	public final fun getKcpAuthVisibility ()Lcom/adyen/checkout/card/KCPAuthVisibility;
 	public final fun getShopperReference ()Ljava/lang/String;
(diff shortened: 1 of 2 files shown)


eps/api/eps.api
```diff
@@ -53,8 +53,6 @@ public final class com/adyen/checkout/eps/EPSConfiguration$Builder : com/adyen/c
 	public fun <init> (Ljava/util/Locale;Lcom/adyen/checkout/core/Environment;Ljava/lang/String;)V
 	public synthetic fun buildInternal ()Lcom/adyen/checkout/components/core/internal/Configuration;
 	public fun buildInternal ()Lcom/adyen/checkout/eps/EPSConfiguration;
-	public fun setHideIssuerLogos (Z)Lcom/adyen/checkout/eps/EPSConfiguration$Builder;
```
jreij, line 56: Are these removed lines a breaking change?

3ds2/api/3ds2.api
jreij, line 1: Release notes: should it be `holderNameRequired` or `isHolderNameRequired`?

OscarSpruit: The builder uses `holderNameRequired`

eps/api/eps.api
OscarSpruit: Should not be the case. The `EPSConfiguration.Builder` overrides this method and just calls super. Making it redundant and since the super type has this method you can still call it on `EPSConfiguration.Builder`

ozgur00 (approved):

3ds2/api/3ds2.api
jreij: Are we gonna keep it `holderNameRequired` in v6? if we're going to change it to `isHolderNameRequired` (to align with other booleans), can we already rename it in v5 so that it doesn't become a breaking change when migrating to v6?

OscarSpruit: Makes sense 👍

PR #2216, opened 2025-07-15
v6 - Checkout controller

OscarSpruit:
## Description
Introduce basic implementation of checkout controller. This is used by merchants to control the checkout flow.
The `ObjectInPublicSealedClass` was updated as well to correctly check parent visibility.

COSDK-511

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
@@ -30,6 +30,8 @@ class CheckoutController {
         _events.tryEmit(Event.Submit)
     }
 
+    // TODO - discuss if this should be in the controller or somewhere else, for if the merchant wants to use our
+    //   component to only handle actions
     fun handleAction(action: Action) {
         _events.tryEmit(Event.HandleAction(action))
     }
--- a/core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 15/7/2025.
+ */
+
+package com.adyen.checkout.core.components
+
+import android.content.Intent
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import com.adyen.checkout.core.action.data.Action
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
```diff
@@ -25,31 +27,49 @@ internal class PaymentFacilitator(
...
+        // TODO - what if we are handling an action?
+        paymentDelegate.submit()
+    }
+
+    private fun handleAction(action: Action) {
         // TODO - Store the actionDelegate
         actionProvider.get(
             action = action,
             coroutineScope = coroutineScope,
         )
```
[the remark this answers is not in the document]
OscarSpruit: This TODO wasn't added in this PR and will be solved later

core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+
+// TODO - KDocs
+class CheckoutController {
+
+    private val _events = MutableSharedFlow<Event>(
+        replay = 0,
```
araratthehero: Nits: This line could be removed from here, since the default value for a replay is 0

core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
```diff
@@ -0,0 +1,50 @@
...
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST,
+    )
+    internal val events: SharedFlow<Event> = _events.asSharedFlow()
+
+    fun submit() {
+        _events.tryEmit(Event.Submit)
+    }
+
+    fun handleAction(action: Action) {
```
ozgur00, line 35: I don't think we need handle action here, since we will be getting that as a return type from the merchants in the callback.

OscarSpruit: What if a merchant just wants to use our component to handle actions? This would give the possibility to do that, but maybe we could think about other solutions as well

ozgur00: I suggest we discuss the alternatives for only handling actions. For now you can leave a `TODO` here for that maybe?

ozgur00 (approved):

PR #2217, opened 2025-07-15
V6 - Action - Add await action

araratthehero:
## Description
Add support for Await action and provide a simple `ViewFactory` composable

## Checklist 

COSDK-480

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/action/data/AwaitAction.kt
+++ b/core/src/main/java/com/adyen/checkout/core/action/data/AwaitAction.kt
@@ -14,11 +14,11 @@ import org.json.JSONException
 import org.json.JSONObject
 
 @Parcelize
-class AwaitAction(
-    override var type: String? = null,
-    override var paymentData: String? = null,
-    override var paymentMethodType: String? = null,
-    var url: String? = null,
+data class AwaitAction(
+    override val type: String?,
+    override val paymentData: String?,
+    override val paymentMethodType: String?,
+    val url: String?,
 ) : Action() {
 
     companion object {
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -35,17 +35,20 @@ public final class com/adyen/checkout/core/action/data/AwaitAction : com/adyen/c
 	public static final field CREATOR Landroid/os/Parcelable$Creator;
 	public static final field Companion Lcom/adyen/checkout/core/action/data/AwaitAction$Companion;
 	public static final field SERIALIZER Lcom/adyen/checkout/core/common/internal/model/ModelObject$Serializer;
-	public fun <init> ()V
 	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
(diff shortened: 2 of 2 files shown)


await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitFactory.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.adyen.checkout.core.components.CheckoutConfiguration
+import kotlinx.coroutines.CoroutineScope
+
+internal class AwaitFactory : ActionFactory<AwaitDelegate> {
+
+    override fun create(
+        coroutineScope: CoroutineScope,
+        checkoutConfiguration: CheckoutConfiguration,
+        savedStateHandle: SavedStateHandle
+    ): AwaitDelegate = AwaitDelegate()
```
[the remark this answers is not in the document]
araratthehero: Will be done in a future PR

core/src/main/java/com/adyen/checkout/core/action/internal/ActionDelegateProvider.kt
```diff
@@ -8,24 +8,21 @@
...
-        action: Action,
+        actionType: String,
         factory: ActionFactory<*>,
     ) {
-        val actionType = action.type
-        require(!actionType.isNullOrEmpty()) {
-            "The action type cannot be empty or null"
-        }
-
         factories[actionType] = factory
```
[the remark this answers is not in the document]
araratthehero: From the point of `ActionDelegateProvider` an empty should be accepted

core/src/main/java/com/adyen/checkout/core/action/data/AwaitAction.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import com.adyen.checkout.core.common.exception.ModelSerializationException
+import com.adyen.checkout.core.common.internal.model.getStringOrNull
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+class AwaitAction(
+    override var type: String? = null,
```
OscarSpruit: These should be `val` I think. And what do you think about removing the default values? The default values only make it easier to make a faulty await action object

core/src/main/java/com/adyen/checkout/core/action/data/AwaitAction.kt
```diff
@@ -0,0 +1,53 @@
...
+package com.adyen.checkout.core.action.data
+
+import com.adyen.checkout.core.common.exception.ModelSerializationException
+import com.adyen.checkout.core.common.internal.model.getStringOrNull
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+class AwaitAction(
```
OscarSpruit: Could this be a `data class`? For easier testing

core/src/main/java/com/adyen/checkout/core/action/data/AwaitAction.kt
araratthehero: I noticed and fixed it in [this PR](https://github.com/Adyen/adyen-android/pull/2218), but I will include it here, since it should be part of this PR.

araratthehero: Oh I cannot find it there either. Will fix it, thanks for pointing out!

OscarSpruit (approved):

PR #2218, opened 2025-07-16
V6 - Rename delegate to component

araratthehero:
## Description
- Renamed `Delegate` to `Component` in the `core` and `mbway` modules.
- Renamed `ComponentState` to `PaymentComponentState`

## Checklist 

COSDK-480


mbway/src/test/java/com/adyen/checkout/mbway/internal/ui/state/MBWayPaymentComponentStateFactoryTest.kt
```diff
@@ -21,31 +21,31 @@ import org.mockito.kotlin.whenever
 import java.util.Locale
 
 @ExtendWith(MockitoExtension::class)
-internal class MBWayDelegateStateFactoryTest(
+internal class MBWayPaymentComponentStateFactoryTest(
```
ozgur00: Shouldn't this class be named `MBWayComponentStateFactoryTest`?

core/src/main/java/com/adyen/checkout/core/components/internal/PaymentMethodFactory.kt
```diff
@@ -10,21 +10,21 @@ package com.adyen.checkout.core.components.internal
...
 import androidx.annotation.RestrictTo
 import com.adyen.checkout.core.components.CheckoutConfiguration
-import com.adyen.checkout.core.components.internal.ui.PaymentDelegate
+import com.adyen.checkout.core.components.internal.ui.PaymentComponent
 import com.adyen.checkout.core.sessions.internal.model.SessionParams
 import kotlinx.coroutines.CoroutineScope
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-interface PaymentMethodFactory<CS : BaseComponentState, T : PaymentDelegate<CS>> {
+interface PaymentMethodFactory<CS : BaseComponentState, T : PaymentComponent<CS>> {
```
ozgur00: I think we should rename `BaseComponentState` to `BasePaymentComponentState` as well.

mbway/src/test/java/com/adyen/checkout/mbway/internal/ui/state/MBWayPaymentComponentStateFactoryTest.kt
araratthehero: Good catch!

PR #2232, opened 2025-07-18
v6 - Sessions example

OscarSpruit:
## Description
Create example integration for v6 components with sessions

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsViewModel.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsViewModel.kt
@@ -31,6 +31,7 @@ internal class V6SessionsViewModel @Inject constructor(
     private val keyValueStorage: KeyValueStorage,
 ) : ViewModel() {
 
+    // TODO - Replace with checkoutConfigurationProvider once it's updated COSDK-563
     private val configuration = CheckoutConfiguration(
         Environment.TEST,
         BuildConfig.CLIENT_KEY,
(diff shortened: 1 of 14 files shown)


araratthehero (approved):

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsViewModel.kt
```diff
@@ -0,0 +1,77 @@
...
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+@HiltViewModel
+internal class V6SessionsViewModel @Inject constructor(
+    private val paymentsRepository: PaymentsRepository,
+    private val keyValueStorage: KeyValueStorage,
+) : ViewModel() {
+
+    private val configuration = CheckoutConfiguration(
```
ozgur00, line 35: We can inject `checkoutConfigurationProvider` here.

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsViewModel.kt
```diff
@@ -0,0 +1,77 @@
...
+        val session = paymentsRepository.createSession(
+            getSessionRequest(
+                merchantAccount = keyValueStorage.getMerchantAccount(),
+                shopperReference = keyValueStorage.getShopperReference(),
+                amount = keyValueStorage.getAmount(),
+                countryCode = keyValueStorage.getCountry(),
+                shopperLocale = keyValueStorage.getShopperLocale(),
+                splitCardFundingSources = keyValueStorage.isSplitCardFundingSources(),
+                threeDSMode = keyValueStorage.getThreeDSMode(),
+                // TODO - We don't need this for now
```
ozgur00: Nit: We can leave a better TODO message about what to do when we support redirect.

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsViewModel.kt
OscarSpruit: `CheckoutConfigurationProvider ` uses the old `CheckoutConfiguration` class, so we can either add support for the new class there or add a TODO here to change it later. WDYT?

ozgur00: Leaving a TODO is good enough at this stage then. I have created a ticket for updating example integrations. So we can leave a TODO and maybe mention that ticket here?

PR #2245, opened 2025-07-23
v5 - Deprecate old 3DS2 actions

OscarSpruit:
## Description
The old 2 action 3DS2 flow is deprecated for v5. To communicate this clearly the specific action classes are now marked deprecated.

COSDK-569

### Deprecated
- The `Threeds2FingerprintAction` and `Threeds2ChallengeAction` actions are now deprecated. Use `Threeds2Action` instead.

### DIFF
--- a/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/provider/Adyen3DS2ComponentProvider.kt
+++ b/3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/provider/Adyen3DS2ComponentProvider.kt
@@ -132,6 +132,7 @@ constructor(
     }
 
     override val supportedActionTypes: List<String>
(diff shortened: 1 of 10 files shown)


3ds2/src/main/java/com/adyen/checkout/adyen3ds2/internal/provider/Adyen3DS2ComponentProvider.kt
jreij, line 1: Shall we deprecate those for v6 as well? Or even remove them completely already?

OscarSpruit: I would say remove them completely, but we can do that as soon as we work on 3DS2 for v6

PR #2260, opened 2025-07-31
V6 - Await - Await factory

araratthehero:
## Description
Add await factory implementation

COSDK-547

### DIFF
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
@@ -8,6 +8,7 @@
 
 package com.adyen.checkout.await.internal.ui
 
+import android.content.Context
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import com.adyen.checkout.await.internal.ui.view.AwaitComponent
@@ -21,6 +22,10 @@ internal class AwaitComponent : ActionComponent {
 
     private val eventChannel = bufferedChannel<ActionComponentEvent>()
     override val eventFlow: Flow<ActionComponentEvent> = eventChannel.receiveAsFlow()
+
+    override fun handleAction(context: Context) {
+        // TODO - Implement handleAction
+    }
 
     @Composable
     override fun ViewFactory(modifier: Modifier) {
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
@@ -12,13 +12,29 @@ import android.content.Context
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import com.adyen.checkout.await.internal.ui.view.AwaitComponent
+import com.adyen.checkout.core.action.data.AwaitAction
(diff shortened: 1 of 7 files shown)


await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitFactory.kt
```diff
@@ -9,15 +9,70 @@
...
+            deviceLocale = Locale.US,
+            dropInOverrideParams = null,
+            componentSessionParams = null,
+        ).commonComponentParams
+
+        // TODO - Analytics. We might need to change the logic on AnalyticsManager creation.
+        val analyticsManager = AnalyticsManagerFactory().provide(
+            componentParams = componentParams,
+            application = null,
+            source = AnalyticsSource.PaymentComponent(PaymentMethodTypes.MB_WAY),
```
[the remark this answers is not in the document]
araratthehero: I added a comment to make sure to change it.

await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
```diff
@@ -8,20 +8,41 @@
...
+import com.adyen.checkout.core.components.internal.ui.model.ComponentParams
+import com.adyen.checkout.core.redirect.internal.RedirectHandler
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.receiveAsFlow
 
-internal class AwaitComponent : ActionComponent {
+@Suppress("UnusedPrivateProperty", "LongParameterList")
+internal class AwaitComponent(
+    private val action: AwaitAction,
```
OscarSpruit, line 30: Wouldn't it be better to pass the action in `handleAction`? This would offer support for a action component to handle multiple actions while keeping state. The old 3DS2 flow requires this for example, though we will not support this flow anymore in v6. Maybe in the future we will need it again. WDYT?

await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
```diff
@@ -8,20 +8,41 @@
...
+    private val analyticsManager: AnalyticsManager,
+    private val redirectHandler: RedirectHandler,
+    private val statusRepository: StatusRepository,
+    private val paymentDataRepository: PaymentDataRepository,
+) : ActionComponent {
 
     private val eventChannel = bufferedChannel<ActionComponentEvent>()
     override val eventFlow: Flow<ActionComponentEvent> = eventChannel.receiveAsFlow()
 
+    override fun handleAction(context: Context) {
```
OscarSpruit, line 42: Is `context` really needed here? I think we could probably send an event to the UI in case of navigation and then we don't have to pass around the context

araratthehero: As discussed I can do this in a future PR

await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
araratthehero: As discussed I added a TODO to discuss this later

OscarSpruit (approved):

PR #2262, opened 2025-07-31
V6 - Await - Await component

araratthehero:
## Description
Await component implementation

## Checklist 

COSDK-547

### DIFF
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
@@ -30,7 +30,6 @@ import kotlinx.coroutines.flow.receiveAsFlow
 internal class AwaitComponent(
     private val action: AwaitAction,
     private val coroutineScope: CoroutineScope,
-    private val componentParams: ComponentParams,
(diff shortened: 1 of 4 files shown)


await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
```diff
@@ -9,42 +9,180 @@
...
+
+        // Redirect flow starts polling after it launched a redirect
+        if (!shouldLaunchRedirect(action)) {
+            startStatusPolling(paymentData)
+        }
+    }
+
+    private fun shouldLaunchRedirect(action: AwaitAction) = !action.url.isNullOrEmpty()
+
+    private fun startStatusPolling(paymentData: String) {
```
OscarSpruit, line 111: Something for a future PR, I think we could extract this polling logic into a separate class as it is somewhat duplicated in some other components

araratthehero: Will add a TODO to address it in a future PR

OscarSpruit (approved):

PR #2273, opened 2025-08-05
Specify release workflow permissions

OscarSpruit:
Set minimal permissions for all workflows related to creating releases.

[6 alerts left](https://github.com/Adyen/adyen-android/security/code-scanning?query=is%3Aopen+branch%3Achore%2Frelease_workflow_permissions)

COSDK-583

### DIFF
--- a/.github/workflows/create_github_release.yml
+++ b/.github/workflows/create_github_release.yml
@@ -7,6 +7,9 @@ on:
         required: true
         type: string
 
+permissions:
+  contents: write
+
 jobs:
   create_github_release:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


.github/workflows/create_github_release.yml
```diff
@@ -7,6 +7,9 @@ on:
         required: true
         type: string
 
+permissions:
+  contents: write
```
araratthehero, line 11: Should this also have `deployments: write`, since it is creating a Github release?

OscarSpruit: As I understand `deployments` is only needed for example for when we deploy our Dokka pages to Github pages

araratthehero: I see. In that case let's keep as it is and we can add if release process fails.

PR #2275, opened 2025-08-06
v5 - UPI improvements

OscarSpruit:
## Description
- Remove QR code flow
- Move UPI ID input to it's own tab instead in the list with apps
- Update copy

| Intent | ID |
|--------|--------|
| Screenshot_20250806_165253 | Screenshot_20250806_165325 | 

## Checklist 
- [x] Code is unit tested

COSDK-451

### Changed
- For UPI, the QR code option is removed and the UPI ID input moved to a more intuitive place.

### DIFF
--- a/upi/src/main/res/template/values/strings.xml.tt
+++ b/upi/src/main/res/template/values/strings.xml.tt
@@ -11,10 +11,11 @@
     <string name="checkout_upi_mode_selection">%%upi.modeSelection%%</string>
     <string name="checkout_upi_vpa_validation">%%upi.collect.field.invalid-id-error%%</string>
     <string name="checkout_upi_continue">%%continue%%</string>
-    <string name="checkout_upi_id">%%upi.mode.enterUpiId%%</string>
+    <string name="checkout_upi_id_label">%%upi.collect.field.label%%</string>
     <string name="checkout_upi_app_other_upi">%%upi.mode.otherUpi%%</string>
     <string name="checkout_upi_no_app_selected">%%upi.error.noAppSelected%%</string>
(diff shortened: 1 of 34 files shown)


upi/src/main/res/template/values/strings.xml.tt
```diff
@@ -9,17 +9,12 @@
 
 <resources>
     <string name="checkout_upi_mode_selection">%%upi.modeSelection%%</string>
-    <string name="checkout_upi_vpa_validation">%%upi.vpa.validationMessage%%</string>
-    <string name="checkout_upi_qr_generation">%%upi.qrcode.generationMessage%%</string>
+    <string name="checkout_upi_vpa_validation">%%upi.collect.field.invalid-id-error%%</string>
     <string name="checkout_upi_continue">%%continue%%</string>
-    <string name="checkout_upi_app_manual_input">%%upi.mode.enterUpiId%%</string>
+    <string name="checkout_upi_id">%%upi.mode.enterUpiId%%</string>
```
araratthehero: Does `checkout_upi_id` need to be translated? I see in some languages they change to `UPI-ID`, `ID UPI`, etc.

Similarly for other strings, they contain values like `UPI app`, `UPI ID`, etc. Those values change based on localization. I wonder if it is okay to translate those values or to change them for other languages.

OscarSpruit: Good point, let me check with the team

OscarSpruit: We will keep these strings translated as the terms "app" and "ID" might not be known in every language

araratthehero (dismissed):

araratthehero (approved):

PR #2277, opened 2025-08-07
V6 - Await - Improvements

araratthehero:
## Description
- Removed `context` from `handleAction`.
- Implemented necessary logic in `PaymentFacilitator`
- Moved redirect logic to its own `RedirectEvent` composable

## Checklist 

COSDK-547

### DIFF
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
@@ -28,8 +28,8 @@ import com.adyen.checkout.core.components.internal.data.api.helper.isFinalResult
 import com.adyen.checkout.core.components.internal.data.model.StatusResponse
 import com.adyen.checkout.core.components.internal.ui.StatusPollingComponent
 import com.adyen.checkout.core.redirect.internal.RedirectHandler
-import com.adyen.checkout.core.redirect.internal.ui.RedirectEvent
 import com.adyen.checkout.core.redirect.internal.ui.RedirectViewEvent
+import com.adyen.checkout.core.redirect.internal.ui.redirectEvent
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.Flow
@@ -59,7 +59,6 @@ internal class AwaitComponent(
     private val redirectEventChannel = bufferedChannel<RedirectViewEvent>()
     private val redirectEventFlow: Flow<RedirectViewEvent> = redirectEventChannel.receiveAsFlow()
 
-    // TODO - Remove context from here and launch redirect from the ViewFactory composable
     override fun handleAction() {
         paymentDataRepository.paymentData = action.paymentData
 
@@ -187,7 +186,7 @@ internal class AwaitComponent(
 
     @Composable
     override fun ViewFactory(modifier: Modifier) {
-        RedirectEvent(
+        redirectEvent(
             redirectHandler = redirectHandler,
             viewEventFlow = redirectEventFlow,
             onError = ::emitError,
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
@@ -80,7 +80,7 @@ internal class AwaitComponent(
         val url = action.url
         try {
             adyenLog(AdyenLogLevel.DEBUG) { "makeRedirect - $url" }
-            redirectHandler.launchUriRedirect(context, url)
+            redirectHandler.launchUriRedirect(context, url!!)
             val paymentData = paymentDataRepository.paymentData
                 // TODO - Error Propagation
                 // ?: throw CheckoutException("Payment data should not be null")
--- a/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
+++ b/await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
@@ -8,7 +8,6 @@
 
 package com.adyen.checkout.await.internal.ui
 
-import android.content.Context
 import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
@@ -29,6 +28,8 @@ import com.adyen.checkout.core.components.internal.data.api.helper.isFinalResult
 import com.adyen.checkout.core.components.internal.data.model.StatusResponse
 import com.adyen.checkout.core.components.internal.ui.StatusPollingComponent
 import com.adyen.checkout.core.redirect.internal.RedirectHandler
+import com.adyen.checkout.core.redirect.internal.ui.RedirectEvent
+import com.adyen.checkout.core.redirect.internal.ui.RedirectViewEvent
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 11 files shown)


await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
```diff
@@ -65,22 +69,22 @@ internal class AwaitComponent(
...
+        if (!url.isNullOrEmpty()) {
+            makeRedirect(url)
         }
     }
 
     @Suppress("TooGenericExceptionCaught", "TooGenericExceptionThrown")
-    private fun makeRedirect(action: AwaitAction, context: Context) {
-        val url = action.url
+    private fun makeRedirect(redirectUrl: String) {
+        redirectEventChannel.trySend(RedirectViewEvent.Redirect(redirectUrl))
```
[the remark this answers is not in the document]
araratthehero: The channel cannot be full, since it is buffered and by default the buffered value is 1

await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
```diff
@@ -55,8 +56,11 @@ internal class AwaitComponent(
     private val eventChannel = bufferedChannel<ActionComponentEvent>()
     override val eventFlow: Flow<ActionComponentEvent> = eventChannel.receiveAsFlow()
 
+    private val redirectEventChannel = bufferedChannel<RedirectViewEvent>()
+    private val redirectEventFlow: Flow<RedirectViewEvent> = redirectEventChannel.receiveAsFlow()
+
     // TODO - Remove context from here and launch redirect from the ViewFactory composable
```
OscarSpruit: This TODO can now be removed, right?

core/src/main/java/com/adyen/checkout/core/redirect/internal/ui/RedirectEvent.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.compose.ui.platform.LocalContext
+import com.adyen.checkout.core.common.AdyenLogLevel
+import com.adyen.checkout.core.common.internal.helper.adyenLog
+import com.adyen.checkout.core.redirect.internal.RedirectHandler
+import kotlinx.coroutines.flow.Flow
+
+@Suppress("TooGenericExceptionCaught")
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Composable
+fun RedirectEvent(
```
OscarSpruit: Because it's not a composable that renders UI, shouldn't it be starting with lowercase?

core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
```diff
@@ -54,24 +56,24 @@ internal class PaymentFacilitator(
...
+                handleResult(result, lifecycle)
             }.launchIn(coroutineScope)
 
         checkoutController.events
             .flowWithLifecycle(lifecycle)
             .onEach { event ->
                 when (event) {
                     CheckoutController.Event.Submit -> submit()
-                    is CheckoutController.Event.HandleAction -> handleAction(event.action)
+                    is CheckoutController.Event.HandleAction -> handleAction(event.action, lifecycle)
```
OscarSpruit, line 67: Wondering if passing the lifecycle here could cause a memory leak 🤔

await/src/main/java/com/adyen/checkout/await/internal/ui/AwaitComponent.kt
araratthehero: Correct. Will remove it.

core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
araratthehero: I do not think so. We do not store Lifecycle object as a long lived class property, especially in a class which is held by a ViewModel. `lifecycle` is being passed through the `observe()` function from the UI directly and it only exists on the call stack for the duration of `observe` and `handleAction` functions.

In the case of the `onEach {}` function, when the lifecycle is not active anymore `flowWithLifecycle()` will stop calling the `onEach {}` function, so we can assume that the `handleAction()` will only be called when the lifecycle is valid.

araratthehero: I thought about cleaning the observers too and we do not need to store the jobs for `paymentComponent.eventFlow` and `checkoutController.events` in `observe()`. Though we need to store the job for `actionComponent.eventFlow` (currently `actionObservationJob`), since every time `handleAction()` is called, a new `actionComponent` is created. And in case the ViewModel gets destroyed, the `viewModelScope` will cause all flow collections to be cancelled too.

araratthehero: Is there something else you think could cause a leak?

OscarSpruit: > In the case of the onEach {} function, when the lifecycle is not active anymore flowWithLifecycle() will stop calling the onEach {} function

But will it also release the reference to the lifecycle? I'm afraid it will keep the reference, because the flow follows the `viewModelScope` which can outlive the activity or fragment

araratthehero: What I understand is that `flowWithLifecycle` creates a new downstream Flow and `onEach` operator collects from that new flow. So when the lifecycle state is `STARTED`, it starts collecting from the upstream flow, creates a new coroutine and passes to the downstream flow to the `onEach` block. When lifecycle goes to the `DESTROYED` state, the `flowWithLifecycle` cancels the collection of the upstream flow and cancels the downstream flow (which should cancel the coroutine at this point). Since the `onEach` lambda is part of the terminated collection, the lambda and the reference to the `lifecycle` should become eligible for garbage collection.

[Here](https://developer.android.com/topic/libraries/architecture/coroutines#:~:text=//%20repeatOnLifecycle%20launches%20the%20block%20in%20a%20new%20coroutine%20every%20time%20the%0A%20%20%20%20%20%20%20%20%20%20%20%20//%20lifecycle%20is%20in%20the%20STARTED%20state%20(or%20above)%20and%20cancels%20it%20when%20it%27s%20STOPPED.) it is mentioned that:

> repeatOnLifecycle launches the block in a new coroutine every time the
> lifecycle is in the STARTED state (or above) and cancels it when it's STOPPED. 

And [here](https://developer.android.com/reference/kotlin/androidx/lifecycle/package-summary#(androidx.lifecycle.Lifecycle).repeatOnLifecycle(androidx.lifecycle.Lifecycle.State,kotlin.coroutines.SuspendFunction1):~:text=Lifecycle.State%20in%20which%20block%20runs%20in%20a%20new%20coroutine.%20That%20coroutine%20will%20cancel%20if%20the%20lifecycle%20falls%20below%20that%20state%2C%20and%20will%20restart%20if%20it%27s%20in%20that%20state%20again.) it is mentioned:

> [Lifecycle.State](https://developer.android.com/reference/kotlin/androidx/lifecycle/Lifecycle.State) in which block runs in a new coroutine. That coroutine will cancel if the lifecycle falls below that state, and will restart if it's in that state again.

The [repeatOnLifecycle](https://developer.android.com/reference/kotlin/androidx/lifecycle/package-summary#(androidx.lifecycle.Lifecycle).repeatOnLifecycle(androidx.lifecycle.Lifecycle.State,kotlin.coroutines.SuspendFunction1)) is called in [flowWithLifecycle](https://developer.android.com/reference/kotlin/androidx/lifecycle/package-summary#flowwithlifecycle).

OscarSpruit: Ah perfect! Thanks for sharing that

PR #2292, opened 2025-08-19
V6 - Move button inside component

araratthehero:
## Description
- Moved button from PaymentFacilitator to Components
- Created StandardComponent class to have a base screen for all Components

COSDK-597


ui/src/main/java/com/adyen/checkout/ui/internal/StandardComponent.kt
```diff
@@ -0,0 +1,59 @@
...
+
+            // TODO - Properly implement the pay button
+            PrimaryButton(
+                onClick = onButtonClick,
+                text = "Pay $13.37",
+                modifier = Modifier.fillMaxWidth(),
+            )
+        }
+    }
+}
```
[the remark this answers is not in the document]
araratthehero: Hmm, we will discuss this

core/src/main/java/com/adyen/checkout/core/components/AdyenPaymentFlow.kt
```diff
@@ -45,21 +38,6 @@ fun AdyenPaymentFlow(
...
-
-            // TODO - Properly implement the pay button
-            // TODO - Should we move the payment button to each component
-            PrimaryButton(
-                onClick = checkoutController::submit,
-                text = "Pay $13.37",
-                modifier = Modifier.fillMaxWidth(),
-            )
-        }
+        adyenComponent.ViewFactory(modifier)
```
ozgur00, line 41: I think we should make all components fill max width here, otherwise we will have to do it in each components' view factory one by one. What do you think?  🤔

araratthehero: What do you think if we do it in `StandardComponent` composable, so it applies to all components?

ozgur00: Good idea. It makes sense to do it there.

ui/src/main/java/com/adyen/checkout/ui/internal/StandardComponent.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun StandardComponent(
```
ozgur00: To me `StandardComponent` sounds like it's a generic type of component, like the else case of our components. I think we should rename it to something like `ComponentContainer` (something that indicates that it wraps around our actual components)

araratthehero: I thought about naming it `Component`, since it's the main composable for all components, but the name was too generic. What do you think of the name `Component`?

OscarSpruit: What about `ComponentScaffold`?

araratthehero: I tried that, but thought of changing it, since it is a base Component and not actually a scaffold. I thought it might be a bit confusing. But if you think it is a better idea I don't mind having it `ComponentScaffold`.

ui/src/main/java/com/adyen/checkout/ui/internal/StandardComponent.kt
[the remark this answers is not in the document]
araratthehero: This is not correct, we have two use-cases. One is when `StandardComponent` does not have a button, thus does not need a `onButtonClick` listener and the other one is when there is a button, which has a value by default, but if necessary, a custom button can be passed. In this second case the `onButtonClick` is required.

ui/src/main/java/com/adyen/checkout/ui/internal/StandardComponent.kt
araratthehero: As discussed changed it to `ComponentScaffold`.

ui/src/main/java/com/adyen/checkout/ui/internal/PayButton.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+@Composable
+fun PayButton(
+    onButtonClick: () -> Unit,
+) {
+    // TODO - Properly implement the pay button
```
ozgur00: We can remove this `TODO` I think. @OscarSpruit the todo was about this change right?

ozgur00 (dismissed):

ozgur00 (approved):

ui/src/main/java/com/adyen/checkout/ui/internal/PayButton.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package com.adyen.checkout.ui.internal
+
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+
+@Composable
+fun PayButton(
+    onButtonClick: () -> Unit,
```
OscarSpruit: Nit: `onButtonClick` could just be `onClick`

PR #2293, opened 2025-08-19
v6 - Unify callbacks for advanced and sessions flows

ozgur00:
## Description
Improve `CheckoutCallbacks` class to support advanced and sessions flow by making it a common callback class that could translate itself into `AdvancedComponentCallbacks` or `SessionsComponentCallbacks` internally. 

## Checklist 

COSDK-144

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/common/PaymentResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/PaymentResult.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/adyen/checkout/core/common/PaymentResult.kt
```diff
@@ -0,0 +1,21 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 15/8/2025.
+ */
+
+package com.adyen.checkout.core.common
+
+import com.adyen.checkout.core.components.data.OrderResponse
+
+data class PaymentResult(
```
OscarSpruit, line 14: Should we add a TODO for KDocs here?

PR #2300, opened 2025-08-25
v6 - Component loading state

OscarSpruit:
## Description
After submitting the component the button now goes into a loading state and the component cannot be interacted with. 

https://github.com/user-attachments/assets/33a181d2-c434-433c-8e47-490419f302f3

COSDK-610

### DIFF
--- a/mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
+++ b/mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
@@ -10,8 +10,6 @@ package com.adyen.checkout.mbway.internal.ui
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.adyen.checkout.core.analytics.internal.AnalyticsManager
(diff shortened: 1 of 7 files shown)


ui/src/main/java/com/adyen/checkout/ui/internal/ComponentScaffold.kt
```diff
@@ -8,27 +8,55 @@
...
-        Spacer(Modifier.size(16.dp))
-        footer()
+    val focusManager = LocalFocusManager.current
+    LaunchedEffect(disableInteraction) {
+        if (disableInteraction) {
+            focusManager.clearFocus()
+        }
+    }
+
+    Box {
```
araratthehero, line 40: Is a `Box` necessary here?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
```diff
@@ -77,8 +80,11 @@ internal class MBWayComponent(
             .stateIn(coroutineScope, SharingStarted.Lazily, stateManager.state.value.toViewState())
     }
 
+    private var isLoading by mutableStateOf(false)
```
ozgur00: I think we can include this in the `ViewState` ( we probably will need to add it to the internal `ComponentState` as well since view state is generated based on that), then get it from there to pass it to the `PayButton`.

ui/src/main/java/com/adyen/checkout/ui/internal/ComponentScaffold.kt
OscarSpruit: Unfortunately, it is. If this `Box` would have a modifier that intercepts the touch event, then it doesn't work. It only works when the `Column` and another container are at the same hierarchy level.

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
OscarSpruit: Nice idea, let me try it out

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
```diff
@@ -77,8 +80,11 @@ internal class MBWayComponent(
...
+
     override fun submit() {
         if (stateManager.isValid) {
+            isLoading = true
```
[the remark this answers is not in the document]
OscarSpruit: We will resolve this in a follow up PR

PR #2304, opened 2025-08-27
v6 - Move ImageLoader related classes to core and ui modules

ozgur00:
## Description
Move ImageLoader from `checkout-core` to core and ui modules. Before, `ui-core` depended on `checkout-core`, and now `ui` is the base module so we need to split the implementation of `ImageLoader`, so the interface is in ui module and other ui classes can depend on it where the concrete implementation `DefaultImageLoader` which makes the api calls would be in `core` module. The disadvantage this brings is that we cannot define the global `localImageLoader` parameter in `ui` module (it will be in `core`), so whenever we initialize an image we will have to pass it manually. 

## Checklist 

COSDK-518


core/src/main/java/com/adyen/checkout/core/common/internal/DefaultImageLoader.kt
```diff
@@ -0,0 +1,115 @@
...
+            }
+            return@withContext
+        }
+
+        val request = Request.Builder()
+            .url(url)
+            .get()
+            .build()
+
+        val call = okHttpClient.newCall(request)
```
araratthehero, line 62: If I undestand correctly we need to place the default implementation of the ImageLoader in `core` because of the `okhttp` right?

An idea I have is that we could perhaps move this class to the ui module, but inject the `okhttp` instance from core. We will still have to create the instance and the static object in core, but the rest of the logic could move to UI. Wdyt?

ozgur00: My main reasoning was more to hide the implementation details of image loading from the `ui` module. If at some point we need to change how we load the image, we could just change the implementation in `core` without the need the change the contract between them, so we won't have to touch the `ui` module.

core/src/main/java/com/adyen/checkout/core/common/internal/DefaultImageLoader.kt
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            response.body?.close()
+        } catch (e: CancellationException) {
+            call.cancel()
+        } catch (e: IOException) {
+            withContext(DispatcherProvider.Main) {
+                onError(e)
+            }
+        }
```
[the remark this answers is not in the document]
OscarSpruit: This is a really good suggestion actually, maybe we can apply it already?

core/src/main/java/com/adyen/checkout/core/common/internal/DefaultImageLoader.kt
araratthehero: Indeed, makes sense

core/src/main/java/com/adyen/checkout/core/common/internal/DefaultImageLoader.kt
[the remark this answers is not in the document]
ozgur00: Agreed, done

PR #2305, opened 2025-08-27
v6 - Add image composable functions

ozgur00:
## Description
The composables included in this PR replaces the `ImageView.load()` functions in `ImageLoadingExtensions` file.

## Checklist 

COSDK-518


ui/src/main/java/com/adyen/checkout/ui/internal/Images.kt
```diff
@@ -0,0 +1,181 @@
...
+ * @param placeholder A painter to show while the image is loading.
+ * @param errorFallback A painter to show if the image fails to load.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Composable
+fun NetworkImage(
+    url: String,
+    modifier: Modifier = Modifier,
+    contentDescription: String?,
+    imageLoader: ImageLoader,
```
OscarSpruit, line 54: Maybe we can obtain the `ImageLoader` inside of this composable? We always have to same one anyway and it would be quite some boilerplate code needed everywhere this composable would be used.

ui/src/main/java/com/adyen/checkout/ui/internal/Images.kt
```diff
@@ -0,0 +1,181 @@
...
+    txVariant: String,
+    modifier: Modifier = Modifier,
+    txSubVariant: String = "",
+    size: LogoSize = LogoSize.SMALL,
+    contentDescription: String? = "Payment method logo",
+    imageLoader: ImageLoader,
+    @DrawableRes placeholder: Int = R.drawable.ic_placeholder_image,
+    @DrawableRes errorFallback: Int = R.drawable.ic_placeholder_image,
+) {
+    // We can reuse the same logo path building logic.
```
OscarSpruit: These comments can be removed I think

ui/src/main/java/com/adyen/checkout/ui/internal/Images.kt
```diff
@@ -0,0 +1,181 @@
...
+    // NOTE: You might need to make these utility functions public if they are private.
+    val densityExtension = LocalContext.current.resources.displayMetrics.densityDpi.getDensityExtension()
+    val logoPath = buildLogoPath(size, txVariant, txSubVariant, densityExtension)
+    val fullUrl = baseUrl + logoPath
+
+    NetworkImage(
+        url = fullUrl,
+        modifier = modifier,
+        contentDescription = contentDescription,
```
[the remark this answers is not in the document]
OscarSpruit: The content description should be null imo. Most of our images are decorative and don't need a description

ui/src/main/java/com/adyen/checkout/ui/internal/Images.kt
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * A composable to load a logo using the Adyen environment and txVariant.
+ * This mirrors the functionality of the existing ImageView.loadLogo extension function.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Suppress("LongParameterList")
+@Composable
+fun AdyenLogo(
```
OscarSpruit: What do you think about naming this `PaymentMethodLogo`?

ui/src/main/java/com/adyen/checkout/ui/internal/Images.kt
ozgur00: Since `DefaultImageLoader` is in `core` I cannot instantiate it here. Without that I'm not sure if it's possible to do it. Do you have something specific in mind?

OscarSpruit: Oh I see. In that case I can think of 2 ways around this:
- Write a wrapper composable in `core`
- Use `staticCompositionLocalOf` and provide the value somewhere in a top level composable (in `core`)

ui/src/main/java/com/adyen/checkout/ui/internal/Images.kt
```diff
@@ -0,0 +1,182 @@
...
+ * @param placeholder A painter to show while the image is loading.
+ * @param errorFallback A painter to show if the image fails to load.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Composable
+fun NetworkImage(
+    url: String,
+    modifier: Modifier = Modifier,
+    contentDescription: String?,
+    imageLoader: ImageLoader,
```
OscarSpruit, line 54: Nit: since `imageLoader` has no default value it should be placed above this `modifier`

OscarSpruit (approved):

PR #2306, opened 2025-08-28
v6 - Move component params to payment facilitator

araratthehero:
## Description
Move component params to payment facilitator layer

COSDK-613


core/src/main/java/com/adyen/checkout/core/components/internal/AdvancedPaymentFacilitatorFactory.kt
```diff
@@ -23,10 +25,21 @@ internal class AdvancedPaymentFacilitatorFactory(
...
+            deviceLocale = Locale.US,
+            dropInOverrideParams = null,
+            componentSessionParams = null,
+        )
+
         val paymentComponent = PaymentMethodProvider.get(
             txVariant = txVariant,
             coroutineScope = coroutineScope,
             checkoutConfiguration = checkoutConfiguration,
+            commonComponentParams = componentParamsData.commonComponentParams,
```
ozgur00: We can pass the whole `componentParamsData` object instead of separate params and then access `commonComponentParams` and `sessionsParams` inside the creation functions in order to reduce the # params in the functions.

PR #2310, opened 2025-08-29
v6 - Move analytics manager creation to payment facilitator 

araratthehero:
## Description
Move analytics manager to the payment facilitator layer

COSDK-613


core/src/main/java/com/adyen/checkout/core/components/internal/AdvancedPaymentFacilitatorFactory.kt
```diff
@@ -35,9 +37,18 @@ internal class AdvancedPaymentFacilitatorFactory(
             componentSessionParams = null,
         )
 
+        // TODO - Analytics. We might need to change the logic on AnalyticsManager creation.
+        val analyticsManager = AnalyticsManagerFactory().provide(
+            componentParams = componentParamsBundle.commonComponentParams,
+            application = null,
+            source = AnalyticsSource.PaymentComponent("AwaitAction"),
```
OscarSpruit, line 45: Should we place a TODO on the source as well?

OscarSpruit (approved):

PR #2321, opened 2025-09-05
v6 - Animate visibility of text field errors

OscarSpruit:
## Description
Animate visibility of text field errors.

https://github.com/user-attachments/assets/48db8da0-98e5-49eb-917b-5d755dda616d

### DIFF
--- a/ui/src/main/java/com/adyen/checkout/ui/internal/CheckoutTextFieldDecorationBox.kt
+++ b/ui/src/main/java/com/adyen/checkout/ui/internal/CheckoutTextFieldDecorationBox.kt
@@ -8,6 +8,9 @@
 
 package com.adyen.checkout.ui.internal
 
+import androidx.compose.animation.AnimatedVisibility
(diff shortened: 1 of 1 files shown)


ui/src/main/java/com/adyen/checkout/ui/internal/CheckoutTextFieldDecorationBox.kt
```diff
@@ -108,12 +111,19 @@ internal fun CheckoutTextFieldDecorationBox(
...
+            exit = shrinkVertically(),
+        ) {
             val supportingTextColor = if (isError) style.errorColor else CheckoutThemeProvider.colors.textSecondary
-            Footnote(
-                text = supportingText,
-                color = supportingTextColor,
-            )
+            supportingText?.let {
+                Footnote(
+                    text = supportingText,
```
araratthehero: I'd use it here
```suggestion
                    text = it,
```

PR #2323, opened 2025-09-08
v6 - Simplified state management

OscarSpruit:
## Description
The goal of this PR is to make sure the internal state management is strongly typed and less complex. A similar structure as before is used, but the implementation of the state management lies with the component.


ui/src/main/java/com/adyen/checkout/ui/internal/CheckoutTextField.kt
```diff
@@ -129,12 +130,17 @@ fun CheckoutTextField(
         val currentOnValueChange by rememberUpdatedState(onValueChange)
         LaunchedEffect(state) {
             snapshotFlow { state.text }
-                .drop(1)
                 .collectLatest { value ->
```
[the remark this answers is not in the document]
OscarSpruit: We decided that this is ok

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayFieldId.kt
```diff
@@ -1,16 +0,0 @@
...
- * This file is open source and available under the MIT license. See the LICENSE file for more info.
- *
- * Created by ararat on 16/4/2025.
- */
-
-package com.adyen.checkout.mbway.internal.ui.state
-
-import com.adyen.checkout.core.components.internal.ui.state.model.FieldId
-
-internal enum class MBWayFieldId : FieldId {
```
araratthehero, line 13: `FieldId` could also be removed

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayViewStateValidator.kt
```diff
@@ -0,0 +1,43 @@
...
+package com.adyen.checkout.mbway.internal.ui.state
+
+import com.adyen.checkout.components.core.internal.util.ValidationUtils
+import com.adyen.checkout.core.components.internal.ui.state.ViewStateValidator
+import com.adyen.checkout.mbway.R
+
+internal class MBWayViewStateValidator : ViewStateValidator<MBWayViewState> {
+
+    override fun validate(viewState: MBWayViewState): MBWayViewState {
+        val telephoneNumber = viewState.countryCode.callingCode + viewState.phoneNumber.text
```
araratthehero, line 19: Should we not remove the `0` at the start before doing the validation?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayViewState.kt
```diff
@@ -9,19 +9,12 @@
 package com.adyen.checkout.mbway.internal.ui.state
 
 import com.adyen.checkout.core.components.internal.ui.model.CountryModel
-import com.adyen.checkout.core.components.internal.ui.state.model.ViewFieldState
-import com.adyen.checkout.core.components.internal.ui.state.model.toViewFieldState
+import com.adyen.checkout.core.components.internal.ui.state.ViewState
+import com.adyen.checkout.core.components.internal.ui.state.model.TextInputState
 
 internal data class MBWayViewState(
```
araratthehero, line 15: It is not necessary for this component, but did you think about separating delegate state from the view state, in case we want to store more fields in the delegate state which are not required for the view, but they are required to generate the component state?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MbWayComponent.kt
```diff
@@ -61,46 +60,39 @@ internal fun MbWayComponent(
...
             modifier = Modifier
                 .fillMaxWidth()
                 .onFocusChanged { focusState ->
-                    fieldChangeListener.onFieldFocusChanged(MBWayFieldId.PHONE_NUMBER, focusState.hasFocus)
+                    changeListener.onPhoneNumberFocusChanged(focusState.hasFocus)
                 },
             label = "Phone Number",
-            isError = viewState.phoneNumberFieldState.errorMessageId != null,
+            isError = isError,
             supportingText = supportingTextPhoneNumber,
```
araratthehero, line 90: What do you think of adding the following here, in case we want to provide a prefilled text field?
```suggestion
            supportingText = supportingTextPhoneNumber,
            initialValue = viewState.phoneNumber.text,
```

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MbWayComponent.kt
```diff
@@ -61,46 +60,39 @@ internal fun MbWayComponent(
...
         // PhoneNumber
-        val supportingTextPhoneNumber = if (viewState.phoneNumberFieldState.errorMessageId != null) {
-            "The phone number is invalid"
-        } else {
-            null
-        }
+        val isError = viewState.phoneNumber.errorMessage != null && viewState.phoneNumber.showError
+        val supportingTextPhoneNumber =
+            if (isError) {
+                "The phone number is invalid"
```
araratthehero: Here we could get the `viewState.phoneNumber.errorMessage` right?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
```diff
@@ -15,85 +15,104 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
...
+        } else {
+            viewStateManager.highlightAllFieldValidationErrors()
+        }
     }
 
-    private val viewStateFlow: StateFlow<MBWayViewState> by lazy {
-        stateManager.state
-            .map(MBWayComponentState::toViewState)
-            .stateIn(coroutineScope, SharingStarted.Lazily, stateManager.state.value.toViewState())
+    private fun MBWayViewState.toPaymentComponentState(
```
araratthehero, line 64: What do you think to moving this back to `MBWayPaymentComponentState.kt`?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayViewStateFactory.kt
```diff
@@ -11,17 +11,18 @@ package com.adyen.checkout.mbway.internal.ui.state
...
+internal class MBWayViewStateFactory(
     private val componentParams: ComponentParams,
-) : ComponentStateFactory<MBWayComponentState, MBWayFieldId> {
+) : ViewStateFactory<MBWayViewState> {
 
-    override fun createDefaultComponentState() = MBWayComponentState(
+    override fun createDefaultViewState() = MBWayViewState(
         countries = getSupportedCountries(componentParams),
+        countryCode = getInitiallySelectedCountry(componentParams),
+        phoneNumber = TextInputState(isFocused = true),
```
araratthehero, line 24: When I set a default text `TextInputState(text = "asgsag", isFocused = true)`, the text field gets stuck and you cannot edit the value, since it only allows digits but we set it to a text. I am not sure how to fix this, but it would be nice if we could somehow limit this.

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayViewStateValidator.kt
OscarSpruit: Our current validation doesn't really mind this. It basically just checks if the phone number is between 9 to 14 digits.  Apparently, phone numbers can be longer than 14 digits, so I think we should revisit our validation logic. I'll place a TODO for now

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayViewState.kt
OscarSpruit: Good question! I was thinking something similar indeed. To keep a separate state for such cases and then we can combine it and the view state to produce the data we share with the merchant. When we work on card we will figure this out I think

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/state/MBWayViewStateFactory.kt
OscarSpruit: Hmm interesting, not sure if we can do something about it. It sounds more like a bug in this TextFieldState API 🤔

araratthehero (approved):

PR #2287, opened 2025-08-14
v6 -Include information about integration path choices to understand and serve developer preferences

JlUgia:
## Description
Today, developers using Adyen can add Google Pay to their applications in multiple ways (Google Pay SDK, Adyen Drop-In, components, etc). Understanding trends and developer preferences is key to improve integration offerings to satisfy ever-changing programming styles and tooling. This PR adds basic versioning telemetry to help understand these patterns and focus efforts on the preferred integration paths.

This change is intentionally incomplete, and meant as a foundational proposal about how to add library and versioning information to Google Pay's `loadPaymentData` request. See how these [fields are populated in another open source library](https://github.com/google-pay/flutter-plugin/blob/4fc0c8b76448a76b26b9fe2a3514270e977b91a4/pay_platform_interface/lib/util/configurations.dart#L44).

## Outstanding items
- [x] Decide for a location in the business logic to transparently include this information (proposing `GooglePayUtils.kt`)
- [x] Procure the version of the library in a programmatic way
- [x] Add tests

PR #2324, opened 2025-09-11
v5 - Add software info to Google Pay request

OscarSpruit:
## Description
Ported #2287 to v5 in this PR

## Checklist 
- [x] Code is unit tested

COSDK-626

### DIFF
--- a/googlepay/src/main/java/com/adyen/checkout/googlepay/MerchantInfo.kt
+++ b/googlepay/src/main/java/com/adyen/checkout/googlepay/MerchantInfo.kt
@@ -21,6 +21,7 @@ import org.json.JSONObject
  *
  * @param merchantName The name of the merchant.
  * @param merchantId The id of the merchant.
+ * @param softwareInfo Information associated with the caller of the request.
  */
 @Parcelize
 data class MerchantInfo(


googlepay/src/main/java/com/adyen/checkout/googlepay/MerchantInfo.kt
```diff
@@ -25,11 +26,13 @@ import org.json.JSONObject
 data class MerchantInfo(
     var merchantName: String? = null,
     var merchantId: String? = null,
+    var softwareInfo: SoftwareInfo? = null,
```
araratthehero, line 30: Could you add the `@param softwareInfo` in the Kdoc?

components-core/src/main/java/com/adyen/checkout/components/core/internal/util/CheckoutPlatformParams.kt
```diff
@@ -0,0 +1,45 @@
...
+object CheckoutPlatformParams {
+
+    var platform = CheckoutPlatform.ANDROID
+        private set
+
+    var version = BuildConfig.CHECKOUT_VERSION
+        private set
+
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    fun overrideForCrossPlatform(
```
araratthehero, line 25: Let's notify cross platform teams about this change too, since they would need to change where they override the platform params?

OscarSpruit: Good one, I already discussed with both and remind them once we released it

PR #2357, opened 2025-10-06
v5 - Pix code improvements

OscarSpruit:
## Description
The QR code view for Pix now has:
- A QR code displayed
- A label with the actual code, so it can be copied by selecting it
- Updated texts 

Screenshot_20251006_152935

## Checklist 

COSDK-552

### Improved
- For PIX, a QR code is now displayed and the copiable code is displayed above the copy button.


qr-code/src/main/java/com/adyen/checkout/qrcode/internal/ui/view/SimpleQRCodeView.kt
```diff
@@ -132,16 +161,26 @@ internal class SimpleQRCodeView @JvmOverloads constructor(
...
     }
 
-    private fun copyCode(qrCodeData: String?) {
+    private fun onCopyClicked(qrCodeData: String?) {
         qrCodeData ?: return
-        context.copyTextToClipboard(
-            "Pix Code",
-            qrCodeData,
-            localizedContext.getString(R.string.checkout_qr_code_copied_toast),
+        binding.copyButton.setText(R.string.checkout_qr_code_pix_code_copied)
```
jreij: Should you use `localizedContext` here?

PR #2365, opened 2025-10-16
V5 - SdkData - Create models

araratthehero:
## Description

- Create SDK Data models .
- Add `sdkData` field in PaymentComponentData as a string, since it will get encoded

## Checklist 
- [x] Code is unit tested

COSDK-513

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
@@ -40,6 +40,7 @@ data class PaymentComponentData<PaymentMethodDetailsT : PaymentMethodDetails>(
     var socialSecurityNumber: String? = null,
     var installments: Installments? = null,
     var supportNativeRedirect: Boolean? = true,
+    val sdkData: String? = null
 ) : ModelObject() {
 
     companion object {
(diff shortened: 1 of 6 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
```diff
@@ -40,6 +40,7 @@ data class PaymentComponentData<PaymentMethodDetailsT : PaymentMethodDetails>(
     var socialSecurityNumber: String? = null,
     var installments: Installments? = null,
     var supportNativeRedirect: Boolean? = true,
+    val sdkData: String? = null
```
jreij, line 43: Did you consider using the `SdkData` type here instead of `String` and then creating a custom serializer/deserializer that turns it into a json then base64 string?

I myself am not super sure about this, as it might give the impression that this attribute is an object not a String but it's something to consider if you don't want to serialize to json then base64 inside each component separately.

jreij: Nevermind, this is part of the public API so it should be exposed as a String not as an object, in case they are mapping the fields of `PaymentComponentData` one by one (which is not a good practice but we can't solve it)

araratthehero: Initially that is how I did and it looked quite clean. But today I double checked with the team and converted it to a String.

The issue is, if we use the `SdkData` here, then we expose the content of the `SdkData` to merchants, so they get the full `SdkData` object. The `SdkData` will be encoded only when they use our Serializers to serialize the `PaymentComponentData` object. We do not want to expose the content of the `SdkData` to the merchants and we want to pass them already encoded string.

I know that if they want to, it is still possible to decode it, adjust it and encode it again, but this is something we are aware of.

araratthehero: In the next PR I will push an `SdkDataProvider`, which centralizes the serialization and encoding logic.

PR #2366, opened 2025-10-16
V5 - SdkData - Provider

araratthehero:
## Description

- Created a `SdkDataProvider` to generate and encode the `SdkData`
- Created custom serializer helpers to encode and decode values.
- Implemented `SdkData` for `Card` and `MbWay` to showcase the differences between implementation.

## Checklist 
- [x] Code is unit tested

COSDK-513

### DIFF
--- a/mbway/src/main/java/com/adyen/checkout/mbway/internal/provider/MBWayComponentProvider.kt
+++ b/mbway/src/main/java/com/adyen/checkout/mbway/internal/provider/MBWayComponentProvider.kt
@@ -26,6 +26,7 @@ import com.adyen.checkout.components.core.internal.analytics.AnalyticsManager
 import com.adyen.checkout.components.core.internal.analytics.AnalyticsManagerFactory
 import com.adyen.checkout.components.core.internal.analytics.AnalyticsSource
 import com.adyen.checkout.components.core.internal.provider.PaymentComponentProvider
+import com.adyen.checkout.components.core.internal.provider.SdkDataProvider
 import com.adyen.checkout.components.core.internal.ui.model.ButtonComponentParamsMapper
 import com.adyen.checkout.components.core.internal.ui.model.CommonComponentParamsMapper
 import com.adyen.checkout.components.core.internal.ui.model.DropInOverrideParams
@@ -107,6 +108,7 @@ constructor(
                 componentParams = componentParams,
                 analyticsManager = analyticsManager,
                 submitHandler = SubmitHandler(savedStateHandle),
+                sdkDataProvider = SdkDataProvider(analyticsManager),
             )
 
             val genericActionDelegate =
@@ -195,6 +197,7 @@ constructor(
                 componentParams = componentParams,
                 analyticsManager = analyticsManager,
                 submitHandler = SubmitHandler(savedStateHandle),
+                sdkDataProvider = SdkDataProvider(analyticsManager),
             )
 
             val genericActionDelegate =
--- a/mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/DefaultMBWayDelegate.kt
+++ b/mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/DefaultMBWayDelegate.kt
@@ -18,6 +18,7 @@ import com.adyen.checkout.components.core.internal.PaymentComponentEvent
 import com.adyen.checkout.components.core.internal.PaymentObserverRepository
 import com.adyen.checkout.components.core.internal.analytics.AnalyticsManager
 import com.adyen.checkout.components.core.internal.analytics.GenericEvents
+import com.adyen.checkout.components.core.internal.provider.SdkDataProvider
 import com.adyen.checkout.components.core.internal.ui.model.ButtonComponentParams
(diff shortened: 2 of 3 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/internal/provider/SdkDataProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.adyen.checkout.core.internal.data.model.ModelUtils
+import java.util.Date
+
+/**
+ * A provider that creates an [SdkData] object with the necessary information.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class SdkDataProvider(
+    private val analyticsManager: AnalyticsManager
+) {
```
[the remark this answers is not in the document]
araratthehero: I don't think this is truly necessary. We do the same for analytics.

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -823,6 +825,9 @@ class DefaultCardDelegate(
             shopperReference = componentParams.shopperReference,
             order = order,
             amount = componentParams.amount,
+            sdkData = sdkDataProvider.createEncodedSdkData(
+                threeDS2SdkVersion = runCompileOnly { ThreeDS2Service.INSTANCE.sdkVersion },
+            ),
```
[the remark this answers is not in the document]
araratthehero: This is done on purpose

checkout-core/src/main/java/com/adyen/checkout/core/internal/data/model/ModelUtils.kt
```diff
@@ -47,6 +53,40 @@ object ModelUtils {
...
+        } catch (e: IllegalArgumentException) {
+            adyenLog(AdyenLogLevel.ERROR, e) { "Failed to deserialize the model object" }
+            null
+        } catch (e: JSONException) {
+            adyenLog(AdyenLogLevel.ERROR, e) { "Failed to deserialize the model object" }
+            null
+        } catch (e: ModelSerializationException) {
+            adyenLog(AdyenLogLevel.ERROR, e) { "Failed to deserialize the model object" }
+            null
+        }
```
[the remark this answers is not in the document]
araratthehero: I thought about it, but I don't want to catch generic Exception. In case the function content changes and a new exception is being thrown, we could easily miss it.

components-core/src/main/java/com/adyen/checkout/components/core/internal/provider/SdkDataProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+     */
+    fun createEncodedSdkData() = createEncodedSdkData(threeDS2SdkVersion = null)
+
+    /**
+     * Creates the [SdkData] object.
+     *
+     * @param threeDS2SdkVersion The version of the 3DS2 SDK.
+     * @return The created encoded [String] from [SdkData] object.
+     */
+    fun createEncodedSdkData(threeDS2SdkVersion: String?): String? {
```
OscarSpruit: If we add a default value of `null` here, we could remove the other function

araratthehero: Initially I wanted to make the `threeDS2SdkVersion` non-null, but since it is now nullable, I will remove the other function.

components-core/src/main/java/com/adyen/checkout/components/core/internal/provider/SdkDataProvider.kt
```diff
@@ -0,0 +1,52 @@
...
+                checkoutAttemptId = analyticsManager.getCheckoutAttemptId(),
+            ),
+            authentication = Authentication(
+                threeDS2SdkVersion = threeDS2SdkVersion,
+            ),
+            createdAt = Date().time,
+            supportNativeRedirect = true,
+        )
+
+        return ModelUtils.serializeAndEncodeOpt(sdkData, SdkData.SERIALIZER)
```
OscarSpruit: IMO it would be better to separate serialisation and encoding by moving these lines here:
```kotlin
val jsonObject = serializer.serialize(modelObject)
return Base64.encode(jsonObject.toString().toByteArray(Charsets.UTF_8))
```
While writing this I also wonder if `SdkData` should extend `ModelObject`, because we are not exposing the class externally

araratthehero: It is true that the encoding is only used in `SdkData`, but I added the `serializeAndEncodeOpt` in `ModelUtils`, to allow other components to use it too if necessary and to have a single place where you can encode a serialized value.

Regarding `SdkData` extending `ModelObject`, we need to still serialize it before encoding it and if necessary deserialize it in the future. By extending from the `ModelObject`, we keep it constant with the serialization logic and treat the `SdkData` as any other model object.

If you want, we could discuss this in the team tomorrow.

PR #2368, opened 2025-10-20
v6: Card - Add card number

ozgur00:
## Description
Implement card number. 

## Checklist 

COSDK-570

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
@@ -8,16 +8,102 @@
 
 package com.adyen.checkout.card.internal.ui.view
 
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.foundation.text.input.maxLength
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.onFocusChanged
+import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.tooling.preview.Preview
+import com.adyen.checkout.card.internal.ui.state.CardChangeListener
+import com.adyen.checkout.card.internal.ui.state.CardViewState
+import com.adyen.checkout.core.common.helper.CardNumberValidator
+import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
+import com.adyen.checkout.core.common.localization.internal.helper.resolveString
+import com.adyen.checkout.core.components.internal.ui.state.model.TextInputState
+import com.adyen.checkout.ui.internal.CheckoutTextField
+import com.adyen.checkout.ui.internal.DigitOnlyInputTransformation
+import com.adyen.checkout.ui.internal.Dimensions
 
 @Composable
-internal fun CardComponent() {
-    // TODO - Card UI
+internal fun CardComponent(
+    viewState: CardViewState,
+    changeListener: CardChangeListener,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier.fillMaxWidth(),
+        verticalArrangement = Arrangement.spacedBy(Dimensions.Large),
+    ) {
+        CardNumberField(
+            cardNumberState = viewState.cardNumber,
+            isAmex = viewState.isAmex,
+            onCardNumberChanged = changeListener::onCardNumberChanged,
+            onCardNumberFocusChanged = changeListener::onCardNumberFocusChanged,
+        )
+    }
+    // TODO - Card Full UI
 }
 
-@Preview
+@Composable
+private fun CardNumberField(
+    cardNumberState: TextInputState,
+    isAmex: Boolean,
+    onCardNumberChanged: (String) -> Unit,
+    onCardNumberFocusChanged: (Boolean) -> Unit,
+) {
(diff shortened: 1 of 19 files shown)


core/src/main/java/com/adyen/checkout/core/common/helper/CardNumberValidationResult.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+sealed interface CardNumberValidationResult {
+    class Valid : CardNumberValidationResult
+    interface Invalid : CardNumberValidationResult {
+        class IllegalCharacters : Invalid
+        class TooLong : Invalid
+        class TooShort : Invalid
+        class LuhnCheck : Invalid
+    }
+}
```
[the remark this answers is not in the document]
ozgur00: These intentionally use `class` instead `object`, since they're public and in case we need to include a parameter we want to avoid breaking changes.

core/src/main/java/com/adyen/checkout/core/common/internal/helper/StringUtil.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Check if the string only contains number and the specified separator characters.
+     *
+     * @param value The string to be checked.
+     * @param separators The optional accepted separators.
+     * @return If the string is only numbers and separators.
+     */
+    fun isDigitsAndSeparatorsOnly(value: String, vararg separators: Char): Boolean {
```
araratthehero, line 35: I would also add tests for this function

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
```diff
@@ -8,16 +8,87 @@
...
+import com.adyen.checkout.core.common.localization.internal.helper.resolveString
+import com.adyen.checkout.core.components.internal.ui.state.model.TextInputState
+import com.adyen.checkout.ui.internal.CheckoutTextField
+import com.adyen.checkout.ui.internal.DigitOnlyInputTransformation
+import com.adyen.checkout.ui.internal.Dimensions
 
 @Composable
-internal fun CardComponent() {
-    // TODO - Card UI
+internal fun CardComponent(
```
araratthehero, line 33: Since the CardComponent will be a big Composable. What do you think if we create composable functions for each input, e.g. `CardNumberField()`, `ExpiryDateField()` and keep the `CardComponent()` simple by using those composables to construct a UI.
e.g.
```
internal fun CardComponent(
...
) {
    Column(
        modifier = modifier.fillMaxWidth(),
        verticalArrangement = Arrangement.spacedBy(Dimensions.Large),
    ) {
        CardNumberField(...)

        Row(...) {
            ExpiryDateField(...)
            SecurityCodeField(...)
        }
    }
}
```

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberOutputTransformation.kt
```diff
@@ -0,0 +1,42 @@
...
+ * Created by ozgur on 13/10/2025.
+ */
+
+package com.adyen.checkout.card.internal.ui.view
+
+import androidx.compose.foundation.text.input.OutputTransformation
+import androidx.compose.foundation.text.input.TextFieldBuffer
+import androidx.compose.foundation.text.input.insert
+
+internal class CardNumberOutputTransformation(
```
OscarSpruit, line 17: They say it's better for the compose performance if these `OutputTransformation`s are `data class`es and are marked with `@Stable`

core/src/main/java/com/adyen/checkout/core/common/helper/CardNumberValidator.kt
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Created by ozgur on 4/10/2024.
+ */
+
+package com.adyen.checkout.core.common.helper
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.common.internal.helper.StringUtil
+
+object CardNumberValidator {
```
OscarSpruit, line 14: Shouldn't this class be moved to the card module?

core/src/test/java/com/adyen/checkout/core/common/helper/CardNumberValidatorTest.kt
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+package com.adyen.checkout.core.common.helper
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments.arguments
+import org.junit.jupiter.params.provider.MethodSource
+
+class CardNumberValidatorTest {
```
OscarSpruit: Could be `internal`

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
OscarSpruit: I like this idea!

core/src/main/java/com/adyen/checkout/core/common/helper/CardNumberValidator.kt
ozgur00: Since these public validation functions are for api only merchants, we hadn't wanted to include them in `card` module at the time. However I think we can create it's own module like `cse`, so merchants can import only that module when they need these. I will create a ticket to do that separately though, also to double check with product.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
ozgur00: Good one! Done.

PR #2370, opened 2025-10-20
v6 - Set up base for navigation

OscarSpruit:
## Description
Implement the base for navigation  using AndroidX navigation3. It supports displaying content inline and in full screen dialogs. Because of navigation3 the target and min sdk versions had to be increased.

As follow ups we can create:
- A reusable `ValuePicker` composable that replaces the country code specific one
- Expose navigation customization, so for example the display strategy and transitions can be changed

COSDK-564

### DIFF
--- a/card/build.gradle
+++ b/card/build.gradle
@@ -10,6 +10,7 @@ plugins {
     alias libs.plugins.checkout.library
     alias libs.plugins.kotlin.compose
     alias libs.plugins.kotlin.parcelize
+    alias libs.plugins.kotlin.serialization
 }
 
 checkoutPublishing {
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
@@ -10,8 +10,9 @@ package com.adyen.checkout.card.internal.ui
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
-import androidx.compose.ui.Modifier
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.navigation3.runtime.NavBackStack
+import androidx.navigation3.runtime.NavKey
 import com.adyen.checkout.card.internal.ui.state.CardChangeListener
 import com.adyen.checkout.card.internal.ui.state.CardComponentState
 import com.adyen.checkout.card.internal.ui.state.CardPaymentComponentState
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
```diff
@@ -42,20 +54,58 @@ internal class PaymentFacilitator(
...
         ) {
-            if (actionComponent != null) {
-                actionComponent.ViewFactory(modifier)
-            } else {
-                paymentComponent.ViewFactory(modifier)
+            NavDisplay(
+                backStack = backStack,
+                sceneStrategy = DialogSceneStrategy(),
+            ) { key ->
+                val entries = paymentComponent.navigation + actionComponent?.navigation.orEmpty()
```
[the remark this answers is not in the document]
OscarSpruit: This is not possible as we are not in a composable context at this point

core/build.gradle
```diff
@@ -37,6 +38,7 @@ dependencies {
 
     implementation platform(libs.compose.bom)
     implementation libs.bundles.compose.ui
+    api libs.bundles.androidx.navigation3
```
araratthehero: Nits: What do you think of moving this up, below `api libs.bundles.androidx.lifecycle`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/PaymentComponent.kt
```diff
@@ -9,18 +9,19 @@
...
+import com.adyen.checkout.core.components.internal.ui.navigation.CheckoutNavEntry
 
 // TODO - Some components might not be composable,
 //  Move ComposableComponent to PaymentMethod specific component later
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 interface PaymentComponent<T : BasePaymentComponentState> : EventComponent<PaymentComponentEvent<T>> {
 
-    @Composable
-    fun ViewFactory(modifier: Modifier)
+    val navigation: Map<NavKey, CheckoutNavEntry>
```
ozgur00: What do you think about using composition here, i.e. putting these two navigation fields in a `NavigationComponent` interface and implementing it in `PaymentComponent` & `ActionComponent`?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
```diff
@@ -112,20 +126,33 @@ internal class MBWayComponent(
...
-            MbWayComponent(
-                viewState = viewState,
-                changeListener = this,
-            )
-        }
+        )
     }
 }
+
+@Serializable
```
ozgur00, line 155: It could make it easier to read how many screens/navigation items are there for each component if we put them in a separate file such as `MbWayNavKeys`, we would also keep the # of lines slightly lower in this file.

OscarSpruit: In the next PR (#2376) this is already done

ozgur00 (approved):

PR #2371, opened 2025-10-21
V5 - SdkData - Implement for all components

araratthehero:
## Description

- Implemented SDK Data for all delegates
- The task is quite repetitive and has been done in all delegates

Note: We should pay attention to not miss the implementation for any of the payment delegates.

## Checklist 
- [x] Code is unit tested

COSDK-513


googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/DefaultGooglePayDelegate.kt
```diff
@@ -184,6 +186,7 @@ internal class DefaultGooglePayDelegate(
             paymentMethod = paymentMethod,
             order = order,
             amount = componentParams.amount,
+            sdkData = sdkDataProvider.createEncodedSdkData(),
```
OscarSpruit: It's a bit hidden, but for Google Pay we also send the 3DS2 version

araratthehero: Good catch!

PR #2374, opened 2025-10-27
 V5 - SdkData - Deprecate fields

araratthehero:
## Description
Deprecated fields which moved to SdkData

## Checklist 

COSDK-513

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/ACHDirectDebitPaymentMethod.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/ACHDirectDebitPaymentMethod.kt
@@ -18,6 +18,7 @@ import org.json.JSONObject
 @Parcelize
 data class ACHDirectDebitPaymentMethod(
     override var type: String?,
+    @Deprecated("This property is deprecated. Use the SERIALIZER to send the payment data to your backend.")
     override var checkoutAttemptId: String?,
     var encryptedBankAccountNumber: String? = null,
     var encryptedBankLocationId: String? = null,
--- a/components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/PaymentComponentData.kt
@@ -39,6 +39,7 @@ data class PaymentComponentData<PaymentMethodDetailsT : PaymentMethodDetails>(
     var dateOfBirth: String? = null,
     var socialSecurityNumber: String? = null,
     var installments: Installments? = null,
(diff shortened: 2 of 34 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/ACHDirectDebitPaymentMethod.kt
```diff
@@ -18,6 +18,7 @@ import org.json.JSONObject
 @Parcelize
 data class ACHDirectDebitPaymentMethod(
     override var type: String?,
+    @Deprecated("This property is deprecated. Use the new sdkData property instead.")
```
OscarSpruit: `sdkData` is not really accessible as it is encoded, so I don't think merchants can use it. It seems like we don't have an alternative to offer and it's probably not necessary to offer that. Should we change the message to something like below?
`This property is deprecated and will be removed in future versions`

araratthehero: Makes sense

components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/ACHDirectDebitPaymentMethod.kt
```diff
@@ -18,6 +18,7 @@ import org.json.JSONObject
 @Parcelize
 data class ACHDirectDebitPaymentMethod(
     override var type: String?,
+    @Deprecated("This property is deprecated and will be removed in future versions.")
```
jreij: Shall we explain what merchants can do to replace the deprecated fields? Or maybe we can just say "use the serializer instead directly accessing the field"?

jreij: I just saw Oscar's comment and the change, however I still believe that a deprecation warning without a replacement action is confusing

araratthehero: I don't think there is a case when merchants need the deprecated fields, other than sending it to us in the `/payments` call. We could mention in the deprecation warning, that if they were sending those in the `/payments` call, then they should now send the `sdkData` field instead. How does it sound?

araratthehero: Updated the comment as discussed

jreij (approved):

PR #2378, opened 2025-10-29
v6: Card - Add CardConfiguration

ozgur00:
## Description
Add `CardConfiguration` to be used in new architecture.

## Checklist 

COSDK-570


card/src/main/java/com/adyen/checkout/card/CardBrand.kt
```diff
@@ -0,0 +1,13 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 8/10/2024.
+ */
+
+package com.adyen.checkout.card
+
+import com.adyen.checkout.core.common.CardBrand
+
+typealias CardBrand = CardBrand
```
araratthehero: Is this typealias necessary?

card/src/test/java/com/adyen/checkout/card/CardConfigurationTest.kt
```diff
@@ -1,147 +0,0 @@
...
-import com.adyen.checkout.components.core.AnalyticsConfiguration
-import com.adyen.checkout.components.core.AnalyticsLevel
-import com.adyen.checkout.components.core.CheckoutConfiguration
-import com.adyen.checkout.core.old.CardType
-import com.adyen.checkout.core.old.Environment
-import org.junit.jupiter.api.Assertions.assertEquals
-import org.junit.jupiter.api.Test
-import java.util.Locale
-
-internal class CardConfigurationTest {
```
araratthehero, line 21: Shouldn't we keep the tests for CardConfigurationTest?

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.adyen.checkout.card.KCPAuthVisibility
+import com.adyen.checkout.card.SocialSecurityNumberVisibility
+import com.adyen.checkout.components.core.PaymentMethod
+import com.adyen.checkout.core.common.AdyenLogLevel
+import com.adyen.checkout.core.common.CardType
+import com.adyen.checkout.core.common.internal.helper.adyenLog
+import com.adyen.checkout.core.components.internal.ui.model.ComponentParamsBundle
+import com.adyen.checkout.core.sessions.internal.model.SessionParams
+
+internal class CardComponentParamsMapper {
```
araratthehero, line 24: Could you also move the tests for this mapper or if it is not final you could leave a TODO.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -127,7 +127,7 @@ internal class CardComponent(
 //                )
 //            }
 
-            cardEncryptor.encryptFields(unencryptedCardBuilder.build(), publicKey)
+            cardEncryptor.encryptFields(unencryptedCardBuilder.build(), componentParams.publicKey)
```
araratthehero: You could also just use `publicKey` here, since the there is a nullability check done before?

card/src/test/java/com/adyen/checkout/card/CardConfigurationTest.kt
ozgur00: This class was testing only the mapping from old config to new config and vice versa. Since we are removing old builders, it's not necessary anymore.

card/src/main/java/com/adyen/checkout/card/CardBrand.kt
ozgur00: I don't remember why we had added it tbh, so removing it for now and we can add it if we find necessary later.

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapper.kt
ozgur00: Left a TODO, since the existing tests don't apply anymore to the new mapper.

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
```diff
@@ -0,0 +1,38 @@
...
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+/**
+ * This class represents a card scheme. The constructor allows for creating a [CardBrand] with a scheme that is not in
+ * the predefined list of [CardType]. Can be used to configure the supported card schemes with
+ * [CardConfiguration.Builder.setSupportedCardTypes].
+ */
+@Parcelize
+data class CardBrand constructor(val txVariant: String) : Parcelable {
```
OscarSpruit: The `constructor` keyword is unnecessary here

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Use this constructor when defining the supported card brand predefined inside [CardType] enum
+     * inside your component
+     */
+    constructor(cardType: CardType) : this(txVariant = cardType.txVariant)
```
OscarSpruit: Wonder if we should still support this constructor. Also thinking if `CardType` should be internal

ozgur00: I've made `CardType` restricted since it's used in card module, and removed this constructor. We should also mark it deprecated in v5.

PR #2379, opened 2025-10-29
v6: Card - Add DetectedCardTypeRepository to new implementation

ozgur00:
## Description
Add bin lookup and local card detection related code to new implementation.

## Checklist 

COSDK-570

### DIFF
--- a/card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
+++ b/card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
@@ -1,0 +1,284 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ozgur on 5/11/2025.
+ */
+
+package com.adyen.checkout.card.internal.data.api
+
+import app.cash.turbine.test
+import com.adyen.checkout.card.internal.data.model.BinLookupResponse
+import com.adyen.checkout.card.internal.data.model.Brand
+import com.adyen.checkout.card.internal.data.model.DetectedCardType
+import com.adyen.checkout.components.core.PaymentMethodTypes
+import com.adyen.checkout.core.common.CardBrand
+import com.adyen.checkout.core.common.CardType
+import com.adyen.checkout.cse.internal.TestCardEncryptor
+import com.adyen.checkout.test.TestDispatcherExtension
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.filter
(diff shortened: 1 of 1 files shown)


card/src/main/java/com/adyen/checkout/card/internal/data/api/DefaultDetectCardTypeRepository.kt
```diff
@@ -0,0 +1,217 @@
...
+import com.adyen.checkout.core.common.internal.helper.runSuspendCatching
+import com.adyen.checkout.cse.internal.BaseCardEncryptor
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.launch
+import java.util.UUID
+
+internal class DefaultDetectCardTypeRepository(
```
araratthehero, line 31: Could you add unit tests for this class or add a TODO to pick up later?

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -57,6 +69,11 @@ internal class CardComponent(
 
     override val navigationStartingPoint: NavKey = CardNavKey
 
+    internal fun initialize(coroutineScope: CoroutineScope) {
```
araratthehero: The `CardComponent` is already internal, it could be removed for this function

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -177,6 +202,32 @@ internal class CardComponent(
...
+                    copy(detectedCardTypes = detectedCardTypes)
+                }
+            }
+            .map { detectedCardTypes ->
+                detectedCardTypes.filter { it.isReliable && it.isSupported }.map { it.cardBrand }
+            }
+            .distinctUntilChanged()
+            .onEach {
+                // TODO - Card. Dual brands reset brand
+                // inputData.selectedCardBrand = null
```
araratthehero, line 227: We don't have inputData anymore. Could you adjust or remove this comment?

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
araratthehero: Also I don't see this function being called from anywhere. Could you add that too?

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
ozgur00: I left this here as a reminder of what was the previous purpose of this, together with the TODO it's clear what we will do when we're working on Dual Branded Cards.

araratthehero (dismissed):

PR #2384, opened 2025-10-29
v6: Card - Add security code

ozgur00:
## Description
Add security code input for card component

## Checklist 
- [x] Code is unit tested

COSDK-570

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
@@ -26,6 +26,7 @@ internal fun SecurityCodeField(
     securityCodeState: TextInputState,
     onSecurityCodeChanged: (String) -> Unit,
     onSecurityCodeFocusChanged: (Boolean) -> Unit,
+    isAmex: Boolean?,
     modifier: Modifier = Modifier,
 ) {
     val showSecurityCodeError =
@@ -49,10 +50,17 @@ internal fun SecurityCodeField(
         onValueChange = { value ->
             onSecurityCodeChanged(value)
         },
-        inputTransformation = DigitOnlyInputTransformation().maxLength(MAX_LENGTH_SECURITY_CODE),
+        inputTransformation = DigitOnlyInputTransformation().maxLength(
+            if (isAmex == false) {
+                MAX_LENGTH_SECURITY_CODE_DEFAULT
+            } else {
+                MAX_LENGTH_SECURITY_CODE_AMEX
+            }
+        ),
         keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
         shouldFocus = securityCodeState.isFocused,
     )
 }
 
-private const val MAX_LENGTH_SECURITY_CODE = 4
+private const val MAX_LENGTH_SECURITY_CODE_DEFAULT = 3
+private const val MAX_LENGTH_SECURITY_CODE_AMEX = 4
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
@@ -9,6 +9,7 @@
 package com.adyen.checkout.card.internal.ui.state
 
 import com.adyen.checkout.core.common.CardBrand
+import com.adyen.checkout.core.common.CardType
 import com.adyen.checkout.core.components.internal.ui.state.ViewState
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputState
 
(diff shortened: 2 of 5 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
```diff
@@ -0,0 +1,58 @@
...
+        onValueChange = { value ->
+            onSecurityCodeChanged(value)
+        },
+        inputTransformation = DigitOnlyInputTransformation().maxLength(MAX_LENGTH_SECURITY_CODE),
+        keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
+        shouldFocus = securityCodeState.isFocused,
+    )
+}
+
+private const val MAX_LENGTH_SECURITY_CODE = 4
```
araratthehero: What do you think of changing the length requirement based on the card type?

Given that the hint mentions 3 digits, we could change the hint based on the card type too. I saw that web does it already.

core/src/main/java/com/adyen/checkout/core/common/localization/internal/DefaultLocalizationSource.kt
```diff
@@ -32,6 +32,8 @@ internal class DefaultLocalizationSource {
                 "Expiry date too far in the future"
             CheckoutLocalizationKey.CARD_NUMBER_INVALID -> "Invalid card number"
             CheckoutLocalizationKey.CARD_NUMBER_INVALID_UNSUPPORTED_BRAND -> "The entered card brand isn't supported"
+            CheckoutLocalizationKey.CARD_SECURITY_CODE -> "Security code"
+            CheckoutLocalizationKey.CARD_SECURITY_CODE_INVALID -> "Invalid security code"
```
araratthehero: You can add these copies in the `strings` resource file.

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InputFieldUIState.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Created by ozgur on 6/10/2025.
+ */
+
+package com.adyen.checkout.card.internal.ui.model
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+enum class InputFieldUIState {
```
araratthehero: Can this be `internal`?

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardValidationUtils.kt
```diff
@@ -11,10 +11,15 @@ package com.adyen.checkout.card.internal.ui.state
...
 import com.adyen.checkout.core.common.helper.CardExpiryDateValidationResult
 import com.adyen.checkout.core.common.helper.CardExpiryDateValidator
 import com.adyen.checkout.core.common.helper.CardNumberValidationResult
 import com.adyen.checkout.core.common.helper.CardNumberValidator
+import com.adyen.checkout.core.common.helper.CardSecurityCodeValidationResult
+import com.adyen.checkout.core.common.helper.CardSecurityCodeValidator
+import com.adyen.checkout.core.common.internal.helper.StringUtil
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 object CardValidationUtils {
```
araratthehero: Can this become internal too?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
ozgur00: Sounds good, I changed the logic for max length limitation. 

The hint I'll apply the change in the next PR where we update the supporting texts.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
```diff
@@ -0,0 +1,66 @@
...
+        supportingText = supportingTextSecurityCode,
+        onValueChange = { value ->
+            onSecurityCodeChanged(value)
+        },
+        inputTransformation = DigitOnlyInputTransformation().maxLength(
+            if (isAmex == false) {
+                MAX_LENGTH_SECURITY_CODE_DEFAULT
+            } else {
+                MAX_LENGTH_SECURITY_CODE_AMEX
+            }
```
araratthehero, line 58: If the card number field is empty then `isAmex` is null, thus the limit for the CVC will be 4, but when a card is detected the limit will most probably change to 3, and then when that card is Amex, the limit will be changed to 4 again.

Isn't it better to have 3 digits by default and only change it to 4 when Amex is detected?

```suggestion
            if (isAmex == true) {
                MAX_LENGTH_SECURITY_CODE_AMEX
            } else {
                MAX_LENGTH_SECURITY_CODE_DEFAULT
            }
```

ozgur00: I followed the Web behavior, thinking it's based on if the form is fully empty, then we should allow entering 4 digits since the shopper might enter an Amex card. AMEX can only start with 34 or 37, and locally we will detect it so I don't think it will be a problem.

However I noticed now that even though they do allow 4 digits initially, the hint still says "3 digits". Even though that's sort of confusing, showing 4 digit text by default will be more confusing for all shoppers since mostly people will use 3 digit security code cards. I think we should follow what web does here: 
- Allow 4 digits initially with 3 digit hint.
- Whenever a card is detected update the hint and max length

araratthehero: Good points. I agree keeping consistent with web and change it later if necessary.

PR #2388, opened 2025-10-30
v6: Redirect - Implement new redirect component

ozgur00:
## Description
Implement the new `RedirectComponent`.

## Checklist 

COSDK-690

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
@@ -19,6 +19,7 @@ import androidx.navigation3.scene.DialogSceneStrategy
 import androidx.navigation3.ui.NavDisplay
 import com.adyen.checkout.core.action.data.Action
 import com.adyen.checkout.core.action.internal.ActionComponent
+import com.adyen.checkout.core.action.internal.ActionComponentEvent
 import com.adyen.checkout.core.action.internal.ActionProvider
 import com.adyen.checkout.core.common.AdyenLogLevel
 import com.adyen.checkout.core.common.internal.helper.CheckoutCompositionLocalProvider
@@ -26,6 +27,7 @@ import com.adyen.checkout.core.common.internal.helper.adyenLog
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationProvider
 import com.adyen.checkout.core.components.CheckoutController
 import com.adyen.checkout.core.components.CheckoutResult
+import com.adyen.checkout.core.components.ComponentError
 import com.adyen.checkout.core.components.internal.ui.IntentHandlingComponent
 import com.adyen.checkout.core.components.internal.ui.PaymentComponent
 import com.adyen.checkout.core.components.internal.ui.model.CommonComponentParams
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/adyen/checkout/core/action/data/RedirectAction.kt
```diff
@@ -0,0 +1,61 @@
...
+
+import com.adyen.checkout.core.common.exception.ModelSerializationException
+import com.adyen.checkout.core.common.internal.model.getStringOrNull
+import kotlinx.parcelize.Parcelize
+import org.json.JSONException
+import org.json.JSONObject
+
+@Parcelize
+data class RedirectAction(
+    override var type: String? = null,
```
OscarSpruit: We should make all these `val`s

core/src/main/java/com/adyen/checkout/core/components/internal/PaymentFacilitator.kt
```diff
@@ -140,8 +141,14 @@ internal class PaymentFacilitator(
...
 
-    @Suppress("UNUSED_PARAMETER")
     private fun handleIntent(intent: Intent) {
-        // TODO - handle intent with action component
+        val actionComponent = actionComponent
+        if (actionComponent !is IntentHandlingComponent) {
+            adyenLog(AdyenLogLevel.DEBUG) {
+                "Action component ${actionComponent?.javaClass?.simpleName} is not type of IntentHandlingComponent"
+            }
+            error("Action component does not implement IntentHandlingComponent")
```
OscarSpruit: Wouldn't it be better if we propagate an error to the merchant? This shouldn't happen, but I'm afraid that if anything goes wrong with an intent filter or anything this could crash in production

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsActivity.kt
```diff
@@ -8,13 +8,15 @@
...
 
+import android.content.Intent
 import android.os.Bundle
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import com.adyen.checkout.example.ui.theme.ExampleTheme
 import com.adyen.checkout.example.ui.theme.UIThemeRepository
+import com.adyen.checkout.redirect.old.RedirectComponent
```
OscarSpruit, line 20: Is it intentional to import the old component?

example-app/src/main/AndroidManifest.xml
```diff
@@ -218,6 +218,22 @@
             android:exported="true"
             android:launchMode="singleTop">
 
+            <!-- Use a different intent filter then the one used to handle the Drop-In result
+            (if you have both Drop-In and standalone components in your app). If you don't do this
+            the user will have to choose which Activity is used to handle the result, which is a
+            confusing user experience. -->
+            <intent-filter>
```
OscarSpruit, line 225: Isn't this also needed for the `V6SessionsActivity`?

redirect/build.gradle
```diff
@@ -8,6 +8,7 @@
 
 plugins {
     alias libs.plugins.checkout.library
+    alias libs.plugins.kotlin.compose
     alias libs.plugins.kotlin.parcelize
```
OscarSpruit, line 12: Isn't `alias libs.plugins.kotlin.serialization` also needed here?

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsActivity.kt
ozgur00: I am thinking of extracting return url creation to some helper class since our components are internal now. Will update it in a different PR.

PR #2392, opened 2025-11-04
v6 - Extract value picker

OscarSpruit:
## Description
- Create base `ValuePicker`, which displays a list of items
- Create base `SearchableValuePicker`, which displays a list of items with a search bar on top
- Create `SearchField`, which is an input field specifically styled as a search bar

https://github.com/user-attachments/assets/dee6e816-ef3e-46d1-b133-354c460fdc57

COSDK-147

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -595,6 +595,7 @@ public final class com/adyen/checkout/core/common/localization/CheckoutLocalizat
 	public static final field CARD_SECURITY_CODE_HINT_3_DIGITS Lcom/adyen/checkout/core/common/localization/CheckoutLocalizationKey;
 	public static final field CARD_SECURITY_CODE_HINT_4_DIGITS Lcom/adyen/checkout/core/common/localization/CheckoutLocalizationKey;
 	public static final field CARD_SECURITY_CODE_INVALID Lcom/adyen/checkout/core/common/localization/CheckoutLocalizationKey;
+	public static final field GENERAL_SEARCH_HINT Lcom/adyen/checkout/core/common/localization/CheckoutLocalizationKey;
 	public static final field MBWAY_COUNTRY_CODE Lcom/adyen/checkout/core/common/localization/CheckoutLocalizationKey;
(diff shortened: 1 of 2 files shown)


ui/src/main/java/com/adyen/checkout/ui/internal/SearchableValuePicker.kt
```diff
@@ -0,0 +1,64 @@
...
+    onItemClick: (ValuePickerItem) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(modifier) {
+        var query by remember { mutableStateOf("") }
+        SearchField(
+            onValueChange = { query = it },
+        )
+
+        Spacer(Modifier.size(Dimensions.Large))
```
araratthehero: This could move to the `Column` as `verticalArrangement = Arrangement.spacedBy(Dimensions.Large)`

ui/src/main/java/com/adyen/checkout/ui/internal/SearchField.kt
```diff
@@ -0,0 +1,175 @@
...
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.unit.sp
+import com.adyen.checkout.test.R
+import com.adyen.checkout.ui.theme.CheckoutTheme
+import kotlinx.coroutines.flow.collectLatest
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Composable
+fun SearchField(
```
araratthehero, line 38: Since `SearchField` is quite identical to `CheckoutTextField`, is it possible to use the `CheckoutTextField` internally?

OscarSpruit: At first I didn't really like this while working on it, but it would save us quite some duplication. I did it anyway by making `state` and `hint` parameters of `CheckoutTextField`. Let me know what you think about the solution

araratthehero: Looks good to me. Now `CheckoutTextField` can be used in more custom flows.

PR #2393, opened 2025-11-05
v6 - Switch payment method integrations

OscarSpruit:
## Description
- Add button to change payment method of the v6 integrations
- Reuse same UI for both sessions and advanced implementations

https://github.com/user-attachments/assets/60b97e62-149d-4726-902e-eac18e38e758


example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6Screen.kt
```diff
@@ -73,12 +93,108 @@ private fun Component(
...
+                bottom = ExampleTheme.dimensions.grid_1,
+            ),
+    ) {
+        Text(text)
+        ExposedDropdownMenuDefaults.TrailingIcon(expanded = isExpanded)
+    }
+}
+
+@Composable
+private fun PaymentMethodDialog(
```
ozgur00: Nit: I think `PaymentMethodOptionsDialog` could be more descriptive for this composable.

PR #2396, opened 2025-11-06
v6 - Drop-in setup base for new activity

OscarSpruit:
## Description
Drop-in setup base for new activity and add entries to example app.

COSDK-719


drop-in/src/main/java/com/adyen/checkout/dropin/CheckoutExt.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.content.Intent
+import com.adyen.checkout.core.components.Checkout
+import com.adyen.checkout.core.components.CheckoutConfiguration
+import com.adyen.checkout.core.components.data.model.PaymentMethodsApiResponse
+import com.adyen.checkout.core.sessions.SessionModel
+import com.adyen.checkout.dropin.internal.ui.DropInActivity
+
+// TODO - Add launcher and service
+@Suppress("UnusedReceiverParameter", "UnusedParameter")
+fun Checkout.startDropIn(
```
OscarSpruit, line 21: This is just a temporary way to start the new activity. In a future PR this will be changed

PR #2397, opened 2025-11-06
v6 - Drop-in setup navigation

OscarSpruit:
## Description
Setup navigation for drop-in.

https://github.com/user-attachments/assets/7d9243e8-5a26-44f1-b92f-9369bb23d7e8

COSDK-719


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/BottomSheetSceneStrategy.kt
```diff
@@ -0,0 +1,78 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+private class BottomSheetScene<T : Any>(
+    override val overlaidEntries: List<NavEntry<T>>,
+    override val key: Any,
+    override val previousEntries: List<NavEntry<T>>,
+    private val entry: NavEntry<T>,
+    private val modalBottomSheetProperties: ModalBottomSheetProperties,
+    private val onBack: () -> Unit,
+) : OverlayScene<T> {
```
ozgur00, line 62: Did you check whether `OverlayScene` is necessary here?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
```diff
@@ -13,33 +13,67 @@ import androidx.activity.compose.setContent
...
-                        modifier = Modifier.padding(innerPadding),
-                    ) {
-                        Title("Drop-in")
-                    }
-                }
+                val backStack = rememberNavBackStack(EmptyNavKey, PreselectedPaymentMethodNavKey)
+                NavDisplay(
+                    backStack = backStack,
+                    sceneStrategy = remember { BottomSheetSceneStrategy() },
+                    onBack = {
```
ozgur00, line 36: We can extract `onBack` code to a private function to make it more readable.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
```diff
@@ -13,33 +13,67 @@ import androidx.activity.compose.setContent
...
+                        }
+
+                        entry<PreselectedPaymentMethodNavKey>(
+                            metadata = BottomSheetSceneStrategy.bottomSheet(),
+                        ) {
+                            PreselectedPaymentMethodScreen(backStack)
+                        }
+
+                        entry<PaymentMethodListNavKey>(
+                            metadata = NavDisplay.transitionSpec {
```
ozgur00: This metadata for transition spec creation could also come from a `TransitionSpecProvider` sort of helper. And the usage would be:
```kotlin
metadata = TransitionSpecProvider.slideInAndOut()
```
Wdyt?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
```diff
@@ -0,0 +1,60 @@
...
+import androidx.compose.material3.TopAppBarDefaults
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import com.adyen.checkout.ui.internal.CheckoutThemeProvider
+import com.adyen.checkout.ui.internal.Dimensions
+import com.adyen.checkout.ui.internal.Title
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+internal fun PaymentMethodListScreen() {
```
ozgur00, line 30: We can pass the `onNavigateClick: () -> Unit` to the `PaymentMethodListScreen` composable and call that once the button is clicked. Then we don't have to interact with back stack from here. 

The example from [android docs](https://developer.android.com/develop/ui/compose/components/app-bars-navigate#example) also does that.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PreselectedPaymentMethodScreen.kt
ozgur00, line 1: Nit: `PreSelected` in the file name should be `Preselected`.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/BottomSheetSceneStrategy.kt
OscarSpruit: It actually is necessary. When using just `Scene` the transitions don't work and everything just jumps around

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
OscarSpruit: I was actually thinking about moving the back stack and all navigation to a `Navigator` class and pass that to all the screens in a follow up PR. What do you think about that?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInActivity.kt
OscarSpruit: Could be solved with the `Navigator` I suggested in the other comment

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
ozgur00: I like the idea!

PR #2398, opened 2025-11-06
V6 - Card - Supported brands list

araratthehero:
## Description
Added supported card brands list, which looks like this:

https://github.com/user-attachments/assets/61d72e8e-9c2c-4d70-b032-70ed46e19c95

## Checklist 

COSDK-660

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
@@ -11,23 +11,13 @@ package com.adyen.checkout.card.internal.ui.view
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.text.KeyboardOptions
-import androidx.compose.foundation.text.input.maxLength
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.focus.onFocusChanged
-import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.tooling.preview.Preview
 import com.adyen.checkout.card.internal.ui.state.CardChangeListener
 import com.adyen.checkout.card.internal.ui.state.CardViewState
-import com.adyen.checkout.core.common.helper.CardNumberValidator
-import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
-import com.adyen.checkout.core.common.localization.internal.helper.resolveString
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputState
-import com.adyen.checkout.ui.internal.CheckoutTextField
 import com.adyen.checkout.ui.internal.ComponentScaffold
-import com.adyen.checkout.ui.internal.DigitOnlyInputTransformation
 import com.adyen.checkout.ui.internal.Dimensions
 import com.adyen.checkout.ui.internal.PayButton
 
@@ -45,11 +35,13 @@ internal fun CardComponent(
         },
     ) {
         Column(
-            modifier = modifier.fillMaxWidth(),
+            modifier = Modifier.fillMaxWidth(),
             verticalArrangement = Arrangement.spacedBy(Dimensions.Large),
         ) {
             CardNumberField(
                 cardNumberState = viewState.cardNumber,
+                supportedCardBrands = viewState.supportedCardBrands,
+                isSupportedCardBrandsShown = viewState.isSupportedCardBrandsShown,
                 isAmex = viewState.isAmex,
                 onCardNumberChanged = changeListener::onCardNumberChanged,
                 onCardNumberFocusChanged = changeListener::onCardNumberFocusChanged,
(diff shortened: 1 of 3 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
```diff
@@ -78,26 +90,47 @@ private fun CardNumberField(
...
+            modifier = Modifier.fillMaxWidth(),
+        ) {
+            FlowRow(
+                modifier = Modifier.padding(top = Dimensions.ExtraSmall),
+                horizontalArrangement = Arrangement.spacedBy(Dimensions.ExtraSmall),
+                verticalArrangement = Arrangement.spacedBy(Dimensions.ExtraSmall),
+            ) {
+                for (cardBrandItem in supportedCardBrands) {
+                    CheckoutNetworkLogo(
+                        modifier = Modifier.size(24.dp, 16.dp),
```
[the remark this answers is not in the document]
araratthehero: We decided for now to leave this hardcoded for the brand icons, since they have exceptions.

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
```diff
@@ -0,0 +1,86 @@
...
+ * @param imageLoader The ImageLoader instance to use. Defaults to the one provided by the Context.
+ * @param size The [LogoSize] required to download the correct sized image.
+ * @param contentDescription The content description for accessibility.
+ * @param placeholder A drawable resource to show while the image is loading.
+ * @param errorFallback A drawable resource to show if the image fails to load.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Suppress("LongParameterList")
+@Composable
+fun CheckoutNetworkLogo(
```
ozgur00, line 41: It's good to add `Checkout` prefix to it, but I think `CheckoutPaymentMethodLogo` is more suited to this composable since it's not for any network logo, but specifically accepts a `txVariant`.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
```diff
@@ -78,26 +96,72 @@ private fun CardNumberField(
         CardNumberOutputTransformation(isAmex = isAmex)
     }
 
-    CheckoutTextField(
-        modifier = Modifier
-            .fillMaxWidth()
-            .onFocusChanged { focusState ->
-                onCardNumberFocusChanged(focusState.isFocused)
+    Column(
```
ozgur00: Maybe we can extract the text input what was previously `CardNumberField` into a private composable and call it `CardNumberInputField`. And in this composable we would have
```kotlin
Column() {
    CardNumberInputField(...)
    CardBrandList(...)
}
```
Wdyt?

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
```diff
@@ -0,0 +1,86 @@
...
+    txVariant: String,
+    environment: Environment,
```
ozgur00: I like that you brought back environment! Good catch! 👍

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
```diff
@@ -0,0 +1,86 @@
...
+    imageLoader: ImageLoader = LocalContext.current.imageLoader,
+    txSubVariant: String = "",
+    size: LogoSize = LogoSize.SMALL,
+    contentDescription: String? = null,
+    @DrawableRes placeholder: Int = R.drawable.ic_placeholder_image,
+    @DrawableRes errorFallback: Int = R.drawable.ic_placeholder_image,
+) {
+    val context = LocalContext.current
+    val fullUrl = remember(context, txVariant, environment, txSubVariant, size) {
+        val densityExtension = context.resources.displayMetrics.densityDpi.getDensityExtension()
```
ozgur00: I just noticed a warning in android studio about using `context.resources`, instead it's suggested to use `LocalResources`. We can get the densityDpi as follows:
```kotlin
val densityDpi = LocalResources.current.displayMetrics.densityDpi
```

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
OscarSpruit: Would it be an idea to provide the `Environment` through a `CompositionLocalProvider`? This way we don't have to pass the environment to the UI when we want to render logos, so it takes away some boilerplate code

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardComponent.kt
araratthehero: I will also move it to another file

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
araratthehero: Hmm, I see what you mean. I was thinking to have different implementations of the `CheckoutNetworkLogo`. If you pass a `txVariant`, it will generate a URL based on the `txVariant`. And we could also have other function declarations which would accept other objects and construct URLs based on those values. But the name would be the same for all Logo fetching composables.

ozgur00: I like what you had in mind! We can keep it as `CheckoutNetworkLogo`.

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
araratthehero: We could achieve that, but in order to create a `CompositionLocal` we need to pass a default value to it. I do not think it is a good idea to use any of the `Environment` values as default.
Another approach is to make the `CompositionLocal` nullable, but it adds extra checks at the usage sights.

Do you know a better way to avoid these problems?

OscarSpruit: The default could just throw an error IMO. If the error is thrown it means we actually made a mistake on our side, so we can catch it early. I have seen other libraries do this as well. Something like:
```Kotlin
private val LocalEnvironment = staticCompositionLocalOf { error("No environment defined") }
```

araratthehero: Makes sense!

PR #2399, opened 2025-11-10
v5 - Ignore UnknownHostException in StatusPollingRepository

ozgur00:
## Description
From API level 35, OS pauses connectivity for apps in background causing an UnknownHostException to be thrown. This PR ignores that error, so this change results in `onError` not being triggered unnecessarily when the app goes to background and connectivity is paused temporarily.

## Checklist 

COSDK-749

## Release notes
[//]: # (### New)
[//]: # (### Fixed)
[//]: # (List fixes here)

### Fixed 
- For devices running Android 15 and above, `onError()` callback is not triggered anymore for Await, QrCode and Twint actions while in background.

### DIFF
--- a/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/StatusRepository.kt
+++ b/components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/StatusRepository.kt
@@ -82,6 +82,8 @@ class DefaultStatusRepository(
                 fetchStatus(it)
             }
             .filterNot {
+                // For API >= 35, apps lose connectivity if in background. That results in an UnknownHostException here.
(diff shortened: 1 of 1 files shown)


components-core/src/main/java/com/adyen/checkout/components/core/internal/data/api/StatusRepository.kt
```diff
@@ -79,6 +81,9 @@ class DefaultStatusRepository(
             .map {
                 fetchStatus(it)
             }
+            .filterNot {
+                it.exceptionOrNull() is UnknownHostException
```
araratthehero, line 87: I would add a comment here to give more context why we are ignoring this exception

PR #2405, opened 2025-11-12
V6 - Card - Drop shadow for network logo

araratthehero:
## Description
- Drop shadow to CheckoutNetworkLogo, since it should be shown for all payment method/brand icons.
- Added background inverse color from the design system

## Checklist 

COSDK-660

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
@@ -13,24 +13,16 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.width
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.foundation.text.input.maxLength
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.clip
-import androidx.compose.ui.draw.dropShadow
 import androidx.compose.ui.focus.onFocusChanged
-import androidx.compose.ui.graphics.shadow.Shadow
 import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.tooling.preview.Preview
-import androidx.compose.ui.unit.DpOffset
-import androidx.compose.ui.unit.dp
 import com.adyen.checkout.card.R
 import com.adyen.checkout.core.common.CardBrand
 import com.adyen.checkout.core.common.CardType
@@ -40,7 +32,6 @@ import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
 import com.adyen.checkout.core.common.localization.internal.helper.resolveString
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputState
 import com.adyen.checkout.ui.internal.CheckoutTextField
-import com.adyen.checkout.ui.internal.CheckoutThemeProvider
 import com.adyen.checkout.ui.internal.DigitOnlyInputTransformation
 import com.adyen.checkout.ui.internal.Dimensions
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
```diff
@@ -58,7 +64,17 @@ fun CheckoutNetworkLogo(
...
+        modifier = modifier
+            .dropShadow(
+                shape = RoundedCornerShape(Dimensions.CornerRadius),
+                shadow = Shadow(
+                    radius = 1.dp,
+                    offset = DpOffset(x = 0.dp, 2.dp),
+                    color = CheckoutThemeProvider.colors.backgroundInverse,
+                    alpha = 0.04f,
+                ),
+            )
```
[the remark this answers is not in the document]
araratthehero: It is better to keep it hardcoded in this composable for now, since in the future we will have more NetworkLogos which accept different types.

ozgur00 (dismissed):

ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutColors.kt
```diff
@@ -20,6 +20,7 @@ value class CheckoutColor(val value: Long)
 @Immutable
 data class CheckoutColors(
     val background: CheckoutColor,
+    val backgroundInverse: CheckoutColor,
```
OscarSpruit: IMO we shouldn't add this color to the public structure. We only use it for shadows and I don't think shadow color is something a lot of people will customize. Let's leave it out to keep the public API minimal and we can add it later if people are asking for it

ui/src/main/java/com/adyen/checkout/ui/internal/InternalColors.kt
```diff
@@ -18,6 +18,7 @@ import com.adyen.checkout.ui.theme.CheckoutColors
 @Immutable
 data class InternalColors(
     val background: Color,
+    val backgroundInverse: Color,
```
OscarSpruit: What do you think of naming this `shadow` to make it more explicit? Or are there other use cases for this color?

ui/src/main/java/com/adyen/checkout/ui/internal/DefaultColors.kt
```diff
@@ -12,6 +12,7 @@ import com.adyen.checkout.ui.theme.CheckoutColor
 
 internal object DefaultColorsLight {
     val BackgroundPrimary = CheckoutColor(0xFFFFFFFF)
+    val BackgroundPrimaryInverse = CheckoutColor(0xFF121212)
```
OscarSpruit: The names here follow the naming of the Bento system. There it's called `BackgroundInversePrimary` and it's values are `#001222` for light and `#EDEDED` for dark

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
```diff
@@ -58,7 +64,17 @@ fun CheckoutNetworkLogo(
 
     NetworkImage(
         url = fullUrl,
-        modifier = modifier,
+        modifier = modifier
```
OscarSpruit, line 68: Figma suggests to use this 👇 Didn't test it, but it seems like it would create this harder shadow to define the border first and add a wider softer shadow after.
```Kotlin
Modifier
    .shadow(elevation = 2.dp, spotColor = Color(0x05001222), ambientColor = Color(0x05001222))
    .shadow(elevation = 4.dp, spotColor = Color(0x0A001222), ambientColor = Color(0x0A001222))
    .padding(0.dp)
    .width(24.dp)
    .height(16.dp)
```

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -115,10 +107,7 @@ private fun CardNumberInputField(
         shouldFocus = cardNumberState.isFocused,
         trailingIcon = {
             CheckoutNetworkLogo(
-                modifier = Modifier
-                    .height(16.dp)
-                    .width(24.dp)
-                    .clip(RoundedCornerShape(Dimensions.CornerRadius)),
+                modifier = Modifier.size(width = 24.dp, height = 16.dp),
```
OscarSpruit: I think we will need this size in more and more places. Should we define a constant?

ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutColors.kt
araratthehero: As discussed, will hardcode this for now

core/src/main/java/com/adyen/checkout/core/common/internal/ui/CheckoutNetworkLogo.kt
araratthehero: Tried, didn't work. Used `dropShadow` instead as discussed with design.

ui/api/ui.api
```diff
@@ -52,6 +52,13 @@ public final class com/adyen/checkout/ui/internal/ComposableSingletons$TextKt {
 	public final fun getLambda$-888577064$ui_release ()Lkotlin/jvm/functions/Function2;
 }
 
+public final class com/adyen/checkout/ui/internal/Dimensions$LogoSize {
```
OscarSpruit: This shouldn't be part of the public API. Could it that nested objects have to be annotated with `@RestrictTo` as well?

araratthehero: Good catch, yes they also need to be annotated.

PR #2407, opened 2025-11-13
v6 - Drop-in ViewModel

OscarSpruit:
## Description
Add a ViewModel to `DropInActivity` and initialize it with the parameters passed by the merchant.

COSDK-719

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PreselectedPaymentMethodScreen.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PreselectedPaymentMethodScreen.kt
@@ -20,6 +20,7 @@ import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.mutableStateListOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.snapshots.SnapshotStateList
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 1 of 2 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PreselectedPaymentMethodScreen.kt
```diff
@@ -33,7 +34,7 @@ import com.adyen.checkout.ui.internal.Title
 
 @Composable
 internal fun PreselectedPaymentMethodScreen(
-    backStack: NavBackStack<NavKey>,
+    backStack: SnapshotStateList<NavKey>,
```
araratthehero, line 37: What is the benefit of using `SnapshotStateList` over the `NavBackStack`?

OscarSpruit: `NavBackStack` is sort of a layer on top of `SnapshotStateList`. I moved the back stack to the ViewModel and opted to use `SnapshotStateList`, because it felt weird to manually create `NavBackStack(mutableStateListOf())` there

OscarSpruit: Adding to that:
`NavBackStack` doesn't do anything, it just delegates everything to the `SnapshotStateList`

araratthehero: Makes sense. Thanks for the explanation.

PR #2420, opened 2025-11-17
Automatically label PRs based on diff size

OscarSpruit:
## Description
Automatically label PRs based on diff size. Line change thresholds:
- Tiny: <= 100
- Small: <= 200
- Medium: <= 500
- Large: <= 1000
- Huge: > 1000

During the diffing deleted files, `.api` files and the verification metadata file are ignored.

### DIFF
--- a/.github/workflows/label_pr_size.yml
+++ b/.github/workflows/label_pr_size.yml
@@ -9,6 +9,7 @@ permissions:
 
 jobs:
   label_pr_size:
+    runs-on: ubuntu-latest
     steps:
       - name: Label PR size
         uses: codelytv/pr-size-labeler@4ec67706cd878fbc1c8db0a5dcd28b6bb412e85a


.github/workflows/label_pr_size.yml
```diff
@@ -0,0 +1,30 @@
...
+
+permissions:
+  pull-requests: write
+  contents: read
+
+jobs:
+  label_pr_size:
+    steps:
+      - name: Label PR size
+        uses: codelytv/pr-size-labeler@4ec67706cd878fbc1c8db0a5dcd28b6bb412e85a
```
araratthehero, line 15: Could the version name (`v1`) be used instead of a the commit hash (`4ec67706cd878fbc1c8db0a5dcd28b6bb412e85a`)?

OscarSpruit: I actually changed this as we got a security alert. `v1` would automatically pick up any new v1 release, so if there would be something malicious in that release we have a problem. Sonar and Github themselves recommend pinning the version with a commit hash

araratthehero: I see. Then we should perhaps do it also for other actions we use, such as [here](https://github.com/Adyen/adyen-android/blob/main/.github/workflows/create_github_release.yml).

araratthehero (approved):

.github/workflows/label_pr_size.yml
OscarSpruit: Yes might be a good idea!

atmamont:
Guys, please bring this to the rest of the platforms! It would be nice to see what impact it brings to the Android team. I hope the bot now shames on huge PRs :)

PR #2424, opened 2025-11-17
v5 - Google Pay issuer country config

OscarSpruit:
## Description
Add `allowedIssuerCountryCodes` and `blockedIssuerCountryCodes` to Google Pay configuration. This allows merchant to allow/block card issued in the listed countries.

COSDK-694

### New
- For Google Pay, added the `allowedIssuerCountryCodes` and `blockedIssuerCountryCodes` configurations which you can use to allow/block cards issued in the provided countries.

### DIFF
--- a/googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayConfiguration.kt
+++ b/googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayConfiguration.kt
@@ -242,7 +242,7 @@ class GooglePayConfiguration private constructor(
 
         /**
          * Sets the allowed issuer country codes. This list should contain ISO 3166-1 alpha-2 country codes. When set
-         * the shopper will only be able the select payment methods issued in the provided countries.
+         * the shopper will only be able to select payment methods issued in the provided countries.
          *
          * Note: allowedIssuerCountryCodes and blockedIssuerCountryCodes are mutually exclusive. Only set one at a time.
          *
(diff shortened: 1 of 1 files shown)


googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayConfiguration.kt
```diff
@@ -236,6 +240,38 @@ class GooglePayConfiguration private constructor(
...
+         * Sets the allowed issuer country codes. This list should contain ISO 3166-1 alpha-2 country codes. When set
+         * the shopper will only be able the select payment methods issued in the provided countries.
+         *
+         * Note: allowedIssuerCountryCodes and blockedIssuerCountryCodes are mutually exclusive. Only set one at a time.
+         *
+         * Check the
+         * [Google Pay docs](https://developers.google.com/pay/api/android/reference/request-objects#CardParameters)
+         * for more details.
+         */
+        @Deprecated("Use property access syntax instead.")
```
ozgur00, line 253: I'm thinking whether we should add new deprecated functions even. If (probably yes) they use setters and  these, their code will look a bit weird, but it could potentially give them some incentive to already move toward the property access syntax. WDYT? 🤔

OscarSpruit: For consistency I thought it would be nice to add these functions. Also if merchants are using the builder pattern to create configurations then only offering the property would kind of force them to migrate to the new way of writing configurations. IMO it's better we offer these deprecated functions still on v5

ozgur00: Fair point. We can keep them.

ozgur00 (approved):

PR #2426, opened 2025-11-18
V6 - Card - Dual branded cards

araratthehero:
## Description
- Show the card brand icons next to the card number field
- Create ListOption composable
- Based on designs, align spacing between UI element groups
- When dual branded flow, show the brand selector and pass the result to the component data

image

## Checklist 
- [x] Code is unit tested

COSDK-660

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/DualBrandSelector.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/DualBrandSelector.kt
@@ -14,6 +14,7 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import com.adyen.checkout.card.internal.ui.model.CardBrandItem
 import com.adyen.checkout.card.internal.ui.model.DualBrandData
 import com.adyen.checkout.core.common.CardBrand
 import com.adyen.checkout.core.common.internal.ui.CheckoutNetworkLogo
@@ -26,12 +27,10 @@ import com.adyen.checkout.ui.internal.SelectableListItem
 
 @Composable
 internal fun DualBrandSelector(
-    dualBrandData: DualBrandData?,
+    dualBrandData: DualBrandData,
     onBrandSelected: (CardBrand) -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    if (dualBrandData == null || !dualBrandData.selectable || dualBrandData.brandOptions.size < 2) return
-
     Column(
         modifier = modifier.fillMaxWidth(),
         verticalArrangement = Arrangement.spacedBy(Dimensions.Large),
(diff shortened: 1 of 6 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/DualBrandSelector.kt
```diff
@@ -0,0 +1,84 @@
...
+private fun BrandsListSection(
+    dualBrandData: DualBrandData,
+    onBrandSelected: (CardBrand) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(
+        modifier = modifier.fillMaxWidth(),
+        verticalArrangement = Arrangement.spacedBy(Dimensions.Small),
+    ) {
+        dualBrandData.brandOptions.take(2).map { brandItem ->
```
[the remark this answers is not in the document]
araratthehero: This is intended

card/src/main/java/com/adyen/checkout/card/internal/ui/view/DualBrandSelector.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.adyen.checkout.ui.internal.Dimensions
+import com.adyen.checkout.ui.internal.ListOption
+
+@Composable
+internal fun DualBrandSelector(
+    dualBrandData: DualBrandData?,
+    onBrandSelected: (CardBrand) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    if (dualBrandData == null || !dualBrandData.selectable || dualBrandData.brandOptions.size < 2) return
```
OscarSpruit: I think this logic should be moved to the call site on `CardComponent`. Having it here could be an unexpected side effect and I think right now `CardComponent` will create the space between `CardDetailsSection` and  `DualBrandSelector` even when the brand selector is not shown

ui/src/main/java/com/adyen/checkout/ui/internal/ListOption.kt
```diff
@@ -0,0 +1,240 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.vectorResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.adyen.checkout.test.R
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Composable
+fun ListOption(
```
OscarSpruit: What do you think of splitting this composable up into a plain `ListItem` that doesn't have the selectable logic, background, and checkmark and a `SelectableListItem` that applies the correct modifiers and uses the trailing icon to render the checkmark? It would separate functionality and for more simple use cases the composable will be more performant

card/src/main/java/com/adyen/checkout/card/internal/ui/view/DualBrandSelector.kt
araratthehero: It makes more sense to move it to the call site, but the `DualBrandSelector` is a bit of a tricky Composable. Inside we do `dualBrandData.brandOptions.take(2)` to get the first two items from the `brandOptions` list. Regardless the check on the call sight, it would be good to check the `brandOptions` size before trying to get the first two items, to not crash the app if the list is incorrectly populated and the call site does not have a check for length. `selectable` could also be easily ignored from the call sight.

Regarding the spacing, when the composable is empty, it does not show the space between the `CardDetailsSection` and `DualBrandSelector`.

What kind of unexpected side effects are you thinking of? What do you think if we move the nullability check to the `CardComponent ` but keep the other checks inside the `DualBrandSelector`?

OscarSpruit: What if we change `DualBrandData` a bit? Something like this:
```kotlin
data class DualBrandData(
    val optionOne: CardBrandItem,
    val optionTwo: CardBrandItem,
)
```
This model restricts there to be exactly 2 options. If there are less `DualBrandData` should be null. If there are more we ignore the extras. Instead of the `selectable` flag `DualBrandData` should be null when it's not "selectable". `CardBrandItem` contains the `CardBrand` and a selected flag, so imo the `selectedBrand` is unnecessary.

With that model we could do the null check in `CardComponent` and we will for sure know that the dual brand selection will render

OscarSpruit (approved):

PR #2427, opened 2025-11-18
v6: Card - Bin Lookup Callbacks

ozgur00:
## Description
Add support for optional bin lookup callbacks.

## Checklist 

COSDK-788

### DIFF
--- a/card/api/card.api
+++ b/card/api/card.api
@@ -1,16 +1,14 @@
 public final class com/adyen/checkout/card/BinLookupData {
 	public static final field $stable I
-	public fun <init> (Ljava/lang/String;Ljava/lang/String;Z)V
+	public fun <init> (Ljava/lang/String;Ljava/lang/String;)V
 	public final fun component1 ()Ljava/lang/String;
 	public final fun component2 ()Ljava/lang/String;
-	public final fun component3 ()Z
-	public final fun copy (Ljava/lang/String;Ljava/lang/String;Z)Lcom/adyen/checkout/card/BinLookupData;
-	public static synthetic fun copy$default (Lcom/adyen/checkout/card/BinLookupData;Ljava/lang/String;Ljava/lang/String;ZILjava/lang/Object;)Lcom/adyen/checkout/card/BinLookupData;
+	public final fun copy (Ljava/lang/String;Ljava/lang/String;)Lcom/adyen/checkout/card/BinLookupData;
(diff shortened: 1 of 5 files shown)


card/src/main/java/com/adyen/checkout/card/internal/CardCallbacks.kt
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+package com.adyen.checkout.card.internal
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.card.OnBinLookupCallback
+import com.adyen.checkout.card.OnBinValueCallback
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class CardCallbacks internal constructor() {
```
araratthehero, line 16: Can this be internal?

card/src/main/java/com/adyen/checkout/card/CardCallbackExt.kt
```diff
@@ -0,0 +1,28 @@
...
+
+fun CheckoutCallbacks.card(initBlock: CardCallbacks.() -> Unit) {
+    apply {
+        val callbacks = CardCallbacks().apply(initBlock)
+        callbacks.onBinValue?.let { addCallback(it, OnBinValueCallback::class) }
+        callbacks.onBinLookup?.let { addCallback(it, OnBinLookupCallback::class) }
+    }
+}
+
+fun CardCallbacks.onBinValue(onBinValueCallback: OnBinValueCallback) {
```
OscarSpruit, line 22: Did you align with iOS on what data we provide in these callbacks? IIRC all platforms provide very different things

card/src/main/java/com/adyen/checkout/card/internal/CardCallbacks.kt
ozgur00: In order to be able to expose extension functions that sets callbacks internally, we have to keep this public. I have marked it restricted to add a layer of protection, in case a merchant tries to access it directly.

card/src/main/java/com/adyen/checkout/card/CardCallbackExt.kt
ozgur00: Yes we did, I removed `isReliable` and named the `onBinLookup` parameter to `data` as per those discussions.

PR #2432, opened 2025-11-20
v6: Card - Extract payment component state mapping

ozgur00:
## Description
Extract mapping from `CardViewState` to `CardPaymentComponentState` to a separate file.

## Checklist 

COSDK-570

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
@@ -39,6 +39,7 @@ import com.adyen.checkout.core.components.internal.ui.PaymentComponent
 import com.adyen.checkout.core.components.internal.ui.navigation.CheckoutNavEntry
 import com.adyen.checkout.core.components.internal.ui.state.StateManager
 import com.adyen.checkout.core.components.paymentmethod.CardPaymentMethod
+import com.adyen.checkout.cse.EncryptionException
 import com.adyen.checkout.cse.internal.BaseCardEncryptor
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
@@ -92,11 +93,9 @@ internal class CardComponent(
                 componentParams = componentParams,
                 cardEncryptor = cardEncryptor,
                 checkoutAttemptId = analyticsManager.getCheckoutAttemptId(),
-            ) { e ->
-                val event = GenericEvents.error(CardPaymentMethod.PAYMENT_METHOD_TYPE, ErrorEvent.ENCRYPTION)
-                analyticsManager.trackEvent(event)
-                // exceptionChannel.trySend(e)
-            }
+                onEncryptionFailed = ::onEncryptionError,
+                onPublicKeyNotFound = ::onPublicKeyNotFound,
+            )
             val event = PaymentComponentEvent.Submit(paymentComponentState)
             eventChannel.trySend(event)
         } else {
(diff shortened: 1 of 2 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -79,7 +73,15 @@ internal class CardComponent(
...
         if (stateManager.isValid) {
-            val paymentComponentState = stateManager.viewState.value.toPaymentComponentState()
+            val paymentComponentState = stateManager.viewState.value.toPaymentComponentState(
+                componentParams = componentParams,
+                cardEncryptor = cardEncryptor,
+                checkoutAttemptId = analyticsManager.getCheckoutAttemptId(),
+            ) {
+                val event = GenericEvents.error(CardPaymentMethod.PAYMENT_METHOD_TYPE, ErrorEvent.ENCRYPTION)
+                analyticsManager.trackEvent(event)
+                // exceptionChannel.trySend(e)
```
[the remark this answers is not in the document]
ozgur00: Passed the exception here to later propagate it.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateExt.kt
```diff
@@ -0,0 +1,96 @@
...
+    componentParams: CardComponentParams,
+    cardEncryptor: BaseCardEncryptor,
+    checkoutAttemptId: String,
+    onEncryptionError: (EncryptionException) -> Unit,
+): CardPaymentComponentState {
+    val unencryptedCardBuilder = UnencryptedCard.Builder()
+
+    val publicKey = componentParams.publicKey
+    if (publicKey == null) {
+        return invalidCardPaymentComponentState()
```
OscarSpruit: Should we propagate an error to the merchant as well in this case? The shopper most likely can't move forward in this case, so the merchant would then at least be able to show some error state

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateExt.kt
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    val encryptedCard: EncryptedCard = try {
```
OscarSpruit: Maybe we could extract this into a `encryptCard()` method?

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateExt.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        cardEncryptor.encryptFields(unencryptedCardBuilder.build(), publicKey)
+    } catch (e: EncryptionException) {
+        onEncryptionError(e)
+        return invalidCardPaymentComponentState()
+    }
+    return mapComponentState(encryptedCard, componentParams, checkoutAttemptId)
+}
+
+private fun mapComponentState(
```
OscarSpruit: The name of this method doesn't make much sense to me. If we move the encryption code out of the function above, we could move this code into `toPaymentComponentState` or we should try and split this up into a lot of small functions, like `createPaymentMethod()`, `createPaymentComponentData`, `createCardComponentState` etc

ozgur00: I like the idea. Made the whole function as granular as possible to make the top function more readable.

PR #2433, opened 2025-11-20
v6 - Drop-in preselected landing

OscarSpruit:
## Description
Implement the landing screen of the pre-selected payment flow. The "Use..." will be implemented later.

https://github.com/user-attachments/assets/8eb2cc49-3e80-4354-a20c-2042ddcafbaa

COSDK-801


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PreselectedPaymentMethodScreen.kt
```diff
@@ -35,37 +44,52 @@ import com.adyen.checkout.ui.internal.Title
...
         }
 
-        // TODO - Center text
+        Spacer(Modifier.size(Dimensions.ExtraLarge))
+
+        CheckoutNetworkLogo(
+            txVariant = viewState.logoTxVariant,
+            contentDescription = null,
+            modifier = Modifier
+                .size(80.dp, 52.dp)
```
ozgur00: The same size is likely to be used in other screens, we could define it in `Dimensions`.

PR #2436, opened 2025-11-25
Setup acceptance releases

OscarSpruit:
## Description
Make our example app ready for acceptance testing:
- Create acceptance build type
- Add different app icon per build type
- Add workflow to publish to Firebase

COSDK-765

### DIFF
--- a/.github/workflows/release_acceptance_app.yml
+++ b/.github/workflows/release_acceptance_app.yml
@@ -50,6 +50,7 @@ jobs:
         run: ./gradlew :example-app:assembleAcceptance -Pversion-name=$VERSION_NAME
 
       - name: Upload to Firebase App Distribution
+        # The version SHA refers to v1.7.1
         uses: wzieba/Firebase-Distribution-Github-Action@bd494989dd4bec0343f78adee87fe66e48279ad6
         env:
           ADYEN_ANDROID_FIREBASE_APP_ID: ${{ secrets.ADYEN_ANDROID_FIREBASE_APP_ID }}


.github/workflows/release_acceptance_app.yml
```diff
@@ -0,0 +1,63 @@
+name: Release acceptance app
+
+on:
```
jreij, line 3: Do we also want to publish this app after every release?

.github/workflows/release_acceptance_app.yml
```diff
@@ -0,0 +1,63 @@
...
+          ADYEN_ANDROID_ACCEPTANCE_MERCHANT_SERVER_URL: ${{ secrets.ADYEN_ANDROID_ACCEPTANCE_MERCHANT_SERVER_URL }}
+          ADYEN_ANDROID_ACCEPTANCE_CLIENT_KEY: ${{ secrets.ADYEN_ANDROID_ACCEPTANCE_CLIENT_KEY }}
+          SIGNING_STORE_FILE: ${{ steps.decode_keystore.outputs.keystore_path }}
+          SIGNING_STORE_PASSWORD: ${{ secrets.ADYEN_ANDROID_ACCEPTANCE_SIGNING_PASSWORD }}
+          SIGNING_KEY_ALIAS: ${{ secrets.ADYEN_ANDROID_ACCEPTANCE_SIGNING_KEY_ALIAS }}
+          SIGNING_KEY_PASSWORD: ${{ secrets.ADYEN_ANDROID_ACCEPTANCE_SIGNING_PASSWORD }}
+        run: ./gradlew :example-app:assembleAcceptance -Pversion-name=$VERSION_NAME
+
+      - name: Upload to Firebase App Distribution
+        uses: wzieba/Firebase-Distribution-Github-Action@bd494989dd4bec0343f78adee87fe66e48279ad6
```
ozgur00, line 54: Could we leave a comment here to explain why we use this specific commit, including what it points to? It's good practice for security indeed, but we can make it clear what it points to in a comment to make it easier to understand the reasoning behind using this specific commit.

.github/workflows/release_acceptance_app.yml
OscarSpruit: I think it would be more useful to have a build that uses the actual release, like the MyStore demo app. Let's discuss with the team

.github/workflows/release_acceptance_app.yml
OscarSpruit: Added a comment saying what release it points to. It really is just the latest "safe" version as of now

ozgur00 (approved):

PR #2439, opened 2025-11-26
v6 - Update design system

OscarSpruit:
## Description
Update design system according to latest state:
- Remove `success` color
- Add `containerOutline` color
- Align color usage for text fields
- Sync colors with Bento
- Remove `elements` from theming structure
- Update weight for text styles
- Remove text styling customization options
    - Only the general font is still customizable
- Use emphasized text style for selected state of button group
- Organize `ui` module with improved package structure

COSDK-147

### DIFF
--- a/ui/src/main/java/com/adyen/checkout/ui/internal/element/InternalElements.kt
+++ b/ui/src/main/java/com/adyen/checkout/ui/internal/element/InternalElements.kt
@@ -16,8 +16,8 @@ import com.adyen.checkout.ui.internal.element.button.InternalButtonGroupStyle
 import com.adyen.checkout.ui.internal.element.button.InternalButtonStyle
 import com.adyen.checkout.ui.internal.element.input.CheckoutTextFieldDefaults
 import com.adyen.checkout.ui.internal.element.input.InternalTextFieldStyle
+import com.adyen.checkout.ui.internal.theme.InternalAttributes
 import com.adyen.checkout.ui.internal.theme.InternalColors
-import com.adyen.checkout.ui.theme.CheckoutAttributes
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 @Immutable
@@ -32,7 +32,7 @@ data class InternalElements(
 
         fun from(
             colors: InternalColors,
-            attributes: CheckoutAttributes,
+            attributes: InternalAttributes,
         ): InternalElements {
             return InternalElements(
                 buttons = InternalButtons(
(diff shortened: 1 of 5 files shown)


ui/src/main/java/com/adyen/checkout/ui/internal/theme/InternalColors.kt
```diff
@@ -19,11 +19,11 @@ import com.adyen.checkout.ui.theme.CheckoutColors
 data class InternalColors(
     val background: Color,
     val container: Color,
+    val containerOutline: Color,
```
ozgur00, line 22: I don't see this new `containerOutline` in the Components section in Figma. For my understanding, how is it different from the outline field?

ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutTextStyles.kt
```diff
@@ -8,51 +8,21 @@
...
 data class CheckoutTextStyles(
-    val title: CheckoutTextStyle,
-    val subtitle: CheckoutTextStyle,
-    val body: CheckoutTextStyle,
-    val bodyEmphasized: CheckoutTextStyle,
-    val subHeadline: CheckoutTextStyle,
-    val subHeadlineEmphasized: CheckoutTextStyle,
-    val footnote: CheckoutTextStyle,
-    val footnoteEmphasized: CheckoutTextStyle,
+    @FontRes val font: Int?,
```
ozgur00, line 17: Android Studio warns about ambiguity about  `@FontRes` usage. Since we don't want merchants to pass something that's not a font rest, should we replace it with `@param:FontRes`? Wdyt?

ozgur00:
What do you think about adding a package under `internal` to encapsulate `button`, `image`, `input` and `text` packages? For name I can think of `elements`, `view`, `components` (this one not so much), but I'm not sold on any of them. 

Then we'd have:
```
> internal
>> elements (view)
>> helper
>> theme
...
```

There'd be less packages under internal and that could make it easier to navigate the files imo.

ui/src/main/java/com/adyen/checkout/ui/internal/theme/InternalColors.kt
OscarSpruit: Oh I guess they forgot to update Figma. The `containerOutline` is used to give every container possibly a different border color. Think about the text input, stepper, etc. Since `elements` are now removed, this color covers the styling options that `elements` provided.

ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutTextStyles.kt
OscarSpruit: Hmm I don't get this warning. Also I think this should be solved when we update to kotlin 2.2.0. Kotlin is then able to detect by itself how to apply the annotation.

PR #2441, opened 2025-11-27
Agents - Add initial version of AGENTS.md file

araratthehero:
## Description
Added initial version of AGENTS.md file

COSDK-825

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,14 +14,71 @@ This document outlines important patterns, practices, and rules to follow when w
 
 **Do not proceed with uncertain changes.** It's better to ask and get clarification than to implement incorrect solutions.
 
-### 2. Keep Implementation Plans Updated
-**When working on multi-phase tasks with a plan document (e.g., `*_IMPLEMENTATION_PLAN.md`):**
+### 2. Always Work with a Plan
+
+**CRITICAL: Create plan FIRST, implement AFTER approval:**
+- **NEVER start implementation without a plan**
+- First, create a comprehensive implementation plan document (e.g., `*_IMPLEMENTATION_PLAN.md`)
+- Present the plan to the user for review and approval
+- Only after the plan is approved, begin implementation
+- Do not execute any code changes during the planning phase
+
+**Before starting any task:**
+- Ensure there is an implementation plan document (e.g., `*_IMPLEMENTATION_PLAN.md`)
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,302 @@
...
+- Code patterns or conventions
+- Expected behavior
+- Implementation details
+- Migration strategies
+- Whether a change is needed or correct
+
+**Do not proceed with uncertain changes.** It's better to ask and get clarification than to implement incorrect solutions.
+
+### 2. Keep Implementation Plans Updated
+**When working on multi-phase tasks with a plan document (e.g., `*_IMPLEMENTATION_PLAN.md`):**
```
atmamont: I'd try thinking ahead and maybe guide an agent into creating such plans and put them into a specific folder. It's up to the team to decide then if you want to commit those plans for features like a big migration or refactor. Reviewing the initial plan for a bigger task could be beneficial as it allows to collect early feedback from the team with the approach that you and agent plan to take.
Speaking from my experience, I have now bunch of local md files that I use when switching across several branches, and I notice it would be great to have a regulated approach and agreement what do we do with such files.

AGENTS.md
```diff
@@ -0,0 +1,302 @@
...
+// Public API
+class CardConfiguration { }
+```
+
+**When making public API changes:**
+
+**Carefully review all changes:**
+- Review each change carefully to ensure it's necessary and correct. If in doubt, ask questions.
+
+**Be wary of breaking changes:**
```
atmamont: I'd be explicit here and tell not ever do breaking changes until discussed and confirmed. Maybe you can save up some context space by just being more strict on public API and then elaborate more on every specific case.

AGENTS.md
```diff
@@ -0,0 +1,302 @@
...
+
+**For sealed classes that are safe to use:**
+- If adding a new entry with no arguments to an existing sealed class, always make it a `class` and not an `object`
+- This allows adding optional arguments in the future without a breaking change
+- Example: `AddressConfiguration` is safe because merchants use it as a parameter, not in `when` expressions
+
+```kotlin
+// This is okay - merchants don't use when with it
+sealed class AddressConfiguration : Parcelable {
+    object None : AddressConfiguration()
```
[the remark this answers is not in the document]
araratthehero: We mention that this is an example of an acceptable case above.

PR #2447, opened 2025-12-02
V6 - Merge queue - Workflow adjustments

araratthehero:
Adjust actions which run on `pull_request` trigger

COSDK-837

### DIFF
--- a/.github/workflows/check_labels.yml
+++ b/.github/workflows/check_labels.yml
@@ -18,6 +18,8 @@ concurrency:
 jobs:
   get_allowed_labels:
     runs-on: ubuntu-latest
+    # This step will only run on pull_request events.
+    # PR information is only available when workflow gets triggered with pull_request event.
     if: github.event_name == 'pull_request'
 
     outputs:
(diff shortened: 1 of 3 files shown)


.github/workflows/check_labels.yml
jreij, line 67: We might need to skip this job (mark it as passed) if the trigger was the merge queue, this line will fail because there is no `github.event.pull_request` since there is not PR to begin with, it's just a branch. But maybe we can enable the merge queue to verify this first.

araratthehero: Indeed, this will probably cause an issue since this step is required to merge a PR.
I made an adjustment to run the jobs only if the event type is `pull_request`.

jreij: I don't think your fix will work, because the if condition will skip the job altogether for the merge queue, which means that the status check won't be marked as passed and the merge queue will just get stuck.

I think the `if` condition should be actually inside the `run` block of the required job (`labels-check`) and should mark the job as successful if the trigger in the merge queue.

OscarSpruit (approved):

PR #2449, opened 2025-12-02
v6: Card - Input field icons

ozgur00:
## Description
Add trailing icons to the card input fields to display placeholder, invalid and valid states.

## Checklist 

COSDK-811

### DIFF
--- a/ui/src/main/java/com/adyen/checkout/ui/internal/helper/Icons.kt
+++ b/ui/src/main/java/com/adyen/checkout/ui/internal/helper/Icons.kt
@@ -22,8 +22,8 @@ fun getThemedIcon(
     lightDrawableId: Int,
     darkDrawableId: Int,
 ): Int {
-    val isDark = backgroundColor.isDark()
     return remember(backgroundColor) {
+        val isDark = backgroundColor.isDark()
         if (isDark) darkDrawableId else lightDrawableId
     }
 }
--- a/ui/src/main/java/com/adyen/checkout/ui/internal/helper/ColorExt.kt
+++ b/ui/src/main/java/com/adyen/checkout/ui/internal/helper/ColorExt.kt
@@ -9,11 +9,9 @@
 package com.adyen.checkout.ui.internal.helper
 
 import androidx.annotation.RestrictTo
-import androidx.compose.runtime.Composable
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.luminance
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-@Composable
 fun Color.isDark() = luminance() < LUMINANCE_THRESHOLD
 private const val LUMINANCE_THRESHOLD = 0.5f


ui/src/main/java/com/adyen/checkout/ui/internal/InternalCheckoutTheme.kt
```diff
@@ -64,3 +65,8 @@ private val LocalColors = staticCompositionLocalOf { InternalColors.from(Checkou
 private val LocalTextStyles = staticCompositionLocalOf { CheckoutTextStyles.default() }
 private val LocalAttributes = staticCompositionLocalOf { CheckoutAttributes.default() }
 private val LocalElements = staticCompositionLocalOf { CheckoutElements.default() }
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Composable
+fun isDarkMode() = CheckoutThemeProvider.colors.background.luminance() < LUMINANCE_THRESHOLD
```
OscarSpruit: I think we should check this based on the color that is used. For example, it might be that in dark mode the input field is customized to have a white background, so we should use the light icon in that case. By checking the global background color we would use the dark icon.

Maybe we can also fix this by passing a color to this function?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardIcons.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.adyen.checkout.card.R
+import com.adyen.checkout.ui.internal.isDarkMode
+
+@Composable
+internal fun rememberCardIcons(): CardIcons {
+    val isDark = isDarkMode()
+    return remember(isDark) {
+        if (isDark) CardIcons.Dark else CardIcons.Light
+    }
+}
```
OscarSpruit: Together with the other comment I think we could change this function into something like:
```suggestion
@Stable
@Composable
internal fun getThemedIcon(
    backgroundColor: Color,
    lightDrawableId: Int,
    darkDrawableId: Int,
): Int {
    return remember(backgroundColor) {
        if (backgroundColor.isDark()) darkDrawableId else lightDrawableId
    }
}
```
This would make the function more generic and reusable in other places. It would also make sure you don't get more res ids than you really need

ui/src/main/java/com/adyen/checkout/ui/internal/InternalCheckoutTheme.kt
ozgur00: I created an extension to `Color` function compose.

ui/src/main/java/com/adyen/checkout/ui/internal/helper/Icons.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@Stable
+@Composable
+fun getThemedIcon(
+    backgroundColor: Color,
+    lightDrawableId: Int,
+    darkDrawableId: Int,
+): Int {
+    val isDark = backgroundColor.isDark()
```
OscarSpruit: Small improvement: if we move this line inside of the `remember`, then the luminance is not calculated every on every recomposition

ui/src/main/java/com/adyen/checkout/ui/internal/theme/Dimensions.kt
```diff
@@ -36,7 +36,9 @@ object Dimensions {
 
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     object LogoSize {
-        val small = DpSize(width = 27.dp, height = 18.dp)
+        val small = DpSize(width = 24.dp, height = 16.dp)
```
OscarSpruit: Did you discuss this change with design? I believe @araratthehero changed it to 27x18 based on feedback from design a while ago

ui/src/main/java/com/adyen/checkout/ui/internal/helper/Icons.kt
ozgur00: It's not possible to call a `@Composable` inside remember, that's why I left it out.

ui/src/main/java/com/adyen/checkout/ui/internal/theme/Dimensions.kt
ozgur00: Good catch! Wasn't aware of the change, reverted it.

PR #2451, opened 2025-12-04
v5 - Kover coverage

OscarSpruit:
## Description
Replace Jacoco with Kover for coverage reports. Kover is Kotlin first, it's better in understanding Kotlin code coverage and it allows to exclude annotations (useful for composables for example). 

## Motivation
With v6 using compose it will seem like our coverage will drop significantly. Using Kover for v6 allows us to have a more realistic report. To have a better comparison between v5 and v6, it makes sense to use Kover on v5 as well.

### DIFF
--- a/config/gradle/kover.gradle
+++ b/config/gradle/kover.gradle
@@ -2,14 +2,6 @@
 // the `plugins` block cannot be used in gradle files such as this one, which are imported into other gradle files with the `apply from` command
 apply plugin: 'org.jetbrains.kotlinx.kover'
 
-if (project.hasProperty('android')) {
-    android.buildTypes {
-        debug {
-            testCoverageEnabled = true
-        }
-    }
-}
-
 kover {
     reports {
         filters {


build.gradle
araratthehero: Can the jacoco usage be removed from here and from `libs.versions.toml`?

config/gradle/sonarcloud.gradle
```diff
@@ -42,10 +42,7 @@ subprojects {
         properties {
             property 'sonar.androidLint.reportPaths', "${layout.buildDirectory.get().asFile}/reports/lint-results-debug.xml"
             property 'sonar.kotlin.detekt.reportPaths', "${layout.buildDirectory.get().asFile}/reports/detekt/detekt-results.xml"
-            property 'sonar.coverage.jacoco.xmlReportPaths', "${layout.buildDirectory.get().asFile}/reports/jacoco/jacocoDebugTestReport/jacocoDebugTestReport.xml"
-
-            def sonarExclusions = coverageExclusions.join(', ')
-            property 'sonar.coverage.exclusions', sonarExclusions
+            property 'sonar.coverage.jacoco.xmlReportPaths', "${layout.buildDirectory.get().asFile}/reports/kover/report.xml"
```
araratthehero, line 45: Can this be changed to the following?

```suggestion
            property 'sonar.coverage.xmlReportPaths', "${layout.buildDirectory.get().asFile}/reports/kover/report.xml"
```

or


```suggestion
            property 'sonar.coverage.kover.xmlReportPaths', "${layout.buildDirectory.get().asFile}/reports/kover/report.xml"
```

OscarSpruit: Unfortunately not. Sonar only support the property we have ([source](https://docs.sonarsource.com/sonarqube-server/9.9/analyzing-source-code/test-coverage/test-coverage-parameters#java))

config/gradle/kover.gradle
```diff
@@ -0,0 +1,80 @@
+// cannot reference the plugins declared in the version catalogs here because it requires the use of the `plugins` block instead of the `apply plugin` command
+// the `plugins` block cannot be used in gradle files such as this one, which are imported into other gradle files with the `apply from` command
+apply plugin: 'org.jetbrains.kotlinx.kover'
+
+kover {
```
araratthehero, line 5: I see some exclusions are not included in this file.
 
- All generated dependency injection code (e.g., *_Factory, Dagger*, *_HiltComponents).
- R classes, Manifest files, and specific ViewBinding/BindingAdapter classes (I see only `databinding.*` being excluded).
- Adapters, Dialogs, Drawables (only Activities and Fragments are currently excluded).
- Specific files including ComposeExtensions, DropInExt, FileDownloader, BuildUtils, and ActivityResult contracts.

Is this intended?

OscarSpruit: - Dagger and Hilt code is already excluded because we exclude the entirety of the example app
- R classes and manifest related things I could not find in the klover reports, so I figured it's safe to remove from this list. Excluding the `databinding` package also excludes the specific binding classes
- These are tricky, because we have a few classes ending in dialog and adapter that can actually be unit tested. I thought to not put those in the list, so that once we remove the old view code we can see which classes are not covered with tests.
- ComposeExtensions (can't find DropInExt) is covered by excluding `@Composable` functions. The other files can actually be tested by mocking android context I think, so I didn't add them for a more realistic coverage report

Does all of that make sense? Let me know if you have different thoughts on this

araratthehero: Makes sense!

PR #2453, opened 2025-12-05
v6 - Drop-in: navigator

OscarSpruit:
## Description
Extract navigation logic into a navigator object. This streamlines navigation across different drop-in screen and eliminates the need to manually handle back presses.

COSDK-719

### DIFF
--- a/drop-in/src/test/java/com/adyen/checkout/dropin/internal/ui/DropInNavigatorTest.kt
+++ b/drop-in/src/test/java/com/adyen/checkout/dropin/internal/ui/DropInNavigatorTest.kt
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2025 Adyen N.V.
+ *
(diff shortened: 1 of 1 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInNavigator.kt
```diff
@@ -0,0 +1,42 @@
...
+
+package com.adyen.checkout.dropin.internal.ui
+
+import androidx.compose.runtime.mutableStateListOf
+import androidx.compose.runtime.snapshots.SnapshotStateList
+import androidx.navigation3.runtime.NavKey
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+internal class DropInNavigator {
```
ozgur00, line 17: We could write tests for this class.

PR #2461, opened 2025-12-11
v6: Card - Display optional tag for Expiry Date & Security Code

ozgur00:
## Description
Based on the bin lookup response, mark expiry date and security code optional if it applies. Display an `Optional` tag in the title if the field is optional.

## Checklist 

COSDK-862


core/src/main/res/values/strings.xml
```diff
@@ -16,11 +16,13 @@
     <string name="checkout_card_number_invalid" translatable="false">Invalid card number</string>
     <string name="checkout_card_number_invalid_unsupported_brand" translatable="false">The entered card brand isn\'t supported</string>
     <string name="checkout_card_expiry_date" translatable="false">Expiry Date</string>
+    <string name="checkout_card_expiry_date_optional" translatable="false">Expiry Date (Optional)</string>
```
OscarSpruit: What do you think about splitting the "(Optional)" into it's own string? This way we wouldn't have to duplicate all labels that can be optional. This could also simplify the state management of input fields as we can just pass an `isOptional` boolean and the input itself displays the optional text behind the label

ozgur00: I like the idea to have `(Optional)` as a separate resource 👍 . I've updated those parts. However since a field can have 3 states, we cannot simplify `requirementPolicy` into a boolean.

card/src/main/java/com/adyen/checkout/card/internal/ui/helper/RequirementPolicyExt.kt
```diff
@@ -0,0 +1,67 @@
...
+package com.adyen.checkout.card.internal.ui.helper
+
+import com.adyen.checkout.card.internal.data.model.Brand
+import com.adyen.checkout.card.internal.data.model.DetectedCardType
+import com.adyen.checkout.card.internal.ui.model.CVCVisibility
+import com.adyen.checkout.card.internal.ui.model.CardComponentParams
+import com.adyen.checkout.card.internal.ui.model.StoredCVCVisibility
+import com.adyen.checkout.core.components.internal.ui.state.model.RequirementPolicy
+
+internal fun CVCVisibility.requirementPolicy() = when (this) {
```
OscarSpruit: I looks like all these extension functions are unused. Can they be removed?

OscarSpruit (approved):

PR #2463, opened 2025-12-16
v6 - Card State Migration

ozgur00:
## Description
Migrate Card Component state management to MVI based state management using Reducers and Producers.

## Checklist 
- [x] Code is unit tested

COSDK-877


card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -99,118 +110,50 @@ internal class CardComponent(
...
+        )
 
-    override fun onStorePaymentMethodChanged(checked: Boolean) {
-        stateManager.updateViewState {
-            copy(
-                storePaymentMethod = checked,
-            )
-        }
-    }
+        val oldBinValue = viewState.value.binValue
```
OscarSpruit: Wouldn't it be better to use the component state here instead of the view state?

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -99,118 +110,50 @@ internal class CardComponent(
...
+        onIntent(CardIntent.UpdateCardNumber(newCardNumber))
+        val newBinValue = viewState.value.binValue
 
-    override fun onBrandSelected(cardBrand: CardBrand) {
-        stateManager.updateComponentState {
-            copy(
-                selectedCardBrand = cardBrand,
-            )
-        }
+        onBinValueChange(oldBin = oldBinValue, newBin = newBinValue)
```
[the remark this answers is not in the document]
OscarSpruit: This is a good suggestion actually! It also makes the `onIntent` flow more simple

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -180,22 +173,20 @@ private fun CardBrandsList(
...
+    state: TextInputViewState,
     detectedBrands: List<CardBrand>,
     modifier: Modifier = Modifier,
 ) {
-    val isInvalid = state.errorMessage != null && state.showError
-
-    AnimatedContent(targetState = isInvalid, modifier = modifier) { isInvalid ->
-        if (isInvalid) {
-            Icon(
+    val state = state.trailingIcon as? CardNumberTrailingIcon
```
OscarSpruit: The name `state` is a bit confusing as it is used before already. Maybe we can just call this `trailingIcon`?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -180,22 +173,20 @@ private fun CardBrandsList(
...
+    AnimatedContent(targetState = state, modifier = modifier) { state ->
+        when (state) {
```
OscarSpruit: Also here it's confusing what `state` is referring to. Maybe we can rename this one to `animatedState`?

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardComponentStateReducer.kt
```diff
@@ -0,0 +1,94 @@
...
+            ),
+            securityCode = state.securityCode.copy(
+                showError = hasSecurityCodeError,
+                isFocused = hasSecurityCodeError && !hasCardNumberError && !hasExpiryDateError,
+            ),
+            holderName = state.holderName.copy(
+                showError = hasHolderNameError,
+                isFocused = hasHolderNameError && !hasCardNumberError && !hasExpiryDateError && !hasSecurityCodeError,
+            ),
+        )
```
OscarSpruit, line 100: To make the isFocused check more robust we could add a mechanism that keeps track of if a field is already focused. Something like:

```kotlin
private fun highlightValidationErrors(state: CardComponentState): CardComponentState {
    var isFocusConsumed = false

    fun shouldFocus(hasError: Boolean): Boolean {
        return if (hasError && !isFocusConsumed) {
            isFocusConsumed = true
            true
        } else {
            false
        }
    }

    val hasCardNumberError = state.cardNumber.errorMessage != null
    // Add other error checks

    return state.copy(
        cardNumber = state.cardNumber.copy(
             showError = hasCardNumberError,
             isFocused = shouldFocus(hasCardNumberError),
        ),
        // Do the same for other input states
    )
}
```

ozgur00: Good one!

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -229,26 +169,27 @@ internal class CardComponent(
             .launchIn(coroutineScope)
     }
 
+    private fun subscribeToBinChanges() {
+        componentState
+            .map { it.binValue }
+            .distinctUntilChanged()
```
araratthehero, line 175: I see in the docs of `distinctUntilChanged()` that `Note that any instance of StateFlow already behaves as if distinctUntilChanged operator is applied to it, so applying distinctUntilChanged to a StateFlow has no effect.`

Can we safely remove it?

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -229,26 +169,27 @@ internal class CardComponent(
...
+            .drop(1)
```
araratthehero, line 176: Why is `drop(1)` used here?

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardComponentStateExt.kt
```diff
@@ -9,7 +9,6 @@
 package com.adyen.checkout.card.internal.ui.state
```
araratthehero, line 9: The name of this file could be changes, since it does not reference to card view state anymore

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
ozgur00: What happens is, after map is invoked on `componentState` the resulting flow is not a state flow. Hence we need distinct until changed.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
ozgur00: It is used to drop the initial empty text value

PR #2471, opened 2025-12-19
v6 - Drop-in: favorite removal

OscarSpruit:
## Description
Implement removing of favorites. Note that it only removes the favorite from this screen. To achieve full removal within the drop-in we need to keep the state in a shared place. This will be done in a later stage.

https://github.com/user-attachments/assets/8f059aa1-6bd7-4459-be88-b93546b0e317

COSDK-812

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ConfirmationDialog.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ConfirmationDialog.kt
@@ -52,7 +52,7 @@ internal fun ConfirmationDialog(
             modifier = Modifier
                 .fillMaxSize()
                 .systemBarsPadding()
-                .padding(Dimensions.Large),
+                .padding(Dimensions.Spacing.Large),
         ) {
             DestructiveButton(
                 text = confirmationText,
@@ -60,7 +60,7 @@ internal fun ConfirmationDialog(
                 modifier = Modifier.fillMaxWidth(),
             )
 
-            Spacer(Modifier.size(Dimensions.Large))
+            Spacer(Modifier.size(Dimensions.Spacing.Large))
 
             SecondaryButton(
                 text = cancellationText,


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesScreen.kt
```diff
@@ -64,59 +68,97 @@ private fun ManageFavoritesContent(
...
+                onDismissRequest = { selectedItem = null },
             )
         }
     }
 }
 
-@Composable
-private fun Section(
-    title: String,
+private fun LazyListScope.section(
```
ozgur00, line 107: What's the benefit of using an extension function rather than a composable?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesScreen.kt
```diff
@@ -64,59 +68,97 @@ private fun ManageFavoritesContent(
...
+    ListItem(
+        leadingIcon = {
+            CheckoutNetworkLogo(
+                txVariant = item.icon,
+                modifier = Modifier.size(Dimensions.LogoSize.medium),
+            )
+        },
+        title = item.title,
+        subtitle = item.subtitle,
+        trailingIcon = {
```
ozgur00: Now that we're passing a text composable, `trailingIcon` name doesn't apply anymore. We could use `trailingContent` as name? Also for consistency we could change the `leadingIcon` as well.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesViewModel.kt
```diff
@@ -62,12 +66,18 @@ internal class ManageFavoritesViewModel(
...
 
         return FavoriteListItem(
+            id = id.orEmpty(),
             icon = icon,
             title = title,
             subtitle = subtitle,
         )
     }
 
+    fun removeFavorite(id: String) {
```
ozgur00, line 76: Should we add a TODO for triggering the api call?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesScreen.kt
OscarSpruit: A `LazyColumn` cannot have composables directly nested, instead you have to use `item {}` and put a composable in there. The `LazyListScope` extension makes sure we can add `item`s to the `LazyColumn`. The reason I changed it from `Column` to `LazyColumn` is to easily animate the removal of items

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesScreen.kt
OscarSpruit: `trailingContent` makes a lot of sense. Not sure if we should rename `leadingIcon` as IMO this should really only be used for icons and not for text or anything like that

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesViewModel.kt
OscarSpruit: I think it's not necessary. After the TODO on line 26 this will change a lot and it could probably be part of that PR

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/ManageFavoritesScreen.kt
ozgur00: Fair, let's leave `leadingIcon` as is to at least indicate that it's for icons only.

ozgur00 (approved):

PR #2472, opened 2025-12-22
Add module migration template for LLM usage

ozgur00:
## Description
Add the initial version of migration template file which will be loaded into LLM tool as context and guide while migrating a payment component module. This is an iteration based on the migration of `blik` module. It doesn't include UI migration and stored version of the component. Phases for stored parts of a module will be added, however the plan for those parts will be based on stored cards changes, therefore it's still pending.

### DIFF
--- a/MODULE_MIGRATION_TEMPLATE.md
+++ b/MODULE_MIGRATION_TEMPLATE.md
@@ -46,9 +46,9 @@ This template outlines the implementation plan for migrating a payment method mo
   - Add {Module}PaymentMethod to `getChildSerializer` when block
 
 **Reference**: 
-- `core/.../MBWayPaymentMethod.kt` (new pattern)
-- `core/.../CardPaymentMethod.kt` (new pattern)
-- `components-core/.../{Module}PaymentMethod.kt` (old implementation for field reference)
+- `core/src/main/java/com/adyen/checkout/core/components/paymentmethod/MBWayPaymentMethod.kt` (new pattern)
+- `core/src/main/java/com/adyen/checkout/core/components/paymentmethod/CardPaymentMethod.kt` (new pattern)
+- `components-core/src/main/java/com/adyen/checkout/components/core/paymentmethod/{Module}PaymentMethod.kt` (old implementation for field reference)
 
 **Verification**:
 ```bash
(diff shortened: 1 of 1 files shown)


MODULE_MIGRATION_TEMPLATE.md
```diff
@@ -0,0 +1,372 @@
+# Module Migration Template
+
+This template outlines the implementation plan for migrating a payment method module to the new component architecture, following the patterns established in the MBWay module.
+
+> **Prerequisites**: This plan assumes scaffold files already exist in the module (empty class files with TODO comments).
```
araratthehero, line 5: What do you think if we create a new section, where we mention what needs to improve in this doc in the future, such as:
- Scaffold files should also be automatically created.
- Every time we convert a new component, it should be added to the references list, so every time it gets better at migrating new components.

Also what do you think of adding Card in the list of references?

MODULE_MIGRATION_TEMPLATE.md
```diff
@@ -0,0 +1,372 @@
...
+  - Properties: `type`, `checkoutAttemptId`, plus payment-method-specific fields
+  - Include `SERIALIZER` companion object
+  - `PAYMENT_METHOD_TYPE` constant
+
+**Files to modify**:
+- [ ] `core/src/main/java/com/adyen/checkout/core/components/paymentmethod/PaymentMethodDetails.kt`
+  - Add {Module}PaymentMethod to `getChildSerializer` when block
+
+**Reference**: 
+- `core/.../MBWayPaymentMethod.kt` (new pattern)
```
atmamont: LLM would spend less tokens if path is full. Otherwise it's enough to specify just a full reference filename, so it can use the `find` tool

MODULE_MIGRATION_TEMPLATE.md
```diff
@@ -0,0 +1,372 @@
...
+- `components-core/.../{Module}PaymentMethod.kt` (old implementation for field reference)
+
+**Verification**:
+```bash
+./gradlew :core:compileDebugKotlin
+./gradlew :core:check
+./gradlew apiDump
+```
+
+**Commit**: "Add {Module}PaymentMethod to core module. COSDK-XXX"
```
atmamont, line 60: Do you guys let LLM commit code, any hiccups or regrets so far?

MODULE_MIGRATION_TEMPLATE.md
```diff
@@ -0,0 +1,372 @@
...
+**Goal**: Create the new configuration class following the new pattern
+
+**Files to create**:
+- [ ] `{module}/src/main/java/com/adyen/checkout/{module}/{Module}Configuration.kt`
+  - `{Module}Configuration` data class implementing `Configuration`
+  - `{Module}ConfigurationBuilder` class (internal constructor)
+  - `CheckoutConfiguration.{module}()` extension function
+  - `CheckoutConfiguration.get{Module}Configuration()` internal function
+
+**Reference**: `card/src/main/java/com/adyen/checkout/card/CardConfiguration.kt`
```
atmamont: 👍

atmamont:
@ozgur00 force push makes it impossible to review changes

MODULE_MIGRATION_TEMPLATE.md
ozgur00: We do. We have a template file for commit messages and we make sure it follows. We have to keep an eye, but it's been useful so far.

PR #2475, opened 2025-12-29
v5 - Fix publish docs action

OscarSpruit:
## Description
The publish docs action has not been working correctly for some time. It was due to having the wrong permissions. I updated the permission according to the [docs](https://github.com/JamesIves/github-pages-deploy-action). So it should work now.

### DIFF
--- a/.github/workflows/publish_docs.yml
+++ b/.github/workflows/publish_docs.yml
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  deployments: write
+  contents: write
 
 jobs:
 


.github/workflows/publish_docs.yml
```diff
@@ -5,7 +5,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  deployments: write
+  contents: write
```
jreij, line 8: Did we receive any notifications with the job failure? If not is there a way to ensure that for the future?

OscarSpruit: An IM pointed out the docs seemed not up-to-date and then I noticed the [last 3 releases](https://github.com/Adyen/adyen-android/actions/workflows/finalize_release.yml) have the finalize release workflow failing. Let's discuss with the team how we missed it and how we could prevent that again 👍

PR #2486, opened 2026-01-07
Rename checkoutConfiguration to configuration to align the parameters with iOS

Robert-SD:


example-app/src/main/java/com/adyen/checkout/example/ui/main/MainViewModel.kt
```diff
@@ -195,7 +195,7 @@ internal class MainViewModel @Inject constructor(
             val result = paymentMethods?.let {
                 Checkout.setup(
                     paymentMethods,
```
OscarSpruit: Because a named argument is used now for the configuration could you do that here as well?
```suggestion
                    paymentMethodsApiResponse = paymentMethods,
```

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainViewModel.kt
```diff
@@ -226,7 +226,7 @@ internal class MainViewModel @Inject constructor(
             val result = sessionModel?.let {
                 Checkout.setup(
                     sessionModel,
```
OscarSpruit: Because a named argument is used now for the configuration could you do that here as well?
```suggestion
                    sessionModel = sessionModel,
```

core/src/main/java/com/adyen/checkout/core/components/Checkout.kt
```diff
@@ -17,10 +17,10 @@ object Checkout {
 
     suspend fun setup(
         sessionModel: SessionModel,
-        checkoutConfiguration: CheckoutConfiguration,
+        configuration: CheckoutConfiguration,
     ): Result<CheckoutContext.Sessions> {
         val initializationData = CheckoutInitializer.initialize(
-            checkoutConfiguration = checkoutConfiguration,
+            checkoutConfiguration = configuration,
```
Robert-SD, line 23: Shall we also rename the parameter of the internal initialize method?

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainViewModel.kt
Robert-SD: Sure!

core/src/main/java/com/adyen/checkout/core/components/Checkout.kt
OscarSpruit: If it is a small change I would say it's okay, but if you have to change a whole chain of classes I don't think it's worth the effort

PR #2492, opened 2026-01-09
Create foundation for ADRs

OscarSpruit:
## Description
This PR sets up the base to start documenting our decision making and have it in a place where we can easily find it. The README explains how our ADR process works and the template can be used the quickly create a new ADR.

### DIFF
--- a/ADR/ADR_TEMPLATE.md
+++ b/ADR/ADR_TEMPLATE.md
@@ -36,11 +36,15 @@
 
 **Cons:**
 
--
+
+## Proposed solution
+
+[//]: # (Write down the initially proposed solution and explain why.)
 
 ## Final decision
 
-[//]: # (Write down what solution is chosen and why.)
+[//]: # (After receiving feedback on the proposed solution, write down the final decision and 
+explain why.)
 
 ## Concerns and follow-up actions
 


ADR/README.md
```diff
@@ -0,0 +1,33 @@
...
+Once agreed a final decision can be made. The decision can either be accepted or declined. When an
+ADR is accepted development can start.
+
+The status of an ADR must be kept up-to-date by changing the metadata table in the header of the
+document. Use the following statuses:
+
+- Created
+- In progress
+- Accepted
+- Declined
```
ozgur00, line 31: What does `Declined` mean? Does it mean the PR that creates the ADR will be rejected?

ADR/README.md
```diff
@@ -0,0 +1,33 @@
...
+made.
+
+### 3. Final decision
+
+Once agreed a final decision can be made. The decision can either be accepted or declined. When an
+ADR is accepted development can start.
+
+The status of an ADR must be kept up-to-date by changing the metadata table in the header of the
+document. Use the following statuses:
+
```
ozgur00, line 27: We could add a section `Advice & Concerns` like web for future reference as we discussed yesterday.

ADR/README.md
OscarSpruit: An ADR can also be a proposal to use a new tool or fix some tech dept for example. In the end it might be that we decide that it's not worth it for example, so then the status would be "declined". In the future we can then look back and easily see we already had this idea, but we decided not to move forward with it. Does that make sense?

atmamont:
Does it make sense to scale across all SDKs or is this purely meant to be used for Android?

ADR/README.md
OscarSpruit: Good one! I added a section to the template called `Concerns and follow-up actions`

ozgur00 (dismissed):

PR #2495, opened 2026-01-12
v6 - Fetch checkoutAttemptId in initialize

araratthehero:
## Description
Fetch `checkoutAttemptId` on Checkout initialize.

## Checklist 
- [x] Code is unit tested

COSDK-443

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutInitializer.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutInitializer.kt
@@ -21,6 +21,8 @@ import com.adyen.checkout.core.sessions.CheckoutSession
 import com.adyen.checkout.core.sessions.CheckoutSessionResult
 import com.adyen.checkout.core.sessions.SessionModel
 import com.adyen.checkout.core.sessions.internal.CheckoutSessionProvider
+import kotlinx.coroutines.async
+import kotlinx.coroutines.supervisorScope
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 object CheckoutInitializer {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/adyen/checkout/core/analytics/internal/data/remote/api/AnalyticsService.kt
```diff
@@ -25,7 +25,20 @@ class AnalyticsService(
...
+        clientKey: String,
+    ): AnalyticsSetupResponse = withContext(coroutineDispatcher) {
+        httpClient.post(
+            path = "v3/analytics",
+            queryParameters = mapOf("clientKey" to clientKey),
+            body = request,
+            requestSerializer = AnalyticsSetupRequest.SERIALIZER,
+            responseSerializer = AnalyticsSetupResponse.SERIALIZER,
+        )
+    }
```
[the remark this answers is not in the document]
araratthehero: We want to keep them separate on purpose. We might need to change the implementation in the future.

core/src/main/java/com/adyen/checkout/core/analytics/internal/data/remote/api/AnalyticsService.kt
OscarSpruit: This class can actually be internal

core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutInitializer.kt
```diff
@@ -27,12 +29,16 @@ object CheckoutInitializer {
         checkoutConfiguration: CheckoutConfiguration,
         sessionModel: SessionModel?,
     ): InitializationData {
-        // TODO - Fetch checkoutAttemptId
         val checkoutSession = sessionModel?.let { getCheckoutSession(sessionModel, checkoutConfiguration) }
         val publicKey = fetchPublicKey(checkoutConfiguration)
+        val checkoutAttemptId = fetchCheckoutAttemptId(
```
OscarSpruit: Currently, these network request will be executed sequentially, which could potentially make it quite slow. By using a `supervisorScope` and `async` we can execute these network calls independently and async, which makes it take as long as the slowest network call. Something like:
```Kotlin
suspend fun initialize(
    checkoutConfiguration: CheckoutConfiguration,
    sessionModel: SessionModel?,
): InitializationData {
    return supervisorScope {
        val checkoutSession = async { sessionModel?.let { getCheckoutSession(sessionModel, checkoutConfiguration) } }
        val publicKey = async { fetchPublicKey(checkoutConfiguration) }
        val checkoutAttemptId = async { fetchCheckoutAttemptId(checkoutConfiguration) }
        InitializationData(
            checkoutSession = checkoutSession.await(),
            publicKey = publicKey.await,
            checkoutAttemptId = checkoutAttemptId.await(),
        )
    }
}
```

araratthehero: Great tip. Will implement it! 👍

OscarSpruit (approved):

PR #2501, opened 2026-01-15
v6 - Analytics - Pass payment method name to AnalyticsManager

araratthehero:
## Description
Pass payment method name to the analyticsManager

COSDK-443

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/data/model/PaymentMethodResponse.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/data/model/PaymentMethodResponse.kt
@@ -15,5 +15,5 @@ import com.adyen.checkout.core.common.internal.model.ModelObject
  */
 abstract class PaymentMethodResponse : ModelObject() {
 
-    abstract val type: String?
+    abstract val type: String
 }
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/adyen/checkout/core/components/data/model/PaymentMethodResponse.kt
```diff
@@ -8,7 +8,12 @@
...
+import com.adyen.checkout.core.common.internal.model.ModelObject
+
 /**
- * Marker interface for [PaymentMethod] and [StoredPaymentMethod].
+ * Parent class for [PaymentMethod] and [StoredPaymentMethod].
  */
-interface PaymentMethodResponse
+abstract class PaymentMethodResponse : ModelObject() {
+
+    abstract val type: String?
```
ozgur00: Should we make this required already? iOS for reference: https://github.com/Adyen/adyen-ios/blob/69729e8ca0b4fae7980d5f38e55f627ff9742eff/Adyen/Core/Core%20Protocols/PaymentMethod.swift#L13

It would address the comment by gemini too.

araratthehero: I made the `type` field non nullable. We can do the same for other fields in future PRs.

PR #2502, opened 2026-01-15
v6 - Rename isSubmitButtonVisible to showSubmitButton

ozgur00:
## Description
Rename `isSubmitButtonVisible` parameter in `CheckoutConfiguration` to `showSubmitButton` for better alignment.

## Checklist 
- [x] Code is unit tested

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/ui/model/CommonComponentParamsMapper.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/ui/model/CommonComponentParamsMapper.kt
@@ -36,7 +36,7 @@ class CommonComponentParamsMapper {
             amount = sessionParams?.amount
                 ?: dropInOverrideParams?.amount
                 ?: checkoutConfiguration.amount,
-            showSubmitButton = dropInOverrideParams?.isSubmitButtonVisible
+            showSubmitButton = dropInOverrideParams?.showSubmitButton
                 ?: checkoutConfiguration.showSubmitButton ?: true,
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/ui/model/CommonComponentParamsMapper.kt
```diff
@@ -36,8 +36,8 @@ class CommonComponentParamsMapper {
             amount = sessionParams?.amount
                 ?: dropInOverrideParams?.amount
                 ?: checkoutConfiguration.amount,
-            isSubmitButtonVisible = dropInOverrideParams?.isSubmitButtonVisible
-                ?: checkoutConfiguration.isSubmitButtonVisible ?: true,
+            showSubmitButton = dropInOverrideParams?.isSubmitButtonVisible
```
araratthehero: Shouldn't we change the `isSubmitButtonVisible` to `showSubmitButton` in `DropInOverrideParams`?

ozgur00: I've changed it, but we might need to remove DropInParams as we have all the common information that needs to be overridden already in the common part where we get from `CheckoutConfiguration`. We can figure this out while working on drop-in.

araratthehero: Could you leave a comment to check later and remove if necessary?

PR #2508, opened 2026-01-22
Provide full branch history to sonar

OscarSpruit:
## Description

This will make sure the git blame reports get generated correctly and Sonar will be able to identify new changes in a PR correctly.


OscarSpruit:
Sonar still seems to give the same warning, but I think this needs to run on the main branch as well for it to resolve

PR #2512, opened 2026-01-23
Fix key parameter propagation in SessionPaymentComponentProvider.get

ivanz5:
## Description
Some overloads of `SessionPaymentComponentProvider.get` were accepting the `key` parameter, but it was silently ignored and not passed to delegated `get` calls with other overloads.

## Checklist 


OscarSpruit:
Hi @ivanz5, thank you very much for finding and fixing this issue! We highly appreciate it!

PR #2530, opened 2026-02-02
v6 - Errors - CheckoutConfiguration - Locale validation

araratthehero:
## Description
Add locale validation to `CheckoutConfiguration` using the new error handling pattern.

### Changes
- Add `CheckoutException` class that wraps `CheckoutError` for throwable configuration errors
- Add `LocaleValidationResult` sealed class for validation results
- Update `LocaleUtil.validateLocale()` to return `LocaleValidationResult` with detailed error info
- Update `CheckoutConfiguration.validateContents()` to throw `CheckoutException` for invalid locales

## Checklist
- [x] Code is unit tested

## Ticket
COSDK-976

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/common/internal/helper/LocaleUtil.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/internal/helper/LocaleUtil.kt
@@ -9,6 +9,7 @@
 package com.adyen.checkout.core.common.internal.helper
 
 import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.error.CheckoutError
 import java.util.IllformedLocaleException
 import java.util.Locale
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/adyen/checkout/core/common/internal/helper/LocaleUtil.kt
```diff
@@ -19,21 +20,30 @@ import java.util.Locale
...
-    fun isValidLocale(locale: Locale): Boolean {
-        @Suppress("SwallowedException")
-        return try {
-            Locale.Builder().setLocale(locale).build()
-            true
-        } catch (ex: IllformedLocaleException) {
-            false
-        }
-    }
+    fun validateLocale(locale: Locale): LocaleValidationResult = runCatching {
```
OscarSpruit, line 33: Would it make sense to keep this method more generic by returning a boolean and then where this function is called we create the error if necessary?

araratthehero: I went with this approach, since wanted to propagate the `IllformedLocaleException` in the `CheckoutError`. Otherwise I had to return an exception as a result in the `isValidLocale` function.

I think it could be useful for merchants to get the exception propagated in `CheckoutException` and it would log it in the crash log too.

PR #2533, opened 2026-02-03
v6 - Errors - CheckoutConfiguration - Client key validation

araratthehero:
## Description
Add client key validation to `CheckoutConfiguration` following the same pattern as locale validation.

- Create `ClientKeyUtil` with validation logic (regex pattern, length constraints, aligned with iOS)
- Add `ClientKeyValidationResult` sealed class for validation results
- Update `CheckoutConfiguration` to validate `clientKey` in `validateContents()`

## Checklist
- [x] Code is unit tested

COSDK-976

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -691,30 +691,6 @@ public final class com/adyen/checkout/core/common/internal/data/model/PublicKeyR
 	public synthetic fun newArray (I)[Ljava/lang/Object;
 }
 
-public final class com/adyen/checkout/core/common/internal/helper/ClientKeyValidationResult$Invalid : com/adyen/checkout/core/common/internal/helper/ClientKeyValidationResult {
-	public static final field $stable I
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/adyen/checkout/core/common/internal/helper/ClientKeyUtil.kt
```diff
@@ -0,0 +1,67 @@
...
+package com.adyen.checkout.core.common.internal.helper
+
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.error.CheckoutError
+
+/**
+ * Utility class for client key validation.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object ClientKeyUtil {
```
OscarSpruit: What do you think of naming this `ClientKeyValidator`? This would make it more specific and it will less likely end up being one of those classes that have a lot of unrelated functions

PR #2569, opened 2026-02-05
v6: GooglePay - Configuration

ozgur00:
## Description
In this PR `GooglePayConfiguration` and `GooglePayComponentParamsMapper` are implemented for further use. Displaying the button logic based on whether it's drop-in or standalone part is left out and will be added later. A TODO has been added. Unit tests for mapper will be added in a follow up PR.

## Ticket Number
COSDK-936

### DIFF
--- a/googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/model/GooglePayComponentParamsMapper.kt
+++ b/googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/model/GooglePayComponentParamsMapper.kt
@@ -15,6 +15,7 @@ import com.adyen.checkout.core.common.Environment
 import com.adyen.checkout.core.common.internal.helper.adyenLog
 import com.adyen.checkout.core.components.data.model.Amount
 import com.adyen.checkout.core.components.internal.ui.model.ComponentParamsBundle
+import com.adyen.checkout.core.error.internal.ComponentError
 import com.adyen.checkout.googlepay.AllowedAuthMethods
 import com.adyen.checkout.googlepay.AllowedCardNetworks
 import com.adyen.checkout.googlepay.GooglePayConfiguration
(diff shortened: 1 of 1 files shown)


googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/model/GooglePayComponentParamsMapper.kt
```diff
@@ -0,0 +1,129 @@
...
+        // TODO - Error propagation - Update the code
+        return this?.merchantAccount
+            ?: paymentMethod.configuration?.gatewayMerchantId
+            ?: throw CheckoutException(
+                CheckoutError(
+                    message = "GooglePay merchantAccount not found. Update your API version or pass it " +
+                        "manually inside your GooglePayConfiguration",
+                    code = CheckoutError.ErrorCode.UNKNOWN
+                )
+            )
```
araratthehero:
```suggestion
            ?: throw ComponentError(
                message = "GooglePay merchantAccount not found. Update your API version or pass it " +
                    "manually inside your GooglePayConfiguration",
            )
```

araratthehero: Since this is an internal error, you could use an `InternalCheckoutError` type instead, since it extend the `Exception` class and can be thrown. For now you can use `com.adyen.checkout.core.error.internal.ComponentError` for all component errors. At a later stage we will clean up the internal error types and adjust where necessary.

In this case the `UNKNOWN` type can be changed back to internal in `core/src/main/java/com/adyen/checkout/core/error/CheckoutError.kt`.

araratthehero: We have the `CheckoutErrorMapper`, which maps the `InternalCheckoutError` to a `CheckoutError`. This allows us to separate the public errors from the internal errors. We do not need to create a `CheckoutException` or a `CheckoutError` in the internal code, unless really necessary, like for the cases when something happens during the SDK setup.

ozgur00: Good catch! Fixed now.

PR #2611, opened 2026-02-25
v6 - Payment method models - Regular - Extra fields

araratthehero:
## Description
Made `id` and `name` fields in AppData and Issuer non-nullable.
Added 8 regular payment method response models with extra fields.
- BCMCPaymentMethod
- IssuerListPaymentMethod
- OnlineBankingPaymentMethod
- GiftCardPaymentMethod
- UPIPaymentMethod
- GooglePayPaymentMethod
- CashAppPayPaymentMethod

## Checklist

COSDK-997


core/src/main/java/com/adyen/checkout/core/components/data/model/paymentmethod/BCMCPaymentMethod.kt
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Payment method model for Bancontact (BCMC).
+ */
+@Parcelize
+data class BCMCPaymentMethod(
+    override val type: String,
+    override val name: String,
+    val brands: List<String>,
+    val fundingSource: String? = null,
```
OscarSpruit: All the nullable values seem to have a default value. Why is that? Wouldn't it be better if all values are always set intentionally?

PR #2612, opened 2026-02-25
v6 - String resource arguments

OscarSpruit:
## Description
This PR adds support to use formatting arguments while resolving strings. Additionally, an implementation is provided that allows merchants to easily override our string resources programmaticly.

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/common/localization/CheckoutLocalizationProvider.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/localization/CheckoutLocalizationProvider.kt
@@ -15,7 +15,7 @@ interface CheckoutLocalizationProvider {
     fun getLocalizedString(context: Context, locale: Locale, key: CheckoutLocalizationKey): String?
 }
 
-class XmlLocalizationProvider(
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/adyen/checkout/core/common/localization/CheckoutLocalizationProvider.kt
```diff
@@ -12,5 +12,20 @@ import android.content.Context
 import java.util.Locale
 
 interface CheckoutLocalizationProvider {
-    fun getString(context: Context, locale: Locale, key: CheckoutLocalizationKey): String?
+    fun getLocalizedString(context: Context, locale: Locale, key: CheckoutLocalizationKey): String?
+}
+
+class XmlLocalizationProvider(
```
araratthehero: We could name this `ResourceLocalizationProvider` or perhaps even better `StringResourceLocalizationProvider` to make it clear that the Int in the map is a string resource id.

OscarSpruit: That's a way better name!

PR #2613, opened 2026-02-25
v6 - Payment method models - Stored - No extra fields

araratthehero:
## Description
Added 2 stored payment method response models (base-only) with no extra fields.
- StoredBLIKPaymentMethod
- StoredTwintPaymentMethod

## Checklist

COSDK-997

### DIFF
--- a/core/api/core.api
+++ b/core/api/core.api
@@ -1697,6 +1697,31 @@ public final class com/adyen/checkout/core/components/data/model/paymentmethod/S
 	public final fun getChildSerializer (Ljava/lang/String;)Lcom/adyen/checkout/core/common/internal/model/ModelObject$Serializer;
 }
 
+public final class com/adyen/checkout/core/components/data/model/paymentmethod/StoredTwintPaymentMethod : com/adyen/checkout/core/components/data/model/paymentmethod/StoredPaymentMethod {
+	public static final field $stable I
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/adyen/checkout/core/components/data/model/paymentmethod/StoredBLIKPaymentMethod.kt
```diff
@@ -0,0 +1,64 @@
...
+    companion object {
+        @JvmField
+        val SERIALIZER: Serializer<StoredBLIKPaymentMethod> = object : Serializer<StoredBLIKPaymentMethod> {
+            override fun serialize(modelObject: StoredBLIKPaymentMethod): JSONObject {
+                return try {
+                    JSONObject().apply {
+                        put(TYPE, modelObject.type)
+                        put(NAME, modelObject.name)
+                        put(ID, modelObject.id)
+                        putOpt(
```
OscarSpruit: In both objects `putOpt` is used, but `supportedShopperInteractions` is not optional. Should it be optional?

araratthehero: Good point. It is not necessary.

PR #2615, opened 2026-02-26
Lower JVM memory setting

OscarSpruit:
## Description
The Renovate runners have very little memory. The previous setting was pushing the limits and caused Renovate to fail a lot.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,7 +17,7 @@ android.defaults.buildfeatures.shaders=false
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx3072m
+org.gradle.jvmargs=-Xmx2048m
 
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit


gradle.properties
```diff
@@ -17,7 +17,7 @@ android.defaults.buildfeatures.shaders=false
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx3072m
+org.gradle.jvmargs=-Xmx2048m
```
araratthehero, line 20: This changes the default for all developers, not just Renovate. Will it be enough for some heavy tasks like check, dokka, etc.?

Is it possible to configure a different limit only for Renovate in `renovate.json` or creating a file similar to `.github/ci-gradle.properties` for Renovate and keeping the default one 3GB?

OscarSpruit: Unfortunately, there is no other way to configure Renovate. In the future they might add this ([discussion](https://github.com/renovatebot/renovate/discussions/41492)). For now, this seems to be the best solution and locally you can change this setting in `~/.gradle/gradle.properties`. Maybe it's even better to have this configured locally as every machine possibly has different specs

araratthehero: Does [this](https://github.com/renovatebot/renovate/issues/39559#issuecomment-3968360179) mean that they already implemented it?

OscarSpruit: Oh what a luck! Not sure when that version will be available to us, but maybe it's better to wait for it

OscarSpruit:
Renovate released a new version where we can directly configure the memory settings. There is no need for this anymore.

PR #2627, opened 2026-03-10
v6 - Supported stored payment methods

OscarSpruit:
## Description
This PR makes sure only supported stored payment methods are displayed in drop-in by:
- Exposing a list of supported stored payment methods publicly
- Extract payment method support logic into a single place
- Updating support logic for stored payment methods


drop-in/src/main/java/com/adyen/checkout/dropin/internal/helper/PaymentMethodSupportCheck.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun isSupported(paymentMethod: PaymentMethod): Boolean {
+        // Only checks if the payment method type is explicitly not supported. This allows unknown redirect payment
+        // methods to be supported.
+        return !PaymentMethodTypes.UNSUPPORTED_PAYMENT_METHODS.contains(paymentMethod.type)
+    }
+
+    fun isSupported(paymentMethod: StoredPaymentMethod): Boolean {
+        return paymentMethod.id.isNotEmpty() &&
+            PaymentMethodTypes.SUPPORTED_STORED_PAYMENT_METHODS.contains(paymentMethod.type) &&
```
araratthehero, line 25: What is the decision behind the choice of creating an explicitly supported list for stored payment methods?

OscarSpruit: Web also has an explicit list and since we no longer have individual components a merchant could use to check with I figured this would be most straightforward. What do you think?

PR #2638, opened 2026-03-17
v6 - DropInService memory leak

OscarSpruit:
## Description
This PR fixes a memory caused by DropInService by simplifying it's lifecycle management. Using a combination of a started and bound service is overly complex for what we are trying to achieve and there is no need since we are not doing cross process communication. Now DropInService is a started service that we can access through a static registry and we are in full control of it's lifecycle.

### DIFF
--- a/drop-in/src/test/java/com/adyen/checkout/dropin/internal/service/DropInServiceRegistryTest.kt
+++ b/drop-in/src/test/java/com/adyen/checkout/dropin/internal/service/DropInServiceRegistryTest.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 20/3/2026.
+ */
+
+package com.adyen.checkout.dropin.internal.service
+
+import com.adyen.checkout.dropin.DropInService
+import kotlinx.coroutines.async
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.assertSame
+import org.junit.jupiter.api.Test
+import org.mockito.kotlin.mock
+
+internal class DropInServiceRegistryTest {
+
+    @AfterEach
+    fun tearDown() {
+        DropInServiceRegistry.unregister()
+    }
+
+    @Test
+    fun `when service is registered then awaitService returns it immediately`() = runTest {
+        val service = mock<DropInService>()
+
+        DropInServiceRegistry.register(service)
+
+        val result = DropInServiceRegistry.awaitService()
+        assertSame(service, result)
+    }
+
+    @Test
+    fun `when service is registered after awaitService is called then awaitService resumes`() = runTest {
+        val service = mock<DropInService>()
+        val deferred = async { DropInServiceRegistry.awaitService() }
+
+        DropInServiceRegistry.register(service)
+
+        assertSame(service, deferred.await())
+    }
+
+    @Test
+    fun `when service is re-registered after unregister then awaitService returns new service`() = runTest {
+        val firstService = mock<DropInService>()
+        DropInServiceRegistry.register(firstService)
+        DropInServiceRegistry.unregister()
+
+        val secondService = mock<DropInService>()
+        val deferred = async { DropInServiceRegistry.awaitService() }
+
+        DropInServiceRegistry.register(secondService)
+
+        assertSame(secondService, deferred.await())
+    }
+}


drop-in/src/main/java/com/adyen/checkout/dropin/DropInService.kt
```diff
@@ -9,36 +9,34 @@
...
 import androidx.lifecycle.LifecycleService
 import com.adyen.checkout.core.action.data.ActionComponentData
+import com.adyen.checkout.core.common.AdyenLogLevel
+import com.adyen.checkout.core.common.internal.helper.adyenLog
 import com.adyen.checkout.core.components.CheckoutResult
 import com.adyen.checkout.core.components.paymentmethod.PaymentComponentState
-import com.adyen.checkout.dropin.internal.service.DropInBinder
+import com.adyen.checkout.dropin.internal.service.DropInServiceRegistry
 
 abstract class DropInService : LifecycleService() {
```
araratthehero, line 20: Won't this service get destroyed by the system if it is not bound to an activity and the app is in the background?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/service/DropInServiceManager.kt
```diff
@@ -29,55 +24,28 @@ internal class DropInServiceManager(
...
     fun stop(context: Context) {
         val intent = Intent(context, serviceClass)
         context.stopService(intent)
     }
 
-    // TODO - Binder could be null when service is not connected yet.
-    //  Implement a queue similar in functionality to one in v5 DropInActivity.
     suspend fun requestOnSubmit(state: PaymentComponentState<*>): CheckoutResult {
-        return binder?.requestOnSubmit(state) ?: error("Service is not bound")
+        return DropInServiceRegistry.get()?.onSubmit(state) ?: error("Service is not available")
```
araratthehero: What do you think of throwing a `CheckoutException` here?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/service/DropInServiceRegistry.kt
```diff
@@ -0,0 +1,25 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 16/3/2026.
+ */
+
+package com.adyen.checkout.dropin.internal.service
+
+import com.adyen.checkout.dropin.DropInService
+
+internal object DropInServiceRegistry {
```
araratthehero, line 16: Could you add a TODO to not forget to write tests for these classes?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/service/DropInServiceRegistry.kt
```diff
@@ -0,0 +1,25 @@
...
+    private var service: DropInService? = null
```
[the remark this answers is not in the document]
araratthehero: I agree with this suggestion. Even though the operations are all performed in the main thread, it is always good to keep the access to this field safe. Also all operations with the `service` are single field reads or writes, so no iteration or collection mutation is happening with the `service` field. So no further safeguards are necessary like `Mutex`. The `@Volatile` should be enough to protect us from the visibility issue that can happen in case the `service` is accessed from a different thread.

drop-in/src/main/java/com/adyen/checkout/dropin/DropInService.kt
OscarSpruit: From the documentation it doesn't seem so

> A started service is one that another component starts by calling [startService()](https://developer.android.com/reference/android/content/Context#startService(android.content.Intent)), which results in a call to the service's [onStartCommand()](https://developer.android.com/reference/android/app/Service#onStartCommand(android.content.Intent,%20int,%20int)) method.
> 
> When a service is started, it has a lifecycle that's independent of the component that started it. The service can run in the background indefinitely, even if the component that started it is destroyed. As such, the service should stop itself when its job is complete by calling [stopSelf()](https://developer.android.com/reference/android/app/Service#stopSelf()), or another component can stop it by calling [stopService()](https://developer.android.com/reference/android/content/Context#stopService(android.content.Intent)).

drop-in/src/main/java/com/adyen/checkout/dropin/internal/service/DropInServiceManager.kt
OscarSpruit: That would already be better, but I didn't change it because I was hesitating. Should we call `onError` or should we throw or should we queue and try again later? Maybe I should put a TODO to think about it later. WDYT?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/service/DropInServiceRegistry.kt
OscarSpruit: I actually added tests for this class

drop-in/src/main/java/com/adyen/checkout/dropin/internal/service/DropInServiceManager.kt
araratthehero: Agreed to add a TODO.

OscarSpruit: Since we are waiting for a service to be available now, there is no longer the possibility for the service to be null at this point. No need for a TODO

PR #2639, opened 2026-03-17
v6 - Improve input transformation for digits only text fields

jreij:
## Description
Introduces common functionality that can be used by all text fields that only support entering digits and formatting them with separators.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-921

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/SocialSecurityNumberField.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/SocialSecurityNumberField.kt
@@ -9,12 +9,10 @@
 package com.adyen.checkout.card.internal.ui.view
 
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.focus.onFocusChanged
-import androidx.compose.ui.text.input.KeyboardType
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/adyen/checkout/core/common/internal/properties/CardNumberProperties.kt
```diff
@@ -0,0 +1,25 @@
...
+ * - Regular, formatted as such 1234 1234 1234 1234
+ * - Amex, formatted as such 1234 123456 12345 1234
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object CardNumberProperties {
+
+    const val CARD_NUMBER_MINIMUM_LENGTH = 12
+    const val CARD_NUMBER_MAXIMUM_LENGTH = 19
+
+    const val CARD_NUMBER_SEPARATOR = ' '
```
OscarSpruit, line 24: Should we allow `-` as a separator as well? I can image some shoppers will try to copy paste a number from somewhere where it's formatted with dashes instead of spaces

jreij: Hmmm I've never seen a card number formatted with dashes, can you share an example if you have one? Otherwise I'm more inclined to allow spaces only

OscarSpruit: I feel like I have seen it before, but can't find an example. There also don't seem to be any standards on this. I guess just the space is fine. We can always improve later

PR #2641, opened 2026-03-18
v6 - Drop-in - persist back stack

OscarSpruit:
## Description
This PR ensures drop-in will be able to survive process-death by storing the back stack in the SavedStateHandle.

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInNavigator.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInNavigator.kt
@@ -26,9 +26,12 @@ internal class DropInNavigator(
     private val _finishFlow = MutableStateFlow(false)
     val finishFlow = _finishFlow.asStateFlow()
 
+    val didRestoreState: Boolean
+
     init {
         val restored = backStackPersister.restore()
-        if (restored != null) {
+        didRestoreState = restored != null
+        if (didRestoreState) {
             _backStack.clear()
             _backStack.addAll(restored)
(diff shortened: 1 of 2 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/helper/SavedStateBackStackPersister.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.os.Bundle
+import androidx.compose.runtime.snapshots.SnapshotStateList
+import androidx.lifecycle.SavedStateHandle
+import androidx.navigation3.runtime.NavKey
+import androidx.navigation3.runtime.serialization.NavKeySerializer
+import androidx.savedstate.compose.serialization.serializers.SnapshotStateListSerializer
+import androidx.savedstate.serialization.decodeFromSavedState
+import androidx.savedstate.serialization.encodeToSavedState
+
+internal class SavedStateBackStackPersister(
```
araratthehero, line 20: I would add tests for this class, to make sure the store and restore function as expected

OscarSpruit: Because it's using Android's `Bundle` it does not work in regular unit tests. Do you think it's worth it to create a Robolectric test for it?

araratthehero: I see. It is a simple class, no need.

PR #2643, opened 2026-03-20
v6 - Payment method list - brand icons

OscarSpruit:
## Description
This PR will make the payment method list display brand icons for the card and pay by bank US payment methods. Also, every item now displays a chevron

image

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
@@ -53,6 +53,8 @@ import com.adyen.checkout.ui.internal.theme.CheckoutThemeProvider
 import com.adyen.checkout.ui.internal.theme.Dimensions
 import java.util.Locale
 
+private const val AMOUNT_VISIBLE_BRANDS = 3
+
 @Composable
 internal fun PaymentMethodListScreen(
     navigator: DropInNavigator,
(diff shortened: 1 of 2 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
```diff
@@ -203,13 +209,47 @@ private fun PaymentMethodItemList(
...
+                    PaymentMethodItemTrailingContent(item)
+                },
                 onClick = { onItemClick(item) },
                 modifier = Modifier.padding(horizontal = Dimensions.Spacing.ExtraSmall),
             )
         }
     }
 }
 
+private const val AMOUNT_VISIBLE_BRANDS = 3
```
jreij: Is it intended that this constant is in the middle of the file?

OscarSpruit: Put it there to keep it close to where it's used. I can put it on top as well to align with standard conventions. What do you prefer?

jreij: Up to you!

OscarSpruit: I moved it up 👍

jreij (approved):

PR #2654, opened 2026-03-26
v6 - Improve expiry date input/output transformations and validation

jreij:
## Description
- Improve expiry date input transformation.
- Improve expiry date output transformation.
- Simplify expiry date validation.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-921

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
@@ -15,10 +15,10 @@ import java.util.Date
 import java.util.Locale
 
 internal object ExpiryDateParser {
-    private const val MONTH_YEAR_FORMAT = "MMyy"
-    private const val MONTH_FORMAT = "MM"
-    private const val SHORT_YEAR_FORMAT = "yy"
-    private const val FULL_YEAR_FORMAT = "yyyy"
+    private val PARSING_FORMATTER = getFormatter("MMyy")
+    private val MONTH_FORMATTER = getFormatter("MM")
+    private val SHORT_YEAR_FORMATTER = getFormatter("yy")
+    private val FULL_YEAR_FORMATTER = getFormatter("yyyy")
 
     /**
      * Parses digit only input into a pair of expiryMonth and expiryYear
@@ -41,19 +41,15 @@ internal object ExpiryDateParser {
 
     private fun parseToDate(expiryDate: String): Date? {
         return try {
-            val parsingFormatter = getFormatter(MONTH_YEAR_FORMAT)
-            parsingFormatter.parse(expiryDate)
+            PARSING_FORMATTER.parse(expiryDate)
         } catch (_: ParseException) {
             null
         }
     }
 
     private fun getMonthAndYear(date: Date, returnFullYear: Boolean): Pair<String, String> {
-        val monthFormatter = getFormatter(MONTH_FORMAT)
-        val yearFormat = if (returnFullYear) FULL_YEAR_FORMAT else SHORT_YEAR_FORMAT
-        val yearFormatter = getFormatter(yearFormat)
-
-        return monthFormatter.format(date) to yearFormatter.format(date)
+        val yearFormatter = if (returnFullYear) FULL_YEAR_FORMATTER else SHORT_YEAR_FORMATTER
+        return MONTH_FORMATTER.format(date) to yearFormatter.format(date)
     }
 
     private fun getFormatter(format: String): SimpleDateFormat {
--- a/core/src/main/java/com/adyen/checkout/core/common/helper/CardExpiryDateValidator.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/helper/CardExpiryDateValidator.kt
@@ -59,14 +59,7 @@ object CardExpiryDateValidator {
         }
 
         val date = try {
-            val parsingFormatter = SimpleDateFormat(MONTH_YEAR_FORMAT, Locale.ROOT).apply {
-                isLenient = false
-                // if the expiryYear is more than 20 years in the future, SimpleDateFormat will use the previous century
-                // which means "52" gets parsed to 1952 instead of 2052
-                // this call ensures that every parsed date is after the year 2000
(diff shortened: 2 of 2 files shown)


card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Parses digit only input into a pair of expiryMonth and expiryYear
+     * The input must be 4 digits long, 2-digit month followed by 2-digit year (MMYY)
+     * Automatically adds the year prefix if returnFullYear is true (26 -> 2026)
+     * Returns null if input is invalid
+     */
+    fun parseToMonthAndYear(expiryDate: String, returnFullYear: Boolean): Pair<String, String>? {
+        if (expiryDate.length != EXPIRY_DATE_MAX_LENGTH_NO_SEPARATORS) {
+            return null
+        }
```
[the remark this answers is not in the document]
jreij: Since this function is internal, we assume the string has already been validated before being passed here.

card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
```diff
@@ -0,0 +1,32 @@
...
+        val expiryMonth = expiryDate.take(2)
+        val shortYear = expiryDate.takeLast(2)
+        val expiryYear = if (returnFullYear) "$YEAR_PREFIX$shortYear" else shortYear
```
OscarSpruit: This will work for a long time, but wouldn't it be better to make this dynamic? Afaik `SimpleDateFormat` automatically handles century rollovers. You could do something like:

```kotlin
val expiryYear = if (returnFullYear) {
    try {
        val date = SimpleDateFormat("yy", Locale.ROOT).parse(shortYear)
        date?.let { SimpleDateFormat("yyyy", Locale.ROOT).format(it) } ?: return null
    } catch (e: ParseException) {
        return null
    }
} else {
    shortYear
}
```

core/src/main/java/com/adyen/checkout/core/common/helper/CardExpiryDateValidator.kt
```diff
@@ -9,89 +9,104 @@
...
+        month: Int,
+        year: Int,
+    ): Calendar {
+        return GregorianCalendar().apply {
+            // Clear all fields to avoid unexpected results from current time
+            clear()
+            // Calendar months are 0-based (January is 0), so we subtract 1
+            set(Calendar.MONTH, month - 1)
+            // add 2000 to the 2-digit year
+            set(Calendar.YEAR, YEAR_2000 + year)
```
OscarSpruit: Same here. The `ExpiryDate` class has some logic to create the expiry year dynamically. Maybe we could use that for inspiration?

card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
OscarSpruit: Maybe we can even replace this whole piece with `SimpleDateFormat("MM/yy", Locale.ROOT)` and use that to retrieve the month and year

jreij: Good point! I switched to `SimpleDateFormat` and while it has two limitations that requires us to still hardcode some stuff, it's more robust now. I added comments explaining the limitations, check out the last commit and let me know your thoughts.

core/src/main/java/com/adyen/checkout/core/common/helper/CardExpiryDateValidator.kt
jreij: Fixed, check my previous reply in the other review comment

core/src/main/java/com/adyen/checkout/core/common/helper/CardExpiryDateValidator.kt
```diff
@@ -49,28 +52,31 @@ object CardExpiryDateValidator {
...
-    ): Calendar {
-        return GregorianCalendar().apply {
-            // Clear all fields to avoid unexpected results from current time
-            clear()
-            // Calendar months are 0-based (January is 0), so we subtract 1
-            set(Calendar.MONTH, month - 1)
-            // add 2000 to the 2-digit year
-            set(Calendar.YEAR, YEAR_2000 + year)
+        val date = try {
+            val parsingFormatter = SimpleDateFormat(MONTH_YEAR_FORMAT, Locale.ROOT).apply {
```
OscarSpruit: Maybe we can move this `SimpleDateFormat` to a class level field, so it doesn't have to created all the time

card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
```diff
@@ -9,24 +9,56 @@
...
+        val monthFormatter = getFormatter(MONTH_FORMAT)
+        val yearFormat = if (returnFullYear) FULL_YEAR_FORMAT else SHORT_YEAR_FORMAT
+        val yearFormatter = getFormatter(yearFormat)
+
+        return monthFormatter.format(date) to yearFormatter.format(date)
     }
 
-    private const val YEAR_PREFIX = "20"
+    private fun getFormatter(format: String): SimpleDateFormat {
+        return SimpleDateFormat(format, Locale.ROOT).apply {
```
OscarSpruit, line 56: Same here actually

core/src/main/java/com/adyen/checkout/core/common/helper/CardExpiryDateValidator.kt
jreij: Good one, fixed

card/src/main/java/com/adyen/checkout/card/internal/helper/ExpiryDateParser.kt
jreij: Fixed, but note that this has 4 different formats so I have to initialize 4 different formatters

PR #2658, opened 2026-03-30
v5 - Co-badged cards - Dual branded disclaimer

araratthehero:
## Description
- Disclaimer has been added below the card number field which appears only when the card is dual branded and selectable.
- Disclaimer hides when there is an error in the card number field
- BrandVIew state has been added and the logic moved to CardView to allow controlling the CardView behavior

## Checklist 

## Ticket Number
COSDK-1062

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
@@ -339,6 +339,8 @@ class CardView @JvmOverloads constructor(
             else -> BrandState.Placeholder
         }
         binding.brandView.isVisible = !isCardScanningVisible
+        binding.textViewDualBrandedDisclaimer.isVisible = brandState is BrandState.DualBrand &&
+            brandState.dualBrandData.selectable
         cardScanningFragment?.setScanButtonVisibility(isCardScanningVisible)
         binding.brandView.update(brandState)
     }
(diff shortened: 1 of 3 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -388,9 +396,12 @@ class CardView @JvmOverloads constructor(
         if (stringResId == null) {
             binding.textInputLayoutCardNumber.hideError()
             binding.cardBrandLogoContainer.isVisible = true
+            val dualBrandData = cardDelegate.outputData.dualBrandData
+            binding.textViewDualBrandedDisclaimer.isVisible = dualBrandData != null && dualBrandData.selectable
```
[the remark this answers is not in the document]
araratthehero: The `setCardBrands` case does not cover the focus changed state.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt
```diff
@@ -332,9 +333,16 @@ class CardView @JvmOverloads constructor(
         environment: Environment,
         isCardScanningVisible: Boolean,
     ) {
+        val brandState = when {
+            dualBrandData != null -> BrandState.DualBrand(dualBrandData, environment)
+            detectedCardTypes.isNotEmpty() -> BrandState.SingleBrand(detectedCardTypes.first(), environment)
```
jreij, line 338: Can `detectedCardTypes.first()` be unsupported? Or did we already filter those out at this stage?

araratthehero: In the delegate we already filter. The `detectedCardTypes` has only supported brands.

PR #2660, opened 2026-03-31
v5- UPI auto detect - Detect installed UPI apps

OscarSpruit:
## Description
This PR enables the UPI component to detect installed UPI apps. When there are apps detected the list from the backend is compared and if the app occurs in both, then it will be displayed. If no apps are detected, then the list from the backend is displayed.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1092

### DIFF
--- a/upi/src/main/java/com/adyen/checkout/upi/internal/ui/DefaultUPIDelegate.kt
+++ b/upi/src/main/java/com/adyen/checkout/upi/internal/ui/DefaultUPIDelegate.kt
@@ -60,6 +60,12 @@ internal class DefaultUPIDelegate(
     private val packageManager: PackageManager,
 ) : UPIDelegate {
 
+    private val detectedPackageNames by lazy {
+        val intent = Intent(Intent.ACTION_VIEW, GENERIC_UPI_URI)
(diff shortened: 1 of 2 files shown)


upi/src/main/java/com/adyen/checkout/upi/internal/ui/DefaultUPIDelegate.kt
```diff
@@ -157,10 +161,17 @@ internal class DefaultUPIDelegate(
         environment: Environment,
         selectedUPIIntentItem: UPIIntentItem?,
     ): List<UPIIntentItem> {
-        return upiApps.mapToPaymentApp(
-            environment = environment,
-            selectedAppId = (selectedUPIIntentItem as? UPIIntentItem.PaymentApp)?.id,
-        )
+        val intent = Intent(Intent.ACTION_VIEW, GENERIC_UPI_URI)
+        val detectedApps = packageManager.queryIntentActivities(intent, PackageManager.MATCH_DEFAULT_ONLY)
```
jreij: Nit: could something like `detectedPackageNames` be clearer?

PR #2680, opened 2026-04-03
v6 - Add compose previews to text fields

jreij:
## Description
Add compose previews to all existing text fields.

## Checklist 

## Ticket Number
COSDK-921

### DIFF
--- a/blik/api/blik.api
+++ b/blik/api/blik.api
@@ -47,6 +47,12 @@ public final class com/adyen/checkout/blik/internal/ui/ComposableSingletons$Stor
 	public final fun getLambda$-1061264656$blik_release ()Lkotlin/jvm/functions/Function3;
 }
 
+public final class com/adyen/checkout/blik/internal/ui/view/ComposableSingletons$BlikCodeFieldKt {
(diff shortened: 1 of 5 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -193,21 +199,46 @@ private fun CardNumberFieldIcon(
...
-        ),
-        isSupportedCardBrandsShown = true,
-        detectedCardBrands = listOf(CardBrand(CardType.MASTERCARD.txVariant)),
-        isAmex = false,
-        onIntent = {},
-    )
+private fun CardNumberFieldPreview(
+    @PreviewParameter(TextFieldStylePreviewParameterProvider::class) theme: CheckoutTheme,
+) {
+    InternalCheckoutTheme(theme) {
```
OscarSpruit: Not sure if it is a great idea, but maybe we can extract this in a `CheckoutPreview(theme)` composable. Would reduce the boilerplate a bit

jreij: Fixed

PR #2690, opened 2026-04-09
v6 - Extract common logic for validation and requirement policy

jreij:
## Description
Extracted common logic to:
- Handle text input field validation based on requirement policy.
- Extract payment data from text input fields based on requirement policy.
- Hide text input fields on the UI based on requirement policy.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-921

### DIFF
--- a/blik/src/main/java/com/adyen/checkout/blik/internal/ui/view/BlikComponent.kt
+++ b/blik/src/main/java/com/adyen/checkout/blik/internal/ui/view/BlikComponent.kt
@@ -43,10 +43,12 @@ internal fun BlikComponent(
             // Helper text
             Body(text = resolveString(CheckoutLocalizationKey.BLIK_HELPER_TEXT))
 
-            BlikCodeField(
-                blikCodeState = viewState.blikCode,
-                onIntent = onIntent,
-            )
+            if (viewState.blikCode != null) {
+                BlikCodeField(
+                    blikCodeState = viewState.blikCode,
+                    onIntent = onIntent,
+                )
+            }
         }
     }
 }
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateExt.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateExt.kt
@@ -1,0 +1,20 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 27/3/2026.
+ */
+
+package com.adyen.checkout.core.components.internal.ui.state.model
+
+import androidx.annotation.RestrictTo
+
+/**
+ * Returns the raw string value of a text input based on its requirement policy. This value is usually set inside the
+ * `PaymentMethodDetails` class and sent as it is through the `onSubmit` callback.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun TextInputComponentState.getPaymentDataValue(): String? {
+    return text.takeIf { requirementPolicy != RequirementPolicy.Hidden && it.isNotBlank() }
+}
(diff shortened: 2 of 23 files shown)


blik/src/main/java/com/adyen/checkout/blik/internal/ui/view/BlikComponent.kt
```diff
@@ -43,10 +43,12 @@ internal fun BlikComponent(
             // Helper text
             Body(text = resolveString(CheckoutLocalizationKey.BLIK_HELPER_TEXT))
```
OscarSpruit, line 44: Should this text be shown if we don't show the field?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateExt.kt
```diff
@@ -0,0 +1,33 @@
...
+fun TextInputComponentState.getPaymentDataValue(): String? {
+    return text.takeIf { requirementPolicy != RequirementPolicy.Hidden && it.isNotBlank() }
+}
+
+/**
+ * Returns true if the text field is hidden or optional and empty -> field is automatically valid
+ * Returns false if the text field is required or optional and not empty -> further validation is needed
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun TextInputComponentState.isNotRequiredAndValid(): Boolean {
```
OscarSpruit: The name of this function is a bit misleading IMO. It sounds like it does actual validation on the value. Would `isNotRequiredAndNotEmpty` be a better name?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MBWayContent.kt
```diff
@@ -53,8 +53,14 @@ internal fun MBWayContent(
                     .fillMaxWidth(),
             )
 
-            // PhoneNumber
-            MBWayPhoneNumberField(viewState.phoneNumber, country.callingCode, onIntent)
+            if (viewState.phoneNumber != null) {
```
OscarSpruit, line 56: Although, I understand that generalizing the requirement policy causes this, I'm not sure about these cases where the field is always required. We have to add this if statement that will basically always be true. Do you think we could somehow force these to be non-nullable?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateExt.kt
OscarSpruit: On second thought - maybe something like `canSkipValidation` or `isValidationNeeded` would be even better

jreij: I was looking for a better name, I like `isValidationNeeded` or something similar. Thanks!

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MBWayContent.kt
jreij: There is nothing "technical" preventing a field from being hidden, other than business rules. On one side having every field behave the same way makes it easier to generate new fields (with AI for example) but on the other side if you read the compose code you might be under the impression that the phone number, card number, etc are fields that can be hidden in some cases, while is reality they can't. I don't know which approach is better.

As for how to achieve that technically, maybe in the `MBWayViewStateProducer`, we can be very explicit that we never expect this field to be hidden, maybe something like this:
```kotlin
phoneNumber = state.phoneNumber.toViewState()
// change to
phoneNumber = requireNotNull(state.phoneNumber.toViewState())
```

Let me know your thoughts on both points.

blik/src/main/java/com/adyen/checkout/blik/internal/ui/view/BlikComponent.kt
jreij: Technically it's not, it's a helper text for the whole component. But the field can never be hidden so it doesn't really matter I guess?

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/view/MBWayContent.kt
OscarSpruit: After discussing we decided none of the options are great and we will keep it as is. Using `requireNotNull` could potentially crash and creating a nullable and non-nullable version of `toViewState` spreads the decision making to the call site. Better to play it safe and have the decision making in one place

OscarSpruit (approved):

PR #2692, opened 2026-04-10
v6 - Architecture - actions

OscarSpruit:
## Description
This PR adds a composable to display actions and adds action support to the `CheckoutPaymentFlow` composable. Additionally, the `NewCheckoutController` now triggers the additional details callback.

## Ticket Number
COSDK-1115

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
@@ -9,41 +9,60 @@
 package com.adyen.checkout.core.components
 
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.platform.LocalLocale
-import com.adyen.checkout.core.common.Environment
-import com.adyen.checkout.core.common.internal.helper.CheckoutCompositionLocalProvider
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationProvider
-import com.adyen.checkout.ui.internal.theme.InternalCheckoutTheme
 import com.adyen.checkout.ui.theme.CheckoutTheme
 
-// TODO - Architecture: Adjust this to make it work
 @Composable
 fun CheckoutPaymentFlow(
-//    controller: NewCheckoutController,
-//    modifier: Modifier = Modifier,
+    controller: NewCheckoutController,
+    modifier: Modifier = Modifier,
     theme: CheckoutTheme = CheckoutTheme(),
     localizationProvider: CheckoutLocalizationProvider? = null,
 ) {
-//    val state by controller.state.collectAsStateWithLifecycle()
-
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/adyen/checkout/core/components/NewCheckoutController.kt
```diff
@@ -206,6 +210,24 @@ class NewCheckoutController(
...
+                    is ActionComponentEvent.ActionDetails -> {
+                        callbacks.onAdditionalDetails?.onAdditionalDetails(event.data)
+                    }
+
+                    is ActionComponentEvent.Error -> {
+                        callbacks.onError?.onError(event.error.toCheckoutError())
+                    }
+                }
+            }
+            .launchIn(coroutineScope)
```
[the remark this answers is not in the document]
OscarSpruit: This will be address in a later PR

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6Screen.kt
araratthehero, line 125: This is unrelated to this PR, but we can discuss this later.

What if `selectedPaymentMethod` changes, we might have ongoing operations happening in the old controller which can leak resources. Should we create a cleanup function inside of the controller, and use `DisposableEffect` to call that somehow?

araratthehero, line 123: This is unrelated to this PR, but we can discuss this later.

`rememberCoroutineScope()` is tied to the composition here, but we have a key of `selectedPaymentMethod` when creating the `checkoutController`. When `selectedPaymentMethod` changes, a new controller is created with the same coroutine scope we had before, and the old coroutine scope does not get cancelled. This might lead to duplicate event collections?

What do you think of creating the coroutine inside of the `remember` block for the `checkoutController`?

core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
```diff
@@ -9,41 +9,60 @@
...
 fun CheckoutPaymentFlow(
-//    controller: NewCheckoutController,
-//    modifier: Modifier = Modifier,
+    controller: NewCheckoutController,
+    modifier: Modifier = Modifier,
     theme: CheckoutTheme = CheckoutTheme(),
     localizationProvider: CheckoutLocalizationProvider? = null,
 ) {
-//    val state by controller.state.collectAsStateWithLifecycle()
+    var state by remember {
```
araratthehero: What do you think of adding `controller` as a parameter for remember, to reset the state if the controller changes?

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6Screen.kt
OscarSpruit: Good point! If we would add a clean up method it might be better to create our own coroutine scope as well, so we can cancel it without affecting the scope of a merchant. Let's discuss later indeed

OscarSpruit: Let's discuss together with the other comment. I think they are related in a way

araratthehero (approved):

PR #2696, opened 2026-04-13
v6 - Refactor DetectCardTypeRepository

jreij:
## Description
This is the first part of a major refactor to how we handle card brand detection. This PR focuses on two main improvements:
- Extracting local brand detection, network brand detection and caching logic into separate classes to make `DefaultDetectCardTypeRepository` more readable and maintainable.
- Replacing the flow and the separate function call from `DetectCardTypeRepository` with one function that returns a flow, making its interface simpler and `DefaultDetectCardTypeRepository` stateless.

The next refactors will address the following points:
- Being more explicit in `CardComponentState` on which flow are we currently in (local vs network - single vs dual branded - etc).
- How the detected brands are mapped and propagated to the UI.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-921


card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -159,31 +159,41 @@ internal class CardComponent(
...
+                DetectCardTypesResult.FetchingFromNetwork -> {
+                    // do nothing, wait for the previous call to resolve
+                }
+
+                is DetectCardTypesResult.Local -> {
+                    onIntent(CardIntent.UpdateDetectedCardTypes(detectCardTypeResult.detectedCardTypes))
+                }
+
+                is DetectCardTypesResult.Network -> {
+                    onIntent(CardIntent.UpdateDetectedCardTypes(detectCardTypeResult.detectedCardTypes))
```
jreij: At the moment we do not clear the component state when the BIN changes and the current `detectedCardTypes` corresponds to previous results from BIN lookup for a different BIN.

Realistically this can only cause a bug in a scenario such as this:
- Shopper enters card number, bin lookup call is made and returns some `detectedCardTypes`.
- Shopper replaces the card number with another one having a different BIN, for example by using Select All + Paste. BIN lookup is directly triggered here but until it returns a result, the UI (and state) is still using the old results from the first BIN.
- Shopper quickly presses pay, a payments call is made with the wrong brand field causing the payment to fail.

Some notes here:
- This bug will not happen if the shopper manually enters their second card number digit by digit, because the repo will emit the local detected brands while typing the first few digits. This would avoid breaking the API call.
- This bug relies the shopper being really quick and BIN lookup being really slow.
- Before this refactor, this bug could not occur because `DetectCardTypeRepository` always returned the local brands first and then the network brands if available. This was possible because the repository had a flow which meant it could emit multiple results for one card number. But with a suspend function, this is not possible anymore.

I am considering caching the last used BIN somewhere and clearing the detected card types from the state if the BIN changes, or returning the local detected brands. I am open to suggestions on where to cache this or completely new ideas on how to handle this bug.

jreij: Following up on the previous comment, one effect of replacing the flow with a suspend function is with autofill or copy/paste. Before this refactor, autofilling a card number would cause the flow to emit two values: the local brand then the BIN lookup brand. After this refactor, only one value can be emitted from the suspend function which will be the BIN lookup brand. So while waiting for the BIN lookup response, no brand will be displayed on the UI.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -159,31 +159,41 @@ internal class CardComponent(
...
+                }
             }
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist it is incredibly unlikely that an older request completes after a newer one, because we trim the card number (into a BIN) and only call the API when the BIN changes. The user would need to very quickly enter two different card numbers in a row and somehow the second call should finish faster than the first.

As for combining handling the Local and Network results, this is not possible because while they both have a `detectedCardTypes` list field, it is not the same field in both because this is a sealed class and not all of its subclasses have this field (FetchingFromNetwork does not have it)

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -195,18 +205,22 @@ internal class CardComponent(
...
+        componentState
+            .map { it.detectedCardTypes }
+            .distinctUntilChanged()
+            .drop(1)
+            .onEach(::onDetectedCardTypesChanged)
+            .launchIn(coroutineScope)
+    }
+
+    private fun onDetectedCardTypesChanged(detectedCardTypes: List<DetectedCardType>?) {
+        val isReliable = !detectedCardTypes.isNullOrEmpty() && detectedCardTypes.any { it.isReliable }
```
araratthehero: The `detectedCardTypes` list is not null. What do you think of doing the following?
```suggestion
    private fun onDetectedCardTypesChanged(detectedCardTypes: List<DetectedCardType>) {
        val isReliable = detectedCardTypes.any { it.isReliable }
```

card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
```diff
@@ -185,100 +193,164 @@ internal class DetectCardTypeRepositoryTest(
...
-                                enableLuhnCheck = true,
-                                supported = true,
-                                cvcPolicy = "required",
-                                expiryDatePolicy = "required",
-                                panLength = 16,
-                                paymentMethodVariant = "scheme",
-                                localizedBrand = "MasterCard",
-                            ),
+    @Test
+    fun `and detectCardTypes is called more than once with the same bin, only one bin lookup call should be made`() =
```
araratthehero: Shouldn't this test be moved inside of the `BinLookupTest`?

card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
```diff
@@ -185,100 +193,164 @@ internal class DetectCardTypeRepositoryTest(
...
-                            supportedCardBrands = listOf(),
-                            clientKey = "",
-                            coroutineScope = coroutineScope,
-                            type = "",
-                        )
-                        val expected = DetectedCardType(
-                            cardBrand = CardBrand(CardType.MASTERCARD.txVariant),
-                            isReliable = true,
+    @Test
+    fun `and detectCardTypes is called again with the same bin before bin lookup call returns, result should indicate that its still fetching from network`() =
```
araratthehero: Shouldn't this test be moved inside of the `BinLookupTest`?

card/src/main/java/com/adyen/checkout/card/internal/data/api/BinLookupCache.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.adyen.checkout.card.internal.data.api
+
+import com.adyen.checkout.card.internal.data.model.BinLookupCacheResult
+import com.adyen.checkout.card.internal.data.model.DetectedCardType
+import com.adyen.checkout.core.common.internal.helper.Sha256
+import java.util.concurrent.ConcurrentHashMap
+
+internal class BinLookupCache {
```
araratthehero, line 16: Could you add unit tests for this class too?

card/src/main/java/com/adyen/checkout/card/internal/data/api/LocalCardBrandDetectionService.kt
```diff
@@ -0,0 +1,51 @@
...
+package com.adyen.checkout.card.internal.data.api
+
+import com.adyen.checkout.card.internal.data.model.Brand
+import com.adyen.checkout.card.internal.data.model.DetectedCardType
+import com.adyen.checkout.core.common.AdyenLogLevel
+import com.adyen.checkout.core.common.CardBrand
+import com.adyen.checkout.core.common.CardType
+import com.adyen.checkout.core.common.internal.helper.adyenLog
+
+internal class LocalCardBrandDetectionService {
```
araratthehero: Could you add tests for this class?

card/src/main/java/com/adyen/checkout/card/internal/data/api/NetworkCardBrandDetectionService.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.adyen.checkout.card.internal.data.model.Brand
+import com.adyen.checkout.card.internal.data.model.DetectedCardType
+import com.adyen.checkout.core.common.AdyenLogLevel
+import com.adyen.checkout.core.common.CardBrand
+import com.adyen.checkout.core.common.internal.helper.adyenLog
+import com.adyen.checkout.core.common.internal.helper.runSuspendCatching
+import com.adyen.checkout.cse.internal.BaseCardEncryptor
+import java.util.UUID
+
+internal class NetworkCardBrandDetectionService(
```
araratthehero, line 23: Could you add tests for this class, especially for the mapping?

card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
araratthehero: I would also add tests for the following scenarios:
- Network failure -> fallback to local detection
- Empty card number input
- Different BINs in sequence (race condition scenario)

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -159,31 +159,41 @@ internal class CardComponent(
...
+            when (detectCardTypeResult) {
+                DetectCardTypesResult.FetchingFromNetwork -> {
+                    // do nothing, wait for the previous call to resolve
+                }
+
+                is DetectCardTypesResult.Local -> {
+                    onIntent(CardIntent.UpdateDetectedCardTypes(detectCardTypeResult.detectedCardTypes))
+                }
+
+                is DetectCardTypesResult.Network -> {
```
araratthehero: Since the logic in both of the cases (`Local` and `Network`) is identical, is it necessary to have two different result types? They could be combined and a single list could be returned.

araratthehero: If we treat `FetchingFromNetwork` as null, the approach could even optiimse further to:
```
detectCardTypeResult.detectedCardTypes?.let { 
    onIntent(CardIntent.UpdateDetectedCardTypes(it)) 
}
```

Not sure how easy it is to read this though

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
araratthehero: I think this is fine, since showing no logo can be better than showing a brand logo which might change after a second no? Also how long does it take to do the bin lookup and see a brand icon?

If it does not look good, we could then show the locally detected card first before making a network call. Though this will require some adjustments or even breaking the flow of the logic encapsulation inside of the repository.

araratthehero: I like the idea of clearing the detected card types if the bin changes. That way we don't have stale information, which can cause issues. Though this can cause other issues, like:
- Not sending a brand to the `payments` call, when user clicks Submit before the brand is detected.
- Flickering of fields which depend on the card brand, such as security code and expiry date.
- Showing a placeholder icon for the brand. (If we find a way to always show a locally detected card before doing a network call, it would solve the next comment too.)

jreij: I couldn't find a clean way to orchestrate all that logic in the repo alone so I decided to move it to the card component with as little logic as possible. Pushed a commit for this change

jreij: My last commit should address this as well

card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
jreij: Should be fixed in the last commit

card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
jreij: Should be fixed in the last commit

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
jreij: With the last commit, we don't need this sealed class anymore

card/src/test/java/com/adyen/checkout/card/internal/data/api/DetectCardTypeRepositoryTest.kt
jreij: All your test suggestions are great! What do you think doing all this in the next PR? I don't want this one to become huge and this whole refactor is under one feature branch anyway

araratthehero: Sure, sounds good!

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
jreij: Fixed

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -159,31 +157,14 @@ internal class CardComponent(
...
-            .onEach {
-                // TODO - Card. Dual brands reset brand
-                // inputData.selectedCardBrand = null
-            }
-            .launchIn(coroutineScope)
+    private fun detectCardTypes(cardNumber: String) {
+        detectCardTypeRepository.detectCardTypes(cardNumber).onEach { result ->
+            onIntent(CardIntent.UpdateDetectedCardTypes(result))
+        }.launchIn(coroutineScope)
     }
```
araratthehero, line 168: Every time the card number changes, we call the `detectCardTypeRepository.detectCardTypes`, which creates a new flow. This causes multiple flows to run simultaneously, which can cause a race condition. To be safe, I would store the job and cancel it once a new value has been submitted.

```suggestion
    private var detectCardTypesJob: Job? = null

    private fun detectCardTypes(cardNumber: String) {
        detectCardTypesJob?.cancel()
        detectCardTypesJob = detectCardTypeRepository.detectCardTypes(cardNumber).onEach { result ->
            onIntent(CardIntent.UpdateDetectedCardTypes(result))
        }.launchIn(coroutineScope)
    }
```

jreij: This works if the latest job is guaranteed to return all the results we need - the problem here is that it doesn't.
Take the example where you input 11 digits then you input another digit. This will trigger two `detectCardTypes` calls:
- The first call/flow will emit the local detected cards then call bin lookup and emit the network cards.
- The second flow will emit the local detected cards and skip bin lookup because there's a request in progress for the same BIN already.

So if we cancel the first flow as soon as the second is triggered, we will lose to the BIN lookup results.

As for the race condition:
- Between two network calls for the same BIN: we ensure we only call BIN lookup once per BIN, so this cannot happen.
- Between two network calls for two different BINs: I feel like the only way it could happen is if the user enters a card number then very quickly enters another one with a different BIN and somehow the second call finishes faster than the first one. Even there, the next PR should address this by making sure we don't apply the results from the flow if they don't match the current BIN in the card number field.
- Between two local results: shouldn't happen because results are returned synchronously.
- Between a local result and a network result: super unlikely but maybe worth looking at? We can change the second call in the example above to not return anything (not even local brands). That way nothing can interfere with the first flow that emits the bin lookup results.

Let me know if I missed anything

jreij: Pushed a commit to fix the last point

araratthehero: Makes sense. Looks good to me now!

araratthehero (approved):

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -159,31 +157,14 @@ internal class CardComponent(
...
-                detectedCardTypes.filter { it.isReliable && it.isSupported }.map { it.cardBrand }
-            }
-            .distinctUntilChanged()
-            .onEach {
-                // TODO - Card. Dual brands reset brand
-                // inputData.selectedCardBrand = null
-            }
-            .launchIn(coroutineScope)
+    private fun detectCardTypes(cardNumber: String) {
+        detectCardTypeRepository.detectCardTypes(cardNumber).onEach { result ->
```
OscarSpruit, line 165: This will launch a new coroutine every time the card number changes. Technically this could result in a concurrency issue where if 2 coroutines are launched around the same time and the first one is slower it could overwrite the second one. To mitigate this I think the only way is to save the `Job` and cancel it every time this function is called

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -195,18 +176,22 @@ internal class CardComponent(
             .launchIn(coroutineScope)
     }
 
-    private fun onDetectedCardTypes(detectedCardTypes: List<DetectedCardType>) {
-        adyenLog(AdyenLogLevel.DEBUG) {
-            "New detected card types emitted - detectedCardTypes: ${detectedCardTypes.map { it.cardBrand }} " +
-                "- isReliable: ${detectedCardTypes.firstOrNull()?.isReliable}"
-        }
+    private fun subscribeToDetectedCardTypesChanges() {
```
OscarSpruit, line 179: I think this is superfluous. We could just call `onDetectedCardTypesChanged` from line 166

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
jreij: Can you check out the discussion we had [here](https://github.com/Adyen/adyen-android/pull/2696#discussion_r3110111362)? I feel like it's the same topic, happy to hear your ideas about it

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
jreij: I already renamed it in the next PR, is it okay to fix it there only?

PR #2701, opened 2026-04-17
chore: Align v6 checkout callbacks (SubmitResult / AdditionalDetailsResult)

Robert-SD:
## Description

Aligns the v6 advanced-flow checkout callbacks with the new stage-specific result types (`SubmitResult` / `AdditionalDetailsResult`) and clarifies how thrown exceptions are handled at each callback boundary.

### Public API (v6 advanced flow)

- New `SubmitResult` sealed interface: `Completion(resultCode)` | `Action(action)` | `Retry(errorMessage?)` | `PartialPayment(order, paymentMethods)`.
- New `AdditionalDetailsResult` sealed interface: `Completion(resultCode)` (single variant — `/payments/details` no longer returns chained actions).
- `AdvancedCheckoutCallbacks.onSubmit` now returns `SubmitResult`; `AdvancedCheckoutCallbacks.onAdditionalDetails` now returns `AdditionalDetailsResult`.
- `DropInService.onSubmit` / `DropInService.onAdditionalDetails` retyped accordingly.
- The `actionComponentData → data` parameter rename applies **only** to the new v6 surface (`AdvancedCheckoutCallbacks`, `DropInService`, v6 `PaymentMethodViewModel`). The v5 legacy callback contracts (`ComponentCallback`, `ActionComponentCallback`, `SessionComponentCallback`, `DropInServiceContract`, `SessionDropInServiceContract`) keep `actionComponentData` to avoid a source-break for Kotlin merchants relying on named arguments.
- Cross-platform parity verified with iOS — variant set, payload names, and nullability all match (see `SUBMIT_AND_ADDITIONAL_DETAILS_RESULT_DECISION` plan).

### Error-handling contract

Per Android team review on 2026-04-30, the dispatcher layer treats merchant-callback throws and SDK-internal API failures asymmetrically:

| Path | Failure source | Behavior |
|---|---|---|
| Advanced — `AdvancedComponentRequestDispatcher.submit/additionalDetails` | Merchant `onSubmit` / `onAdditionalDetails` throws | **Not caught.** Throw propagates and crashes the process. Treated as a programming error; recoverable infrastructure failures should be modeled by the merchant returning `SubmitResult.Retry(errorMessage = ...)`. |
| Session — `SessionComponentRequestDispatcher.submit/additionalDetails` | `SessionRepository` HTTP / parsing failure | Caught. `callbacks.onError(error.toCheckoutError())` then `throw CancellationException(...)` to abort the launched coroutine cleanly without invoking `handleResult`. **Mirrors iOS `Session.didSubmit` / `Session.didProvide`** — both platforms terminate session API failures with `onError` only. |

This is intentionally divergent from iOS, which catches merchant throws in `Checkout+DelegateCallbacks.swift` and routes them to `onError`. The asymmetry is documented for Flutter / React Native wrapper authors: on Android they cannot rely on the SDK to convert merchant-callback throws because the JVM process will be gone before the bridge sees the throw.

### Internal

- New `Throwable.toCheckoutError()` mapper in `core/error/CheckoutErrorMapper.kt` — delegates to the existing `InternalCheckoutError` mapping when applicable; otherwise wraps as `ErrorCode.UNKNOWN` preserving message and cause. Used **only** by the session dispatcher's API-failure path.
- `SessionInteractor.makeDetailsCallInternal` defensively maps an unexpected `action` payload from `/sessions/{id}/payments/details` to `SessionCallResult.Details.Error(IllegalStateException(UNEXPECTED_ACTION_RESPONSE_MESSAGE))` (`SessionCallResult.Details.Action` is removed). Per backend contract, `/details` no longer returns chained actions.
- `FullCheckoutFlow.submit` and `ActionHandler.ActionDetails` are **not** wrapped in `runSuspendCatching`. Plain delegation; merchant throws propagate uncaught.

### Tests

- `SessionComponentRequestDispatcherTest` — 4 tests covering the failure paths:
  - `submit()` failure → `onError` invoked with mapped `CheckoutError` (cause preserved), `CancellationException` thrown with original cause.
  - `additionalDetails()` failure → same.
  - `submit()` failure → `onFinished` is **not** invoked.
  - `additionalDetails()` failure → `onFinished` is **not** invoked.

### Follow-ups (not in this PR)

- `SubmitResult.Completion` / `Retry` / `PartialPayment` and `AdditionalDetailsResult.Completion` are wired through the dispatcher but `FullCheckoutFlow.handleResult` and `ActionHandler` still consume them as `// TODO` — the merchant-facing `onComplete` / `onFinished(PaymentResult)` plumbing lands in a follow-up PR. Empty/null `resultCode` handling on `Completion` is also deferred until that consumer exists.
- A typed `CheckoutResultCode` enum on Android (parity with iOS) is deferred.
- Contract tests for `AdvancedComponentRequestDispatcher` (locking in the "no catch" rule) are deferred.

## Checklist
- [X] If applicable, make sure Breaking change label is added.
- [X] Code is unit tested
- [X] Changes are tested manually

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
@@ -12,7 +12,7 @@ import com.adyen.checkout.core.error.CheckoutError
 
 sealed interface AdditionalDetailsResult {
 
-    data class Finished(val temp: String? = null) : AdditionalDetailsResult
+    data class Finished(val resultCode: String) : AdditionalDetailsResult
 
     data class Error(val error: CheckoutError) : AdditionalDetailsResult
 }
--- a/core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by robertsc on 17/4/2026.
+ */
+
+package com.adyen.checkout.core.components
+
+import com.adyen.checkout.core.error.CheckoutError
+
+sealed interface AdditionalDetailsResult {
+
+    data class Finished(val temp: String? = null) : AdditionalDetailsResult
+
+    data class Error(val error: CheckoutError) : AdditionalDetailsResult
+}
--- a/core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
@@ -8,11 +8,8 @@
 
 package com.adyen.checkout.core.components
 
-import com.adyen.checkout.core.error.CheckoutError
-
 sealed interface AdditionalDetailsResult {
 
-    data class Finished(val resultCode: String) : AdditionalDetailsResult
-
-    data class Error(val error: CheckoutError) : AdditionalDetailsResult
+    /** Indicates the additional-details exchange has completed and carries a final `resultCode`. */
+    data class Completion(val resultCode: String) : AdditionalDetailsResult
 }
--- a/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
@@ -13,22 +13,21 @@ import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethod
 import com.adyen.checkout.core.error.CheckoutError
 import com.adyen.checkout.core.action.data.Action as ActionResponse
 
-// TODO - KDocs, revisit later after having parameters
 /**
  * Represents the outcome of a checkout operation.
  */
 sealed interface SubmitResult {
 
     /** Indicates the payment process has finished successfully. */
-    // TODO - Replace temp parameter with actual value
-    data class Finished(val temp: String? = null) : SubmitResult
+    data class Finished(val resultCode: String) : SubmitResult
 
     /** Indicates that an additional action is required from the shopper. */
     data class Action(val action: ActionResponse) : SubmitResult
 
+    /** Indicates that a partial payment has been made. */
     data class PartialPayment(
         val order: OrderResponse?,
-        val paymentMethods: PaymentMethods,
+        val paymentMethods: PaymentMethods? = null,
     ) : SubmitResult
 
     /** Indicates an error occurred during the payment process. */
--- a/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
@@ -8,10 +8,10 @@
 
 package com.adyen.checkout.core.components
 
-import com.adyen.checkout.core.error.CheckoutError
 import com.adyen.checkout.core.components.data.OrderResponse
-import com.adyen.checkout.core.action.data.Action as ActionResponse
 import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethods
+import com.adyen.checkout.core.error.CheckoutError
+import com.adyen.checkout.core.action.data.Action as ActionResponse
 
 // TODO - KDocs, revisit later after having parameters
 /**
--- a/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
@@ -29,7 +29,7 @@ sealed interface SubmitResult {
 
     /** Indicates that a partial payment has been made. */
     data class PartialPayment(
-        val order: OrderResponse?,
+        val order: OrderResponse,
         val paymentMethods: PaymentMethods? = null,
     ) : SubmitResult
 }
--- a/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
@@ -8,22 +8,29 @@
 
 package com.adyen.checkout.core.components
 
+import com.adyen.checkout.core.error.CheckoutError
+import com.adyen.checkout.core.components.data.OrderResponse
 import com.adyen.checkout.core.action.data.Action as ActionResponse
+import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethods
 
 // TODO - KDocs, revisit later after having parameters
 /**
  * Represents the outcome of a checkout operation.
  */
-sealed interface CheckoutResult {
+sealed interface SubmitResult {
 
     /** Indicates the payment process has finished successfully. */
     // TODO - Replace temp parameter with actual value
-    data class Finished(val temp: String? = null) : CheckoutResult
+    data class Finished(val temp: String? = null) : SubmitResult
 
     /** Indicates that an additional action is required from the shopper. */
-    data class Action(val action: ActionResponse) : CheckoutResult
+    data class Action(val action: ActionResponse) : SubmitResult
+
+    data class PartialPayment(
+        val order: OrderResponse?,
+        val paymentMethods: PaymentMethods,
+    ) : SubmitResult
 
     /** Indicates an error occurred during the payment process. */
-    // TODO - Error propagation: Revisit error type.
-    data class Error(val errorMessage: String) : CheckoutResult
+    data class Error(val error: CheckoutError) : SubmitResult
 }
--- a/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
@@ -24,12 +24,12 @@ sealed interface SubmitResult {
     /** Indicates that an additional action is required from the shopper. */
     data class Action(val action: ActionResponse) : SubmitResult
 
+    /** Indicates an error occurred during the payment process. */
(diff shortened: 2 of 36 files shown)


core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
```diff
@@ -0,0 +1,18 @@
...
+package com.adyen.checkout.core.components
+
+import com.adyen.checkout.core.common.exception.CheckoutError
+
+sealed interface AdditionalDetailsResult {
+
+    data class Finished(val temp: String? = null) : AdditionalDetailsResult
+
+    data class Error(val error: CheckoutError) : AdditionalDetailsResult
+}
```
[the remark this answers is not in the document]
Robert-SD: We will not add the Action for /payments/details response. This is no longer needed.

core/src/main/java/com/adyen/checkout/core/components/AdditionalDetailsResult.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package com.adyen.checkout.core.components
+
+import com.adyen.checkout.core.error.CheckoutError
+
+sealed interface AdditionalDetailsResult {
+
+    data class Finished(val resultCode: String) : AdditionalDetailsResult
+
+    data class Error(val error: CheckoutError) : AdditionalDetailsResult
```
OscarSpruit: Are you sure about using `CheckoutError` here? I feel like `CheckoutError` is something we give to merchants and not receive, because to provide the correct error code etc can easily go wrong

core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
```diff
@@ -0,0 +1,35 @@
...
+sealed interface SubmitResult {
+
+    /** Indicates the payment process has finished successfully. */
+    data class Finished(val resultCode: String) : SubmitResult
+
+    /** Indicates that an additional action is required from the shopper. */
+    data class Action(val action: ActionResponse) : SubmitResult
+
+    /** Indicates an error occurred during the payment process. */
+    data class Error(val error: CheckoutError) : SubmitResult
```
OscarSpruit: Same here

core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Created by ozgur on 16/5/2025.
+ */
+
+package com.adyen.checkout.core.components
+
+import com.adyen.checkout.core.components.data.OrderResponse
+import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethods
+import com.adyen.checkout.core.error.CheckoutError
+import com.adyen.checkout.core.action.data.Action as ActionResponse
```
Robert-SD, line 14: Is using the alias here ok?

core/src/main/java/com/adyen/checkout/core/components/internal/ActionHandler.kt
```diff
@@ -55,7 +56,8 @@ internal class ActionHandler(
             .onEach { event ->
                 when (event) {
                     is ActionComponentEvent.ActionDetails -> {
-                        componentRequestDispatcher.additionalDetails(event.data)
+                        runSuspendCatching { componentRequestDispatcher.additionalDetails(event.data) }
+                            .onFailure { componentRequestDispatcher.error(it.toCheckoutError()) }
```
Robert-SD: I added the .onFailure to catch errors and map them to the .error path. Would that be okay? 
I've checked the codebase and found other usages of `runSuspendCatching`.

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
```diff
@@ -58,8 +59,9 @@ internal class FullCheckoutFlow(
                 when (event) {
                     is PaymentComponentEvent.Submit -> {
                         paymentComponent.setLoading(true)
-                        val result = componentRequestDispatcher.submit(event.state.data)
-                        handleResult(result)
+                        runSuspendCatching { componentRequestDispatcher.submit(event.state.data) }
+                            .onSuccess { handleResult(it) }
+                            .onFailure { componentRequestDispatcher.error(it.toCheckoutError()) }
```
Robert-SD: Same here

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
```diff
@@ -48,12 +48,13 @@ internal class SessionComponentRequestDispatcher(
 //                    event = ErrorEvent.API_PAYMENTS,
 //                )
 //                analyticsManager.trackEvent(event)
-                return CheckoutResult.Error(error.message ?: "Failed to submit payment")
+//                callbacks.onError(error.toCheckoutError())
+                throw error
```
Robert-SD: We pass the error through instead of wrapping it with the `Error` case.

example-app/src/main/java/com/adyen/checkout/example/service/ExampleV6DropInService.kt
```diff
@@ -81,7 +82,22 @@ class ExampleV6DropInService : DropInService() {
...
+                SubmitResult.Completion(resultCode)
+            }
+        }
+    }
+
+    private fun handleAdditionalDetailsResponse(jsonResponse: JSONObject?): AdditionalDetailsResult {
+        return when {
+            jsonResponse == null -> {
+                Log.e(TAG, "FAILED")
+                throw IOException("Empty payments/details response")
```
Robert-SD: Would it be okay here to throw in case the json is null? It is the merchants code, can we prevent throwing or shall we do that?

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6ViewModel.kt
```diff
@@ -135,27 +138,40 @@ internal class V6ViewModel @Inject constructor(
...
         val response = paymentsRepository.makeDetailsRequest(request)
-        return handleResponse(response)
+        return handleAdditionalDetailsResponse(response)
     }
 
-    private fun handleResponse(json: JSONObject?): CheckoutResult {
+    private fun handleSubmitResponse(json: JSONObject?): SubmitResult {
         return when {
-            json == null -> CheckoutResult.Error("Network error")
+            json == null -> throw IOException("Empty payments response")
```
Robert-SD: same here

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
```diff
@@ -62,15 +66,11 @@ internal class SessionComponentRequestDispatcher(
...
             onFailure = { error ->
-                callbacks.onError(
-                    CheckoutError(
-                        code = ErrorCode.HTTP,
-                        message = "Failed to submit details",
-                        cause = error,
-                    ),
-                )
+                callbacks.onError(error.toCheckoutError())
+                throw CancellationException("Session additional details failed.", error)
```
Robert-SD: This was the recommendation to stop the coroutine. We could send `Completion("Error")` instead. What do you think?

core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
OscarSpruit: Yes that's fine 👍

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
OscarSpruit: Cancelling the coroutine prevents us from doing anything after it. We can't reset the state of the pay button for example. I think it would be okay to return `SubmitResult.Retry` for the payments call, so we allow shoppers to try again. For the additional details call I think we can just return `AdditionalDetailsResult.Completion("Error")` and maybe we shouldn't trigger the `onError` callback 🤔

core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Indicates the SDK should re-prompt the shopper. Loops back into the next `onSubmit()`.
+     *
+     * @param errorMessage Optional shopper-facing message the SDK can surface before re-prompting.
+     */
+    data class Retry(val errorMessage: String? = null) : SubmitResult
+
+    /** Indicates that a partial payment has been made. */
+    data class PartialPayment(
```
OscarSpruit, line 40: In the first alpha we will not support partial payments yet. Should we maybe comment this out for now or put a `RestrictTo` on it to warn people to not use it?

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6ViewModel.kt
Robert-SD: Updated this to avoid throwing and return `Completion("Error")` for empty responses instead.

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
Robert-SD: Updated: payments failures now call `onError` and return `SubmitResult.Retry`, while details failures call `onError` and return `AdditionalDetailsResult.Completion("Error")` without cancelling the coroutine.

core/src/main/java/com/adyen/checkout/core/components/SubmitResult.kt
Robert-SD: Kept `PartialPayment`, but marked it with `@RestrictTo(LIBRARY_GROUP)` and noted that it is not available in alpha01.

OscarSpruit (approved):

PR #2719, opened 2026-04-23
v6 - Fix public api validation false positives

araratthehero:
## Description

The `Validate public API` workflow used Gradle stderr as the API-change signal:

```bash
( ./gradlew apiCheck --continue 2> api_changes.txt ) || true
...
if [ -s api_changes.txt ]; then exit 1; fi
```

Any stderr output from the build failed the check, regardless of whether `.api`
files actually differ. When Kotlin classes annotated with `@Deprecated` are
referenced from Java sources (e.g. generated DataBinding / ViewBinding /
`R.java`), `javac` emits:

```
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```

These notes go to stderr and caused false positives on recent deprecation PRs
(e.g. #2708).

### Fix

- Run `./gradlew apiDump --quiet` to regenerate all `.api` files from source,
  then use `git diff` against the committed `.api` files as the failure signal.
  This is robust to any stderr noise from the build.
- Add `git add --intent-to-add -- '**/*.api'` so newly-created `.api` files for
  a new module (previously a silent false negative) are now detected.
- Rewrite `scripts/process_api_changes.sh` to render the `git diff` output as
  per-module fenced diff blocks. The previous implementation had a latent bug
  where it pairwise-diffed unrelated modules' `.api` files.

## Manual end-to-end testing

Each scenario was a scratch PR targeting this branch so the workflow comment is
preserved on-PR for review:

- #2720 — **add** scenario: new public functions added in `await` and `blik` without running `apiDump`. CI posted 🚫 with two `## <module>` sections as expected; job failed as intended.
- #2721 — **remove** scenario: orphan entry added to `await.api` that does not exist in source. CI posted 🚫 with a single `## await` removal hunk as expected; job failed as intended.
- #2722 — **new file (intent-to-add)** scenario: committed `sepa/api/sepa.api` removed so CI must regenerate it. CI posted 🚫 with a `## sepa` new-file addition hunk as expected; job failed as intended.
- #2723 — **green path** scenario: public API added together with the matching `apiDump` output. CI posted ✅ "No public API changes" and the job passed.

## Checklist

## Ticket Number
COSDK-1121

### DIFF
--- a/.github/workflows/validate_public_api.yml
+++ b/.github/workflows/validate_public_api.yml
@@ -32,7 +32,15 @@ jobs:
 
       - name: Check public API
         run: |
-          ( ./gradlew apiCheck --continue 2> "${{ github.workspace }}/api_changes.txt" ) || true
+          # Regenerate all .api files from the current source.
+          ./gradlew apiDump --quiet
+          # Mark any newly-created .api files (e.g. from a new module) as
+          # intent-to-add so they show up in `git diff` even though untracked.
+          git add --intent-to-add -- '**/*.api'
+          # Capture only actual changes to .api files. This ignores unrelated
+          # stderr output from the build (e.g. javac "uses deprecated API" notes,
+          # Gradle deprecation warnings) that previously caused false positives.
+          git diff --no-color -- '**/*.api' > "${{ github.workspace }}/api_changes.txt"
           bash ./scripts/process_api_changes.sh
 
       - name: Comment on PR
--- a/.github/workflows/validate_public_api.yml
+++ b/.github/workflows/validate_public_api.yml
@@ -37,11 +37,7 @@ jobs:
           # Mark any newly-created .api files (e.g. from a new module) as
           # intent-to-add so they show up in `git diff` even though untracked.
           git add --intent-to-add -- '**/*.api'
-          # Capture only actual changes to .api files. This ignores unrelated
-          # stderr output from the build (e.g. javac "uses deprecated API" notes,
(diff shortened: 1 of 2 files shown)


.github/workflows/validate_public_api.yml
```diff
@@ -32,7 +32,15 @@ jobs:
...
-          ( ./gradlew apiCheck --continue 2> "${{ github.workspace }}/api_changes.txt" ) || true
+          # Regenerate all .api files from the current source.
+          ./gradlew apiDump --quiet
+          # Mark any newly-created .api files (e.g. from a new module) as
+          # intent-to-add so they show up in `git diff` even though untracked.
+          git add --intent-to-add -- '**/*.api'
+          # Capture only actual changes to .api files. This ignores unrelated
+          # stderr output from the build (e.g. javac "uses deprecated API" notes,
+          # Gradle deprecation warnings) that previously caused false positives.
+          git diff --no-color -- '**/*.api' > "${{ github.workspace }}/api_changes.txt"
```
OscarSpruit: I think this git diff + the `api_changes.txt` is now redundant. The `process_api_changes.sh` script already does a git diff and could use that output

araratthehero: Fixed it. You can also check the test PRs opened for different test scenarios. Links are in the body of the PR.

OscarSpruit (approved):

PR #2721, opened 2026-04-23
[DO NOT MERGE] Test - validate_public_api workflow - remove scenario

araratthehero:
Scratch PR used to exercise the updated `validate_public_api` workflow end-to-end on CI. **Do not merge.**

### Scenario
Single commit that inserts an orphan entry into `await/api/await.api` — it claims a public function (`awaitApiTestStubRemove`) that does not exist in source. This reproduces the state that results when a developer removes a public API and forgets to run `./gradlew apiDump`.

CI's `./gradlew apiDump` should regenerate `await.api` without the orphan entry. The workflow should then post a PR comment with 🚫 and one `## await` section showing the entry as a `-` hunk.

Linked to #2719.

### DIFF
--- a/await/api/await.api
+++ b/await/api/await.api
@@ -6,6 +6,10 @@ public final class com/adyen/checkout/await/BuildConfig {
 	public fun <init> ()V
 }
 
+public final class com/adyen/checkout/await/old/AwaitApiTestStubRemoveKt {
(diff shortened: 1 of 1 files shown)


araratthehero:
Test scenario verified — closing. See #2719 for the fix.

PR #2723, opened 2026-04-23
[DO NOT MERGE] Test - validate_public_api workflow - green path scenario

araratthehero:
Scratch PR used to exercise the updated `validate_public_api` workflow end-to-end on CI. **Do not merge.**

### Scenario
Add a public stub function in `await` **and** commit the matching `await.api` entry in the same commit (i.e. the developer ran `./gradlew apiDump` before pushing).

CI should detect no .api drift: the workflow should post a PR comment with ✅ "No public API changes" and the job should pass.

Linked to #2719.

### DIFF
--- a/await/api/await.api
+++ b/await/api/await.api
@@ -6,6 +6,10 @@ public final class com/adyen/checkout/await/BuildConfig {
 	public fun <init> ()V
 }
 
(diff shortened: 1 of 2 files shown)


araratthehero:
Test scenario verified — closing. See #2719 for the fix.

PR #2724, opened 2026-04-23
v5 - Co-badged cards - Fix tracking for dual branded selector visibility

araratthehero:
## Description
Fixes the analytics `displayed` event for the dual brand button so it is only tracked when the dual brand UI is actually user-selectable.

Previously the event fired whenever `dualBrandData` had any `brandOptions`, including the non-selectable display-only case. The flow now:
- Maps `dualBrandData` to `null` when `selectable == false` so only selectable states propagate downstream.
- Uses a custom `distinctUntilChanged` comparator on `brandOptions` so `null → selectable` transitions reset the de-duplication (the event re-fires correctly after a brand sequence clears and reappears).
- Reads `selectedBrand` from the emitted `dualBrandData` instead of the stale `outputData` snapshot, so the tracked `brand` always matches the emission being recorded.

## Checklist
- [x] Code is unit tested

## Ticket number
COSDK-1062

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
@@ -89,6 +89,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.distinctUntilChangedBy
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
@@ -306,14 +307,16 @@ class DefaultCardDelegate(
             }
             .launchIn(coroutineScope)
 
-        outputDataFlow.map { it.dualBrandData?.brandOptions?.map { it.brand.txVariant } }
-            .distinctUntilChanged()
+        outputDataFlow
+            .map { it.dualBrandData?.takeIf { data -> data.selectable } }
+            .distinctUntilChangedBy { it?.brandOptions?.map { option -> option.brand.txVariant } }
             .filterNotNull()
-            .map { brandOptions ->
+            .onEach { dualBrandData ->
+                val brandOptions = dualBrandData.brandOptions.map { it.brand.txVariant }
                 val event = GenericEvents.displayed(
                     component = paymentMethod.type.orEmpty(),
                     target = DUAL_BRAND_ANALYTICS_TARGET,
-                    brand = outputData.dualBrandData?.selectedBrand?.txVariant.orEmpty(),
+                    brand = dualBrandData.selectedBrand?.txVariant.orEmpty(),
                     configData = cardConfigDataGenerator.generateDualBrandConfigData(brandOptions),
                 )
                 analyticsManager.trackEvent(event)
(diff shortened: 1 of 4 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -306,14 +306,18 @@ class DefaultCardDelegate(
...
-            .distinctUntilChanged()
+        outputDataFlow
+            .map { it.dualBrandData?.takeIf { data -> data.selectable } }
+            .distinctUntilChanged { old, new ->
+                old?.brandOptions?.map { it.brand.txVariant } == new?.brandOptions?.map { it.brand.txVariant }
+            }
             .filterNotNull()
-            .map { brandOptions ->
+            .onEach { dualBrandData ->
+                val brandOptions = dualBrandData.brandOptions.map { it.brand.txVariant }
```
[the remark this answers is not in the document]
araratthehero: Hmm, this should not be a big issue, but I can see what I can do

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -306,14 +307,16 @@ class DefaultCardDelegate(
             }
             .launchIn(coroutineScope)
 
-        outputDataFlow.map { it.dualBrandData?.brandOptions?.map { it.brand.txVariant } }
-            .distinctUntilChanged()
+        outputDataFlow
+            .map { it.dualBrandData?.takeIf { data -> data.selectable } }
```
OscarSpruit, line 311: Wouldn't a filter be better here?
```kotlin
.filter { it.dualBrandData?.selectable == true }
```

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
```diff
@@ -306,14 +307,16 @@ class DefaultCardDelegate(
...
+            .distinctUntilChangedBy { it?.brandOptions?.map { option -> option.brand.txVariant } }
```
OscarSpruit, line 312: Maybe I'm missing something, but is it really needed to map here? I believe all these objects are `data class`es, so just comparing them should do to same trick

araratthehero: `DualBrandData.selectedBrand` (and `CardBrandItem.isSelected` inside `brandOptions`) change when the shopper picks a brand, so plain `distinctUntilChanged()` via data class equality would re-fire the displayed event on every selection, but that case is already covered by the selected event tracked just above. The `brandOptions.map { txVariant }` key eliminate duplicates and leaves the brands which are on screen.

card/src/main/java/com/adyen/checkout/card/internal/ui/DefaultCardDelegate.kt
araratthehero: If we use `filter` then the stream will be typed to `CardOutputData` and in the `onEach` block we need to access the `dualBrandData` from the outputData, which makes it nullable. In the end we need to ignore and not execute the onEach block content if it is null.

Do you have a solution in mind?

OscarSpruit: Okay thanks for explaining! In that case I think everything is fine. One small thing could be to use `mapNotNull` here and then the `filterNotNull` (2 lines down) is not needed

araratthehero: It won't actually work, since the `mapNotNull` will filter the null before `distinctUntilChangedBy`. 

What happens now is `map { takeIf }` → `distinctUntilChangedBy` → `filterNotNull` keeps null in the stream. It means when user removes the card number (the selector hides) and then enters characters again (the case of showing the dual branded selector again) we are sending analytics event. In the case of `filter` or `mapNotNull` the event won't be sent again when re-showing the selector.

araratthehero: If you want, I can leave a comment to explain why this is important.

OscarSpruit (approved):

PR #2737, opened 2026-04-29
v6 - Architecture - Secondary screens

OscarSpruit:
## Description
This PR introduces secondary screens for payment methods.

## Ticket Number
COSDK-1115

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
@@ -45,6 +45,7 @@ fun CheckoutPaymentFlow(
                     controller = controller,
                     onNavigate = { route ->
                         state = when (route) {
+                            // TODO - Move PaymentMethod case to onNavigate callback in secondary composable
                             CheckoutRoute.PaymentMethod -> CheckoutPaymentFlowState.PaymentMethod
                             CheckoutRoute.Action -> CheckoutPaymentFlowState.Action
                             is CheckoutRoute.Secondary -> CheckoutPaymentFlowState.Secondary(route.identifier)
--- a/core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
@@ -43,6 +43,7 @@ fun CheckoutPaymentFlow(
                 controller = controller,
                 onNavigate = { route ->
                     state = when (route) {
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentFlow.kt
```diff
@@ -32,37 +34,52 @@ fun CheckoutPaymentFlow(
...
+                CheckoutAction(
+                    controller = controller,
+                    modifier = modifier,
+                    theme = theme,
+                    localizationProvider = localizationProvider,
+                )
+            }
+
+            is CheckoutPaymentFlowState.Secondary -> {
+                CheckoutSecondary(
```
OscarSpruit, line 70: Add a TODO to add a `onNavigate` callback here that would comeback to the payment method composable

mbway/src/main/java/com/adyen/checkout/mbway/internal/ui/MBWayComponent.kt
```diff
@@ -72,13 +72,37 @@ internal class MBWayComponent(
...
 
+    @Composable
+    override fun SecondaryContent(identifier: String, modifier: Modifier) {
+        val viewState by viewState.collectAsStateWithLifecycle()
+
+        CountryCodePicker(
+            viewState = viewState,
+            onItemClick = {
+                onIntent(MBWayIntent.UpdateCountry(it))
+                eventChannel.trySend(PaymentComponentEvent.CloseSecondaryScreen)
```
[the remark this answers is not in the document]
OscarSpruit: It is safe in our use case to use `trySend`

ui/src/main/java/com/adyen/checkout/ui/internal/element/ValuePicker.kt
```diff
@@ -25,7 +27,7 @@ internal fun ValuePicker(
 ) {
     LazyColumn(
         verticalArrangement = Arrangement.spacedBy(Dimensions.Spacing.ExtraSmall),
-        modifier = modifier,
+        modifier = modifier.heightIn(max = LocalWindowInfo.current.containerDpSize.height),
```
araratthehero, line 30: Doesn't `fillMaxHeight` work here instead of reading it from the `LocalWindowInfo.current`?

OscarSpruit: Unfortunately not, `fillMaxHeight` will try to match the size of the parent, but if the parent doesn't restrict the height it will result in the infinite height error. The current approach sets a max height of the screen height, which is just a workaround that prevents the `LazyColumn` from trying to get an infinite height

araratthehero (approved):

PR #2749, opened 2026-05-11
v6 - Add card postal code field

tkuntubayev:
## Description
This PR introduces the "Postal Code" field to the Card Component, allowing merchants to collect the cardholder's postal code as part of the payment process. 

 Key changes include:
 - Configuration: Added `billingAddressMode` to `CardConfiguration`, enabling merchants to choose between `None` and `PostalCode` display modes (`Full` and `Lookup` are added for completeness but are commented out in this branch).
 - UI: Implemented `PostalCodeField` using Jetpack Compose and integrated it into the `CardComponent`.
 - Logic:
    - Updated `CardComponentState` and `CardViewState` to include postal code data.
    - Added validation logic for postal codes (limited to 10 characters and not empty) when requirementPolicy is `Required`.
    - Updated mappers and producers to handle the new field in the component's lifecycle.
- Localization: Added `card_postal_code` and related strings to the core module.
- Testing: Added comprehensive unit tests for state management, validation, and mapping logic.

    image
Initial `CardComponent` state when `PostalCode` `BillingAddressMode` requested

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1192


card/src/main/java/com/adyen/checkout/card/BillingAddressMode.kt
```diff
@@ -0,0 +1,43 @@
...
+    data class Full(
+        val supportedCountryCodes: Set<String> = emptySet(),
+        val hideForCardTypes: Set<CardType> = emptySet()
+    ) : BillingAddressMode()
+
+    /**
+     * Address Lookup option will be shown as part of card component.
+     */
+    @Parcelize
+    data object Lookup : BillingAddressMode() */
```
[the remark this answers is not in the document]
tkuntubayev: The team agreed to add the commented code for completentess since v6 code is under active development.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/PostalCodeField.kt
```diff
@@ -0,0 +1,115 @@
...
+}
+
+@Composable
+private fun PostalCodeIcon(
+    state: TextInputViewState,
+    modifier: Modifier = Modifier,
+) {
+    val isInvalid = state.trailingIcon == PostalCodeTrailingIcon.Warning
+    val resourceId = when (state.trailingIcon as? PostalCodeTrailingIcon) {
+        PostalCodeTrailingIcon.Warning -> com.adyen.checkout.test.R.drawable.ic_warning
```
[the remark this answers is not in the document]
tkuntubayev: The `ui` module and `ui-core` module use the same namespace, so that causes clashes. `ui-core` will be removed once we can get rid of v5 code and then we can give ui a proper namespace

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardComponentStateExt.kt
```diff
@@ -145,15 +147,27 @@ private fun createCardDetails(
...
 
+private fun CardComponentState.getBillingAddress() : Address? {
+    postalCode.getPaymentDataValue()?.let {
+        return Address(
+            postalCode = it
+        )
+    }
+
+    return null
+}
```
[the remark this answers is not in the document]
tkuntubayev: The `getPaymentDataValue()` implementation handles it with:
`text.takeIf { requirementPolicy != RequirementPolicy.Hidden && it.isNotBlank() }`

making this concern impossible to happen. Also good to mention that `PostalCode` is not meant to be optional in practice

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardValidationUtils.kt
```diff
@@ -189,6 +190,19 @@ internal object CardValidationUtils {
...
+    ): CardPostalCodeValidation {
+        if (!postalCode.requiresValidation()) return CardPostalCodeValidation.VALID
+        return if (postalCode.text.isBlank()) {
+            CardPostalCodeValidation.INVALID_BLANK
+        } else if (postalCode.text.length > PostalCodeProperties.POSTAL_CODE_MAX_LENGTH) {
+            CardPostalCodeValidation.INVALID_TOO_LONG
+        } else {
+            CardPostalCodeValidation.VALID
+        }
+    }
```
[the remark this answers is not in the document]
tkuntubayev: `requiresValidation()` cannot return `false` for optional field when too long input is provided, same as it cannot return `true` for blank optional field. It forces validation when optional TextInput is not blank, and omits it vice versa. 

```
fun TextInputComponentState.requiresValidation(): Boolean {
    return when (requirementPolicy) {
        RequirementPolicy.Hidden -> false
        RequirementPolicy.Optional -> text.isNotBlank()
        RequirementPolicy.Required -> true
    }
}
```

OscarSpruit (dismissed):

OscarSpruit (dismissed):

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParams.kt
```diff
@@ -23,6 +23,7 @@ data class CardComponentParams(
     val showStorePayment: Boolean,
     val socialSecurityNumberMode: FieldMode,
     val koreanAuthenticationMode: FieldMode,
+    val showPostalCode: Boolean,
```
jreij, line 26: I wonder if we should just use `FieldMode` everywhere internally for params that indicate the state of a field

card/src/main/java/com/adyen/checkout/card/internal/ui/properties/PostalCodeProperties.kt
```diff
@@ -0,0 +1,8 @@
+package com.adyen.checkout.core.common.internal.properties
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+object PostalCodeProperties {
```
jreij: This does not need to be in the `core` module for now, it can be internal in `card` (check out `SocialSecurityNumberProperties` for example). Maybe later we'll have to move it to a common module when we build the full address form but let's decide on that when we reach that point.

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParams.kt
tkuntubayev: It might be a good moment to align the types. We currently use 

`Boolean` for:
- `showHolderName`
- `showStorePayment`
- newly added `showPostalCode`

`FieldMode` for:
- socialSecurityNumberMode
- koreanAuthenticationMode

Custom visibility classes:
- `CVCVisibility` 
- `StoredCVCVisibility` (this one seems be duplicating `FieldMode` definitions)

I think `mode` implies for something more than just visibility and that can be a good option if we foresee new types that won't be only tied to visibility.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardValidationUtils.kt
```diff
@@ -218,6 +232,13 @@ enum class CardSecurityCodeValidation {
     INVALID,
 }
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jreij: I think this is not needed, the enum can be internal. Same for all the other enums in this file.

card/src/main/java/com/adyen/checkout/card/internal/ui/properties/PostalCodeProperties.kt
tkuntubayev: Makes sense, let's think if we'd need to have the same postal code parameters across other modules once we'll get there

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardValidationUtils.kt
tkuntubayev: Indeed, it's not used anywhere outside of the card module. I have replaced it with `internal` visibility modifier

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParams.kt
jreij: Then I suggest you add this as a topic for the next meeting and we can align all fields in a separate PR. For this PR feel free to keep your current solution and resolve this comment 👍

card/src/main/java/com/adyen/checkout/card/internal/ui/properties/PostalCodeProperties.kt
```diff
@@ -0,0 +1,8 @@
+package com.adyen.checkout.card.internal.ui.properties
+
+import androidx.annotation.RestrictTo
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
jreij: This class can be made internal now

tkuntubayev: Right, updated

card/src/main/java/com/adyen/checkout/card/internal/ui/model/CardComponentParams.kt
tkuntubayev: Added this topic to the next meeting's agenda 👍

jreij (approved):

PR #2756, opened 2026-05-15
v6 - Card brands - US debit cards (hidden cards)

araratthehero:
## Description

Hidden-brand handling alignment with Web/iOS.

Hidden brands (`accel`, `pulse`, `star`, `nyce` — US ATM/debit network rails) were stripped from `supportedCardBrands` client-side before the BIN lookup request was built. That caused the backend to return them as `supported=false`, the state handler to drop them, and merchants to never know the card had a hidden rail.

This PR moves hidden-brand handling from a client-side filter to the data-to-state classification boundary, matching Web/iOS behavior:

- `CardComponentParamsMapper.removeRestrictedCards()` is removed. Hidden brands flow through to the BIN lookup request.
- `RestrictedCardType` is renamed to `HiddenCardType`, moves from `internal.ui.model` to `internal.helper`, and is simplified from an enum to a `Set<String>` with a top-level `isHiddenCardType()` function.
- `DetectedCardType` gains an `isHidden: Boolean` flag, set by `NetworkCardBrandDetectionService` at construction time. This aligns with existing `isSupported` and `isShopperSelectionAllowedInDualBranded` flags and simplifies future backend migration.
- Two new `internal` `CardBrandState` variants:
  - `HiddenBrand` — only hidden brand(s) detected. UI mirrors `NoBrandsDetected`. No `brand` sent in `/payments`.
  - `SingleReliableWithHiddenBrand(cardBrandData)` — single supported brand alongside hidden brand(s) (e.g. `[visa, accel]`). Only the supported logo is shown; hidden logos are never rendered. No `brand` sent in `/payments` since the rail is ambiguous.
- Hidden brands are filtered from the supported card brands grid in `CardViewStateProducer`, so they never appear in the UI logo list.

All changes are on `internal` code.

### Behavior matrix

| Scenario | Detected | Logos shown | `brand` in payload | State |
|---|---|---|---|---|
| Standard single | `[visa]` | visa | yes | `SingleReliableBrand` (unchanged) |
| Standard dual | `[visa, mc]` | both | yes | `DualBrand` (unchanged) |
| Standard dual w/ selection | `[visa, cartebancaire]` | both | yes (selected) | `DualBrandWithShopperSelection` (unchanged) |
| **Dual with hidden** | `[visa, accel]` | visa only | **no** | **`SingleReliableWithHiddenBrand`** (new) |
| 3 brands with hidden | `[visa, cartebancaire, accel]` | visa + cartebancaire | yes (selected) | `DualBrandWithShopperSelection` (existing path) |
| **Hidden only** | `[accel]` | placeholder + grid | **no** | **`HiddenBrand`** (new) |
| Supported + unsupported hidden | `[visa(supported), accel(unsupported)]` | visa | yes | `SingleReliableBrand` (unsupported hidden ignored) |
| All unsupported | `[discover]` (merchant=`[visa]`) | error | n/a | `UnsupportedBrand` (unchanged) |
| No detection | `[]` | grid | n/a | `NoBrandsDetected` (unchanged) |

## Checklist

- [x] Code is unit tested

## Ticket Number

COSDK-1204


card/src/main/java/com/adyen/checkout/card/internal/helper/HiddenCardType.kt
jreij, line 1: Do you think using hidden instead of restricted could make this flow clearer?

araratthehero: I agree. Will make the change.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardBrandIntentsHandler.kt
```diff
@@ -76,23 +77,41 @@ internal class CardBrandIntentsHandler(
 
             DetectedCardTypeList.Source.NETWORK,
             DetectedCardTypeList.Source.CACHED -> {
+                val nonRestrictedSupportedBrands = supportedDetectedCardTypes.filterNot {
+                    isRestrictedCardType(it.cardBrand.txVariant)
```
jreij: What do you think of adding a boolean inside `DetectedCardType` that indicates if the brand should be hidden/restricted? IMO it achieves two things:

- It aligns with other similar values such as `isSupported` and `isShopperSelectionAllowedInDualBranded`
- When this value starts coming from the backend, it will be easy to migrate the logic, only inside `NetworkCardBrandDetectionService` (same will happen for `isShopperSelectionAllowedInDualBranded`)

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardBrandIntentsHandler.kt
```diff
@@ -76,23 +77,41 @@ internal class CardBrandIntentsHandler(
...
+                }
+                val anyRestrictedBrandDetected = supportedDetectedCardTypes.any {
+                    isRestrictedCardType(it.cardBrand.txVariant)
+                }
+
                 when {
                     // network detection + no detected brands
                     detectedCardTypes.isEmpty() -> CardBrandState.NoBrandsDetected
 
+                    // network detection + only restricted brands are supported
```
jreij: I think this when block is a bit difficult to read now, I had a similar issue when I was writing it in my PR and here's what I can suggest:

```kotlin
when {
    // network detection + no detected brands
    detectedCardTypes.isEmpty() -> CardBrandState.NoBrandsDetected

    nonRestrictedSupportedBrands.isEmpty() -> {
        if (anyRestrictedBrandDetected) {
            // network detection + only restricted brands are supported
            CardBrandState.RestrictedBrand
        } else {
            // network detection + detected brands but no supported brands
            CardBrandState.UnsupportedBrand
        }
    }

    nonRestrictedSupportedBrands.size == 1 -> {
        if (anyRestrictedBrandDetected) {
            // network detection + 1 non-restricted supported brand + restricted brand(s)
            CardBrandState.SingleReliableWithRestrictedBrand(
                cardBrandData = nonRestrictedSupportedBrands.first().toCardBrandData(),
            )
        } else {
            // network detection + 1 non-restricted supported brand
            CardBrandState.SingleReliableBrand(
                cardBrandData = nonRestrictedSupportedBrands.first().toCardBrandData(),
            )
        }
    }

    // network detection + multiple non-restricted supported brands
    else -> {
        getDualBrandedCardBrandState(currentState.cardBrandState, nonRestrictedSupportedBrands)
    }
}
```

Basically you make it a when with 4 cases:

- No brands at all
- No brands that are supported and not restricted
- 1 brand that is supported and not restricted
- More than 1 brand supported and not restricted

Then inside the blocks you check if anything is restricted. Let me know if you find this easier to read.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardComponentState.kt
```diff
@@ -46,6 +46,9 @@ internal sealed class CardBrandState {
         val cardBrandDataList: List<CardBrandData>,
         val shopperSelectedCardBrandData: CardBrandData,
     ) : CardBrandState()
+
+    data object RestrictedBrand : CardBrandState()
```
jreij: Nit: for readability consider ordering the values of this sealed class by: no brands - single brand -  dual brand instead of putting the cases with restricted at the bottom.

jreij (dismissed):

jreij (approved):

PR #2759, opened 2026-05-18
v6 - Card Scanning - Create new internal API

jreij:
## Description

Create a new Compose-friendly, library-internal API for card scanning in the `card-scanning` module.

- `CardScannerInitializer` — wraps Google Pay `PaymentsClient` initialization, returns `CardScannerController?`
- `CardScannerController` — provides `getIntentSender()`, `parseResult()`, and `terminate()`
- `CardScanResult` — data class with `pan`, `expiryMonth`, `expiryYear`
- All classes annotated with `@RestrictTo(LIBRARY_GROUP)`
- Added `core` module dependency for v6 `Environment` and logging

### Progress

✅ Phase 1 — [Move existing classes to old package](https://github.com/Adyen/adyen-android/pull/2758)
➡️ **Phase 2 — Create new internal API in card-scanning module**
Phase 3 — Add scanning state & intents to v6 card component state layer
Phase 4 — Add scanning analytics events
Phase 5 — Integrate scanning in v6 CardComponent (logic layer)
Phase 6 — Integrate scanning in v6 composable UI
Phase 7 — Tests

## Checklist
- [x] Code is unit tested

COSDK-1195

### DIFF
--- a/card-scanning/src/main/java/com/adyen/checkout/card/scanning/internal/CardScannerController.kt
+++ b/card-scanning/src/main/java/com/adyen/checkout/card/scanning/internal/CardScannerController.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 18/5/2026.
+ */
+
+package com.adyen.checkout.card.scanning.internal
+
+import android.app.PendingIntent
+import android.content.Intent
+import android.content.IntentSender
+import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.common.AdyenLogLevel
+import com.adyen.checkout.core.common.internal.helper.adyenLog
+import com.google.android.gms.wallet.PaymentCardRecognitionResult
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class CardScannerController internal constructor(
+    pendingIntent: PendingIntent,
+) {
+
+    private var paymentCardRecognitionPendingIntent: PendingIntent? = pendingIntent
+
+    fun getIntentSender(): IntentSender? {
+        if (paymentCardRecognitionPendingIntent == null) {
+            adyenLog(AdyenLogLevel.ERROR) { "CardScannerController has been terminated." }
+            return null
+        }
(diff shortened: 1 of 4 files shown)


card-scanning/src/main/java/com/adyen/checkout/card/scanning/internal/CardScannerInitializer.kt
```diff
@@ -0,0 +1,76 @@
...
+            val request = PaymentCardRecognitionIntentRequest.getDefaultInstance()
+            paymentsClient
+                .getPaymentCardRecognitionIntent(request)
+                .addOnSuccessListener { intentResponse ->
+                    if (continuation.isActive) {
+                        continuation.resume(
+                            CardScannerController(intentResponse.paymentCardRecognitionPendingIntent),
+                        )
+                    }
+                }
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist Good point. Made the constructor parameter non-nullable (`PendingIntent` instead of `PendingIntent?`), so the controller can never be in an invalid state. The `paymentCardRecognitionPendingIntent` from the Google API is already non-null, so no additional null check is needed in the initializer.

card-scanning/src/main/java/com/adyen/checkout/card/scanning/internal/CardScannerController.kt
```diff
@@ -0,0 +1,45 @@
...
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class CardScannerController internal constructor(
+    pendingIntent: PendingIntent,
+) {
+
+    private var paymentCardRecognitionPendingIntent: PendingIntent? = pendingIntent
+
+    fun getIntentSender(): IntentSender {
+        val pendingIntent = paymentCardRecognitionPendingIntent
+            ?: error("CardScannerController has been terminated.")
```
OscarSpruit: Maybe throwing an error here is too harsh. Let's see if we can fail gracefully

jreij: Good one! Changed `getIntentSender()` to return `IntentSender?` — it now logs an error and returns `null` instead of throwing when called after `terminate()`.

OscarSpruit (approved):

PR #2770, opened 2026-05-22
v6 - Improve internal param classes

OscarSpruit:
## Description
Refactors the internal component parameter architecture by replacing the `ComponentParams` interface / `CommonComponentParams` / `CommonComponentParamsMapper` / `ComponentParamsBundle` / `DropInOverrideParams` pattern with a simpler, centralized `CheckoutParams` data class and `CheckoutParamsFactory`.

### Motivation
The previous param structure had several issues:

- `ComponentParams` was an interface that each module's params class had to implement via delegation (by commonComponentParams), adding boilerplate.
- `CommonComponentParamsMapper` mixed mapping logic with session/drop-in override resolution, and returned a `ComponentParamsBundle` that bundled `SessionParams` alongside the common params.
- `DropInOverrideParams` was a separate data class carrying amount, session params, and submit button state, adding an extra layer of indirection.
- The `isCreatedByDropIn` flag leaked drop-in awareness into every component's params.

### Next
The `isCreatedByDropIn` flag was removed for simplicity. In a later stage similar functionality will be added.

## Checklist 
- [x] Code is unit tested


core/src/main/java/com/adyen/checkout/core/common/internal/CheckoutParams.kt
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Locale
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class CheckoutParams(
+    val shopperLocale: Locale,
+    val environment: Environment,
+    val clientKey: String,
+    val analyticsParams: AnalyticsParams,
+    val amount: Amount?,
+    val showSubmitButton: Boolean,
```
jreij, line 25: Can we add a TODO to ensure we enforce this as `true` with drop-in?

core/src/main/java/com/adyen/checkout/core/components/CheckoutConfiguration.kt
```diff
@@ -97,25 +97,12 @@ class CheckoutConfiguration(
     }
 
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun addConfiguration(key: String, configuration: Configuration) {
-        availableConfigurations[key] = configuration
-    }
-
-    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    fun addActionConfiguration(configuration: Configuration) {
+    fun addConfiguration(configuration: Configuration) {
```
jreij, line 100: Can we add a TODO to make sure we can support multiple components per configuration class if we need to? Like we do now in v5 with Instant.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -82,6 +83,7 @@ internal class CardComponent(
     private val onBinValueCallback: OnBinValueCallback?,
     private val onBinLookupCallback: OnBinLookupCallback?,
     private val cardScannerWrapper: CardScannerWrapper,
+    private val checkoutParams: CheckoutParams,
```
jreij: Here we pass the whole `checkoutParams` while in the other components we only pass the required fields (`clientKey` / `publicKey`). Can we use either approach for consistency?

After digging a bit deeper, I noticed that we don't really need those dependencies inside the component classes anyway. If at the factory level we inject the `clientKey` directly into `SubmitFingerprintRepository`, `NativeRedirectService`, `CardScannerWrapper` and create actual mapper classes instead of extension functions for `PaymentComponentState` and inject `publicKey` directly into them, we could remove `CheckoutParams` / `clientKey` / `publicKey` completely from the component classes. 

What are your thoughts? Could be in another PR but up to you

core/src/main/java/com/adyen/checkout/core/common/internal/CheckoutParams.kt
OscarSpruit: I am not sure if that is needed actually. In drop-in we might want to use a custom button to improve UX.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
OscarSpruit: In general, I think it's good practice to only pass what is necessary, but if all/most fields from `CheckoutParams` would be needed it would be better to pass the whole object. IMO optimizing for the specific situation of an object is more important than consistency. Forcing consistency can increase coupling and adds more boilerplate code when writing unit tests for example

Reasoning aside, it would make more sense to only pass the public key and environment here. I'll change it

core/src/main/java/com/adyen/checkout/core/common/internal/CheckoutParams.kt
jreij: Well then we might need to enforce it to `false`?

OscarSpruit: This is required for drop-in anyway, I don't think it's necessary to leave a comment

araratthehero (approved):

PR #2779, opened 2026-05-27
v6 - Action only flow

OscarSpruit:
## Description
This PR adds explicit support for the action only flow. Key components are the new `Checkout.setup` function, `CheckoutContext.ActionOnly` and `ActionOnlyCheckoutCallbacks`.

## Ticket Number
COSDK-1117

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
@@ -9,6 +9,7 @@
 package com.adyen.checkout.core.common
 
 import android.os.Parcelable
+import com.adyen.checkout.core.action.data.Action
 import com.adyen.checkout.core.components.CheckoutConfiguration
 import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethods
 import com.adyen.checkout.core.sessions.CheckoutSession
@@ -40,6 +41,7 @@ sealed interface CheckoutContext : Parcelable {
     @Parcelize
     @ConsistentCopyVisibility
     data class ActionOnly internal constructor(
+        internal val action: Action,
         override val checkoutConfiguration: CheckoutConfiguration,
         internal val checkoutAttemptId: String?,
         internal val publicKey: String?,
--- a/core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
@@ -32,4 +32,11 @@ sealed interface CheckoutContext : Parcelable {
         internal val checkoutAttemptId: String?,
         internal val publicKey: String?,
     ) : CheckoutContext
+
+    @Parcelize
+    data class ActionOnly internal constructor(
+        val checkoutConfiguration: CheckoutConfiguration,
+        internal val checkoutAttemptId: String?,
+        internal val publicKey: String?,
+    ) : CheckoutContext
 }
--- a/core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
@@ -17,25 +17,30 @@ import kotlinx.parcelize.Parcelize
 // TODO - Kdocs
 sealed interface CheckoutContext : Parcelable {
 
+    val checkoutConfiguration: CheckoutConfiguration
+
     @Parcelize
+    @ConsistentCopyVisibility
     data class Sessions internal constructor(
         val checkoutSession: CheckoutSession,
-        val checkoutConfiguration: CheckoutConfiguration,
+        override val checkoutConfiguration: CheckoutConfiguration,
         internal val checkoutAttemptId: String?,
         internal val publicKey: String?,
     ) : CheckoutContext
 
     @Parcelize
+    @ConsistentCopyVisibility
     data class Advanced internal constructor(
         val paymentMethods: PaymentMethods,
-        val checkoutConfiguration: CheckoutConfiguration,
+        override val checkoutConfiguration: CheckoutConfiguration,
         internal val checkoutAttemptId: String?,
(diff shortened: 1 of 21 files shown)


core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
```diff
@@ -17,18 +17,30 @@ import kotlinx.parcelize.Parcelize
...
+    val checkoutConfiguration: CheckoutConfiguration
+
     @Parcelize
+    @ConsistentCopyVisibility
     data class Sessions internal constructor(
         val checkoutSession: CheckoutSession,
-        val checkoutConfiguration: CheckoutConfiguration,
+        override val checkoutConfiguration: CheckoutConfiguration,
         internal val checkoutAttemptId: String?,
         internal val publicKey: String?,
```
araratthehero, line 29: These two values could also be moved to the `CheckoutContext` level and `CheckoutContextExt` could be removed.

core/src/main/java/com/adyen/checkout/core/components/Checkout.kt
```diff
@@ -72,6 +72,27 @@ object Checkout {
         )
     }
 
+    suspend fun setup(
+        configuration: CheckoutConfiguration,
+    ): Result<CheckoutContext.ActionOnly> {
```
araratthehero, line 79: This overload of the setup function looks a bit confusing. It is not as explicit as other overloads we have for the sessions and advanced flows. I am afraid merchants will call this function by mistake, since it requires only one parameter, which is common in all setup functions, while this setup call is specific for action only flow.

I am not sure what would be best to do here, but what do you think of naming this one `setupActionOnly` to make it more explicit for merchants when calling setup?

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModel.kt
```diff
@@ -100,6 +101,7 @@ internal class DropInViewModel(
         return when (this) {
             is CheckoutContext.Sessions -> checkoutConfiguration
             is CheckoutContext.Advanced -> checkoutConfiguration
+            is CheckoutContext.ActionOnly -> error("Unsupported context: $this")
         }
     }
```
araratthehero: Can we remove this completely and access the `checkoutConfiguration` as `context.checkoutConfiguration`? Since we moved the `checkoutConfiguration` field to the `CheckoutContext` class level.

core/src/test/java/com/adyen/checkout/core/components/internal/FullCheckoutFlowTest.kt
```diff
@@ -45,15 +38,15 @@ import org.mockito.junit.jupiter.MockitoExtension
 import org.mockito.kotlin.any
 import org.mockito.kotlin.doReturn
 import org.mockito.kotlin.whenever
-import java.util.Locale
 
+@OptIn(ExperimentalCoroutinesApi::class)
 @ExtendWith(MockitoExtension::class)
```
araratthehero, line 46: I would add tests for `ActionOnlyCheckoutFlow` too. We could test if the `actionHandler.handleAction` is being executed when the class is initialized.

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
```diff
@@ -29,30 +22,13 @@ import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.receiveAsFlow
 
-@Suppress("LongParameterList")
 internal class FullCheckoutFlow(
```
araratthehero, line 25: What do you think if we log a message when merchants call `ActionOnlyCheckoutFlow.submit()`? Now we just have a code no-op comment

core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
OscarSpruit: The reason I did it like this is that if we move those fields to `CheckoutContext` they will have to be public. Do you think it is okay to have them public?

core/src/main/java/com/adyen/checkout/core/components/Checkout.kt
OscarSpruit: Agreed! iOS has the function defined like this, so I did the same for the sake of alignment. I'll discuss with them and see what we can come up with

core/src/main/java/com/adyen/checkout/core/common/CheckoutContext.kt
araratthehero: Makes sense. I wouldn't make them public, since there is no reason to expose those.

core/src/main/java/com/adyen/checkout/core/components/Checkout.kt
erenbesel: This might be an oversight on our side. Indeed, the action only one looks a bit off compared to the other two. If I didn't miss this, I would probably only change `setup(with congiguration....` instead of `setup(configuration...`. But we can discuss more options

OscarSpruit: @erenbesel Kotlin doesn't have this argument label feature. We would either have to change the function name or the parameter name (`setupWith(...)` or `setup(with: CheckoutConfiguration)`). Both are not great 🤔

OscarSpruit: We decided to change it to the below, cause we think that gives the best guidance
```kotlin
suspend fun setup(
    action: Action,
    configuration: CheckoutConfiguration,
): Result<CheckoutContext.ActionOnly>
```

araratthehero (dismissed):

araratthehero (approved):

PR #2804, opened 2026-06-03
v6 - Minimize customization options

OscarSpruit:
## Description
This PR minimizes the UI customization options to only colors. The other options can easily added back later if there is demand for it. Also, KDocs are added to the public customization classes.

## Checklist 

### DIFF
--- a/ui/src/main/java/com/adyen/checkout/ui/internal/theme/InternalCheckoutTheme.kt
+++ b/ui/src/main/java/com/adyen/checkout/ui/internal/theme/InternalCheckoutTheme.kt
@@ -16,9 +16,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.staticCompositionLocalOf
 import com.adyen.checkout.ui.internal.element.InternalElements
 import com.adyen.checkout.ui.internal.text.InternalTextStyles
-import com.adyen.checkout.ui.theme.CheckoutAttributes
 import com.adyen.checkout.ui.theme.CheckoutColors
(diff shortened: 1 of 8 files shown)


ui/src/main/java/com/adyen/checkout/ui/internal/theme/InternalCheckoutTheme.kt
```diff
@@ -28,8 +26,8 @@ fun InternalCheckoutTheme(
     content: @Composable () -> Unit,
 ) {
     val colors = remember(theme.colors) { InternalColors.from(theme.colors) }
-    val textStyles = remember(theme.textStyles) { InternalTextStyles.from(theme.textStyles) }
-    val attributes = remember(theme.attributes) { InternalAttributes.from(CheckoutAttributes.default()) }
+    val textStyles = remember { InternalTextStyles.from(CheckoutTextStyles.default()) }
```
jreij, line 29: Out of curiosity: the previous implementation had `InternalTextStyles.from(theme.textStyles)` for this line and `InternalAttributes.from(CheckoutAttributes.default())` for the other one, was that a bug?

jreij (approved):

ui/src/main/java/com/adyen/checkout/ui/internal/theme/InternalCheckoutTheme.kt
OscarSpruit: Yes that was a bug 😅

PR #2809, opened 2026-06-08
v6 - Card - Brands - CardBrandViewState

araratthehero:
## Description

Introduce `CardBrandViewState` sealed class to replace raw brand lists in `CardViewState`. Replace `isAmex: Boolean` with a `CardNumberFormat` enum for better extensibility. Extract `CardBrand?.toCardNumberFormat()` extension for reuse across producers.

### Progress

✅ Phase 1 — [Remove old dual brand selector UI](https://github.com/Adyen/adyen-android/pull/2807)
➡️ **Phase 2 — Introduce `CardBrandViewState` and refactor `CardViewState` (this PR)**

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1253

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
@@ -9,7 +9,6 @@
 package com.adyen.checkout.card.internal.ui.state
 
 import com.adyen.checkout.core.common.CardBrand
-import com.adyen.checkout.core.common.CardType
 import com.adyen.checkout.core.components.internal.ui.state.ViewState
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputViewState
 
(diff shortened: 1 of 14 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
```diff
@@ -26,12 +25,9 @@ internal data class CardViewState(
     val isStorePaymentFieldVisible: Boolean,
     val supportedCardBrands: List<CardBrand>,
     val isSupportedCardBrandsShown: Boolean,
-    val detectedCardBrands: List<CardBrand>,
+    val cardBrandViewState: CardBrandViewState,
+    // TODO - Move isAmex to another layer if necessary/possible
+    val isAmex: Boolean,
```
jreij: Currently the only use of `isAmex` on the UI is inside `CardNumberOutputTransformation` to indicate where the spaces should be inserted when formatting the text field. While I feel like it's weird having a field named `isAmex` here (or anywhere) I feel like this is indeed where it belongs.

Maybe we could change it into something like a `CardNumberFormat` enum with `Default` and `Amex` as values? That way it doesn't read like _tech debt_ anymore and we could extend it in the future if we need to.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateProducer.kt
```diff
@@ -47,42 +47,75 @@ internal class CardViewStateProducer : ViewStateProducer<CardComponentState, Car
...
+                        brand = cardBrandData.cardBrand,
+                        isSelected = cardBrandData == cardBrandState.shopperSelectedCardBrandData,
+                    )
+                },
+            )
         }
     }
 
+    private fun getIsAmex(cardBrandState: CardBrandState): Boolean {
+        return when (cardBrandState) {
```
jreij: Nit: if you first get the `cardBrandData` in this `when`, then return `txVariant == CardType.AMERICAN_EXPRESS.txVariant` you wouldn't need to call `isAmex()` multiple times and you wouldn't even need to extract a function

jreij (approved):

PR #2810, opened 2026-06-08
v6 - Sessions - Final callback invocation

OscarSpruit:
## Description
This PR ensures the final callback is invoked when the payment request is successful without an action. In `SessionComponentRequestDispatcherTest` missing test cases were added and the tests are re-organized per function.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1252

### DIFF
--- a/core/src/test/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcherTest.kt
+++ b/core/src/test/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcherTest.kt
@@ -8,6 +8,7 @@
 
 package com.adyen.checkout.core.components.internal
 
+import com.adyen.checkout.core.action.data.Action
 import com.adyen.checkout.core.action.data.ActionComponentData
 import com.adyen.checkout.core.components.AdditionalDetailsResult
 import com.adyen.checkout.core.components.SessionCheckoutCallbacks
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcherTest.kt
```diff
@@ -16,74 +17,177 @@ import com.adyen.checkout.core.components.data.PaymentComponentData
...
 import org.mockito.kotlin.any
 import org.mockito.kotlin.doReturn
+import org.mockito.kotlin.mock
 import org.mockito.kotlin.whenever
 import java.io.IOException
 
 @ExtendWith(MockitoExtension::class)
 internal class SessionComponentRequestDispatcherTest(
-    @Mock private val sessionRepository: SessionRepository,
+    @param:Mock private val sessionRepository: SessionRepository,
```
araratthehero, line 37: Is the `@param` necessary here?

OscarSpruit: Currently, when applying annotations to constructor fields it will only be applied to the constructor param and not the backing field. The default will change soon (I believe in Kotlin 2.4.0) to apply to both. Now just using `@Mock` will still work fine then, but the annotation is not doing anything on the backing field. In short: for correctness I added that `param:`

araratthehero (approved):

PR #2811, opened 2026-06-08
v6 - Requires user interaction check

OscarSpruit:
## Description
This PR adds a function to `CheckoutController` which lets merchants check if a payment components needs users input before `submit()` can be called. This will be useful for components that have no UI (known as instant) or components with conditional UI.

## Ticket Number
COSDK-1119

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
@@ -59,4 +59,6 @@ class CheckoutController internal constructor(
     fun submit() {
         flow.submit()
     }
+
+    fun requiresUserInteraction(): Boolean = flow.requiresUserInteraction()
 }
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
```diff
@@ -95,6 +95,9 @@ internal class FullCheckoutFlow(
         paymentComponent?.submit()
     }
 
+    override fun requiresUserInteraction(): Boolean =
```
araratthehero, line 95: Could you add tests for this?

core/src/main/java/com/adyen/checkout/core/components/CheckoutController.kt
```diff
@@ -59,4 +59,6 @@ class CheckoutController internal constructor(
     fun submit() {
         flow.submit()
     }
+
+    fun requiresUserInteraction(): Boolean = flow.requiresUserInteraction()
```
araratthehero, line 83: I would add KDocs for the `requiresUserInteraction` and the `submit` functions

araratthehero (approved):

nauaros (commented):
Aligned with the new property in `CheckoutPaymentComponent` in iOS

PR #2816, opened 2026-06-09
v6 - Docs - Scaffold base migration and card guides

atmamont:
## Description
This PR introduces the first Android v6 documentation set for merchants who want to start a new integration or migrate from v5. It adds a v6 foundations guide, card configuration/session/advanced guides, theme documentation, migration notes, and README links, while trimming duplicated samples so the top-level guides stay focused on SDK instantiation.

This pass follows the structure and intent of the iOS documentation PR: [Adyen/adyen-ios#2568](https://github.com/Adyen/adyen-ios/pull/2568).

## To be addressed
- Clarify the real sessions success path: `SessionCheckoutCallbacks.onFinished` does not currently cover direct `SubmitResult.Completion` flows, so the docs should not imply it covers every successful session payment. Related Android PR: [#2810 v6 - Sessions - Final callback invocation](https://github.com/Adyen/adyen-android/pull/2810).
- Explain where final completion is handled in the advanced flow, including the current merchant responsibility around `SubmitResult` / `AdditionalDetailsResult`.
- Fix the localization guidance so it reflects current behavior: Compose currently renders with the device locale, while `CheckoutLocalizationProvider` can override selected strings.

## Checklist

## Ticket Number
COSDK-1268


docs/v6/README.md
```diff
@@ -0,0 +1,191 @@
...
+## Overview
+
+The public integration surface is centered around four concepts:
+
+- `Checkout.setup(...)` initializes the checkout flow.
+- `CheckoutConfiguration` is the shared configuration container.
+- `CheckoutController(...)` binds a checkout target to callbacks for either sessions or advanced flow.
+- `CheckoutPaymentFlow(...)` renders the Compose UI for the active checkout controller.
+
+Depending on the flow, `Checkout.setup(...)` returns either `CheckoutContext.Sessions` or `CheckoutContext.Advanced`.
```
OscarSpruit, line 25: There will also be `CheckoutContext.ActionOnly` soon. Should we document that as well or leave it out (as it's quite an advanced use case)?

docs/v6/theme.md
```diff
@@ -0,0 +1,73 @@
...
+import com.adyen.checkout.ui.theme.CheckoutColors
+import com.adyen.checkout.ui.theme.CheckoutTheme
+```
+
+## Apply a theme
+
+Create the theme in your UI layer and pass it to `CheckoutPaymentFlow(...)`:
+
+```kotlin
+val theme = CheckoutTheme(
```
OscarSpruit, line 21: To handle dark mode there are some extra steps to be done. Should we add those?

MIGRATION.md
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+}
+```
+
+For a complete card example, see [docs/v6/card-session-flow.md](docs/v6/card-session-flow.md).
+
+#### Advanced flow
+
+##### Before (v5)
```
OscarSpruit, line 68: What is supposed to be shown here exactly? The before section shows how to use callbacks on v5 and the after shows how to do the setup on v6.

docs/v6/README.md
atmamont: I left it out initially for iOS, so this one follows along. I tried keeping this small to establish a foundation for docs.
We can add it for the alpha02 scope

docs/v6/theme.md
atmamont: Sure, I will add dark/light mode specifics

MIGRATION.md
atmamont: This is the new file that Android repository didn't have before.
The target audience of this file are existing merchants that migrate from v5. We help them to get the idea from just comparing code snippets between v5 and v6.

But in this case the advanced flow code snippet is indeed very incomplete, let me expand it.

docs/v6/card-session-flow.md
```diff
@@ -0,0 +1,79 @@
...
+            showError(result.error.message.orEmpty())
+        }
+        is Checkout.Result.Success -> {
+            val controller = CheckoutController(
+                target = CheckoutTarget.PaymentMethod(PaymentMethodTypes.SCHEME),
+                context = result.checkoutContext,
+                callbacks = SessionCheckoutCallbacks(
+                    onFinished = {
+                        showSuccess()
+                    },
```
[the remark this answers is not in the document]
atmamont: @gemini-code-assist This will be addresses later if needed after linked PR is merged.

docs/v6/card-advanced-flow.md
```diff
@@ -0,0 +1,84 @@
...
+                controller = controller,
+                theme = theme,
+                localizationProvider = localizationProvider,
+            )
+        }
+    }
+}
+```
+
+`callPayments(...)` should return `SubmitResult`, and `callDetails(...)` should return `AdditionalDetailsResult`. `renderCheckout(...)` should render `CheckoutPaymentFlow(...)` from your Compose UI layer.
```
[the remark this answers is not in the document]
atmamont: Same

MIGRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+```kotlin
+val checkoutConfiguration = CheckoutConfiguration(
+    environment = environment,
+    shopperLocale = shopperLocale,
+    private fun setupCardView(
+        paymentMethod: PaymentMethod,
+        checkoutConfiguration: CheckoutConfiguration,
+    ) {
+}
```
araratthehero: Here we could show how merchants would pass configurations for card or threeds2, like we have mentioned inthe `v6` block. To keep the Configuration example identical.

MIGRATION.md
```diff
@@ -0,0 +1,219 @@
...
+        val cardComponent = CardComponent.PROVIDER.get(
+            activity = this,
+            paymentMethod = paymentMethod,
+            checkoutConfiguration = checkoutConfiguration,
+            callback = viewModel,
+        )
+
+        binding.cardView.attach(cardComponent, this)
+    }
+}
```
araratthehero, line 87: This part looks cleaner than the `v6` block. Could you make the `v6` block also look cleaner and for both of them to have the same set of callbacks (e.g. `v5` is missing `onError`)?

MIGRATION.md
```diff
@@ -0,0 +1,219 @@
...
+                    card()
+                },
+                coroutineScope = lifecycleScope,
+            )
+
+            renderCheckout(
+                controller = controller,
+                theme = theme,
+                localizationProvider = localizationProvider,
+            )
```
araratthehero: This should be done in Compose, could you make it clearer that this is compose part?

araratthehero: The `renderCheckout` is also not the best reference for this. You could use `CheckoutPaymentFlow` as mentioned in the Summary section.

MIGRATION.md
```diff
@@ -0,0 +1,219 @@
...
+- `CheckoutPaymentFlow(...)` becomes the main Compose rendering entry point.
+
+### Card component
+
+#### Configuration object
+
+##### Before (v5)
+
+```kotlin
+checkoutConfiguration.card {
```
araratthehero: Here you could also create the `CheckoutConfiguration` object as it is done for v6, to keep the code blocks consistent.

MIGRATION.md
```diff
@@ -0,0 +1,219 @@
...
+
+##### Before (v5)
+
+```kotlin
+val cardComponent = CardComponent.PROVIDER.get(
+    owner = activity,
+    paymentMethod = paymentMethod,
+    checkoutConfiguration = checkoutConfiguration,
+    callback = callback,
+)
```
araratthehero, line 218: I would also mention here that you need to attach the component to the view, like it is done in the `Advanced flow` section above. 
```
binding.cardView.attach(cardComponent, this)
```

docs/v6/theme.md
atmamont: Added a simple example for light/dark mode using default theme colors.

MIGRATION.md
atmamont: Refined the code and will do even more based on Ararat's comments.

MIGRATION.md
atmamont: Addressed in the latest commit: the v5 Sessions flow example now shows a full `CheckoutConfiguration(...)` with both `card { ... }` and `adyen3DS2 { ... }` so it mirrors the v6 snippet more closely.

MIGRATION.md
atmamont: Addressed in the latest commit: the v5 advanced-flow example now includes `onError(...)`, and the v6 example was expanded into `CardActivity` plus a dedicated Compose `CheckoutScreen` to make the structure and callback mapping clearer.

MIGRATION.md
atmamont: Addressed in the latest commit: the v6 advanced-flow example now makes the Compose boundary explicit and renders `CheckoutPaymentFlow(...)` directly from `CheckoutScreen` instead of referring to `renderCheckout(...)`.

MIGRATION.md
atmamont: Addressed in the latest commit: the v5 card configuration section now creates a full `CheckoutConfiguration(...)` object so the before/after code blocks stay consistent.

MIGRATION.md
atmamont: Addressed in the latest commit: the v5 rendering snippet now explicitly includes `binding.cardView.attach(cardComponent, activity)` after `CardComponent.PROVIDER.get(...)`.

docs/v6/README.md
```diff
@@ -0,0 +1,191 @@
...
+```kotlin
+val configuration = CheckoutConfiguration(
+    environment = Environment.TEST,
+    clientKey = clientKey,
+    shopperLocale = Locale.forLanguageTag("en-US"),
+) {
+    card(showCardholderName = true)
+    threeDS2(threeDSRequestorAppURL = "https://your-app.example/adyen")
+}
+
```
araratthehero: You already described the configuration in the block above. I would remove it from here.

docs/v6/README.md
```diff
@@ -0,0 +1,191 @@
...
+import com.adyen.checkout.core.components.CheckoutController
+import com.adyen.checkout.core.components.CheckoutPaymentFlow
+import com.adyen.checkout.core.components.data.model.Amount
+import com.adyen.checkout.threeds2.threeDS2
+import com.adyen.checkout.ui.theme.CheckoutTheme
+import java.util.Locale
+import kotlinx.coroutines.launch
+```
+
+## CheckoutConfiguration
```
araratthehero, line 53: What do you think to cover the callbacks in this doc too? The advanced and sessions callbacks.

docs/v6/README.md
```diff
@@ -0,0 +1,191 @@
...
+## Theme
+
+Use `CheckoutTheme` to customize the Compose checkout UI. For the full guide, see [theme.md](theme.md).
+
+## Localization
+
+Android v6 uses two separate concepts for localization:
+
+- `shopperLocale` on `CheckoutConfiguration` sets the shopper locale used by the checkout flow.
+- `CheckoutLocalizationProvider` lets you override specific checkout strings at render time.
```
araratthehero, line 149: I am not sure if we want to promote the option, but merchants can also override the strings through xml.

docs/v6/card-advanced-flow.md
```diff
@@ -0,0 +1,84 @@
...
+                        },
+                        onBinLookup = OnBinLookupCallback { data ->
+                            println("BIN lookup: $data")
+                        },
+                    )
+                },
+                coroutineScope = lifecycleScope,
+            )
+
+            renderCheckout(
```
araratthehero: Render is still used here. I would show the Compose integration example.

docs/v6/card-session-flow.md
```diff
@@ -0,0 +1,79 @@
...
+                        },
+                        onBinLookup = OnBinLookupCallback { data ->
+                            println("BIN lookup: $data")
+                        },
+                    )
+                },
+                coroutineScope = lifecycleScope,
+            )
+
+            renderCheckout(
```
araratthehero: Render is still used here. I would show the Compose integration example.

docs/v6/README.md
atmamont: I tried to avoid putting too much of code context into a README.MD and instead point to specific card examples, otherwise every time we have a code sample we have to do it again and again. This approach just could easily maintenance burden as we will have to keep all of them aligned upon future changes. Updating code samples in specific md files seems to be more scalable.

docs/v6/README.md
atmamont: This is a good example of a situation mentioned in [this comment](https://github.com/Adyen/adyen-android/pull/2816#discussion_r3395561895)

docs/v6/README.md
atmamont: Yes, as this belongs to the unified across mobile platforms API that we have control of. 
Early discussions mentioned there is no way to override Compose UI strings via xml, did anything change?

docs/v6/card-advanced-flow.md
atmamont: Good example of how we have code snippets maintenance debt adding up even within this PR scope!

docs/v6/README.md
araratthehero: I agree with your comment, but you included configuration in the example of Sessions and did not include it in the example of Advanced flow. I would prefer having either both of them mentioning or none. Considering that you already explained how to create configuration in the section above in the same file, this part is a duplication and could be removed. Removing it will make the file smaller too.

docs/v6/README.md
araratthehero: The translations strings are still defined in xml files, it does not block merchants to override them through xml. In fact it might be easier for them to do that, but if they do not want to use xml at all, they can use the Compose approach.

atmamont: Should I briefly mention this is still an option and link this from existing documentation to not dive too deep here? https://github.com/Adyen/adyen-android/blob/main/docs/UI_CUSTOMIZATION.md#overriding-string-resources

araratthehero: Yes, it is perfect!

atmamont: Addressed

araratthehero (dismissed):

OscarSpruit (dismissed):

araratthehero (dismissed):

araratthehero (approved):

PR #2819, opened 2026-06-11
v6 - Generic payment component

OscarSpruit:
## Description
This PR add the `GenericPaymentComponent`. This component can handle payment methods that don't require any user input. To use this component from the example implementation a button was added.

https://github.com/user-attachments/assets/39e4b6a5-6c57-40ce-b2ca-25606fea3fad

In the video you can see an example of the iDeal/Wero flow. When returning from the browser, it seems like the redirect component does not continue properly. This will be investigated and fixed in a follow up PR.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1119

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/ui/GenericPaymentComponentFactory.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/ui/GenericPaymentComponentFactory.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by oscars on 10/6/2026.
+ */
+
+package com.adyen.checkout.core.components.internal.ui
+
+import com.adyen.checkout.core.analytics.internal.AnalyticsManager
+import com.adyen.checkout.core.common.internal.CheckoutParams
+import com.adyen.checkout.core.components.CheckoutAdditionalCallback
+import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethod
+import com.adyen.checkout.core.components.internal.PaymentComponentFactory
+import com.adyen.checkout.core.components.internal.data.provider.DefaultSdkDataProvider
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/ui/GenericPaymentComponentFactory.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import com.adyen.checkout.core.analytics.internal.AnalyticsManager
+import com.adyen.checkout.core.common.internal.CheckoutParams
+import com.adyen.checkout.core.components.CheckoutAdditionalCallback
+import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethod
+import com.adyen.checkout.core.components.internal.PaymentComponentFactory
+import com.adyen.checkout.core.components.internal.data.provider.DefaultSdkDataProvider
+import kotlinx.coroutines.CoroutineScope
+
+internal class GenericPaymentComponentFactory : PaymentComponentFactory<GenericPaymentComponent> {
```
araratthehero: Since the `GenericPaymentComponentFactory` doesn't hold a state, could you make it an object?

core/src/test/java/com/adyen/checkout/core/components/internal/PaymentMethodProviderTest.kt
```diff
@@ -198,16 +201,30 @@ internal class PaymentMethodProviderTest {
...
+                coroutineScope = this,
+                analyticsManager = TestAnalyticsManager(),
+                params = generateCheckoutParams(),
+                additionalCallbacks = emptySet(),
+            )
+            assertInstanceOf<GenericPaymentComponent>(actualComponent)
+        }
+
+    @Test
+    fun `when get is called for an unregistered and unsupported payment method, then GenericPaymentComponent is returned`() =
```
araratthehero: Isn't this a better name for this test?

```suggestion
    fun `when get is called for an unregistered and unsupported payment method, then null is returned`() =
```

araratthehero (approved):

PR #2821, opened 2026-06-12
v6 - Redirect Component Intent Handling

OscarSpruit:
## Description
This PR adds automatic intent handling for the `RedirectComponent` and removes the need to externally catch and pass the intent.

**Changes**
- Remove `IntentHandlingComponent` interface — redirect no longer requires callers to forward intents manually
- Add `NewIntentHandler` composable — uses `addOnNewIntentListener` on the activity to automatically capture redirect results
- Inject `CoroutineScope` via constructor — instead of storing it as a mutable field
- Remove intent forwarding from example app — `V6Activity`, `V6SessionsActivity`, and their ViewModels no longer need to handle intents for redirects
- Add unit tests for `RedirectComponent` — 444 lines of new tests covering redirect, native redirect, error, and analytics scenarios
- Move old tests to old package — keeps legacy test files organized alongside deprecated code
- Move `TestAnalyticsManager` to test fixtures — makes it reusable across modules

**Demo**
https://github.com/user-attachments/assets/e6baae84-e790-41f0-8490-df49893f2968

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1119


core/src/main/java/com/adyen/checkout/core/components/internal/ActionHandler.kt
```diff
@@ -74,4 +76,8 @@ internal class ActionHandler(
             }
         }
     }
+
+    fun handleReturn(intent: Intent) {
+        (actionComponent as? ReturningActionComponent)?.handleReturn(intent)
```
araratthehero: What do you think of logging a message here, to track if there is no actionComponent and the handleReturn is called.

redirect/src/test/java/com/adyen/checkout/redirect/internal/ui/RedirectComponentTest.kt
```diff
@@ -0,0 +1,385 @@
...
+            // WHEN
+            component.handleAction()
+
+            // THEN
+            assertEquals("testNativeData", paymentDataRepository.nativeRedirectData)
+        }
+    }
+
+    @Nested
+    inner class OnNewIntentTest {
```
araratthehero: This could be renamed alongside the functions under it.

araratthehero (approved):

PR #2822, opened 2026-06-12
v6 - Sessions onComplete callback

OscarSpruit:
## Description
This PR renames the `onFinished` callback to `onComplete` and passes the agreed upon data object to it.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1252


core/src/main/java/com/adyen/checkout/core/sessions/internal/data/model/SessionDetailsResponse.kt
```diff
@@ -52,7 +52,7 @@ data class SessionDetailsResponse(
                 return SessionDetailsResponse(
                     sessionData = jsonObject.getStringOrNull(SESSION_DATA).orEmpty(),
                     status = jsonObject.getStringOrNull(STATUS),
-                    resultCode = jsonObject.getStringOrNull(RESULT_CODE),
+                    resultCode = jsonObject.getStringOrNull(RESULT_CODE).orEmpty(),
```
araratthehero: I think it is better to keep the `resultCode` as null and do a null check wherever the value is used. That way if necessary we will know when the value was present and when not.

core/src/main/java/com/adyen/checkout/core/sessions/internal/data/model/SessionPaymentsResponse.kt
```diff
@@ -53,7 +53,7 @@ data class SessionPaymentsResponse(
                 return SessionPaymentsResponse(
                     sessionData = jsonObject.getStringOrNull(SESSION_DATA).orEmpty(),
                     status = jsonObject.getStringOrNull(STATUS),
-                    resultCode = jsonObject.getStringOrNull(RESULT_CODE),
+                    resultCode = jsonObject.getStringOrNull(RESULT_CODE).orEmpty(),
```
araratthehero: Same here?

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
```diff
@@ -79,4 +93,8 @@ internal class SessionComponentRequestDispatcher(
     override fun error(error: CheckoutError) {
         callbacks.onError(error)
     }
+
+    companion object {
+        private const val RESULT_CODE_MISSING = "Unknown"
```
araratthehero, line 98: Is this something that other platforms can also implement?

araratthehero (dismissed):

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
OscarSpruit: iOS has `resultCode` as non-nullable, so they don't have to implement this

Robert-SD: In Flutter we also [define](https://github.com/Adyen/adyen-flutter/blob/19ba8c63facaccc2c97b67535821e6edda011e31/lib/src/common/model/result_code.dart#L15) it as the fallback.

araratthehero (approved):

PR #2824, opened 2026-06-15
v6 - Card Installments

tkuntubayev:
## Description
Implemented Installments feature for the Card component following [the aligned v6 API ](https://github.com/Adyen/adyen-checkout-sdk-meta/blob/develop/alignment/installments/INSTALLMENTS_CONFIGURATION.md)

## Checklist 
- [X] Code is unit tested
- [X] Changes are tested manually
- [X] Aligned public API changes with other platforms (if applicable)

## Ticket Number
COSDK-1194


card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentParams.kt
```diff
@@ -0,0 +1,70 @@
...
+    cardBrand: CardBrand? = null,
+): List<InstallmentModel> {
+    val hasOptionsForBrand = cardBrand != null &&
+        cardBasedOptions.containsKey(cardBrand)
+
+    val availableInstallmentOptions = when {
+        hasOptionsForBrand -> cardBasedOptions[cardBrand]
+        !defaultOptions?.values.isNullOrEmpty() -> defaultOptions
+        else -> null
+    }
```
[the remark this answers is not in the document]
tkuntubayev: The suggestion has been applied

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param values List of available installment counts (e.g. [2, 3, 6]).
+ * @param plans The plan types to offer. Defaults to [Plan.REGULAR] only.
+ * @param preselectedValue The installment count pre-selected in the UI.
+ */
+@Parcelize
+data class InstallmentOptions(
+    val values: List<Int>,
+    val plans: List<Plan> = listOf(Plan.REGULAR),
+    val preselectedValue: Int? = null,
```
jreij: If this is not set and the shopper does not select an installment, the component does not validate this and the payment can go through. Is this intended or should we show a validation error?

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentParams.kt
```diff
@@ -0,0 +1,70 @@
...
+            numberOfInstallments <= 0 -> null
+            else -> amount?.copy(value = amount.value / numberOfInstallments)
+        }
+        InstallmentModel(
+            plan = InstallmentPlan.REGULAR,
+            numberOfInstallments = numberOfInstallments,
+            amountPerInstallment = amountPerInstallment,
+            showAmount = showInstallmentAmount,
+        )
+    }
```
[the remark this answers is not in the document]
tkuntubayev: 1. The current implementation on all the 3 platforms (Android, iOS, Web) maps the values array to the "regular" installment plan, regardless if the "regular" plan is in the session/card configuration. This is an established way of mapping the installments.

2. The suggestion has been applied

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardBrandIntentsHandler.kt
```diff
@@ -226,4 +232,39 @@ internal class CardBrandIntentsHandler(
...
+        } else {
+            val installmentParams = componentParams.installmentParams
+            val preselectedValue = installmentParams?.cardBasedOptions?.get(cardBrand)
+                ?: installmentParams?.defaultOptions?.preselectedValue
+            updatedInstallmentOptions.firstOrNull {
+                it.plan == InstallmentPlan.REGULAR && it.numberOfInstallments == preselectedValue
+            }
+        }
+        return InstallmentState(updatedInstallmentOptions, selectedInstallment)
+    }
```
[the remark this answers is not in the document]
tkuntubayev: The suggestion has been applied

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardComponentStateFactory.kt
```diff
@@ -68,6 +75,21 @@ internal class CardComponentStateFactory(
...
+    private fun getPreselectedInstallment(installmentOptions: List<InstallmentModel>) : InstallmentModel? {
+        val preselectedNumberOfInstallments = componentParams.installmentParams
+            ?.defaultOptions
+            ?.preselectedValue
+            ?: return null
+
+        return installmentOptions.firstOrNull {
+            it.plan == InstallmentPlan.REGULAR && it.numberOfInstallments == preselectedNumberOfInstallments
+        }
+    }
```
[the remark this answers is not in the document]
tkuntubayev: The suggestion has been applied

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
tkuntubayev: If the `preselectedValue` isn't set, the selected installment falls back to the `One time payment` option which is an entry that doesn't get converted into an Installment

jreij: Ah ok! Then maybe the UI should reflect that? Currently the installments input is just empty when you load the component
| Component | Installments screen |
| -- | -- |
| Screenshot_1781597061 | Screenshot_1781597065 |

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardComponentStateExt.kt
```diff
@@ -167,14 +170,28 @@ private fun createPaymentComponentData(
...
 
+private fun CardComponentState.getInstallments(): Installments? {
+    return when (installmentState.selectedInstallment?.plan) {
+        InstallmentPlan.REGULAR, InstallmentPlan.REVOLVING ->
+            Installments(
+                plan = installmentState.selectedInstallment.plan.type,
+                value = installmentState.selectedInstallment.numberOfInstallments,
+            )
+
+        else -> null
```
jreij: Consider removing the `else` and using an exhaustive `when`. If we add a new value to the enum later, the compiler will force us to handle it instead of returning null which could cause bugs

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
```diff
@@ -0,0 +1,73 @@
...
+private fun SessionInstallmentOptionsParams.toParams() = InstallmentOptionsParams(
+    values = values ?: emptyList(),
+    plans = plans?.mapNotNull { it.toInstallmentPlan() } ?: listOf(InstallmentPlan.REGULAR),
+    preselectedValue = preselectedValue,
+)
+
+private fun String.toInstallmentPlan() = when (this) {
+    InstallmentPlan.REGULAR.type -> InstallmentPlan.REGULAR
+    InstallmentPlan.REVOLVING.type -> InstallmentPlan.REVOLVING
+    else -> null
```
jreij: Same here, consider replacing `else` with an exhaustive `when`

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardBrandIntentsHandler.kt
```diff
@@ -226,4 +232,39 @@ internal class CardBrandIntentsHandler(
...
+}
+
+private fun CardBrandState.asReliableCardBrand(): CardBrand? {
+    return when (this) {
+        is CardBrandState.SingleReliableBrand -> cardBrandData.cardBrand
+        is CardBrandState.SingleReliableWithHiddenBrand -> cardBrandData.cardBrand
+        is CardBrandState.DualBrand -> cardBrandDataList.first().cardBrand
+        is CardBrandState.DualBrandWithShopperSelection -> shopperSelectedCardBrandData.cardBrand
+        else -> null
```
jreij: Same here, consider replacing else with an exhaustive when

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardBrandIntentsHandler.kt
```diff
@@ -226,4 +232,39 @@ internal class CardBrandIntentsHandler(
...
+    }
+
+    private fun getUpdatedInstallmentOptions(cardBrand: CardBrand?): List<InstallmentModel> {
+        return componentParams.installmentParams?.mapToInstallmentModels(
+            cardBrand = cardBrand,
+        ) ?: emptyList()
+    }
+}
+
+private fun CardBrandState.asReliableCardBrand(): CardBrand? {
```
jreij: This same block exists in `CardBrandIntentsHandler.getUpdatedCardComponentState`, I think we should use the same function in both places or explicitly move the when block where it's used and rename the function

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
jreij: This seems to be fixed, resolving

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
tkuntubayev: The when is performed on `String` type, where that is compared against the enum's String variable. We could iterate on all enum entries until the enum variable and String match. This way all the new enums added in the future will be automatically used by this mapping. Please check the updated version

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardBrandIntentsHandler.kt
tkuntubayev: Nice catch! I have unified 2 blocks with having that as an extension function. Let me know if you'd have any other suggestions on that

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentModel.kt
```diff
@@ -0,0 +1,67 @@
...
+internal fun InstallmentModel.toDisplayText(): String {
+    return when (plan) {
+        InstallmentPlan.ONE_TIME ->
+            resolveString(CheckoutLocalizationKey.CARD_INSTALLMENTS_ONE_TIME)
+
+        InstallmentPlan.REVOLVING ->
+            resolveString(CheckoutLocalizationKey.CARD_INSTALLMENTS_REVOLVING)
+
+        InstallmentPlan.REGULAR -> {
+            val count = numberOfInstallments ?: 1
```
OscarSpruit: The default value of `1` feels like logic that shouldn't be in a composable. Would it be possible to make `numberOfInstallments` non-nullable? So this logic would be moved to the logic layer

Maybe a `sealed interface` like this would help
```kotlin
internal sealed interface InstallmentModel {
    data object OneTime : InstallmentModel
    data object Revolving : InstallmentModel
    data class Regular(
        val numberOfInstallments: Int,
        val amountPerInstallment: Amount?,
        val showAmount: Boolean,
    ) : InstallmentModel
}
```

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentModel.kt
```diff
@@ -0,0 +1,67 @@
...
+    return InstallmentModel(
+        plan = InstallmentPlan.REVOLVING,
+        numberOfInstallments = 1, // The number of installments for revolving is always 1
+        amountPerInstallment = null,
+        showAmount = false,
+    )
+}
+
+@Composable
+internal fun InstallmentModel.toDisplayText(): String {
```
OscarSpruit: I think it would be cleaner to move this function into it's own file. To keep the model and UI separated

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentModel.kt
```diff
@@ -0,0 +1,67 @@
...
+
+internal data class InstallmentModel(
+    val plan: InstallmentPlan,
+    val numberOfInstallments: Int?,
+    val amountPerInstallment: Amount?,
+    val showAmount: Boolean,
+)
+
+@Suppress("FunctionName")
+internal fun OneTimeInstallmentModel(): InstallmentModel {
```
OscarSpruit: Nit: since the linter is not agreeing with the naming I would suggest to do something like:
```kotlin
internal data class InstallmentModel(
    ...
) {
    companion object {
        internal fun oneTime(): InstallmentModel = ...
        internal fun revolving(): InstallmentModel = ...
    }
}
```

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentParams.kt
```diff
@@ -0,0 +1,72 @@
...
+    return availableInstallmentOptions
+        ?.mapToInstallmentModels(amount, showInstallmentAmount)
+        ?: emptyList()
+}
+
+private fun InstallmentOptionsParams.mapToInstallmentModels(
+    amount: Amount?,
+    showInstallmentAmount: Boolean,
+): List<InstallmentModel> {
+    val result = mutableListOf<InstallmentModel>()
```
OscarSpruit: Currently, a `MutableList` is returned, so if a consumer casts it back, they could modify this list. To prevent this you could use a list builder, this ensures a read-only list will be returned
```kotlin
return buildList {
    add(OneTimeInstallmentModel())

    if (plans.contains(InstallmentPlan.REVOLVING)) {
        add(RevolvingInstallmentModel())
    }

    values.mapTo(this) { numberOfInstallments ->
        val amountPerInstallment = when {
            numberOfInstallments <= 0 -> null
            else -> amount?.let {
                amount.copy(value = amount.value / numberOfInstallments)
            }
        }
        InstallmentModel(
            plan = InstallmentPlan.REGULAR,
            numberOfInstallments = numberOfInstallments,
            amountPerInstallment = amountPerInstallment,
            showAmount = showInstallmentAmount,
        )
    }
}
```

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
```diff
@@ -0,0 +1,62 @@
...
+ * Note: All values in [values] must be greater than 1.
+ *
+ * @param values List of available installment counts (e.g. [2, 3, 6]).
+ * @param plans The plan types to offer. Defaults to [Plan.REGULAR] only.
+ * @param preselectedValue The installment count pre-selected in the UI.
+ */
+@Parcelize
+data class InstallmentOptions(
+    val values: List<Int>,
+    val plans: List<Plan> = listOf(Plan.REGULAR),
```
OscarSpruit: Usually we don't set default values in the config classes, but I see we are doing this already on v5. Would it make sense to not set defaults here and move those to the mapping layer? @jreij what do you think?

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
```diff
@@ -0,0 +1,70 @@
...
+            options.toParams()
+        },
+        showInstallmentAmount = showInstallmentAmount,
+        amount = amount,
+    )
+}
+
+private fun SessionInstallmentOptionsParams.toParams() = InstallmentOptionsParams(
+    values = values ?: emptyList(),
+    plans = plans?.mapNotNull { it.toInstallmentPlan() } ?: listOf(InstallmentPlan.REGULAR),
```
OscarSpruit, line 52: There is a slight difference here with how we map the installment options from the configuration. Here we fallback to a list of just regular of we don't get anything from the response. And in the configuration mapping we do not fallback.

Also, `plans?.mapNotNull { it.toInstallmentPlan() }` could result in an empty list. What should we do then? Fallback to the default value or is the empty list okay?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentPicker.kt
```diff
@@ -0,0 +1,81 @@
...
+internal fun InstallmentPicker(
+    installmentOptions: List<InstallmentModel>,
+    selectedInstallment: InstallmentModel?,
+    onItemClick: (InstallmentModel) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Column(modifier = modifier) {
+        Body(resolveString(CheckoutLocalizationKey.CARD_INSTALLMENTS_SUBTITLE))
+        Spacer(Modifier.height(Dimensions.Spacing.Large))
+        Column(
```
OscarSpruit, line 38: You could replace this with the `ValuePicker` composable. However, that would probably need more mapping 🤔 I leave it up to you what's better

core/src/main/res/values/strings.xml
```diff
@@ -38,6 +38,13 @@
     <string name="checkout_card_kcp_password_invalid" translatable="false">Invalid password</string>
     <string name="checkout_card_postal_code" translatable="false">Postal code</string>
     <string name="checkout_card_postal_code_invalid" translatable="false">Invalid postal code</string>
+    <string name="checkout_card_installments" translatable="false">Installments</string>
+    <string name="checkout_card_installments_title" translatable="false">Number of installments</string>
+    <string name="checkout_card_installments_subtitle" translatable="false">Select the number of installments you would like to use</string>
+    <string name="checkout_card_installments_option_one_time" translatable="false">One time payment</string>
+    <string name="checkout_card_installments_option_revolving" translatable="false">Revolving payment</string>
+    <string name="checkout_card_installments_option_regular" translatable="false">%s months</string>
+    <string name="checkout_card_installments_option_regular_with_price" translatable="false">%1$sx %2$s</string>
```
OscarSpruit, line 47: Shouldn't there be a space in between? `%1$sx %2$s` -> `%1$s x %2$s`

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
jreij: Yes I agree, all defaults should be set internally and configs should be nullable (or required like `values` here)

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
```diff
@@ -29,4 +30,6 @@ internal data class CardViewState(
     val cardNumberFormat: CardNumberFormat,
     val isLoading: Boolean,
     val isCardScanButtonVisible: Boolean,
+    val installmentOptions: List<InstallmentModel>,
+    val selectedInstallment: InstallmentModel?,
```
jreij: Consider combining these two fields are they are strongly tied

core/src/main/res/values/strings.xml
tkuntubayev: This hasn't changed from v5 where the 'x' follows the `%1$s` without any spaces

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
tkuntubayev: Agreed, the variables are now placed under `InstallmentViewState`

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentsParamsMapper.kt
tkuntubayev: Indeed, the difference comes with the Session Params variables being nullable, while in the component configuration the plans list is non-nullable, hence doesn't require any fallback.

If that results in an empty list, the mapper function takes all values as for `Regular` InstallmentPlan. This is a unified logic on all 3 platforms.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
jreij: Looks good! One more comment but this is really nitpicky so only take it if you like it yourself.

When there are no installments, you now check `if(viewState.installmentViewState.installmentOptions.isNotEmpty())` which can be not super obvious.

Instead, if you know that there are no installments configured at all, you can make the whole `installmentViewState` null to avoid any doubts or confusion, and then you check `if(viewState.installmentViewState != null)`.

In addition, we use the same pattern with the text fields (we return null if it shouldn't be shown) so it would be consistent with them. The same logic can apply to both `InstallmentViewState` and `InstallmentState`.

I'll leave this for you to decide

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentModel.kt
tkuntubayev: The sealed interface fits well here, thanks for suggesting. That model is a sealed interface now

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentModel.kt
tkuntubayev: Moved to a new `InstallmentModelExtensions`

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentModel.kt
tkuntubayev: Since the model is now a sealed interface, this is no longer applicable

card/src/main/java/com/adyen/checkout/card/internal/ui/model/InstallmentParams.kt
tkuntubayev: Agree, this has been updated

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
tkuntubayev: Thanks, I'd prefer aligning with the existing established patterns, so the code would be consistent.
The mapping function now sets the installmentViewState to null if installmentOptions is empty

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
tkuntubayev: Good to know that this is our default approach. The Installment Configuration parameters no longer set default values

card/src/main/java/com/adyen/checkout/card/internal/ui/view/InstallmentPicker.kt
tkuntubayev: Thanks for sharing the option. After taking a closer look:

- The `ValuePicker` is internal in `ui` module, so would need to have RestrictTo annotation to be accessed in `card`
- The `ValuePickerItem` expects a non-null Subtitle, which is when provided as empty string leaves space, without centralizing the Title vertically properly. The option could be to make the `subtitle` in `ValuePickerItem` optional
- The mapping to `ValuePickerItem` and back adds boilerplate with no clear value
- `SelectableListValue` which is currently used directly is also used by `ValuePicker` internally - so updates to `SelectableListValue` would be reflected in both, so I see no advantage in maintainability as well

With the above I'd prefer keeping the existing implementation. Let me know how this sounds to you

OscarSpruit: Yes, let's keep what you have 👍

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
```diff
@@ -0,0 +1,62 @@
...
+ * to the shopper. [defaultOptions] and [cardBasedOptions] can be combined. In that case [InstallmentOptions]
+ * from [cardBasedOptions] will override the option defined in [defaultOptions].
+ *
+ * @param defaultOptions Options applied to all card brands.
+ * @param cardBasedOptions Brand-specific options. Overrides [defaultOptions] for matching brands.
+ * @param showInstallmentAmount Whether to show the per-installment amount.
+ */
+@Parcelize
+data class InstallmentConfiguration(
+    val defaultOptions: InstallmentOptions?,
```
OscarSpruit, line 28: What happens if you pass null here? Maybe it would be good to explain that in the KDocs

card/src/main/java/com/adyen/checkout/card/InstallmentConfiguration.kt
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param values List of available installment counts (e.g. [2, 3, 6]).
+ * @param plans The plan types to offer. Defaults to [Plan.REGULAR] only.
+ * @param preselectedValue The installment count pre-selected in the UI.
+ */
+@Parcelize
+data class InstallmentOptions(
+    val values: List<Int>,
+    val plans: List<Plan>,
+    val preselectedValue: Int?,
```
OscarSpruit, line 47: Same here - What happens if you pass null here? Maybe it would be good to explain that in the KDocs

OscarSpruit (dismissed):

jreij (dismissed):

OscarSpruit (approved):

PR #2827, opened 2026-06-16
v6 - Sessions onBeforeSubmit callback

Robert-SD:
## Description
This PR adds the `onBeforeSubmit` callback to the sessions flow. It allows merchants to update shopper data, provide patched `sessionData`, or abort the submission before the SDK submits the sessions `/payments` request.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-899


core/src/main/java/com/adyen/checkout/core/components/BeforeSubmitResult.kt
```diff
@@ -0,0 +1,44 @@
...
+        val sessionData: String? = null,
+    ) : BeforeSubmitResult()
+
+    /**
+     * Stop the submission flow and reset the component to the ready state.
+     *
+     * This does not call the error callback.
+     */
+    class Abort : BeforeSubmitResult()
+}
```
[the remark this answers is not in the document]
Robert-SD: Thanks for the suggestion. We intentionally keep this as an abstract class with an internal constructor because this is a merchant-facing result type that can be used in `when` expressions. Our Android SDK guidelines prefer this shape over `sealed` for new public result types so future cases can be added without breaking merchant source code.

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
```diff
@@ -79,4 +94,32 @@ internal class SessionComponentRequestDispatcher(
...
+
+        return try {
+            when (val result = callback(inputData)) {
+                is BeforeSubmitResult.Proceed -> {
+                    result.sessionData?.let { sessionData = it }
+                    data.applyBeforeSubmitData(result.data)
+                }
+                is BeforeSubmitResult.Abort -> null
+                else -> data
+            }
```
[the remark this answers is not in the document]
Robert-SD: Keeping the fallback is intentional while `BeforeSubmitResult` remains an abstract public result type. It preserves forward compatibility if we add another result case later; making this exhaustive would require using `sealed`, which we avoid for this merchant-facing API per our Android SDK guidelines.

core/src/main/java/com/adyen/checkout/core/components/data/PaymentComponentData.kt
```diff
@@ -132,3 +132,15 @@ data class PaymentComponentData<PaymentMethodDetailsT : PaymentMethodDetails>(
             }
     }
 }
+
+@Suppress("UNCHECKED_CAST")
```
OscarSpruit: This suppression is not needed

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
```diff
@@ -29,21 +33,19 @@ internal class SessionComponentRequestDispatcher(
     private var sessionData: String = initialSessionData
 
     override suspend fun submit(data: PaymentComponentData<*>): SubmitResult {
-        // TODO - Session patching
-        callbacks.beforeSubmit?.invoke(data)
-        sessionRepository.submitPayment(
+        val finalData = handleBeforeSubmit(data) ?: return SubmitResult.Retry()
```
OscarSpruit: This looks a bit obscure. I think it would be easier to read/understand if `handleBeforeSubmit` returns `BeforeSubmitResult` and then the check here becomes more explicit

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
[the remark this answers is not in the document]
OscarSpruit: I think it actually better to use a `sealed interface` here. Other result classes also use it, because we agreed some time ago that for incoming objects it's fine to use

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
```diff
@@ -54,11 +56,40 @@ internal class SessionComponentRequestDispatcher(
...
+        val callback = callbacks.onBeforeSubmit ?: return data
+
+        val inputData = BeforeSubmitData(
+            billingAddress = data.billingAddress,
+            deliveryAddress = data.deliveryAddress,
+            shopperName = data.shopperName,
+            shopperEmail = data.shopperEmail,
+        )
+
+        return try {
```
OscarSpruit: For other callbacks we decided not to do the error handling for merchants - why would we do it here?

example-app/src/main/java/com/adyen/checkout/example/data/api/model/SessionRequest.kt
```diff
@@ -34,4 +34,5 @@ data class SessionRequest(
     val installmentOptions: Map<String, SessionSetupInstallmentOptions>?,
     val showInstallmentAmount: Boolean,
     val showRemovePaymentMethodButton: Boolean,
+    val payable: Boolean = true,
```
OscarSpruit: We can remove the default value here, because in `RequestUtils` we already have it defined. Keeps all fields of this model consistent.

example-app/src/main/java/com/adyen/checkout/example/ui/v6/V6SessionsViewModel.kt
```diff
@@ -47,6 +51,8 @@ internal class V6SessionsViewModel @Inject constructor(
 ) : ViewModel() {
 
     private lateinit var checkoutContext: CheckoutContext.Sessions
+    private lateinit var sessionId: String
```
OscarSpruit: Because we already have `checkoutContext` in a field, these 2 are not necessary

core/src/main/java/com/adyen/checkout/core/components/internal/SessionComponentRequestDispatcher.kt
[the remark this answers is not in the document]
Robert-SD: Changed into `sealed interface` .

OscarSpruit (approved):

PR #2830, opened 2026-06-18
v6 - Error - Payment component cannot be created

araratthehero:
## Description
When a payment component cannot be created, `CheckoutControllerFactory` now notifies the merchant via the `onFailure` callback using the existing `CheckoutError.ErrorCode.PAYMENT_METHOD_FAILURE` code with a descriptive reason, instead of silently doing nothing (no UI, no-op `submit()`, zero feedback).

- Added `PaymentComponentResult` sealed class (`Success` / `Failure`) to carry the resolution outcome.
- Extracted target→component resolution into `PaymentComponentResolver`, covering all failure cases: no payment methods response, payment method / stored payment method not found, payment method / stored payment method not supported, and unknown `CheckoutTarget`.
- `createFullCheckoutController()` dispatches `CheckoutError(PAYMENT_METHOD_FAILURE, reason)` via the component request dispatcher on failure, and still passes `null` to `FullCheckoutFlow` (existing no-op behavior preserved).

No public API changes. No breaking changes.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1301

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutControllerFactory.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutControllerFactory.kt
@@ -24,11 +24,12 @@ import com.adyen.checkout.core.components.CheckoutController
 import com.adyen.checkout.core.components.CheckoutTarget
 import com.adyen.checkout.core.components.SessionCheckoutCallbacks
 import com.adyen.checkout.core.components.data.model.paymentmethod.PaymentMethods
-import com.adyen.checkout.core.components.internal.ui.PaymentComponent
+import com.adyen.checkout.core.error.CheckoutError
 import com.adyen.checkout.core.sessions.internal.data.api.SessionRepository
 import com.adyen.checkout.core.sessions.internal.data.api.SessionService
 import kotlinx.coroutines.CoroutineScope
 
+@Suppress("TooManyFunctions")
 internal class CheckoutControllerFactory {
 
     fun create(
@@ -101,7 +102,7 @@ internal class CheckoutControllerFactory {
             params = checkoutParams,
         )
 
-        val paymentComponent = createPaymentComponent(
+        val paymentComponentResult = createPaymentComponent(
             target = target,
             context = context,
             callbacks = callbacks,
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/adyen/checkout/core/components/internal/PaymentComponentResolverTest.kt
```diff
@@ -0,0 +1,231 @@
...
+        val result = resolve(
+            target = object : CheckoutTarget {},
+            paymentMethods = PaymentMethods(),
+        )
+
+        assertEquals(PaymentComponentResult.Failure("Unsupported checkout target."), result)
+    }
+
+    @Test
+    fun `when payment method is supported, then Success with the component is returned`() = runTest {
```
OscarSpruit, line 162: I think it would be good to have the same test, but for a stored payment method as well

core/src/test/java/com/adyen/checkout/core/components/internal/PaymentComponentResolverTest.kt
```diff
@@ -0,0 +1,231 @@
...
+        val result = resolve(
+            target = CheckoutTarget.PaymentMethod("scheme"),
+            paymentMethods = PaymentMethods(
+                paymentMethods = listOf(GenericPaymentMethod(type = "scheme", name = "Card")),
+            ),
+        )
+
+        assertEquals(PaymentComponentResult.Success(component), result)
+    }
+
```
OscarSpruit, line 175: Another test case that would be good is to pass `CheckoutContext.ActionOnly` to document what should happen in this case

core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutControllerFactory.kt
```diff
@@ -110,6 +110,19 @@ internal class CheckoutControllerFactory {
...
+            is PaymentComponentResult.Success -> paymentComponentResult.component
+            is PaymentComponentResult.Failure -> {
+                componentRequestDispatcher.failure(
+                    CheckoutError(
+                        code = CheckoutError.ErrorCode.PAYMENT_METHOD_FAILURE,
+                        message = paymentComponentResult.message,
+                    ),
+                )
+                null
+            }
```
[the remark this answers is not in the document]
araratthehero: No need for this, this is not a valid use-case and the suggestion is not gonna fix the issue.

OscarSpruit (approved):

PR #2834, opened 2026-06-22
Fix Editor crashes by gating flow observations with lifecycle

### LINKED ISSUE #2820: Crashes in card input fields since 5.16.0: IndexOutOfBoundsException in TextLine.measure and IllegalArgumentException in WordIterator — caused by view recreation skip (PR #2428)
### Summary

After upgrading from 5.14.0 to 5.19.0 we started receiving two new fatal crashes in production, both originating from the card input fields in Drop-in. Neither crash existed in our previous release on 5.14.0. We believe both are unmasked by the view-recreation skip introduced in #2428 (commit `688a37ab`, COSDK-794, released in 5.16.0).

### Environment

- Adyen Android SDK: 5.19.0 Drop-in integration (`drop-in-compose`, launched via `DropIn.startPayment`; `card` module only for `CardConfiguration`)
- Crashes observed in production via Crashlytics, multiple devices/OEMs
- Not reproducible on demand (race condition), but crash volume started exactly with the SDK bump

### Crash 1 — cursor anchor computed against stale layout

```
Fatal Exception: java.lang.IndexOutOfBoundsException: offset(1) should be less than line limit(0)
       at android.text.TextLine.measure(TextLine.java:574)
       at android.text.Layout.getHorizontal(Layout.java:1765)
       at android.text.Layout.getHorizontal(Layout.java:1741)
       at android.text.Layout.getPrimaryHorizontal(Layout.java:1711)
       at android.widget.TextView.getCursorAnchorInfo(TextView.java:14788)
       at android.widget.Editor$CursorAnchorInfoNotifier.updatePosition(Editor.java:4979)
       at android.widget.Editor$PositionListener.onPreDraw(Editor.java:3846)
       at android.view.ViewTreeObserver.dispatchOnPreDraw(ViewTreeObserver.java:1179)
       at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:4648)
```

### Crash 2 — word selection against rewritten text

```
Fatal Exception: java.lang.IllegalArgumentException: Invalid offset: 3. Valid range is [0, 0]
       at android.text.method.WordIterator.checkOffsetIsValid(WordIterator.java:401)
       at android.text.method.WordIterator.nextBoundary(WordIterator.java:125)
       at android.widget.Editor.getWordEnd(Editor.java:1311)
       at android.widget.Editor$SelectionModifierCursorController.onTouchEvent(Editor.java:8581)
       at android.widget.Editor.onTouchEvent(Editor.java:1983)
       at android.widget.TextView.onTouchEvent(TextView.java:13993)
       ...
       at android.app.Dialog.dispatchTouchEvent(Dialog.java:1402)   <- Drop-in bottom sheet
```

### Analysis

Both crashes are long-known Android framework bugs in `Editor`, triggered when an `EditText`'s text is replaced while the framework holds stale layout/selection state. They were previously masked by the SDK's own behavior, and #2428 removed that masking:

**Before 5.16.0:** `AdyenComponentView.attach()` collects `viewFlow` with `flowWithLifecycle`. Since `viewFlow` is a `StateFlow`, it re-emits on every `ON_START`, and each emission ran `removeAllViews()` + a fresh inflate. Every background/foreground cycle produced brand-new `EditText` instances, discarding any stale `Editor`/IME/layout state.

**Since 5.16.0:** when the same `ComponentViewType` is re-emitted, recreation is skipped and the `EditText` instances persist across stop/start cycles. Two problems follow:

1. The inner observers (e.g. `outputDataFlow` → `onCardNumberValidated` → `setText` when `rawValue != state.value`, [CardView.kt L303](https://github.com/Adyen/adyen-android/blob/5.19.0/card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardView.kt#L303)) are launched in `lifecycleOwner.lifecycleScope` and are **not** lifecycle-gated. They can call `setText` while the activity is stopped — no traversals run, so the field's `Layout` is left stale (0 lines).
2. On return to foreground, the IME reconnects to the *same* `EditText` and requests cursor-anchor monitoring. The first `onPreDraw` computes `getCursorAnchorInfo` against the stale layout → **Crash 1**.
3. The long-lived field plus the formatting `TextWatcher`s (card number/expiry rewriting content) means the selection controller can hold pre-rewrite offsets; a double-tap/long-press then calls `getWordEnd` with an out-of-range offset → **Crash 2**.

These crashes happen inside platform view dispatch — there is no way for the integrating app to catch or work around them.

### Suggested fix directions

- Gate the per-view observers with `flowWithLifecycle` (or `repeatOnLifecycle`) so text is never mutated while the window is stopped; and/or
- When skipping recreation in `AdyenComponentView`, force a layout re-sync on the input fields (e.g. `requestLayout()` and clamp/clear selection) before the IME reconnects.

## Steps to Reproduce (best effort)

1. Open Drop-in with the card component, start typing a card number.
2. Background the app, wait, return to foreground (Crash 1 window), or double-tap the card number field right after the formatter rewrites the text (Crash 2 window).
3. Race-dependent — easiest to observe at scale in production. Crash volume started immediately with the 5.14.0 → 5.19.0 upgrade.

## Integration Information
1. Client-side integration: Drop-in
2. SDK version: 5.19.0
3. Android version(s) / Device model(s) where issue occurs: 
Image

gourabsingha1:
Fixes #2820. Gates delegate flow observations in views with lifecycle to prevent UI mutations while the lifecycle is stopped.

### DIFF
--- a/bacs/src/main/java/com/adyen/checkout/bacs/internal/ui/view/BacsDirectDebitInputView.kt
+++ b/bacs/src/main/java/com/adyen/checkout/bacs/internal/ui/view/BacsDirectDebitInputView.kt
@@ -35,6 +35,7 @@ import com.adyen.checkout.ui.core.old.internal.util.showError
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
+import com.adyen.checkout.ui.core.old.internal.util.collectWithLifecycle
 import com.adyen.checkout.ui.core.R as UICoreR
 
 @Suppress("TooManyFunctions")
@@ -161,8 +162,7 @@ internal class BacsDirectDebitInputView @JvmOverloads constructor(
 
     private fun observeDelegate(delegate: BacsDirectDebitDelegate, coroutineScope: CoroutineScope) {
         delegate.outputDataFlow
-            .onEach { outputDataChanged(it) }
-            .launchIn(coroutineScope)
+            .collectWithLifecycle(this, coroutineScope) { outputDataChanged(it) }
     }
 
     private fun outputDataChanged(bacsDirectDebitOutputData: BacsDirectDebitOutputData) {
--- a/card/src/main/java/com/adyen/checkout/card/old/internal/ui/view/CardView.kt
+++ b/card/src/main/java/com/adyen/checkout/card/old/internal/ui/view/CardView.kt
@@ -54,6 +54,7 @@ import com.adyen.checkout.ui.core.old.internal.util.showError
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
+import com.adyen.checkout.ui.core.old.internal.util.collectWithLifecycle
 import com.adyen.checkout.ui.core.R as UICoreR
 
 /**
@@ -192,8 +193,7 @@ class CardView @JvmOverloads constructor(
 
     private fun observeDelegate(delegate: CardDelegate, coroutineScope: CoroutineScope) {
         delegate.outputDataFlow
-            .onEach { outputDataChanged(it) }
-            .launchIn(coroutineScope)
+            .collectWithLifecycle(this, coroutineScope) { outputDataChanged(it) }
     }
 
     private fun outputDataChanged(cardOutputData: CardOutputData) {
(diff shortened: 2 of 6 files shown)


OscarSpruit:
Hi @gourabsingha1, thank you for taking the time for this contribution! We appreciate the enthusiasm, but our team is already looking into this. Since we have specific requirements and context around this change, we'll be handling it ourselves.

We're closing this PR, but appreciate you flagging the topic. If you'd like to contribute in the future, feel free to open an issue first or comment on the existing issue so we can coordinate and avoid duplicate effort.

The same goes for #2835

PR #2838, opened 2026-06-23
e2e example - Add runtime SDK configuration injection for e2e tests

Robert-SD:
## Summary
- Add `ExternalConfigurationReader.kt` to read a Base64-encoded JSON config from the `config` intent extra at app launch
- Extend `KeyValueStorage` with `showCardholderName` support (persisted via SharedPreferences)
- Wire up `MainActivity` to read the intent extra and apply the external config before UI creation
- Update `CheckoutConfigurationProvider` to apply the external `showCardholderName` value

This enables the e2e test framework to inject SDK configuration (e.g., `showCardholderName: true`) at launch time via `adb shell am start --es config '<base64>'`, without rebuilding the app.

### JSON Schema
```json
{
  "CARD_CONFIGURATION": {
    "showCardholderName": true
  }
}
```

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
@@ -6,10 +6,7 @@
 
 package com.adyen.checkout.example.data.storage
 
-import com.squareup.moshi.Json
-import com.squareup.moshi.Moshi
-import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
-import java.util.Base64 as JavaBase64
+import org.json.JSONObject
 import javax.inject.Inject
 import javax.inject.Singleton
 
@@ -20,13 +17,14 @@ import javax.inject.Singleton
  *
  * The JSON schema uses unified keys aligned with the native SDKs (e.g. `showCardholderName`).
  * All fields are optional — omitted fields leave the stored default unchanged.
+ *
+ * When [configBase64] is null or empty (normal app launch without test extras), the persisted
+ * configuration is reset to defaults to prevent test pollution from previous e2e runs.
  */
 @Singleton
 class ExternalConfigurationReader @Inject constructor(
     private val keyValueStorage: KeyValueStorage,
 ) {
-
-    private val moshi: Moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
 
     /**
      * Decodes the Base64-encoded JSON config and applies it to storage.
@@ -34,14 +32,22 @@ class ExternalConfigurationReader @Inject constructor(
      * @param configBase64 Base64-encoded JSON string, or null if no config was passed.
      */
     fun apply(configBase64: String?) {
-        if (configBase64.isNullOrEmpty()) return
+        if (configBase64.isNullOrEmpty()) {
+            // Reset to defaults to prevent test pollution from previous e2e runs
+            keyValueStorage.setShowCardholderName(SettingsDefaults.SHOW_CARDHOLDER_NAME)
+            return
+        }
 
+        decodeAndApply(configBase64)
+    }
+
+    private fun decodeAndApply(configBase64: String) {
         val json = runCatching {
-            String(JavaBase64.getDecoder().decode(configBase64))
+            String(android.util.Base64.decode(configBase64, android.util.Base64.DEFAULT), Charsets.UTF_8)
         }.getOrNull() ?: return
 
         val config = runCatching {
-            moshi.adapter(ExternalConfiguration::class.java).fromJson(json)
+            ExternalConfiguration.fromJson(json)
         }.getOrNull() ?: return
 
         config.card?.let { applyCardConfiguration(it) }
@@ -53,9 +59,35 @@ class ExternalConfigurationReader @Inject constructor(
 }
 
 internal data class ExternalConfiguration(
-    @Json(name = "CARD_CONFIGURATION") val card: ExternalCardConfiguration? = null,
-)
+    val card: ExternalCardConfiguration? = null,
+) {
+    companion object {
+        private const val KEY_CARD_CONFIGURATION = "CARD_CONFIGURATION"
+
+        fun fromJson(json: String): ExternalConfiguration {
+            val root = JSONObject(json)
+            val cardJson = root.optJSONObject(KEY_CARD_CONFIGURATION)
+            return ExternalConfiguration(
+                card = cardJson?.let { ExternalCardConfiguration.fromJson(it) },
+            )
+        }
+    }
+}
 
 internal data class ExternalCardConfiguration(
-    @Json(name = "showCardholderName") val showCardholderName: Boolean? = null,
-)
+    val showCardholderName: Boolean? = null,
+) {
+    companion object {
+        private const val KEY_SHOW_CARDHOLDER_NAME = "showCardholderName"
+
+        fun fromJson(json: JSONObject): ExternalCardConfiguration {
+            return ExternalCardConfiguration(
+                showCardholderName = if (json.has(KEY_SHOW_CARDHOLDER_NAME)) {
+                    json.getBoolean(KEY_SHOW_CARDHOLDER_NAME)
+                } else {
+                    null
+                },
+            )
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
```diff
@@ -0,0 +1,61 @@
...
+     * Decodes the Base64-encoded JSON config and applies it to storage.
+     *
+     * @param configBase64 Base64-encoded JSON string, or null if no config was passed.
+     */
+    fun apply(configBase64: String?) {
+        if (configBase64.isNullOrEmpty()) return
+
+        val json = runCatching {
+            String(JavaBase64.getDecoder().decode(configBase64))
+        }.getOrNull() ?: return
```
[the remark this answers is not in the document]
Robert-SD: Fixed in latest commit. Thanks for the review!

example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    private val moshi: Moshi = Moshi.Builder().add(KotlinJsonAdapterFactory()).build()
+
+    /**
+     * Decodes the Base64-encoded JSON config and applies it to storage.
+     *
+     * @param configBase64 Base64-encoded JSON string, or null if no config was passed.
+     */
+    fun apply(configBase64: String?) {
+        if (configBase64.isNullOrEmpty()) return
```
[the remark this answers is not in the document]
Robert-SD: Fixed in latest commit. Thanks for the review!

example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+internal data class ExternalConfiguration(
+    @Json(name = "CARD_CONFIGURATION") val card: ExternalCardConfiguration? = null,
+)
+
+internal data class ExternalCardConfiguration(
+    @Json(name = "showCardholderName") val showCardholderName: Boolean? = null,
+)
```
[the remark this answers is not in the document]
Robert-SD: Fixed in latest commit. Thanks for the review!

example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
```diff
@@ -0,0 +1,90 @@
...
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ */
+
+package com.adyen.checkout.example.data.storage
+
+import org.json.JSONObject
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
```
Robert-SD, line 13: I considered a few options for decoding the Base64 configuration string.
 
- **`android.util.Base64`** — works on all Android API levels, but it is Android-framework code that cannot be called from plain JVM unit tests. Using it would force this logic into instrumented or Robolectric tests, which is heavier and slower for a simple decoder.
- **`java.util.Base64`** — works on the JVM, but is only available from Android API 26 / Java 8. The project `minSdk` is 23 and core library desugaring is not enabled, so using it would crash on older devices.
- **Third-party Base64 library** — possible, but adding a new dependency just to decode a Base64 string is overkill.
 
`kotlin.io.encoding.Base64` is part of the Kotlin stdlib, works on all Android API levels, works in JVM unit tests, and requires no extra dependency. The `@OptIn(ExperimentalEncodingApi::class)` annotation is only an opt-in for a still-experimental API surface; the decode behavior itself is stable and fully functional.

example-app/src/main/java/com/adyen/checkout/example/data/storage/KeyValueStorage.kt
```diff
@@ -65,6 +66,8 @@ interface KeyValueStorage {
     fun setAnalyticsMode(analyticsMode: AnalyticsMode)
     fun getOnBeforeSubmitMode(): OnBeforeSubmitMode
     fun setOnBeforeSubmitMode(onBeforeSubmitMode: OnBeforeSubmitMode)
+    fun isShowCardholderName(): Boolean
+    fun setShowCardholderName(showCardholderName: Boolean)
```
araratthehero: Nits: What do you think if we rename this to `isCardHolderNameShown`, `setCardholderNameShown` to read more naturally. `isShowCardholderName` reads a bit off.

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainActivity.kt
```diff
@@ -266,5 +275,6 @@ class MainActivity : AppCompatActivity() {
         private val TAG = getLogTag()
 
         internal const val RETURN_URL_EXTRA = "RETURN_URL_EXTRA"
+        internal const val CONFIG_EXTRA = "config"
```
araratthehero, line 281: Nits: Could you keep these two values consistent, uppercase or lowercase?

example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
OscarSpruit: It's fine to use indeed. We use it internally as well

example-app/src/main/java/com/adyen/checkout/example/data/storage/ExternalConfigurationReader.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        val config = runCatching {
+            ExternalConfiguration.fromJson(json)
+        }.getOrNull() ?: return
+
+        config.card?.let { applyCardConfiguration(it) }
+    }
+
+    private fun applyCardConfiguration(card: ExternalCardConfiguration) {
+        card.showCardholderName?.let { keyValueStorage.setShowCardholderName(it) }
```
Robert-SD: Instead of writing the value to the settings, we should use constructor injection through memory.

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -79,6 +79,7 @@ internal class CheckoutConfigurationProvider @Inject constructor(
 
             card {
                 setShopperReference(keyValueStorage.getShopperReference())
+                setHolderNameRequired(keyValueStorage.isShowCardholderName())
```
Robert-SD: Pass it via constructor injection.

example-app/src/main/java/com/adyen/checkout/example/ui/main/MainActivity.kt
Robert-SD: Good catch! I've checked and I can't change CONFIG_EXTRA's value because the e2e testing framework already hardcodes it as lowercase "config" when launching the app via --es config <base64>, so renaming it would break that integration. We can discuss this further but then have to adjust all SDKs.

araratthehero: I see, no need then. It is trivial.

example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -56,6 +59,12 @@ internal class CheckoutConfigurationProvider @Inject constructor(
             return shopperLocaleString?.let { Locale.forLanguageTag(it) }
         }
 
+    // Overridable at runtime via ExternalConfigurationReader (e.g. by the e2e test framework),
+    // falling back to the default value used across the example app.
+    private val showCardholderName: Boolean
+        get() = externalConfigurationReader.cardConfiguration?.showCardholderName
+            ?: SettingsDefaults.SHOW_CARDHOLDER_NAME
```
araratthehero, line 66: Since we are not showing the holder name toggle in the settings I would remove the `SettingsDefaults.SHOW_CARDHOLDER_NAME` value and pass null to the `showCardholderName`.

For `setHolderNameRequired` for the old configuration, we could just pass false if null. It will be removed in the future.

araratthehero (approved):

PR #2837, opened 2026-06-23
v6 - Google Pay - Component and factory wiring

araratthehero:
## Description
Phase 1 of the Google Pay v6 migration: assemble the component and factory on top of the existing state machine.

- Implement `GooglePayComponent` against `PaymentComponent`: holds the `ComponentStateFlow` + derived view state, initializes analytics, and implements `setLoading()`, `requiresUserInteraction()` (= `true`), and `onCleared()`. `submit()` and `Content` are placeholders — the real sheet launch + result handling land in Phase 2, and the button UI in Phase 4.
- Implement `GooglePayFactory`: builds params via `GooglePayComponentParamsMapper`, and wires `analyticsManager`, `DefaultSdkDataProvider`, `paymentMethodType`, and the state factory/reducer/validator/view-state producer.
- `componentParams` is carried into the component now (forward-looking wiring) and is consumed in Phases 2 & 3.
- Start the component with the Google Pay button hidden (`isButtonVisible = false`) to avoid a flash before the `isReadyToPay` availability check (Phase 3).

### Progress
➡️ **Phase 1 — Component + factory wiring (this PR)**
Phase 2 — Sheet launch + result handling ([#2839](https://github.com/Adyen/adyen-android/pull/2839))
Phase 3 — Availability (`isReadyToPay`) gating
Phase 4 — Google Pay button UI
Phase 5 — Example app integration & on-device verification

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1309

PR #2828, opened 2026-06-17
v6 - Card / MBWay - Extract composable logic from components

jreij:
## Description
Extracts composable logic from `CardComponent` and `MBWayComponent` into dedicated composable files, keeping component classes free of Compose internals.

Three logical steps:
1. **Rename** `view/CardComponent.kt` → `view/CardContent.kt` to free up the name for the composable entry point.
2. **Move compose logic** — scanner launcher setup, `LaunchedEffect`, and `collectAsStateWithLifecycle` calls are moved from the component classes into the `CardContent`/`MBWayContent` (new `StateFlow` overload) and new `CardSecondaryContent`/`MBWaySecondaryContent` composables. Component classes now simply delegate.
3. **Typed secondary screen entry objects** — replace `private const val` identifiers in companion objects with `CardSecondaryContentEntry` / `MBWaySecondaryContentEntry` objects, co-located with the composables that use them.

## Checklist
- [x] Changes are tested manually

PR #2839, opened 2026-06-23
v6 - Google Pay - Sheet launch and result handling

araratthehero:
## Description
Assembles the Google Pay sheet launch and result handling for the v6 component.

- `submit()` sets loading and emits a `GooglePayViewEvent.Pay`; it no longer emits `Submit` directly.
- A new `googlePayEvent` effect-composable (mirroring `redirectEvent`) owns the `PaymentsClient` and the `GetPaymentDataResult` activity-result launcher, and launches the sheet once the `loadPaymentData` task completes (`addOnCompleteListener(launcher::launch)`, per Google's docs).
- `onPaymentResult` maps the `ApiTaskResult`: success → `PaymentComponentEvent.Submit` (+ submit analytics), cancel → idle (reset loading), error/null-data → `PaymentComponentEvent.Error` (+ third-party error analytics).

Also includes a small `core` fix: `redirectEvent` keyed its `LaunchedEffect` on the unstable `onError` method reference, which could restart the collector on recomposition; it now keys only on the stable `viewEventFlow`.

### Progress
✅ Phase 1 — Component & factory wiring ([#2837](https://github.com/Adyen/adyen-android/pull/2837))
➡️ **Phase 2 — Sheet launch + result handling (this PR)**
Phase 3 — Availability (`isReadyToPay`) gating
Phase 4 — Google Pay button UI
Phase 5 — Example app integration + on-device verification

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1309


googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayEvent.kt
```diff
@@ -0,0 +1,49 @@
...
+                GooglePayViewEvent.Pay -> {
+                    val task = paymentsClient.loadPaymentData(
+                        GooglePayUtils.createPaymentDataRequest(componentParams),
+                    )
+                    task.addOnCompleteListener(launcher::launch)
+                }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
araratthehero: Thanks! Addressed in faa3f17.

Lifecycle safety: instead of an isActive guard inside the listener, we now await the task completion inside the coroutine via a structured Task.awaitTask() (suspendCancellableCoroutine) and then launch. If the composable is disposed, the LaunchedEffect coroutine is cancelled and execution never reaches launcher.launch, so cancellation is handled structurally. This also mirrors the existing awaitTask() precedent in the module.

Stale captures: componentParams is a component val fixed at construction and paymentsClient is remembered from it, so both are stable for the component lifetime and cannot go stale here. We intentionally key the effect on the stable viewEventFlow only, to avoid unnecessary restarts.

core/src/main/java/com/adyen/checkout/core/redirect/internal/ui/RedirectEvent.kt
```diff
@@ -28,7 +28,7 @@ fun redirectEvent(
     onError: (InternalCheckoutError) -> Unit
 ) {
     val context = LocalContext.current
-    LaunchedEffect(redirectHandler, viewEventFlow, onError) {
+    LaunchedEffect(viewEventFlow) {
```
OscarSpruit, line 34: Now that `onError` is no longer keyed it would be good to use `rememberUpdatedState` for it, because the `LaunchedEffect` will now not be aware of when `onError` changes.

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
```diff
@@ -67,19 +78,76 @@ internal class GooglePayComponent(
 
     @Composable
     override fun Content(modifier: Modifier) {
-        // TODO - Render the Google Pay button and launch the payment sheet.
+        googlePayEvent(
```
jreij: I'm trying in PR [#2828](https://github.com/Adyen/adyen-android/pull/2828) to remove as much compose logic as possible from the components. In addition, I'm not sure what the name `googlePayEvent` is supposed to mean. I have a similar kind of logic in the same PR for card scanning, and everything is now done in the top level of `CardContent`. Maybe check it out and let's align on the best approach for both card scanning and google pay?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
```diff
@@ -67,19 +78,76 @@ internal class GooglePayComponent(
...
+        // TODO - Render the Google Pay button.
         Box(modifier = modifier)
     }
 
     override fun submit() {
-        // TODO - Launch the Google Pay sheet and handle the result before emitting.
-        if (componentStateValidator.isValid(componentState.value)) {
-            val paymentComponentState = componentState.value.toPaymentComponentState(
+        setLoading(true)
+        viewEventChannel.trySend(GooglePayViewEvent.Pay)
```
jreij, line 94: Can you add a comment to clarify why we need a channel and to send an event to the view?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
araratthehero: `googlePayEvent` was created, since there was no `GooglePayContent`, but I went ahead and made it consistent with the `CardContent`.

jreij (approved):

PR #2840, opened 2026-06-25
v6 - Analytics - Non Nullable Attempt ID

OscarSpruit:
## Description
Make `checkoutAttemptId` non-nullable across the SDK by resolving it eagerly during initialization.

Previously, `checkoutAttemptId` was nullable (`String?`) throughout `CheckoutContext`, `DefaultAnalyticsManager`, and related classes. The analytics manager maintained its own `CheckoutAttemptIdState` sealed class to track fetching/failed/available states, and fetched the attempt ID during initialize().

This PR:
- Moves the `checkoutAttemptId` fetch to `CheckoutInitializer`, falling back to "fetch-checkoutAttemptId-failed" on failure instead of returning null.
- Makes `checkoutAttemptId` a non-nullable String in `CheckoutContext` (Sessions, Advanced, ActionOnly), `InitializationData`, `AnalyticsManagerFactory`, and `DefaultAnalyticsSetupProvider`.
- Injects checkoutAttemptId directly into `DefaultAnalyticsManager` via constructor, removing the internal `CheckoutAttemptIdState` state machine and simplifying `getCheckoutAttemptId()` to a direct property access.
- Removes now-unnecessary tests for null/failed checkoutAttemptId states in `DefaultAnalyticsManagerTest`.
- Renames `DISPATCH_INTERVAL_MILLIS` to `DISPATCH_INTERVAL_SECONDS` using `kotlin.time.Duration` directly.

## Checklist 
- [x] Code is unit tested

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/analytics/internal/DefaultAnalyticsManager.kt
+++ b/core/src/main/java/com/adyen/checkout/core/analytics/internal/DefaultAnalyticsManager.kt
@@ -27,10 +27,9 @@ import kotlin.time.Duration.Companion.seconds
 internal class DefaultAnalyticsManager(
     private val analyticsRepository: AnalyticsRepository,
     private val analyticsParams: AnalyticsParams,
+    private val checkoutAttemptId: String,
     private val coroutineDispatcher: CoroutineDispatcher = DispatcherProvider.IO,
 ) : AnalyticsManager {
-
-    private var checkoutAttemptIdState: CheckoutAttemptIdState = CheckoutAttemptIdState.NotAvailable
 
     private var isInitialized: Boolean = false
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/adyen/checkout/core/analytics/internal/DefaultAnalyticsManager.kt
```diff
@@ -99,7 +93,7 @@ internal class DefaultAnalyticsManager(
         }
         timerJob = coroutineScope?.launch(coroutineDispatcher) {
             while (isActive) {
-                delay(DISPATCH_INTERVAL_MILLIS)
+                delay(DISPATCH_INTERVAL_SECONDS)
```
araratthehero, line 96: Does this still work properly after switching from millis to seconds?

OscarSpruit: Yes, because it uses `Duration`, which `delay` internally converts to milliseconds

araratthehero (approved):

PR #2841, opened 2026-06-25
v6 - Analytics - Improve Analytics Manager

OscarSpruit:
## Description
Simplify `AnalyticsManager` lifecycle by removing the initialize/clear pattern and injecting the coroutine scope via constructor.

Previously, `AnalyticsManager` required consumers to call `initialize(owner, coroutineScope)` and `clear(owner)` manually, with an owner-reference mechanism to prevent mismatched clears. Each component (`BlikComponent`, `MBWayComponent`, `GenericPaymentComponent`, etc.) had boilerplate init blocks and `onCleared` overrides to manage this lifecycle.

This PR:
- Removes `initialize` and `clear` from the `AnalyticsManager` interface, reducing it to `trackEvent` and `getCheckoutAttemptId`.
- Injects `CoroutineScope` into `DefaultAnalyticsManager` via constructor, moving the analytics setup call to the init block. - The scope is now passed through `AnalyticsManagerFactory` and `CheckoutControllerFactory`.
- Deletes the `CheckoutAttemptIdState` sealed class, which is no longer needed since checkoutAttemptId is non-nullable and the state machine was already removed in the base branch.
- Renames `AnalyticsRepository.fetchCheckoutAttemptId()` to `setup()` to better reflect its purpose.
- Removes lifecycle boilerplate from components: `BlikComponent`, `StoredBlikComponent`, `CardComponent`, `StoredCardComponent`, `MBWayComponent`, and `GenericPaymentComponent` no longer call initialize/clear. Their `onCleared()` is now a no-op.
- Removes `TestAnalyticsManager.assertIsInitialized()` / `assertIsCleared()` and associated test assertions since these lifecycle methods no longer exist.
- Adds new tests for `DefaultAnalyticsManager`: error handling for `storeEvent`/`sendEvents` failures, timer restart on force-send, and `getCheckoutAttemptId` return value.
- Replaces `cannotSendEvents()` with `canTrackEvents()` (inverted logic) for clarity, and guards the setup call in init behind it.

## Checklist 
- [x] Code is unit tested


core/src/main/java/com/adyen/checkout/core/analytics/internal/DefaultAnalyticsManager.kt
```diff
@@ -28,47 +28,35 @@ internal class DefaultAnalyticsManager(
...
-            }.fold(
-                onSuccess = {
-                    startTimer()
-                },
-                onFailure = {
-                    adyenLog(AdyenLogLevel.WARN, it) { "Failed to fetch checkoutAttemptId." }
-                },
-            )
+    init {
+        if (canTrackEvents()) {
```
OscarSpruit: Remove this if check, because the setup call can be made no matter the configuration

araratthehero (dismissed):

araratthehero (approved):

PR #2846, opened 2026-06-26
v6 - Unify Navigation

OscarSpruit:
## Description
This decouples navigation from the composables, which allows for use cases like non-UI components and review page flow to be able to navigate.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1302


core/src/main/java/com/adyen/checkout/core/components/CheckoutRoute.kt
```diff
@@ -9,19 +9,24 @@
...
- * Represents the navigation routes that can be triggered while displaying [CheckoutPaymentMethod].
+ * Represents the navigation routes that can be triggered during the checkout flow.
  */
-abstract class CheckoutPaymentMethodRoute internal constructor() {
+abstract class CheckoutRoute internal constructor() {
+
+    /**
+     * Route to display a payment method component. Use [CheckoutPaymentMethod] to display it.
+     */
+    class PaymentMethod : CheckoutRoute()
```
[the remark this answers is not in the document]
OscarSpruit: We don't do this for the sake of forward-compatibility

core/src/main/java/com/adyen/checkout/core/components/CheckoutRoute.kt
```diff
@@ -9,19 +9,24 @@
...
 
     /**
      * Route to display an action component. Use [CheckoutAction] to display the action.
      */
-    class Action : CheckoutPaymentMethodRoute()
+    class Action : CheckoutRoute()
```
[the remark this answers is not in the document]
OscarSpruit: We don't do this for the sake of forward-compatibility

araratthehero (dismissed):

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
```diff
@@ -33,13 +32,11 @@ internal class FullCheckoutFlow(
...
 
-    private val paymentMethodNavigationChannel: Channel<CheckoutPaymentMethodRoute> = bufferedChannel()
-    override val paymentMethodNavigation: Flow<CheckoutPaymentMethodRoute> =
-        paymentMethodNavigationChannel.receiveAsFlow()
-
-    private val secondaryNavigationChannel: Channel<CheckoutSecondaryRoute> = bufferedChannel()
-    override val secondaryNavigation: Flow<CheckoutSecondaryRoute> =
-        secondaryNavigationChannel.receiveAsFlow()
+    private val navigationChannel = MutableSharedFlow<CheckoutRoute>(
+        extraBufferCapacity = 1,
```
araratthehero, line 36: This will keep the latest value in the flow, until all subscribers receive it and then it will be removed. So new subscribers won't get the latest value. Is this the correct behaviour?

Also shouldn't we rename the navigationChannel to navigationFlow?

OscarSpruit: IMO navigation is an event and it's up to the consumer of the event to change its state based on the event (for example add it to the back stack).

Good catch! I'll rename it

araratthehero: Fair point!

araratthehero (approved):

PR #2845, opened 2026-06-26
v6 - Google Pay - Availability gating

araratthehero:
## Description
Gates the Google Pay v6 component on `isReadyToPay` availability.

- New `GooglePayAvailabilityCheck` (injectable, mockable): Google Play Services pre-check + `PaymentsClient.isReadyToPay()` awaited via the structured `awaitTask()` helper.
- `GooglePayComponent.Content` inlines `LocalContext.current` + `LaunchedEffect(Unit) { checkAvailability(context) }` (CardComponent precedent); `checkAvailability` dispatches `GooglePayIntent.UpdateAvailability`. Rendering is gated on `viewState.isAvailable`, so nothing renders until availability succeeds (no button flash).
- `GooglePayFactory` constructs and injects `GooglePayAvailabilityCheck`.

### Progress
✅ Phase 1 — Component & factory wiring ([#2837](https://github.com/Adyen/adyen-android/pull/2837))
✅ Phase 2 — Sheet launch + result handling ([#2839](https://github.com/Adyen/adyen-android/pull/2839))
➡️ **Phase 3 — Availability (`isReadyToPay`) gating (this PR)**
Phase 4 — Google Pay button UI
Phase 5 — Example app integration & on-device verification

## Checklist
- [x] Code is unit tested
- [x] Changes are tested manually

## Ticket Number
COSDK-1309

PR #2847, opened 2026-06-26
v6 - Google Pay - Button

araratthehero:
## Description
Renders the Google Pay button using the official [`compose-pay-button`](https://github.com/google-pay/compose-pay-button) library.

- Adds `GooglePayButton` (in `internal/ui/view`), wrapping the library's `PayButton` and mapping Adyen's `GooglePayButtonStyling` (theme, type, corner radius) to the library API via `GooglePayButtonStylingMapper`.
- Uses the button's built-in disabled state (`enabled = !isLoading`) during processing instead of swapping in a spinner.
- Renders the button inside `GooglePayContent`, gated on availability.
- Adds `GooglePayUtils.getAllowedPaymentMethodsJson()` (mirrors v5) to feed the button's allowed payment methods; computed in the component to keep SDK logic out of Compose.

### Progress
✅ Phase 1 — Component & factory wiring ([#2837](https://github.com/Adyen/adyen-android/pull/2837))
✅ Phase 2 — Sheet launch + result handling ([#2839](https://github.com/Adyen/adyen-android/pull/2839))
✅ Phase 3 — Availability (`isReadyToPay`) gating ([#2845](https://github.com/Adyen/adyen-android/pull/2845))
➡️ **Phase 4 — Google Pay button UI (this PR)**
Phase 5 — Example app integration & on-device verification

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1309


googlepay/src/main/java/com/adyen/checkout/googlepay/internal/helper/GooglePayUtils.kt
```diff
@@ -102,6 +103,20 @@ internal object GooglePayUtils {
...
+     * `PayButton` from the Google Pay button library.
+     *
+     * @param params The parameters based on the Google Pay component configuration.
+     * @return The allowed payment methods serialized as a JSON array string.
+     */
+    fun getAllowedPaymentMethodsJson(params: GooglePayComponentParams): String {
+        return ModelUtils.serializeOptList(
+            getAllowedPaymentMethods(params),
+            GooglePayPaymentMethodModel.SERIALIZER,
+        )?.toString() ?: "[]"
```
jreij: I'm not a big fan of hardcoding `"[]"`. What if we fallback to `JSONArray()` before calling `toString()`?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/view/GooglePayContent.kt
```diff
@@ -30,8 +30,11 @@ import kotlinx.coroutines.flow.StateFlow
 internal fun GooglePayContent(
     viewStateFlow: StateFlow<GooglePayViewState>,
     viewEventFlow: Flow<GooglePayViewEvent>,
+    allowedPaymentMethods: String,
+    buttonStyling: GooglePayButtonStyling?,
```
jreij: Is it intentional that those fields are not part of the view state?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/view/GooglePayButton.kt
```diff
@@ -0,0 +1,40 @@
...
+@Composable
+internal fun GooglePayButton(
+    isLoading: Boolean,
+    isButtonVisible: Boolean,
+    allowedPaymentMethods: String,
+    buttonStyling: GooglePayButtonStyling?,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    if (isButtonVisible) {
```
jreij: In other UI elements, we would check if visible in the parent composable and if it's not then we wouldn't call `GooglePayButton()` at all, what do youn think of doing the same here?

jreij: Maybe a suggestion for both this comment and the other one (plus the discussion on Slack): wrap all params related to the button inside one object which is set to null if the button shouldn't be shown

.github/release_notes_dependency_list.toml
```diff
@@ -25,6 +25,7 @@
 "com.android.tools.build:gradle" = "[Android Gradle Plugin](https://developer.android.com/build/releases/gradle-plugin)"
 "com.google.android.gms:play-services-wallet" = "[Play Services Wallet](https://developers.google.com/pay/api/android/support/release-notes)"
 "com.google.android.material:material" = "[Material](https://github.com/material-components/material-components-android/releases/tag/{})"
+"com.google.pay.button:compose-pay-button" = "[Google Pay Compose Button](https://github.com/google-pay/compose-pay-button/releases)"
```
jreij: This can be `https://github.com/google-pay/compose-pay-button/releases/tag/v{}`

.github/release_notes_dependency_list.toml
```diff
@@ -25,6 +25,7 @@
 "com.android.tools.build:gradle" = "[Android Gradle Plugin](https://developer.android.com/build/releases/gradle-plugin)"
 "com.google.android.gms:play-services-wallet" = "[Play Services Wallet](https://developers.google.com/pay/api/android/support/release-notes)"
 "com.google.android.material:material" = "[Material](https://github.com/material-components/material-components-android/releases/tag/{})"
+"com.google.pay.button:compose-pay-button" = "[Google Pay Compose Button](https://github.com/google-pay/compose-pay-button/releases/tag/{})"
```
jreij: I think you're missing a `v` before the `{}`

araratthehero: Good point, I mixed it up with other dependencies which did not have `v` before the version name.

jreij (approved):

PR #2848, opened 2026-06-29
v6 - Google Pay - Example app integration

araratthehero:
## Description
Enables Google Pay in the v6 example app so the component can be exercised end-to-end.

- Adds `GOOGLE_PAY` and `GOOGLE_PAY_LEGACY` to the v6 sample's supported payment methods (`CheckoutContextExt.kt`). Both type strings map to the same component, matching v5.
- Adds a v6 `googlePay(countryCode, checkoutOption)` block to the example's `checkoutConfig` (`CheckoutConfigurationProvider.kt`), mirroring the old sample's relevant settings.

### Progress

✅ Component and factory wiring — #2837
✅ Sheet launch and result handling — #2839
✅ Availability gating — #2845
✅ Button — #2847
➡️ **Example app integration (this PR)**

## Checklist

## Ticket Number
COSDK-1309

### DIFF
--- a/example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
+++ b/example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
@@ -24,6 +24,7 @@ import com.adyen.checkout.example.data.storage.CardAddressMode
 import com.adyen.checkout.example.data.storage.CardInstallmentOptionsMode
 import com.adyen.checkout.example.data.storage.KeyValueStorage
 import com.adyen.checkout.giftcard.giftCard
+import com.adyen.checkout.googlepay.googlePay
 import com.adyen.checkout.googlepay.old.googlePay
 import com.adyen.checkout.instant.instantPayment
 import com.adyen.checkout.mealvoucherfr.mealVoucherFR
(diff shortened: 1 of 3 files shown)


example-app/src/main/java/com/adyen/checkout/example/ui/configuration/CheckoutConfigurationProvider.kt
```diff
@@ -123,6 +124,11 @@ internal class CheckoutConfigurationProvider @Inject constructor(
                 installmentConfiguration = getInstallmentConfiguration(),
             )
 
+            googlePay(
+                countryCode = keyValueStorage.getCountry(),
+                checkoutOption = "COMPLETE_IMMEDIATE_PURCHASE",
```
jreij, line 129: Is this needed for the component to work or can merchants not set it? Same for country. Asking because in v6 we want to make sure all configurations are optional and you can just use the SDK with 0 manual configs (unless strictly necessary like `threeDSRequestorAppURL`)

araratthehero: [Here](https://developers.google.com/pay/api/android/reference/request-objects#TransactionInfo) it mentions that the `checkoutOption` is required and the `countryCode` is only required for the EEA countries. Though I test it and the payment succeeds without sending those values. Could be that it can fail on Live.

I will discuss with the team and fix it as part of the Configuration alignment.

jreij (approved):

PR #2850, opened 2026-06-29
v6 - Authentication Rename

OscarSpruit:
## Description
Move "new" 3ds2 classes to `:authentication` module and rename them. Bonus: avoid generic error usage.

Next step: add more tests to these classes.

## Ticket Number
COSDK-1123


jreij (dismissed):

settings.gradle
jreij, line 1: Some gradle imports might be missing, based on where we import `3ds2`:

- Root `build.gradle` -> `dokka(project(':authentication'))` (this is a bit problematic, easy to forget modules there) 
- `action` and `action-core` module - is this still relevant in v6?
- BCMC module - probably didn't even need to have `3ds2` there because we have `card`
- Google Pay module

OscarSpruit: Sharp! `action` and `action-core` are indeed no longer relevant

jreij (approved):

PR #2851, opened 2026-06-29
v6 - Card - Dual Brand Selection

tkuntubayev:
## Description
Adds UI to allow selection of a brand for Dual Branded Cards with Selection and adds corresponding Analytics events.

[Recording](https://github.com/user-attachments/assets/fa2bd1bb-15f7-4657-936b-12a05b1153a6)

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1193

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/analytics/DualBrandCardEvents.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/analytics/DualBrandCardEvents.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by temirlan on 29/6/2026.
+ */
+
+package com.adyen.checkout.card.internal.analytics
+
+import com.adyen.checkout.card.internal.ui.state.CardBrandData
+import com.adyen.checkout.core.analytics.internal.GenericEvents
+import com.adyen.checkout.core.common.CardBrand
+
+internal object DualBrandCardEvents {
+
+    fun dualBrandSelectionDisplayed(
+        component: String,
+        selectedBrand: CardBrand,
+        brandOptions: List<CardBrandData>,
+    ) = GenericEvents.displayed(
+        component = component,
+        target = DUAL_BRAND_ANALYTICS_TARGET,
+        brand = selectedBrand.txVariant,
+        configData = mapOf("dualBrands" to brandOptions.joinToString(",") { it.cardBrand.txVariant }),
+    )
+
+    fun brandSelected(
+        component: String,
+        selectedBrand: CardBrand
+    ) = GenericEvents.selected(
+        component = component,
+        target = DUAL_BRAND_ANALYTICS_TARGET,
+        brand = selectedBrand.txVariant,
+    )
+
+    private const val DUAL_BRAND_ANALYTICS_TARGET = "dual_brand_button"
+}
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
@@ -266,13 +266,19 @@ constructor(
     }
 
     private fun subscribeToDualBrandSelectionAppearAnalyticsEvents() {
+        // Fires dualBrandSelectionDisplayed when DualBrandWithShopperSelection appears.
         componentState
             .map { it.cardBrandState }
+            // Compares bucket membership (DualBrandWithShopperSelection vs. anything else) rather
+            // than state equality: suppresses brand-selection changes within the dual brand UI
+            // (same bucket → "equal") while emitting on appear/disappear transitions.
             .distinctUntilChanged { old, new ->
                 val isOldDualBrandWithSelection = old is CardBrandState.DualBrandWithShopperSelection
                 val isNewDualBrandWithSelection = new is CardBrandState.DualBrandWithShopperSelection
                 isOldDualBrandWithSelection == isNewDualBrandWithSelection
             }
+            // Discards the disappear emissions so only appear transitions reach onEach.
+            // WARNING: must come AFTER distinctUntilChanged — reversing them causes the event to never fire.
             .mapNotNull { it as? CardBrandState.DualBrandWithShopperSelection }
             .onEach { cardBrandState ->
                 val event = DualBrandCardEvents.dualBrandSelectionDisplayed(
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
@@ -56,7 +56,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.drop
-import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapNotNull
@@ -268,13 +267,14 @@ constructor(
 
     private fun subscribeToDualBrandAnalyticsEvents() {
         componentState
-            .map { it.cardBrandState is CardBrandState.DualBrandWithShopperSelection }
-            .distinctUntilChanged()
-            .filter { isVisible -> isVisible }
-            .onEach {
-                val cardBrandState = componentState.value.cardBrandState
-                    as? CardBrandState.DualBrandWithShopperSelection ?: return@onEach
-
+            .map { it.cardBrandState }
+            .distinctUntilChanged { old, new ->
+                val isOldDualBrandWithSelection = old is CardBrandState.DualBrandWithShopperSelection
+                val isNewDualBrandWithSelection = new is CardBrandState.DualBrandWithShopperSelection
+                isOldDualBrandWithSelection == isNewDualBrandWithSelection
+            }
+            .mapNotNull { it as? CardBrandState.DualBrandWithShopperSelection }
+            .onEach { cardBrandState ->
                 val event = DualBrandCardEvents.dualBrandSelectionDisplayed(
                     component = paymentMethodType,
                     selectedBrand = cardBrandState.shopperSelectedCardBrandData.cardBrand,
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
@@ -14,6 +14,7 @@ import android.content.Intent
 import android.content.IntentSender
 import androidx.activity.result.ActivityResultLauncher
 import androidx.activity.result.IntentSenderRequest
+import androidx.annotation.VisibleForTesting
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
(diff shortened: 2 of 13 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -138,13 +154,74 @@ private fun DetectedBrandsList(
...
+                                    shape = RoundedCornerShape(Dimensions.CornerRadius),
+                                )
+                                .background(color = CheckoutThemeProvider.colors.background)
+                        } else {
+                            Modifier
+                        },
+                    )
+                    .clickable { onBrandSelect(brandItem.brand) }
+                    .padding(Dimensions.Spacing.ExtraSmall)
+                    .clip(RoundedCornerShape(Dimensions.CornerRadius)),
```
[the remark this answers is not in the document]
tkuntubayev: The first clip is applied to the brand logo raw image, the second clip is to the combined image with paddings. Having just one will make either the raw image rectangular or paddings being rectangular which will mismatch the UI design.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -244,6 +251,39 @@ constructor(
...
+
+                val event = DualBrandCardEvents.dualBrandSelectionDisplayed(
+                    component = paymentMethodType,
+                    selectedBrand = cardBrandState.shopperSelectedCardBrandData.cardBrand,
+                    brandOptions = cardBrandState.cardBrandDataList,
+                )
+                analyticsManager.trackEvent(event)
+            }
+            .launchIn(coroutineScope)
+    }
```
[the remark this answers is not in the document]
tkuntubayev: The accessing `componentState.value` inside a reactive flow pipeline is replaced with flow operators

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -138,13 +154,74 @@ private fun DetectedBrandsList(
...
             is CardBrandViewState.Placeholder -> BrandLogo(txVariant = null)
             is CardBrandViewState.SingleBrand -> BrandLogo(cardBrandViewState.brand.txVariant)
-            is CardBrandViewState.DualBrand -> cardBrandViewState.brands.forEach { brand ->
-                BrandLogo(brand.txVariant)
-            }
-            // TODO - Co-badged selectable cards [COSDK-1193]
-            is CardBrandViewState.SelectableDualBrand -> cardBrandViewState.brands.forEach { brandItem ->
-                BrandLogo(brandItem.brand.txVariant)
-            }
+            is CardBrandViewState.DualBrand -> DualBrandLogos(cardBrandViewState.brands)
```
jreij, line 157: Can you update the compose preview to showcase something related to dual branding?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -138,13 +154,74 @@ private fun DetectedBrandsList(
...
+                    .semantics {
+                        role = Role.RadioButton
+                        selected = brandItem.isSelected
+                    }
+                    .clip(RoundedCornerShape(Dimensions.CornerRadius))
+                    .then(
+                        if (brandItem.isSelected) {
+                            Modifier
+                                .border(
+                                    width = 1.dp,
```
jreij, line 212: I see this 1dp hardcoded here and in `CheckoutNetworkLogo`, both representing borders. Is this a dimension we should extract?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
```diff
@@ -218,7 +296,7 @@ private fun CardNumberFieldIcon(
                 )
             }
 
-            else -> DetectedBrandsList(cardBrandViewState)
+            else -> DetectedBrandsList(cardBrandViewState, onBrandSelect)
```
jreij: Not directly related to this PR but can we explicitly use `CardNumberTrailingIcon.BrandLogos` instead of `else` here?

tkuntubayev: For that to happen, we'd need to handle the `null` case now as `trailingIcon` is nullable. 
We could:
1. make it non-nullable by exiting early in `CardNumberFieldIcon`:
```val trailingIcon = state.trailingIcon as? CardNumberTrailingIcon ?: return```
2. or come up with a default icon to show for `null` case

 wdyt?

File: card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt (hunk 4 above)
Comment (tkuntubayev) [reply card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt:212]: The one in the `CheckoutNetworkLogo` represents a shadow blur radius, not really a border. Combining them under 1 dimension might be misleading. So far each of them is used only once. Let me know wdyt on whether we'd like to extract them under 2 dimensions, put under 1 common dimension or keep it as-is

File: card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt [hunk 6]
```diff
@@ -245,6 +251,40 @@ constructor(
...
+            )
+            analyticsManager.trackEvent(event)
+        }
+        onIntent(intent)
+    }
+
+    private fun subscribeToDualBrandSelectionAppearAnalyticsEvents() {
+        componentState
+            .map { it.cardBrandState }
+            .distinctUntilChanged { old, new ->
```
jreij, line 275: I struggled to understand this logic here and with quite a bit of help from AI, I now get it. I was sure we could re-write this in a more understandable way but I was wrong, this seems to be the best way.

However, this logic is very fragile to an accidental refactor. For example if someone tries to "optimize" by moving `mapNotNull` before `distinctUntilChanged` or making `distinctUntilChanged` only return if `isOldDualBrandWithSelection == false` and `isNewDualBrandWithSelection == true`, it would cause a bug.

I suggest we add a comment explaining the logic a little bit (and warning against changing it) and adding some unit tests to cover all the edge cases.

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextFieldDecorationBox.kt
```diff
@@ -98,7 +98,7 @@ internal fun CheckoutTextFieldDecorationBox(
                 .styledBackground(style, isFocused, isError)
                 .fillMaxWidth()
                 .heightIn(Dimensions.MinTouchTarget)
-                .padding(horizontal = Dimensions.Spacing.Large, vertical = Dimensions.Spacing.Medium),
+                .padding(horizontal = Dimensions.Spacing.Large),
```
jreij, line 101: Is this change intended?

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextFieldDefaults.kt
```diff
@@ -18,7 +18,7 @@ internal object CheckoutTextFieldDefaults {
         attributes: InternalAttributes,
     ): InternalTextFieldStyle {
         return InternalTextFieldStyle(
-            backgroundColor = colors.container,
+            backgroundColor = colors.background,
```
jreij, line 21: Is this change intended?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
jreij: Ah in that case let's just keep it as it is then

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
jreij: Is null even a possible use case with this field? if not then maybe we can just not return anything and put a comment that null is not expected

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextFieldDefaults.kt
tkuntubayev: Yes, to show proper background for `TextInputField` and have contrast with the container color of Selectable Dual-Brand logos container. With this it gets aligned with the current Figma design

card/src/main/java/com/adyen/checkout/card/internal/analytics/DualBrandCardEvents.kt
```diff
@@ -0,0 +1,38 @@
+/*
```
jreij, line 1: All commits are missing the ticket number, not sure if the agent forgot to add it

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextFieldDecorationBox.kt
tkuntubayev: Yes, the vertical padding adds unnecessary space for the whole view when the content has it's own paddings like the container padding of selectable Dual-Brand logos. The height of the view is constrained by `.heightIn(Dimensions.MinTouchTarget)` and after inner talk, the vertical padding was suggested for removal.

card/src/main/java/com/adyen/checkout/card/internal/analytics/DualBrandCardEvents.kt
tkuntubayev: The agent indeed didn't pick up the commit instructions part of AGENTS.md, so I did commit myself and wasn't aware of the ticket number need to be in each commit.

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
tkuntubayev: Added previews for both types of dual branding

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
tkuntubayev: Sounds good, I've extended Unit tests with more edge case scenarios. Also added supportive/warning comments, let me know how this looks now

card/src/main/java/com/adyen/checkout/card/internal/analytics/DualBrandCardEvents.kt
jreij: Yeah, we include the ticket number in the commit to make it easier for someone doing git blame / history in the future to quickly look up the ticket that introduced a certain change

jreij: Maybe you can ask an agent to rebase all the commits and just add the ticket number to the message?

tkuntubayev: Rebased, all commits now have the ticket number

card/src/main/java/com/adyen/checkout/card/internal/ui/view/CardNumberField.kt
tkuntubayev: Currently this field is always populated by `CardViewStateProducer` with the expected `TrailingIcon` subtype so in practice it should never be `null`. The `else` is replaced with proper subtype check and early exit with supportive comment added

jreij (approved):
🚀

PR #2869, opened 2026-07-02
v6 - Clarify session-controlled card docs

atmamont:
## Description
- clarify that card session flow uses session response values for `showStorePaymentMethod`
- clarify that card session flow uses session response values for `installmentConfiguration` and `showInstallmentAmount`
- remove the misleading `showStorePaymentMethod = true` line from the sessions-specific example
- iOS reference: https://github.com/Adyen/adyen-ios/pull/2622

## Ticket Number
COSDK-1108


docs/v6/README.md
atmamont, line 1: @gemini-code-assist What does it mean? There is no code changed in the PR scope, only markdown files

atmamont: /gemini review-comment-reply

jreij, line 1: Not directly related to cards, but the client side configured amount is also fully ignored in a sessions flow - in case we need to document that as well

atmamont: Good catch, addressed here https://github.com/Adyen/adyen-android/pull/2869/changes/a45dcf76bfa6502111f518f634f425c9d8d3da3f

docs/v6/README.md
```diff
@@ -84,6 +84,8 @@ val configuration = CheckoutConfiguration(
 
 Use the builder block to register payment-method and action configuration that should apply to the flow.
 
+In session flow, the effective `amount` comes from the `/sessions` response. A client-side `amount` on `CheckoutConfiguration` is ignored and only used as a fallback when the session has no amount. `showStorePaymentMethod` and `installmentConfiguration` for cards are likewise controlled by the `/sessions` response through `enableStoreDetails`, `installmentOptions`, and `showInstallmentAmount`. Configure these values in your session request and do not rely on component-level values to override them.
```
jreij: > only used as a fallback when the session has no amount

I don't think this should ever happen, [amount is required for /sessions](https://docs.adyen.com/api-explorer/Checkout/72/post/sessions#request-amount) and if we don't receive it back from the backend we should treat this as a bug not a fallback IMO

atmamont: True, removed the fallback part.

docs/v6/card.md
```diff
@@ -64,6 +64,9 @@ val configuration = CheckoutConfiguration(
 | `socialSecurityNumberVisibility` | Shows or hides the CPF/CNPJ field. |
 | `supportedCardBrands` | Restricts the supported card brands for the flow. |
 | `showCardScanner` | Shows or hides the card scanner entry point. |
+| `installmentConfiguration` | Configures installments. |
+
+When you use [session flow](card-session-flow.md), `showStorePaymentMethod` and `installmentConfiguration` are controlled by the `/sessions` response. Their effective values come from `enableStoreDetails`, `installmentOptions`, and `showInstallmentAmount`.
```
jreij: `enableStoreDetails` is an internal response param coming from sessions setup but it is not merchant facing. I believe [storePaymentMethodMode](https://docs.adyen.com/api-explorer/Checkout/72/post/sessions#request-storePaymentMethodMode) is the right request param on sessions.

As for [installmentOptions](https://docs.adyen.com/api-explorer/Checkout/72/post/sessions#request-installmentOptions) and [showInstallmentAmount](https://docs.adyen.com/api-explorer/Checkout/72/post/sessions#request-showInstallmentAmount), I believe they have the same name in `/sessions`.

atmamont: I made it more explicit, please check out.

jreij: Looks good!

jreij (approved):

PR #2871, opened 2026-07-03
v6 - Submit guard

OscarSpruit:
## Description
Add a canSubmit guard to FullCheckoutFlow.submit() to prevent duplicate payment submissions when the method is called multiple times while an async request is in-flight or when an action component is shown.

## Checklist 
- [x] Code is unit tested

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
@@ -22,6 +22,7 @@ import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
+import java.util.concurrent.atomic.AtomicBoolean
 
 internal class FullCheckoutFlow(
     private val componentRequestDispatcher: SubmittableComponentRequestDispatcher,
@@ -38,8 +39,7 @@ internal class FullCheckoutFlow(
     )
     override val navigation: Flow<CheckoutRoute> = navigationFlow.asSharedFlow()
 
-    @Volatile
-    private var canSubmit = true
+    private val canSubmit = AtomicBoolean(true)
 
     init {
         paymentComponent.eventFlow
@@ -67,8 +67,7 @@ internal class FullCheckoutFlow(
     }
 
     override fun submit() {
-        if (!canSubmit) return
-        canSubmit = false
+        if (!canSubmit.compareAndSet(true, false)) return
         paymentComponent.submit()
         paymentComponent.setLoading(true)
     }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
```diff
@@ -38,15 +39,15 @@ internal class FullCheckoutFlow(
...
+    private val canSubmit = AtomicBoolean(true)
+
     init {
-        paymentComponent?.eventFlow
-            ?.onEach { event ->
+        paymentComponent.eventFlow
+            .onEach { event ->
                 when (event) {
                     is PaymentComponentEvent.Submit -> {
-                        paymentComponent.setLoading(true)
```
araratthehero, line 46: The loading state of the button is not visible anymore when clicking on the Pay button. Is that intended?

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
```diff
@@ -38,15 +39,15 @@ internal class FullCheckoutFlow(
     )
     override val navigation: Flow<CheckoutRoute> = navigationFlow.asSharedFlow()
 
+    private val canSubmit = AtomicBoolean(true)
```
araratthehero, line 42: What are the use-cases that this would be modified from multiple async sources?

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
OscarSpruit: Good catch! I'll fix it

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
OscarSpruit: Technically the `submit()` function can be called multiple times from different threads, there is nothing to restrict someone from doing that. Now I don't know why someone would do that, but thought it would be good to be defensive here

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
araratthehero, line 55: Just to confirm, we do not set `paymentComponent.setLoading(false)` for the Errors, because when an Error happens, the component state is not recoverable and user cannot make a payment anymore right?

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
araratthehero: I see. I guess then we could keep it to be more defensive.

araratthehero (dismissed):

core/src/main/java/com/adyen/checkout/core/components/internal/FullCheckoutFlow.kt
OscarSpruit: Yes indeed

jreij (approved):

PR #2873, opened 2026-07-08
v6 - Google Pay - Emit error if Google Pay is unavailable

jreij:
## Description
Emit an error if Google Pay is unavailable.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1310

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/error/CheckoutErrorMapper.kt
+++ b/core/src/main/java/com/adyen/checkout/core/error/CheckoutErrorMapper.kt
@@ -11,6 +11,7 @@ package com.adyen.checkout.core.error
 import com.adyen.checkout.core.error.internal.GenericError
 import com.adyen.checkout.core.error.internal.HttpError
 import com.adyen.checkout.core.error.internal.InternalCheckoutError
+import com.adyen.checkout.core.error.internal.PaymentMethodUnavailableError
 
 /**
  * Maps internal error hierarchy to public CheckoutError.
@@ -19,6 +20,7 @@ internal fun InternalCheckoutError.toCheckoutError(): CheckoutError {
     val errorCode = when (this) {
         is HttpError -> CheckoutError.ErrorCode.HTTP
         is GenericError -> CheckoutError.ErrorCode.GENERIC
+        is PaymentMethodUnavailableError -> CheckoutError.ErrorCode.PAYMENT_METHOD_FAILURE
     }
 
     return CheckoutError(
--- a/core/src/main/java/com/adyen/checkout/core/error/internal/PaymentMethodUnavailableError.kt
+++ b/core/src/main/java/com/adyen/checkout/core/error/internal/PaymentMethodUnavailableError.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 8/7/2026.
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/adyen/checkout/core/error/internal/PaymentMethodUnavailableError.kt
```diff
@@ -0,0 +1,23 @@
...
+import androidx.annotation.RestrictTo
+
+/**
+ * Indicates that the payment method is not available for this transaction.
+ *
+ * @param message A human-readable description of the error.
+ * @param cause The underlying cause of this error, if any.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+class PaymentMethodUnavailableError(
```
jreij, line 20: Do we need a new error here?

core/src/main/java/com/adyen/checkout/core/error/CheckoutErrorMapper.kt
```diff
@@ -19,6 +20,7 @@ internal fun InternalCheckoutError.toCheckoutError(): CheckoutError {
     val errorCode = when (this) {
         is HttpError -> CheckoutError.ErrorCode.HTTP
         is GenericError -> CheckoutError.ErrorCode.GENERIC
+        is PaymentMethodUnavailableError -> CheckoutError.ErrorCode.PAYMENT_METHOD_FAILURE
```
jreij, line 23: Is this the correct mapping?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
```diff
@@ -99,6 +100,9 @@ internal class GooglePayComponent(
     private fun checkAvailability() {
         coroutineScope.launch {
             val isAvailable = googlePayAvailabilityCheck.isAvailable()
+            if (!isAvailable) {
+                emitError(PaymentMethodUnavailableError("Google Pay is not available"))
```
jreij, line 104: Do we need to send any analytics?

core/src/main/java/com/adyen/checkout/core/error/internal/PaymentMethodUnavailableError.kt
araratthehero: If the error code does not fall under HTTP or Generic errors, then a new Error type should be created, like you did here. We already have the `PaymentMethodFailure` aligned with iOS, they use it for ApplePay and we can reuse this error code for future payment methods which can fail because of availability. So yes, we need a new error type here.

core/src/main/java/com/adyen/checkout/core/error/CheckoutErrorMapper.kt
araratthehero: Yes, this is the correct mapping. We don't expect this list to grow much in the future. We still need to add a mapping for `Cancelled` error and that should be it for now. Most of the errors fall under `GenericError`

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
araratthehero: We do not track analytics event for this case yet. We can discuss in the future if adding an event is necessary here and create a specific errorCode for it.

araratthehero (approved):

PR #2875, opened 2026-07-08
v6 - Google Pay - Fix validation logic

jreij:
## Description

- Remove the validation of `paymentData` inside `GooglePayComponentStateValidator.isValid` since this call is only triggered when the submit button is clicked while `paymentData` only becomes available afterwards.
- Ensure that `GooglePayComponentStateValidator.isValid` validates that Google Pay is available before allowing the submit event to go through. Although the Google Pay button will not be shown is Google Pay is unavailable - which means a submit cannot be triggered through it - the merchant can still manually trigger `submit()` or add their custom button.
- Ensure `GooglePayComponentStateValidator.isValid` is being called inside submit - previously it was completely ignored.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1310


googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
```diff
@@ -27,6 +27,7 @@ internal fun GooglePayComponentState.toPaymentComponentState(
         order = null,
     )
 
+    // TODO check if paymentComponentData should be null if the token is null or empty
```
jreij: Maybe worth returning null the data is not valid, if a merchant forgets to check the valid flag they will forward data that would fail the API call (unless setting `isValid  = false` here somehow prevents onSubmit from being triggered altogether).

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateValidator.kt
```diff
@@ -18,8 +17,10 @@ internal class GooglePayComponentStateValidator : ComponentStateValidator<Google
...
 
     override fun isValid(state: GooglePayComponentState): Boolean {
-        val paymentData = state.paymentData ?: return false
-        return runCatching { GooglePayUtils.findToken(paymentData) }
-            .getOrNull()?.isNotEmpty() == true
+        // this validation is triggered when the component is submitted (pay button is clicked)
+        // at that moment the google pay sheet has not been opened yet so the GooglePayComponentState.paymentData is
+        // still null
+        // so the only parameter that can be validated here is the availability check result
+        return state.isAvailable
```
jreij, line 20: Need a second pair of eyes to confirm this

googlepay/src/test/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponentTest.kt
```diff
@@ -86,6 +86,35 @@ internal class GooglePayComponentTest {
...
+    @Test
+    fun `when availability check returns false and submit is called afterwards, then no event is emitted for the view`() =
+        runTest {
+            val component = createComponent(
+                coroutineScope = CoroutineScope(UnconfinedTestDispatcher(testScheduler)),
+                googlePayAvailabilityCheck = mockGooglePayAvailabilityCheck(isAvailable = false),
+            )
+            val events = component.viewEventFlow.test(testScheduler)
+            // although the flow itself is not tested, this line is needed to ensure the flow gets updated
+            component.viewState.test(testScheduler)
```
jreij, line 113: Any idea why this happens? Is there a better way to fix it?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
```diff
@@ -27,6 +27,7 @@ internal fun GooglePayComponentState.toPaymentComponentState(
...
     val isValid = paymentData != null && GooglePayUtils.findToken(paymentData).isNotEmpty()
```
jreij: Technically we validate in two stages:

- Stage 1: when shopper presses the pay button. This usually happens inside the `GooglePayComponentStateValidator`, but at that moment we need to launch the google pay SDK and we don't receive the payment data until after the payment is done. Which is why we can only validate the availability here.
- Stage 2: when the google sheet is closed and a result is returned. At this point we have the paymentData and we need to validate that as well. Currently the validation is being done here but it's worth discussing if we should keep this "second validation" logic here or move it to the `GooglePayComponentStateValidator` and call it from here.

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
OscarSpruit: Merchants can't even check the `isValid` flag. In `onSubmit` they get `PaymentComponentData` and not `PaymentComponentState`

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
OscarSpruit: IMO there should be no validation here at all. This should be just a simple mapping

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateValidator.kt
OscarSpruit: Maybe we can do the full validation here (stage 2) and when submit is clicked we just check `state.isAvailable`

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
```diff
@@ -118,8 +120,12 @@ internal class GooglePayComponent(
     }
 
     override fun submit() {
-        setLoading(true)
-        viewEventChannel.trySend(GooglePayViewEvent.Pay)
+        if (componentStateValidator.isValid(componentState.value)) {
+            setLoading(true)
```
OscarSpruit, line 130: Usually a component doesn't set it's own loading state, but because actual submission is happening later I guess it's needed. Do we use this loading state to prevent double clicks on the button?

googlepay/src/test/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponentTest.kt
OscarSpruit: The view state flows use `SharingStarted.WhileSubscribed`, so if there are no subscribers there will be no values emitted. What you did is probably the best way to get it to work

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
jreij: So it's better to return null if not valid right?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateValidator.kt
jreij: I'm worried this would cause an inconsistency. In most components the `ComponentStateValidator.isValid` function is called inside the `submit` function (when the button is clicked) but in some components like here, it would be called when the state is being generated to be sent to `onSubmit`.

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
jreij: Ah even better! We can validate when we receive the result back from the google pay SDK and only map this one if everything is valid 👍

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
jreij: Yes indeed, there is a line in the GooglePayButton composable:
```kotlin
PayButton(
    ...
    enabled = !buttonViewState.isLoading,
)
```

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateExt.kt
OscarSpruit: Yes I think so

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/state/GooglePayComponentStateValidator.kt
OscarSpruit: It might look like an inconsistency, but I don't think it actually is. Other components check validity before triggering `onSubmit` (and when the button is clicked, but that's not important). Google Pay would be the outlier if it only check it on button click and not before triggering `onSubmit`

jreij: Ok I'm convinced! A state should only be valid when it's the final state ready to be submitted. I adjusted the code to switch the logic around, let me know your thoughts.

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
```diff
@@ -144,9 +160,9 @@ internal class GooglePayComponent(
     }
 
     private fun handleSuccessResult(paymentData: PaymentData?) {
-        if (paymentData == null) {
-            adyenLog(AdyenLogLevel.ERROR) { "GooglePay payment data is null" }
-            trackThirdPartyErrorEvent("Result is success, but data is missing")
+        if (paymentData == null || GooglePayUtils.findToken(paymentData).isNullOrEmpty()) {
+            adyenLog(AdyenLogLevel.ERROR) { "GooglePay payment data is null or invalid" }
+            trackThirdPartyErrorEvent("Result is success, but data is missing or invalid")
```
jreij, line 161: @araratthehero is this string flexible or can I change it like I did here?

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
```diff
@@ -155,14 +171,6 @@ internal class GooglePayComponent(
         analyticsManager.trackEvent(GenericEvents.submit(paymentMethodType))
 
         componentState.handleIntent(GooglePayIntent.UpdatePaymentData(paymentData))
-
-        val paymentComponentState = componentState.value
```
jreij, line 159: This function was sending an intent, manually updating the state and submitting the component, all in one place - so technically the intent was not being used for anything. I changed it because I think we should either send an intent or ignore the intent and manually trigger onSubmit from here.

I prefer the first option but it means we need to observe the changes of the state and trigger onSubmit when it becomes valid, which is what `submitWhenStateIsValid` is doing.

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
atmamont: Just for info: we already have discrepancies with Twint SDK initialization, the same error comes in two different strings from platforms. This has to be aligned between platforms by the time we release v6.

googlepay/src/main/java/com/adyen/checkout/googlepay/internal/ui/GooglePayComponent.kt
OscarSpruit: Not really a fan of this pattern. There is not a singular, clear control flow. Instead it's spread out over multiple places. Also this flow could theoretically be triggered by other events, even though that might not be the intention.

What I think would be better is to have the whole submission in `handleSuccessResult`. IMO `PaymentData` is not really state, it's more part of an event flow. So what I mean is that we don't have to update the component state with it. It would become something like
```
private fun handleSuccessResult(paymentData: PaymentData?) {
    if (paymentData == null || GooglePayUtils.findToken(paymentData).isNullOrEmpty()) {
        adyenLog(AdyenLogLevel.ERROR) { "GooglePay payment data is null or invalid" }
        trackThirdPartyErrorEvent("Result is success, but data is missing or invalid")
        emitError(GenericError("GooglePay encountered an unexpected error"))
        return
    }
    adyenLog(AdyenLogLevel.INFO) { "GooglePay payment result successful" }

    analyticsManager.trackEvent(GenericEvents.submit(paymentMethodType))

    val paymentComponentState = componentState.value.toPaymentComponentState(
        paymentMethodType = paymentMethodType,
        sdkDataProvider = sdkDataProvider,
        // IMPORTANT: this is non-nullable in this case, so toPaymentComponentState can be cleaner as well
        paymentData = paymentData,
    )
    eventChannel.trySend(PaymentComponentEvent.Submit(paymentComponentState))
}
```

jreij: Makes sense, I completely removed payment data from the intent and state and triggered onSubmit manually when the result is returned, let me know what you think.

OscarSpruit (dismissed):

araratthehero (dismissed):

OscarSpruit (approved):

PR #2876, opened 2026-07-08
v6 - Analytics - Render events

OscarSpruit:
## Description
This PR adds render event tracking to all payment components. The render event is triggered when the component is created to keep the behaviour with v5 consistent.

## Checklist 
- [x] Code is unit tested

## Ticket Number
COSDK-1315

### DIFF
--- a/blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikComponent.kt
+++ b/blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikComponent.kt
@@ -18,19 +18,22 @@ import com.adyen.checkout.blik.internal.ui.state.BlikViewStateProducer
 import com.adyen.checkout.blik.internal.ui.state.toPaymentComponentState
 import com.adyen.checkout.blik.internal.ui.view.BlikContent
 import com.adyen.checkout.core.analytics.internal.AnalyticsManager
+import com.adyen.checkout.core.analytics.internal.GenericEvents
 import com.adyen.checkout.core.common.internal.helper.bufferedChannel
 import com.adyen.checkout.core.components.internal.PaymentComponentEvent
 import com.adyen.checkout.core.components.internal.data.provider.SdkDataProvider
 import com.adyen.checkout.core.components.internal.ui.PaymentComponent
 import com.adyen.checkout.core.components.internal.ui.state.ComponentStateFlow
 import com.adyen.checkout.core.components.internal.ui.state.viewState
+import com.adyen.checkout.core.components.paymentmethod.PaymentMethodTypes
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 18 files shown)


blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikComponent.kt
```diff
@@ -51,6 +54,15 @@ internal class BlikComponent(
 
     private val viewState = componentState.viewState(viewStateProducer, coroutineScope)
 
+    init {
+        trackRenderEvent()
```
jreij, line 58: Technically this is not the moment the component is rendered on the UI right? If a merchant initializes the controller then attaches it to the UI later then the render will have the wrong timestamp. If they don't attach the controller to the UI at all then we'll have a render event without a render.

I know this is the same behavior as v5 and it's a tough problem to solve but I'm brainstorming if there is a more accurate way to track this event.

@atmamont how does this work on iOS now? Is the track event 100% linked to the UI being rendered or is it indirectly linked and the accuracy depends the merchant having a correct integration?

jreij: @OscarSpruit could we somehow use the `Content` function to track the render without having to pass callback to every single composable?

OscarSpruit: We could use something like a `LaunchedEffect` in `Content` to track. However, this could trigger multiple times during the flow (rotating screen, navigating back and forth, etc.). In v5 the render event can be sent only once during the flow, so that's why I tried keeping the behaviour as similar as possible

jreij: Can we trigger it inside `Content` but limit it to one event per instance of each component? I realize it will get complex but if we can encapsulate the logic in a clean and reusable way we might be able to track better data and catch weird bugs if a merchant is not rendering the view for some reason.

atmamont: On the iOS we rely on `UIViewController.vivDidLoad()` to trigger most of the events (with some minor exemption cases). However, as we give away the view controller to merchants, it's on them when they want to "load" it. They might misuse UIViewController lifecycle and just load up it unnecessarily early upon instantiation long before they present it. The other way of sending events on "will appear" UIViewController lifecycle event will cause multiple events sent on every time it "appears", which is not great as when there is any popover or a system alert presented over the view, the iOS triggers "will appear" every time resulting in event duplication.

I think in this case "wrong" timestamp is less evil than duplicated events.

jreij (dismissed):

jreij (approved):

PR #2877, opened 2026-07-10
v6 - Google Pay - Configuration

OscarSpruit:
## Description
Reworks the Google Pay configuration to support multiple payment method types and adds issuer country code filtering. Split into two commits:

- **Support multiple payment methods** — replaces the flat CARD-specific fields on `GooglePayConfiguration` with an extensible `allowedPaymentMethods: List<GooglePayPaymentMethodParameters>`. Introduces the public `GooglePayPaymentMethodParameters` abstract type (with a `Card` subclass) and a distinct internal resolved model (`GooglePayPaymentMethodParams`) so defaults are resolved once and downstream code stays null-safe. `GooglePayUtils` now builds one Google Pay `PaymentMethod` per configured entry.
- **Issuer country codes** — adds `allowedIssuerCountryCodes` / `blockedIssuerCountryCodes` to `GooglePayPaymentMethodParameters.Card`, plumbed through the mapper and serialized into `CardParameters` (JSON keys `allowedIssuerCountryCodes` / `blockedIssuerCountryCodes`).

Default resolution and the generated Google Pay request models were verified against v5; the only intentional difference is that `allowedCardNetworks` now falls back to all networks when the API brands resolve to an empty list.

## Checklist
- [ ] If applicable, make sure Breaking change label is added.
- [x] Code is unit tested
- [ ] Aligned public API changes with other platforms (if applicable)

## Ticket Number
COSDK-1116


googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayPaymentMethodParameters.kt
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Describes a payment method supported by Google Pay and its parameters. Pass a list of these to
+ * [GooglePayConfiguration.allowedPaymentMethods] to configure which payment methods are offered.
+ *
+ * This is a mapping of the
+ * [PaymentMethod](https://developers.google.com/pay/api/android/reference/request-objects#PaymentMethod)
+ * object from the Google Pay SDK. Each subclass corresponds to a supported `type` and carries the
+ * `parameters` specific to that type.
+ */
+abstract class GooglePayPaymentMethodParameters internal constructor() : Parcelable {
```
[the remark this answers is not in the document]
OscarSpruit: This is on purpose to be forward compatible

googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayConfiguration.kt
```diff
@@ -20,17 +20,11 @@ class GooglePayConfiguration internal constructor(
     val totalPriceStatus: String?,
     val countryCode: String?,
     val merchantInfo: MerchantInfo?,
-    val allowedAuthMethods: List<String>?,
-    val allowedCardNetworks: List<String>?,
-    val isAllowPrepaidCards: Boolean?,
-    val isAllowCreditCards: Boolean?,
-    val isAssuranceDetailsRequired: Boolean?,
+    val allowedPaymentMethods: List<GooglePayPaymentMethodParameters>?,
```
araratthehero: Here merchants might pass multiple params for cards, which according to Google Pay docs [here](https://developers.google.com/pay/api/android/reference/request-objects#PaymentDataRequest:~:text=Warning%3A%20Multiple%20payment%20methods%20of%20type%20CARD%20are%20not%20supported.) is not supported. What do you think of adding a logic to internally take the first of each type, or convert the list to a set?

Another idea is to keep this as is, but to document this properly, so merchants know they cannot pass Card twice.

googlepay/src/test/java/com/adyen/checkout/googlepay/internal/ui/model/GooglePayComponentParamsMapperTest.kt
```diff
@@ -216,7 +228,49 @@ internal class GooglePayComponentParamsMapperTest {
...
+                    allowedCardNetworks = listOf("MASTERCARD", "AMEX", "DISCOVER"),
+                ),
+            ),
+        )
+
+        assertEquals(expected, params)
+    }
+
+    @Test
+    fun `when multiple payment methods are configured then all are resolved and preserved`() {
```
araratthehero: Here there is an example of two different cards. In case we make a change this test needs to be adjusted too.

googlepay/src/main/java/com/adyen/checkout/googlepay/GooglePayConfiguration.kt
OscarSpruit: Hmm good point, let's see what we can do to prevent that from happening

araratthehero (approved):

PR #2878, opened 2026-07-13
v6 - Pay Button - Display amount on Pay button

OscarSpruit:
## Description
Consolidates the Pay button in `:core` and makes it amount-aware across v6 components.

- Moved `PayButton` into `:core` so all components share a single implementation.
- `PayButton` now derives its label from the payment `amount`, owning the branching in one place:
  - No amount → "Pay"
  - Zero amount → "Save details"
  - Positive amount → "Pay ${formatted amount}"
- Threaded `amount` through Card, stored Card, BLIK, and MB WAY view states and producers.
- Removed the Pay button and its UI from the stored BLIK component, which renders no UI.
- Added `CheckoutLocalizationKey` entries and default strings for the three button states (additive, non-breaking public API).
- Added unit tests for `Amount.format` and amount pass-through in the component view-state producers.

| non-zero | zero |
|--------|--------|
| image | image |

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1312


core/src/main/res/values/strings.xml
```diff
@@ -66,4 +66,7 @@
     <string name="checkout_blik_code_hint" translatable="false">6-digit code</string>
     <string name="checkout_blik_code_not_valid" translatable="false">Enter 6 numbers</string>
     <string name="checkout_blik_helper_text" translatable="false">Get the code from your banking app.</string>
+    <string name="checkout_pay_button_no_amount" translatable="false">Pay</string>
```
jreij, line 69: These strings need to be updated

OscarSpruit: Done

jreij: Nit: could you update the PR body to reflect the new strings?

jreij: Also `Pay <formatted amount>` is showing as "Pay ", probably `<formatted amount>` being resolved as html

core/src/main/java/com/adyen/checkout/core/components/internal/ui/PayButton.kt
```diff
@@ -0,0 +1,40 @@
...
+fun PayButton(
+    amount: Amount?,
+    onClick: () -> Unit,
+    isLoading: Boolean = false,
+) {
+    val text = when {
+        amount == null -> resolveString(CheckoutLocalizationKey.PAY_BUTTON_NO_AMOUNT)
+        amount.value == 0L -> resolveString(CheckoutLocalizationKey.PAY_BUTTON_ZERO_AMOUNT)
+        else -> resolveString(CheckoutLocalizationKey.PAY_BUTTON_WITH_AMOUNT, amount.format(LocalLocale.current))
+    }
```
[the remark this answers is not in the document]
araratthehero: I do agree that the best option here could be to just show `Pay` label and no amount

araratthehero, line 33: What do you think of moving this logic out of the Composable to a separate function and unit test it.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/StoredCardViewStateProducer.kt
```diff
@@ -9,11 +9,14 @@
 package com.adyen.checkout.card.internal.ui.state
 
 import com.adyen.checkout.card.internal.ui.model.SecurityCodeTrailingIcon
+import com.adyen.checkout.core.components.data.model.Amount
 import com.adyen.checkout.core.components.internal.ui.state.ViewStateProducer
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputComponentState
 import com.adyen.checkout.core.components.internal.ui.state.model.toViewState
 
-internal class StoredCardViewStateProducer : ViewStateProducer<StoredCardComponentState, StoredCardViewState> {
+internal class StoredCardViewStateProducer(
```
araratthehero, line 17: I'd add unit test for this class too, which could check that the amount is passed to the `StoredCardViewState`

core/src/main/java/com/adyen/checkout/core/components/internal/ui/PayButton.kt
OscarSpruit: It will be a bit difficult since `resolveString` needs composable context. I would suggest to not overcomplicate this right now and find a better way to drive strings from the component side later and test it then

card/src/main/java/com/adyen/checkout/card/internal/ui/state/StoredCardViewStateProducer.kt
OscarSpruit: This class does not have any tests yet. I will add them

core/src/main/java/com/adyen/checkout/core/components/internal/ui/PayButton.kt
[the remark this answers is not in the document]
OscarSpruit: Done this by changing the `amount.format()` to return an empty string in case of error

core/src/test/java/com/adyen/checkout/core/components/data/model/AmountTest.kt
```diff
@@ -125,6 +129,85 @@ internal class AmountTest {
...
+            val amount = Amount(currency = "ISK", value = 1337L)
+
+            val result = amount.format(Locale.US)
+
+            assertTrue(result.contains("13.37"), "Expected '13.37' in '$result'")
+        }
+
+        @Test
+        fun `when formatting an amount with invalid currency then empty string is returned`() {
+            // ISK defaults to 0 fraction digits on the platform, but Adyen defines 2
```
araratthehero: Nits: This comment is redundant
```suggestion
```

araratthehero (approved):

PR #2888, opened 2026-07-20
v6 - Core - Clean up error handling TODOs

OscarSpruit:
## Description
Cleans up error-handling TODOs across the action and core modules:
- Removes error propagation TODOs and replaces `RuntimeException` with `CheckoutException`.
- Enforces the base URL to be valid in `OkHttpClient`.
- Makes `ActionFactory` generic over the action type (updating `Authentication`, `Await`, and `Redirect` factories/components accordingly).
- Removes the TODO on resolving the Google Pay merchant account.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1343

### DIFF
--- a/authentication/src/test/java/com/adyen/checkout/authentication/internal/ui/AuthenticationComponentTest.kt
+++ b/authentication/src/test/java/com/adyen/checkout/authentication/internal/ui/AuthenticationComponentTest.kt
@@ -18,7 +18,6 @@ import com.adyen.checkout.authentication.internal.data.api.SubmitFingerprintRepo
 import com.adyen.checkout.authentication.internal.data.model.AuthenticationSerializer
 import com.adyen.checkout.authentication.internal.data.model.SubmitFingerprintResult
 import com.adyen.checkout.authentication.internal.ui.model.AuthenticationComponentParams
-import com.adyen.checkout.core.action.data.Action
 import com.adyen.checkout.core.action.data.ActionComponentData
 import com.adyen.checkout.core.action.data.RedirectAction
 import com.adyen.checkout.core.action.data.Threeds2Action
@@ -100,7 +99,7 @@ internal class AuthenticationComponentTest(
     }
 
     private fun createComponent(
-        action: Action = Threeds2Action(
+        action: Threeds2Action = threeds2Action(
             subtype = Threeds2Action.SubType.FINGERPRINT.value,
             token = Base64.encode(TEST_FINGERPRINT_TOKEN.toByteArray()),
         ),
@@ -130,25 +129,9 @@ internal class AuthenticationComponentTest(
     inner class HandleActionTest {
 
         @Test
-        fun `action is not Threeds2Action, then an error is emitted`() = runTest {
-            // GIVEN
-            val redirectAction = RedirectAction(url = "https://redirect.url")
-            component = createComponent(action = redirectAction)
-            component.initialize(this)
-            val eventFlow = component.eventFlow.test(testScheduler)
-
-            // WHEN
-            component.handleAction(Activity(), mock())
-
-            // THEN
-            val event = assertInstanceOf<ActionComponentEvent.Error>(eventFlow.latestValue)
-            assertEquals("Unsupported action", event.error.message)
-        }
-
-        @Test
         fun `subtype is null, then an error is emitted`() = runTest {
             // GIVEN
-            val action = Threeds2Action(subtype = null, token = "token")
+            val action = threeds2Action(subtype = null, token = "token")
             component = createComponent(action = action)
             component.initialize(this)
             val eventFlow = component.eventFlow.test(testScheduler)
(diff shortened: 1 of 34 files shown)


authentication/src/test/java/com/adyen/checkout/authentication/internal/ui/AuthenticationComponentTest.kt
```diff
@@ -1136,6 +1119,40 @@ internal class AuthenticationComponentTest(
                 AuthenticationEvents.Result.TIMEOUT,
             ),
         )
+
+        @Suppress("LongParameterList")
+        private fun threeds2Action(
```
jreij, line 1124: Small comment regardless for this PR - we use this pattern in many tests but I've recently found a way to reduce the boilerplate if the returned object is a data class and the mapping is fairly simple.

The function becomes something like this:
```kotlin
private fun threeds2Action() = Threeds2Action(
    type = Threeds2Action.ACTION_TYPE,
    paymentData = null,
    paymentMethodType = null,
    token = null,
    subtype = null,
    authorisationToken = null,
)
```

And then you call it like this:

```kotlin
val action = threeds2Action().copy(
    subtype = Threeds2Action.SubType.FINGERPRINT.value,
    token = Base64.encode(TEST_FINGERPRINT_TOKEN.toByteArray()),
    authorisationToken = "test-auth-token",
)
```

Of course you could argue that you're reducing boilerplate but adding an unnecessary `copy` call so I would leave it for you to use it in the future or not.

Also as a side note, can/should these action types be data classes?

core/src/main/java/com/adyen/checkout/core/components/data/model/paymentmethod/PaymentMethod.kt
```diff
@@ -25,30 +24,16 @@ abstract class PaymentMethod
...
         val SERIALIZER: Serializer<PaymentMethod> = object : Serializer<PaymentMethod> {
             override fun serialize(modelObject: PaymentMethod): JSONObject {
-                val paymentMethodType = with(modelObject.type) {
-                    if (isNullOrEmpty()) {
-                        // TODO - Error Propagation
-                        // throw CheckoutException("PaymentMethod type not found")
-                        throw RuntimeException("PaymentMethod type not found")
-                    } else {
-                        this
-                    }
```
araratthehero, line 39: I would remove this for `StoredPaymentMethod` too.

core/src/main/java/com/adyen/checkout/core/components/paymentmethod/PaymentMethodDetails.kt
```diff
@@ -21,37 +20,22 @@ import org.json.JSONObject
...
             }
 
             override fun deserialize(jsonObject: JSONObject): PaymentMethodDetails {
-                val actionType = jsonObject.getStringOrNull(TYPE)
-                if (actionType.isNullOrEmpty()) {
-                    // TODO - Error Propagation
-                    // throw CheckoutException("PaymentMethod type not found")
-                    throw RuntimeException("PaymentMethod type not found")
-                }
+                val actionType = jsonObject.getString(TYPE)
```
araratthehero: This could be renamed to:

```suggestion
                val paymentMethodType = jsonObject.getString(TYPE)
```

core/src/main/java/com/adyen/checkout/core/action/data/Threeds2Action.kt
```diff
@@ -13,12 +13,12 @@ import org.json.JSONObject
 
 @Parcelize
 class Threeds2Action(
```
araratthehero: Can this also be a data class like other action classes?

authentication/src/test/java/com/adyen/checkout/authentication/internal/ui/AuthenticationComponentTest.kt
OscarSpruit: Both ways are valid imo. We can keep it as is and use both patterns

core/src/main/java/com/adyen/checkout/core/components/data/model/paymentmethod/PaymentMethod.kt
OscarSpruit: Done

core/src/main/java/com/adyen/checkout/core/components/paymentmethod/PaymentMethodDetails.kt
OscarSpruit: Done

core/src/main/java/com/adyen/checkout/core/action/data/Threeds2Action.kt
OscarSpruit: Done

araratthehero (approved):

PR #2910, opened 2026-07-29
Align v5 and v6 build scripts

OscarSpruit:
## Description
Align v5 and v6 build scripts


build-logic/convention/src/main/kotlin/com/adyen/checkout/AggregateDependencyListsTask.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.gradle.api.provider.Property
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.Optional
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+
+abstract class AggregateDependencyListsTask : DefaultTask() {
+
+    @get:InputFiles
```
jreij, line 27: `:build-logic:convention:check` still fails in `validatePlugins` with strict validation enabled. This property needs a normalization annotation such as `@PathSensitive`, and the custom task types also need explicit cacheability intent (`@DisableCachingByDefault`, `@CacheableTask`, or `@UntrackedTask`); the current failures cover `AggregateDependencyListsTask`, `GenerateDependencyListTask`, and `ResolveDependenciesTask`. Since this PR introduces the build logic as production build code, can we make the included build check task pass?

build-logic/convention/src/main/kotlin/PublishConventionPlugin.kt
```diff
@@ -0,0 +1,177 @@
...
+                        configurations
+                            .matching {
+                                it.name.startsWith("releaseTestFixturesVariant") &&
+                                    (it.name.endsWith("ApiPublication") || it.name.endsWith("RuntimePublication"))
+                            }
+                            .forEach { testFixturesConfiguration ->
+                                releaseComponent.withVariantsFromConfiguration(testFixturesConfiguration) { skip() }
+                            }
+
+                        from(releaseComponent)
```
jreij, line 105: Publishing from the Android `release` component still changes consumer-visible metadata. A regenerated `:card` POM/GMM includes `org.jetbrains.kotlin:kotlin-stdlib:2.0.20` while this branch uses Kotlin `1.9.25`, and also exposes AGP/view binding metadata such as `androidx.databinding:viewbinding:8.8.1`. Can we preserve the old published dependency metadata for v5, or confirm these new transitive dependencies and versions are intentional?

build-logic/convention/src/main/kotlin/CheckoutAndroidLibraryPlugin.kt
```diff
@@ -0,0 +1,149 @@
...
+                    buildConfigField("String", "CHECKOUT_VERSION", "\"${rootProject.extraProperties["versionName"]}\"")
+                }
+
+                buildFeatures {
+                    buildConfig = true
+                }
+
+                publishing {
+                    singleVariant("release") {
+                        withSourcesJar()
```
jreij, line 54: The AGP `withSourcesJar()` / `withJavadocJar()` artifacts still include generated implementation details such as `BuildConfig` and data binding classes. The previous custom artifacts did not include these generated classes, so this changes published source/javadoc contents for consumers. Can we preserve the old artifact contents for v5, or confirm this is intentional?

build-logic/convention/src/main/kotlin/CheckoutAndroidLibraryPlugin.kt
```diff
@@ -0,0 +1,149 @@
...
+            apply(plugin = "checkout.kover")
+            apply(plugin = "checkout.ktlint")
+            apply(plugin = "checkout.publish")
+            apply(plugin = "checkout.sonar")
+
+            extensions.configure<LibraryExtension> {
+                compileSdk = libs.versions.compile.sdk.get().toInt()
+
+                defaultConfig {
+                    minSdk = libs.versions.min.sdk.get().toInt()
```
jreij, line 39: The new library convention sets `compileSdk` and `minSdk`, but drops `targetSdk` from published Android library modules. Even if this does not currently change the processed manifest, it is a build configuration change from v5; can we add `targetSdk = libs.versions.target.sdk.get().toInt()` here unless the removal is intentional?

build-logic/convention/src/main/kotlin/com/adyen/checkout/AggregateDependencyListsTask.kt
OscarSpruit: Done

build-logic/convention/src/main/kotlin/CheckoutAndroidLibraryPlugin.kt
OscarSpruit: Added back the target SDK. If we ever upgrade to AGP 9 we have to remove it again, because it is no longer part of that API

build-logic/convention/src/main/kotlin/PublishConventionPlugin.kt
OscarSpruit: This new way is much better as it actually uses resolved versions to include in the metadata. The old way was actually including inaccurate data because of this. I would say we keep this new way

build-logic/convention/src/main/kotlin/CheckoutAndroidLibraryPlugin.kt
OscarSpruit: This new way is actually an improvement. Those generated classes ending up in the artefacts is just cosmetic and harmless. Would keep this as is as well

jreij (approved):

PR #2913, opened 2026-07-31
v6 - Agents - Dependency change guidance

jreij:
## Description
Nothing in our agent guidance asked an agent to justify taking a dependency, or to update `.github/release_notes_dependency_list.toml` when adding one — which CI rejects on every PR. Neither file was mentioned in any markdown in the repo.

Guidance is added in three layers so it applies regardless of how a task is framed:

- **`AGENTS.md` — new "Adding or Modifying Dependencies" section.** Always-on context, so it also applies when a dependency change is a side effect of unrelated work. Leads with the approval gate, then the release notes list rules. Adds a verification checklist item.
- **New `android-add-dependency` skill.** Step 1 blocks on an impact assessment (scope, merchant impact, public API exposure, R8/ProGuard, build/CI, tests, licensing, rejected alternatives) before any edit. Then the mechanics: deriving the dependency id, choosing `[included]` vs `[excluded]`, writing the release notes link, and regenerating verification metadata.
- **`android-commit` gate.** If `gradle/libs.versions.toml` is staged, verify the change was approved and that both the dependency list and verification metadata are updated.

Also fixes two stale skill paths left over from the pre-`SKILL.md` layout.

Markdown only — no code, Gradle, or resource changes.

## Checklist

### DIFF
--- a/.agents/skills/android-add-dependency/SKILL.md
+++ b/.agents/skills/android-add-dependency/SKILL.md
@@ -1,0 +1,143 @@
+---
+name: android-add-dependency
+description: Decide on and apply a dependency change — impact assessment, libs.versions.toml, release notes dependency list, and verification metadata.
+---
+
+# android-add-dependency
+
+Decide whether to add, change, or remove a dependency in the Adyen Android SDK, then apply it without breaking release notes generation or dependency verification.
+
+## Usage
+
+Invoke this skill **before** touching `gradle/libs.versions.toml` — as soon as it looks like work will need a new library or plugin, a different version or artifact, or the removal of an existing one. This applies even when the dependency change is a side effect of a larger task, and even when the developer named the dependency themselves.
+
+Renovate handles routine version bumps. This skill is for dependency changes made by hand.
+
+## Steps
+
+### 1. Stop and get approval for the dependency itself
+
+**Do not edit any file yet.** This is a decision, not a mechanical step. We are an SDK: every dependency we take is imposed on every merchant that integrates us, and every one we drop or re-point can break them. The default answer to a new dependency is no.
+
+**First, try to avoid it:**
+
+- Does the Android platform, AndroidX, or something already in `gradle/libs.versions.toml` cover this?
+- Is the part we actually need small enough to implement ourselves? For published modules, prefer a small internal implementation over a new dependency.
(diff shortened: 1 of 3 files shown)


.agents/skills/android-add-dependency/SKILL.md
```diff
@@ -0,0 +1,143 @@
+---
+name: android-add-dependency
+description: Decide on and apply a dependency change — impact assessment, libs.versions.toml, release notes dependency list, and verification metadata.
```
araratthehero: I would keep the description short and more explicit for the agents to not get confused. What needs to be done can go in the description of the skill instead of in its description.
```suggestion
description: Add, remove or update an external dependency in the SDK.

File: AGENTS.md [hunk 2]
```diff
@@ -285,6 +285,33 @@ if (checkCompileOnly("com.external.sdk.SomeClass")) {
...
+
+**Every dependency in `gradle/libs.versions.toml` must also be listed in `.github/release_notes_dependency_list.toml`:**
+- `[included]` — merchant-relevant dependencies: shipped in a published module, or part of the toolchain merchants must be compatible with (AGP and Kotlin). The value is a markdown link to the vendor's release notes, with `{}` where the version goes. These entries are published in our release notes.
+- `[excluded]` — everything else (tests, example app, build logic, CI, code quality tooling), plus merchant-relevant dependencies already covered by another entry such as a BoM. The value is an empty string.
+
+**Always ask the developer to confirm the section and the link before writing the entry.** Never guess a release notes URL.
+
+CI enforces this: a newly added dependency that appears in neither list fails the `validate_dependencies` check on every PR.
+
+Dependency verification is enabled, so a new dependency also requires regenerating `gradle/verification-metadata.xml` or the build fails on a missing checksum.
```
araratthehero: This part of the section looks like a duplicate to what we have in the `android-add-dependency` skill. What do you think of removing this from here, but mentioning that if a new dependency needs to be added, an existing dependency needs to be updated or removed, then the `android-add-dependency` skill should be used. This way we don't need to maintain the same information in two different places.

.agents/skills/android-add-dependency/SKILL.md
jreij: Fixed

AGENTS.md
jreij: Fixed

araratthehero (approved):

PR #2922, opened 2026-08-03
v6 - Clean up state layer

jreij:
## Description
This PR is a set of small cleanups to the form field state layer used by text-input-based components (Card, MBWay, Blik), fixing a couple of latent correctness issues along the way. No behavior change is intended beyond the two bug fixes called out below, and there is no public API change.

- Deduplicated the should-show-error-message rule: it was hand-written in five places alongside the same rule in toViewState(), and is now a single derived property on TextInputComponentState so the state and view layers cannot drift apart. updateFocus is also simplified to assign the boolean directly.
- Deleted StateManager, DefaultStateManager, ViewStateFactory and ViewStateValidator, which had no production callers. Components build their state on ComponentStateFlow instead.
- Added characterisation tests for ComponentStateFlow's reduce-validate-emit pipeline, which previously had no coverage.
- Fixed a bug where the initial component state was never validated (runningFold emitted it without validation), so an untouched form reported itself valid. This only avoided causing visible issues because every component auto-focuses its first field, which happens to trigger validation.
- Unified focus semantics on isFocused: MBWayPhoneNumberField and BlikCodeField read focusState.hasFocus while all card fields read focusState.isFocused. These diverge once a field composable wraps focusable children, which the dual brand card number field already does.
- Stopped crashing when a focus request cannot be honoured (requestFocus() throws if the FocusRequester is not attached to a focusable node, which can happen if the field leaves composition between composition and the effect running). This is only safe now that the initial state is validated on construction.

## Checklist
- [x] Code is unit tested


ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextField.kt
```diff
@@ -159,7 +159,9 @@ fun CheckoutTextField(
 
     LaunchedEffect(shouldFocus) {
         if (shouldFocus) {
-            focusRequester.requestFocus()
+            // Throws if the requester is not attached to a focusable node, which can happen when the field leaves
+            // composition between composition and this effect running. Losing the focus is preferable to crashing.
+            runCatching { focusRequester.requestFocus() }
```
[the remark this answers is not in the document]
OscarSpruit: Since this is not a suspending call and we don't care about the error I think `runCatching` is fine

blik/src/main/java/com/adyen/checkout/blik/internal/ui/view/BlikCodeField.kt
```diff
@@ -52,7 +52,7 @@ internal fun BlikCodeField(
         modifier = modifier
             .fillMaxWidth()
             .onFocusChanged { focusState ->
-                onFocusChange(focusState.hasFocus)
+                onFocusChange(focusState.isFocused)
```
OscarSpruit, line 55: Did you test if this made any difference?

jreij: It only matters if a field composable wraps focusable children, such as the card number field with the brands and camera icon. So for Blik and MBWay there was no issue anyway but this fix ensures we are using the right field, for reference to future components.

OscarSpruit (dismissed):

OscarSpruit (approved):

PR #2923, opened 2026-08-03
v6 - Core - Fix shopper input loss on backgrounding in ComponentStateFlow

jreij:
## Description

`ComponentStateFlow` backed the accumulated form state with `runningFold` over a `Channel`, shared via `stateIn(WhileSubscribed(5s))`. Backgrounding the app (`collectAsStateWithLifecycle` unsubscribes below `STARTED`) tears that subscription down after 5s per layer (~10s total once `viewState()`'s own `WhileSubscribed` is stacked on top). The next collection restarts the fold from `initialState` with no replay, silently discarding everything the shopper typed.

#### How to reproduce
Example app → MB Way → type a phone number → background the app for 15s → return to the app. The phone number field is empty. Returning within 9s preserves it.
`CardComponent` was accidentally immune because a permanent analytics subscriber kept the sharing coroutine alive.

#### Fix
Own the state directly in a `MutableStateFlow`. `handleIntent()` updates it synchronously, so there is no subscription to tear down and nothing to discard. This also drops the now-unused `coroutineScope` parameter from `ComponentStateFlow(...)` — safe to remove because it was only ever used to launch the `stateIn` sharing coroutine for the internal fold, not as a lifecycle guard; every place that legitimately ties its lifetime to the merchant-owned scope (`viewState()`'s own `stateIn`, and `CardComponent`'s `.launchIn(coroutineScope)` analytics/bin-callback collectors) still receives and uses it unchanged.

#### Why `viewState()` still uses `WhileSubscribed` safely
`WhileSubscribed(5s)` is the Android-team-recommended default for `stateIn`, and it was originally applied to both `stateIn` calls in this file without noticing that using it on the state *accumulator* itself was unsafe. `viewState()` doesn't have that problem: its `stateIn` wraps a pure re-derivation (`producer.produce(state)`) of `componentState`, which is now a hot, conflated `MutableStateFlow` that retains its value with no subscribers needed. If `viewState()`'s sharing coroutine is torn down and restarts, it just re-subscribes to `componentState` and recomputes — cheap, pure, no I/O, and nothing to lose because `viewState()` was never the one holding the data. Concretely: while the sharing coroutine is stopped, `viewState.value` simply holds its last emitted value frozen in place; `WhileSubscribed`'s `replayExpirationMillis` defaults to infinite, and we don't override it, so it only ever emits `STOP` (keep the replay cache) and never `STOP_AND_RESET_REPLAY_CACHE` (reset to the initial seed). `componentState` keeps accepting `handleIntent()` calls the entire time regardless, so when a new collector appears there's nothing stale to reconcile — the map just recomputes from the current truth.

#### Discarded alternative
`SharingStarted.Eagerly` instead of `WhileSubscribed` — a one-line fix that keeps the fold alive forever. Rejected because the state would still live inside a flow operator chain "by promise" rather than by construction, and the `Channel` round-trip would leave `componentState.value` briefly stale right after `handleIntent()` returns.

#### Precondition
`MutableStateFlow.update {}` can re-invoke its lambda under contention, so `reducer.reduce`/`validator.validate` must stay pure — true today for all reducers.

## Checklist
- [x] Code is unit tested

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
@@ -11,14 +11,13 @@ package com.adyen.checkout.core.components.internal.ui.state
 import androidx.annotation.RestrictTo
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalForInheritanceCoroutinesApi
-import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.map
-import kotlinx.coroutines.flow.receiveAsFlow
-import kotlinx.coroutines.flow.runningFold
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.flow.update
 
 @Suppress("UnnecessaryOptInAnnotation")
 @OptIn(ExperimentalForInheritanceCoroutinesApi::class)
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
```diff
@@ -32,47 +31,42 @@ fun <C : ComponentState, I : ComponentStateIntent> ComponentStateFlow(
...
-    coroutineScope = coroutineScope,
 )
 
+// Owns its state directly in a MutableStateFlow rather than folding intents over a cold flow. A cold flow's
+// accumulator is torn down by WhileSubscribed once the last collector goes away (e.g. the app is backgrounded for
+// longer than the sharing timeout), and nothing replays the intents handled while nobody was collecting - so
+// shopper input would silently be discarded. Owning the state directly means there is no subscription whose
+// cancellation could lose anything.
+//
+// This requires reducer.reduce and validator.validate to be pure (no side effects), because MutableStateFlow.update
```
OscarSpruit, line 46: How can we ensure there will be no side effects?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
```diff
@@ -32,47 +31,42 @@ fun <C : ComponentState, I : ComponentStateIntent> ComponentStateFlow(
...
+        get() = state.replayCache
 
     override suspend fun collect(collector: FlowCollector<C>): Nothing {
-        stateFlow.collect(collector)
+        state.collect(collector)
     }
 
     override fun handleIntent(intent: I) {
-        intents.trySend(intent)
+        state.update { validator.validate(reducer.reduce(it, intent)) }
```
OscarSpruit, line 69: This now makes updating the state synchronous, so it runs on the callers thread (when called from compose on the UI thread). It should be fine, but if we ever do more expensive operations we might want to reconsider

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
jreij: I think the current interface of the reducer and validator already ensure this because it forces you to return a state - `reducer.reduce` takes a state and intent and returns a state, and `validator .validate` takes a state and returns a state. And they are instantiated in the factory so to introduce a side effect you really need to go out of your way to inject some unrelated object or callback into the validator/reducer. Does that make sense?

Still the comment itself is good for AI agents building and reviewing I think.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
jreij: Given that both the validator and reducer do not have suspendable functions, we should be fine on this point with the current structure right?

OscarSpruit: For now it's fine

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/ComponentStateFlow.kt
OscarSpruit: True

OscarSpruit (approved):

PR #2926, opened 2026-08-03
v6 - Errors - Convert creation errors to CheckoutException and report unsupported actions

araratthehero:
## Description
Internal errors could escape to merchants unconverted, and a missing action module crashed the payment flow. This PR closes both gaps.

**Convert internal errors escaping controller creation.** `CheckoutControllerFactory` now wraps all three `create` overloads in `runCatchingCreation`, which catches `InternalCheckoutError` and rethrows it as the public `CheckoutException`. Only `InternalCheckoutError` is caught on purpose — broken invariants and programming errors still fail fast, so they surface as bugs instead of being reported as checkout errors. This is what makes the Google Pay `InvalidConfigurationError` from the previous PR reach merchants as a `CheckoutException` rather than an internal type.

**Report an unsupported action type instead of crashing.** `ActionComponentProvider.get` threw `IllegalStateException` when no factory was registered for an action type. That crashed the flow mid-payment: it happens after `/payments` has already returned an action, inside an SDK-owned coroutine where no merchant stack frame exists to catch it, so the merchant received no callback at all while a payment was in flight. It now throws `GenericError`, and `ActionHandler.handleAction` catches it and dispatches `onFailure`. This aligns actions with payment methods, which already degrade gracefully through `PaymentComponentResult.Failure`. The usual cause is a merchant not including the module for that action type, so the message points at build dependencies.

**Document the error handling model.** Added KDoc on `InternalCheckoutError` describing the three tiers: configuration validation returning a `CheckoutError`, internal runtime errors converted through `toCheckoutError()`, and plain runtime exceptions reserved for invariants and integration bugs.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1002

### Fixed
- An unsupported action type no longer crashes the payment flow. The error is now delivered through `onFailure`, with a message indicating that the corresponding module is missing from the build dependencies.

### Changed
- Internal errors that occur while creating a checkout controller are now surfaced as a `CheckoutException` instead of an internal error type.

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutControllerFactory.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutControllerFactory.kt
@@ -24,6 +24,9 @@ import com.adyen.checkout.core.components.CheckoutController
 import com.adyen.checkout.core.components.CheckoutTarget
 import com.adyen.checkout.core.components.SessionCheckoutCallbacks
 import com.adyen.checkout.core.components.internal.data.provider.DefaultSdkDataProvider
+import com.adyen.checkout.core.error.CheckoutException
+import com.adyen.checkout.core.error.internal.InternalCheckoutError
+import com.adyen.checkout.core.error.toCheckoutError
 import com.adyen.checkout.core.sessions.internal.data.api.SessionRepository
 import com.adyen.checkout.core.sessions.internal.data.api.SessionService
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutControllerFactory.kt
```diff
@@ -89,13 +92,29 @@ internal class CheckoutControllerFactory {
...
+    /**
+     * Runs the controller creation, converting any [InternalCheckoutError] that escapes into the
+     * public [CheckoutException].
+     *
+     * Only [InternalCheckoutError] is caught on purpose. Programming errors and broken invariants
+     * are expected to fail fast so that they surface as bugs instead of being reported as checkout
+     * errors.
+     */
+    // The caught error is not swallowed, toCheckoutError() keeps it as the cause of the thrown exception.
+    @Suppress("SwallowedException")
```
OscarSpruit, line 111: Is this suppression really needed? It seems like you are using the caught exception

araratthehero: I checked and detekt still fails without it.
```
CheckoutControllerFactory.kt:112:14: The caught exception is swallowed.
The original exception could be lost. [SwallowedException]
```

It looks like the rule only treats the exception as propagated when it's passed directly to the thrown exception.

jreij (approved):

PR #2929, opened 2026-08-03
v6 - Nightly test build

OscarSpruit:
## Description
This PR ensures a nightly test build will be released with the name "nightly" and the commit hash as description.

### DIFF
--- a/.github/workflows/release_acceptance_app.yml
+++ b/.github/workflows/release_acceptance_app.yml
@@ -1,6 +1,9 @@
 name: Release acceptance app
 
 on:
+  schedule:
+    # Every work day (Mon-Fri) at 00:00 UTC
+    - cron: '0 0 * * 1-5'
   workflow_dispatch:
     inputs:
       version_name:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_acceptance_app.yml
```diff
@@ -42,7 +45,7 @@ jobs:
 
       - name: Build with Gradle
         env:
-          VERSION_NAME: ${{ github.event.inputs.version_name }}
+          VERSION_NAME: ${{ github.event.inputs.version_name || 'nightly' }}
```
jreij, line 48: Is this aligned with iOS?

jreij: I feel like the name should reflect that it's the latest changes in development rather than the schedule

OscarSpruit: @atmamont suggested this and I agree it's a common naming used for this purpose

.github/workflows/release_acceptance_app.yml
```diff
@@ -1,6 +1,9 @@
 name: Release acceptance app
 
 on:
+  schedule:
+    # Every night at 00:00 UTC
+    - cron: '0 0 * * *'
```
jreij: Should we care about when we release contains no changes (weekend or when we didn't merge anything?)

jreij (dismissed):

.github/workflows/release_acceptance_app.yml
OscarSpruit: Good point! Detecting changes makes it a bit complicated, but we can easily skip the weekends

nauaros: We could add only from mon to fri

jreij (approved):

PR #2937, opened 2026-08-06
v6 - Core - Centralize text field trailing icon error state handling

jreij:
## Description
Fixes a bug where only 4 text fields showed the generic error icon on validation errors and the rest did not.

- `TextInputViewState.trailingIcon` now resolves to `TrailingIcon.Error` whenever the field is showing an error, so the icon is decided in one place instead of each field opting in.
- `CheckoutTextFieldTrailingIcon` renders it and animates the transitions between icons.
- Field specific icons (brand logos, scan button, security code and expiry date placeholders) are now produced independently of the error state.
- `trailingIcon` is now required on `CheckoutTextField`, so a new field cannot silently omit it. Fields that render their own icon, like the search field, are unaffected.

**Trade-off:** a field whose icon resolves to `TrailingIcon.Empty` still emits an empty `AnimatedContent`, which takes up 8dp of trailing space. Not emitting it would drop the animation between the empty and error states.

## Checklist
- [x] Code is unit tested

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TrailingIcon.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TrailingIcon.kt
@@ -9,6 +9,29 @@
 package com.adyen.checkout.core.components.internal.ui.state.model
 
 import androidx.annotation.RestrictTo
+import com.adyen.checkout.core.common.internal.ui.CheckoutTextFieldTrailingIcon
 
+/**
+ * The trailing icon to be displayed at the end of a text field.
+ *
+ * [Empty] and [Error] are generic states handled by [CheckoutTextFieldTrailingIcon] for every field. Fields that need
+ * their own icons declare additional states by extending this class.
+ *
+ * Implementations must be comparable by equality, as they are used as the target state of an animation. Prefer
+ * `data object` or `data class`.
+ */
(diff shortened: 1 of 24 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TrailingIcon.kt
```diff
@@ -10,5 +10,28 @@ package com.adyen.checkout.core.components.internal.ui.state.model
 
 import androidx.annotation.RestrictTo
 
+/**
+ * The trailing icon to be displayed at the end of a text field.
+ *
+ * [Empty] and [Error] are generic states handled by
+ * [com.adyen.checkout.core.common.internal.ui.CheckoutTextFieldTrailingIcon] for every field. Fields that need their
```
araratthehero: Nits: You could import `com.adyen.checkout.core.common.internal.ui.CheckoutTextFieldTrailingIcon` and write `CheckoutTextFieldTrailingIcon` in here, instead of writing the full path in the comment section.

araratthehero (dismissed):

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TrailingIcon.kt
jreij: Fixed

araratthehero (approved):

PR #2942, opened 2026-08-10
v6 - Core - Improve error visibility naming

jreij:
## Description
`TextInputComponentState` contained an error message and a separate flag for whether to display it. That forced both properties to be read together everywhere and allowed for an illogical state where an error doesn't exist but should be shown.

To address this, `TextInputComponentState.InputError` is introduced. It holds the message and its visibility together, so:
  - The meaningless state above cannot be reached anymore
  - Validators update the error without changing the visibility - by calling `updateError()`
  - Whoever creates an error can now decide whether it is visible as well. So in the future, an error that should appear immediately rather than waiting for a focus change can still be supported

This PR also replaces all `errorMessage == null` / `!= null` checks with `isValid`.

No behaviour changes were introduced.

### Reviewing
Split into four commits so each can be read on its own. Only the last one compiles.

1. **The type** — the change worth reading closely
2. **Its tests** — the three states, and that `updateError` preserves visibility
3. **Validators and reducers** — the same edit repeated across four components
4. **Remaining tests** — likewise

### Progress
➡️ **Phase 1 — Core: the error type (this PR)**
Phase 2 — Core: form unit, focus handshake, IME plumbing
Phase 3 — Card: field ids and order
Phase 4 — Card: reducer and validator
Phase 5 — Card: view state and UI
Phase 6 — Roll out to stored card, MB Way, BLIK
Phase 7 — Cleanup and document

## Checklist
- [x] Code is unit tested

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
+++ b/core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
@@ -9,32 +9,60 @@
 package com.adyen.checkout.core.components.internal.ui.state.model
 
 import androidx.annotation.RestrictTo
+import androidx.annotation.VisibleForTesting
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class TextInputComponentState(
     val text: String = "",
     val description: CheckoutLocalizationKey? = null,
-    val errorMessage: CheckoutLocalizationKey? = null,
+    val error: InputError? = null,
     val isFocused: Boolean = false,
-    val showError: Boolean = false,
     val requirementPolicy: RequirementPolicy = RequirementPolicy.Required,
 ) {
 
+    /**
+     * An error on the field, together with whether the shopper should see it yet.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    data class InputError(
+        val message: CheckoutLocalizationKey,
+        val isVisible: Boolean = false,
+    )
+
     val isValid: Boolean
-        get() = errorMessage == null
+        get() = error == null
+
+    val isErrorVisible: Boolean
+        get() = error?.isVisible == true
+
+    fun updateText(text: String) = copy(text = text).hideErrorIfPresent()
 
     /**
-     * Whether the error message should be displayed, which requires both an error to be present and the field to be
-     * showing its errors.
+     * Replaces the error, keeping whether it is currently visible. Validators call this on every pass, so replacing a
+     * message must not hide an error the shopper is already looking at.
      */
-    val showErrorMessage: Boolean
-        get() = showError && errorMessage != null
+    fun updateError(message: CheckoutLocalizationKey?): TextInputComponentState {
+        return when {
+            message == null -> copy(error = null)
+            error == null -> copy(error = InputError(message, isVisible = false))
+            else -> copy(error = InputError(message, isVisible = error.isVisible))
+        }
+    }
 
-    fun updateText(text: String) = copy(text = text, showError = false)
+    /**
+     * Shows the error, if there is one.
+     */
+    fun showErrorIfPresent() = copy(error = error?.copy(isVisible = true))
 
-    fun updateFocus(hasFocus: Boolean) = copy(
-        isFocused = hasFocus,
-        showError = !hasFocus,
-    )
+    /**
+     * Hides the error, if there is one.
+     */
+    @VisibleForTesting
+    fun hideErrorIfPresent() = copy(error = error?.copy(isVisible = false))
+
+    fun updateFocus(hasFocus: Boolean): TextInputComponentState {
+        val focused = copy(isFocused = hasFocus)
+        return if (hasFocus) focused.hideErrorIfPresent() else focused.showErrorIfPresent()
+    }
 }
--- a/core/src/test/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateTest.kt
+++ b/core/src/test/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateTest.kt
@@ -11,293 +11,326 @@ package com.adyen.checkout.core.components.internal.ui.state.model
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Assertions.assertNull
 import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
 
 /**
  * Tests `TextInputComponentState.kt`: how the state of a field evolves as the shopper types, focuses and blurs, and
- * when that results in an error being displayed. Some tests observe the outcome through `toViewState`, as that is what
- * the UI ends up rendering.
+ * when that results in an error being displayed. A few tests observe the outcome through `toViewState`, where the
+ * point is the transition rather than the mapping.
  *
  * Tests for the mapping itself live in `TextInputViewStateTest`.
  */
 internal class TextInputComponentStateTest {
 
-    // UC1: Error on Explicit Validation
-    @Test
-    fun `when state has error message and showError is true, then error is displayed`() {
-        // GIVEN
-        val state = TextInputComponentState(
-            text = "invalid",
-            errorMessage = CheckoutLocalizationKey.CARD_NUMBER_INVALID,
-            showError = true,
-        )
+    @Nested
+    inner class IsValidTest {
 
-        // WHEN
-        val viewState = state.toViewState()
+        // UC1: Error on Explicit Validation - verify isValid
+        @Test
+        fun `when state has an error, then isValid returns false`() {
+            // GIVEN
+            val state = TextInputComponentState(text = "invalid", error = hiddenError())
 
-        // THEN
-        assertEquals(true, viewState?.isError)
-        assertEquals(CheckoutLocalizationKey.CARD_NUMBER_INVALID, viewState?.supportingText)
+            // WHEN
+            val isValid = state.isValid
+
+            // THEN
+            assertFalse(isValid)
+        }
+
+        // UC14: Empty Field - verify empty is considered valid
+        @Test
+        fun `when field is empty with no error, then isValid returns true`() {
+            // GIVEN
+            val state = TextInputComponentState(text = "", error = null)
+
+            // WHEN
+            val isValid = state.isValid
+
+            // THEN
+            assertTrue(isValid)
+        }
     }
 
-    // UC1: Error on Explicit Validation - verify isValid
-    @Test
-    fun `when state has error message, then isValid returns false`() {
-        // GIVEN
-        val state = TextInputComponentState(
-            text = "invalid",
-            errorMessage = CheckoutLocalizationKey.CARD_NUMBER_INVALID,
-        )
+    /**
+     * There are only three states, because [TextInputComponentState.InputError] holds the message and its visibility
+     * together. A field that shows an error without having one cannot be built.
+     */
+    @Nested
+    inner class IsErrorVisibleTest {
 
-        // WHEN
-        val isValid = state.isValid
+        @Test
+        fun `when there is no error, then it is not visible`() {
+            // GIVEN
+            val state = TextInputComponentState(error = null)
 
-        // THEN
-        assertFalse(isValid)
+            // WHEN
+            val isErrorVisible = state.isErrorVisible
+
+            // THEN
+            assertFalse(isErrorVisible)
+        }
+
+        @Test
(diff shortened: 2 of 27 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -26,15 +31,23 @@ data class TextInputComponentState(
 
     /**
      * Whether the error message should be displayed, which requires both an error to be present and the field to be
-     * showing its errors.
+     * allowed to show it.
+     */
+    val showError: Boolean
+        get() = shouldDisplayErrorIfInvalid && !isValid
+
+    fun updateText(text: String) = copy(text = text).preventErrorDisplay()
```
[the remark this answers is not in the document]
jreij: Not taking this one. The extra `copy()` is a non-issue — it happens once per keystroke on a six-field data class, not in a hot loop — and the point of building `updateText` on `preventErrorDisplay()` is that "stop showing an error" has exactly one definition. Inlining `shouldDisplayErrorIfInvalid = false` here adds a second copy of that rule, which is the duplication the previous cleanup removed for `showError && errorMessage != null`.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -26,15 +31,23 @@ data class TextInputComponentState(
...
+     * Stops the field from showing an error message, hiding any error it is currently showing.
+     */
+    fun preventErrorDisplay() = copy(shouldDisplayErrorIfInvalid = false)
 
-    fun updateFocus(hasFocus: Boolean) = copy(
-        isFocused = hasFocus,
-        showError = !hasFocus,
-    )
+    fun updateFocus(hasFocus: Boolean) = copy(isFocused = hasFocus)
+        .run { if (hasFocus) preventErrorDisplay() else allowErrorDisplay() }
```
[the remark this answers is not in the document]
jreij: Same reasoning as above, plus one more: `shouldDisplayErrorIfInvalid = !hasFocus` hardcodes the assumption that gaining focus always hides the error. That is exactly the assumption an upcoming phase has to break, because focus moved by pressing Pay must keep the error it just revealed. Writing it as `!hasFocus` makes that look like an invariant when it is not.

I did take the readability half of the point — the `.run { if … }` was awkward, so it is now an explicit local and an if/else.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -26,15 +31,25 @@ data class TextInputComponentState(
...
      */
-    val showErrorMessage: Boolean
-        get() = showError && errorMessage != null
+    fun allowErrorDisplay() = copy(shouldDisplayErrorIfInvalid = true)
 
-    fun updateText(text: String) = copy(text = text, showError = false)
+    /**
+     * Stops the field from showing an error message, hiding any error it is currently showing.
+     */
+    fun preventErrorDisplay() = copy(shouldDisplayErrorIfInvalid = false)
```
araratthehero: Do these functions need to be public, if they are only used inside of the `updateFocus`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -26,15 +31,25 @@ data class TextInputComponentState(
...
 
-    fun updateFocus(hasFocus: Boolean) = copy(
-        isFocused = hasFocus,
-        showError = !hasFocus,
-    )
+    fun updateFocus(hasFocus: Boolean): TextInputComponentState {
+        val focused = copy(isFocused = hasFocus)
+        return if (hasFocus) focused.preventErrorDisplay() else focused.allowErrorDisplay()
```
araratthehero: What do you think of writing this like:
```suggestion
    fun updateFocus(hasFocus: Boolean) = copy(
        isFocused = hasFocus,
        shouldDisplayErrorIfInvalid = !hasFocus,
    )
```

araratthehero: I see that this is related to Gemini's suggestion [here](https://github.com/Adyen/adyen-android/pull/2942/changes#r3748251705).

core/src/test/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateTest.kt
```diff
@@ -134,19 +241,19 @@ internal class TextInputComponentStateTest {
...
         )
 
         // WHEN
         val updatedState = state.updateText("new")
 
         // THEN
-        assertFalse(updatedState.showError)
+        assertFalse(updatedState.shouldDisplayErrorIfInvalid)
         assertEquals("new", updatedState.text)
     }
```
araratthehero: This tests looks like a duplicate of `when the value changes, then the field is not allowed to show an error`. Also some tests are nested and some not in this file. Could you check if the tests could be cleaned up here?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -26,15 +31,25 @@ data class TextInputComponentState(
 
     /**
      * Whether the error message should be displayed, which requires both an error to be present and the field to be
-     * showing its errors.
+     * allowed to show it.
+     */
+    val showError: Boolean
+        get() = shouldDisplayErrorIfInvalid && !isValid
```
araratthehero: The name `showError` to me looks odd. It sounds like a command, but it is just a boolean. What do you think of renaming this to something like `isError` or `isErrorVisible`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: They can be private for this PR but they'll need to become public in the next one(s)

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: Yes indeed, I think this PR by itself is not enough to see the full picture, would you prefer if I push the next phase in this same PR instead of a separate one?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: what do you think of `shouldDisplayError` to match `shouldDisplayErrorIfInvalid`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: Actually I made it visible for testing only, the next PR can remove it

core/src/test/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentStateTest.kt
jreij: Fixed the whole file into nested classes with more clarity and less duplication

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: Actually this whole structure makes little sense, we don't need 2 booleans. Technically we have 3 states:
- No error
- Error that should be shown now
- Error that should not be shown yet

I reconsidered this and decided to go with an Error class containing a message and a boolean.
With two booleans we could have a scenario where the error must be shown but there is no actual error. I really can't think of any use case where this is valid, so I think the Error class can actually cover all our needs.

Let me know if this is easier to read now.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
araratthehero: No need, we can leave it as is and I can pay attention to this in the next PR

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -9,32 +9,60 @@
...
     val isFocused: Boolean = false,
-    val showError: Boolean = false,
     val requirementPolicy: RequirementPolicy = RequirementPolicy.Required,
 ) {
 
+    /**
+     * An error on the field, together with whether the shopper should see it yet.
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    data class Error(
```
araratthehero: This name might shadow the Kotlin Error class. What do you think of renaming it to `ErrorState`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -9,32 +9,60 @@
...
+            error == null -> copy(error = Error(message, isVisible = false))
+            else -> copy(error = Error(message, isVisible = error.isVisible))
+        }
+    }
 
-    fun updateText(text: String) = copy(text = text, showError = false)
+    /**
+     * Shows the error, if there is one.
+     */
+    fun showErrorIfPresent() = copy(error = error?.copy(isVisible = true))
```
araratthehero, line 56: What if we add the `isFocused` as a parameter to this function. This was we don't need to update the focus from the call site and the logic could move here.
e.g. Before:
```
phoneNumber = state.phoneNumber.showErrorIfPresent()
                        .copy(isFocused = hasPhoneNumberError),
```
After:
```
phoneNumber = state.phoneNumber.showErrorIfPresent(isFocused = hasPhoneNumberError)
```

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
```diff
@@ -9,32 +9,60 @@
...
 
     /**
-     * Whether the error message should be displayed, which requires both an error to be present and the field to be
-     * showing its errors.
+     * Replaces the error, keeping whether it is currently visible. Validators call this on every pass, so replacing a
+     * message must not hide an error the shopper is already looking at.
      */
-    val showErrorMessage: Boolean
-        get() = showError && errorMessage != null
+    fun withError(message: CheckoutLocalizationKey?): TextInputComponentState {
```
araratthehero: We could change the name of this function to `updateError` to match with `updateText`

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
araratthehero: What do you think if this becomes a sealed class called ErrorState, with 3 potential types.
- No error
- Error that should be shown now
- Error that should not be shown

This way we can remove the `isErrorVisible` boolean and use the state directly where we need to access the errors. Not sure if this would be beneficial, but an idea to think about.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: I think this might become irrelevant with the refactor (the focus flag)

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: I get it, it's a nice solid idea but I'm worried about the code complexity it creates. Imagine every test that needs to check the error message needs to cast to one of the 2 types that have an error message, since we can't have the error message in the base sealed class (the no error state won't have it). Do you feel strongly about the boolean?

araratthehero: I see. And having the sealed class only with two types would just remove the `isVisible` flag. Okay in this stage let's leave it as is, we can review reassess this later if necessary.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: Fixed, chose `InputError`

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputComponentState.kt
jreij: Fixed

araratthehero (approved):

PR #2944, opened 2026-08-14
v5 - Agents - Dependency update skill

jreij:
## Description

Renovate opens an "All v5 dependencies" PR against `v5` every three months. The work it needs is repeatable but not obvious, and until now it lived only in the heads of whoever did it last. This adds a skill so the next person, or their agent, starts from the same place.

Two files, split because they change at different rates:

- **`SKILL.md`** — the process. Three rules and their ordering, the ten steps, how to derive the `minSdk` ceilings from AAR manifests rather than trusting a stored table, how to triage the API dump churn a Kotlin major produces, and the verification and reporting expected before the PR is called done.
- **`DIFFERENCES.md`** — the decision record. What is deliberately different from `main`, why, and which PR wrote it down. The process should be near-static; this changes whenever a difference is accepted or resolved.

General commit, branch, PR and check guidance stays on `main` and is read from there rather than duplicated, so only these two files live on v5.

Stacked on #2886 — the differences recorded here are the ones settled while working that PR.

## Checklist

### DIFF
--- a/.agents/skills/android-update-dependencies-v5/DIFFERENCES.md
+++ b/.agents/skills/android-update-dependencies-v5/DIFFERENCES.md
@@ -1,0 +1,47 @@
+# Deliberate differences between v5 and main
+
+Everything in the build system should be identical to `main` **except** what is listed here. If you find a difference that is not on this list, port it from `main` rather than justifying it. If it should not be ported, add it here with the reason.
+
+Remove an entry as soon as its difference is resolved. A stale entry is worse than a missing one, because it causes the next reader to preserve something that no longer needs preserving.
+
+The *Introduced* column records the PR that added the entry to this file, which is not necessarily when the difference itself first appeared. Commit hashes are omitted because the branch is rebased.
+
+## Versions and build system
+
+| Difference | Why | Introduced |
+| --- | --- | --- |
+| `min-sdk` is lower than main's | v5 is still being released and cannot raise it. Every version ceiling below follows from this; if `min-sdk` changes, re-derive them all | #2886 |
+| Versions capped by that ceiling | See step 3 of the skill. Derive them from the AAR manifests each time; the set changes as AndroidX publishes | #2886 |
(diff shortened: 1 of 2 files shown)


.agents/skills/android-update-dependencies-v5/DIFFERENCES.md
```diff
@@ -0,0 +1,47 @@
+# Deliberate differences between v5 and main
+
+Everything in the build system should be identical to `main` **except** what is listed here. If you find a difference that is not on this list, port it from `main` rather than justifying it. If it should not be ported, add it here with the reason.
+
+Remove an entry as soon as its difference is resolved. A stale entry is worse than a missing one, because it causes the next reader to preserve something that no longer needs preserving.
+
+The *Introduced* column records the PR that added the entry to this file, which is not necessarily when the difference itself first appeared. Commit hashes are omitted because the branch is rebased.
+
+## Versions and build system
```
OscarSpruit: Should we add the difference that was added to `DokkaConventionPlugin` here (the Github URL)

.agents/skills/android-update-dependencies-v5/SKILL.md
```diff
@@ -0,0 +1,197 @@
...
+- no `public static final field CREATOR` removed
+- every member that became `final` belongs to a `final class`, so nobody could have overridden it
+- `$DefaultImpls` classes retained — that is what keeps already-compiled merchant code linking
+- no change to any method's parameters or return type
+- no class removed other than generated `$Creator` helpers
+
+Account for every removed line. If something does not fit a known codegen category, it is a real API change and needs a decision, not a dump.
+
+### 6. Fix tests and the example app
+
```
araratthehero, line 142: This section is empty

jreij: I think it's on purpose, nothing to add in there

araratthehero (approved):

PR #2947, opened 2026-08-17
v6 - POC - Form field architecture

jreij:
Field order currently lives in three places that nothing keeps in sync: the sequence of `if` blocks in `CardContent`, the sequence of a `copy(...)` call in the reducer for focus priority, and implicitly in Compose's focus search for the keyboard's `Next` key. This makes three things impossible: the IME action cannot be derived at all (Compose fixes it when a field is created and offers no way to ask whether another focusable field follows), prefill cannot know where to send focus, and pay focuses the first invalid field in declaration order rather than the order the shopper reads.

This publishes the order from the state layer and has everything derive from it.

What that gives us, all working on a device:
- Pay highlights every error and focuses the first invalid field **in visual order**, and that field now keeps the error it just revealed. It used to hide it, because a requested focus was indistinguishable from a shopper tap.
- Every text field shows `Next` except the last visible one, which shows `Done`. `Done` closes the keyboard and does not submit.
- Card scanning moves focus to the first field it did not fill.
- Which fields are visible and in which order is answered once, so the screen reads no configuration at all.

### Reviewing
Six commits in four layers, bottom-up. Each compiles and, except where stated, changes no behaviour.

1. **Add the form unit that owns field order and focus requests** — `FormState`, `FormFieldId`, `FocusRequest`, `KeyboardAction` and four pure functions. Nothing reads them yet.
2. **Give the card form an explicit field order** — card's field ids, the canonical sequence, and visibility derived from the same properties the producer reads. `CardComponentState.form` is computed on every read, so the order cannot go stale.
3. **Keep the error on the field pay focuses** — the one behaviour change in the stack, and the bug this work exists for. Also collapses eight `UpdateXFocus` intents into one.
4. **Carry the pending focus request in a field's view state**
5. **Publish the card field order to the view state** — still nothing reads it.
6. **Render the card form from its field order** — nine `if` blocks become one loop; fields move onto the focus request and the IME action.

### Deliberately not here
- **Tests were trimmed to get to a demo sooner.** The configuration matrix, the exhaustive focus-rule tests and the Compose UI tests all come back before the layer PRs go up.
- Stored card, MB Way and BLIK still use the old focus flag, so `isFocused` is carried unchanged and marked. `rg "TODO - Form fields cleanup"` and `rg "TODO - Form fields rollout"` are the two worklists.
- No field is reordered. The mechanism exists; the sequence is the one we ship today.

### Progress
✅ Phase 1 — Core: the error type ([#2942](https://github.com/Adyen/adyen-android/pull/2942))
➡️ **Phases 2 to 5 — the POC (this PR)**
Phase 6 — Roll out to stored card, MB Way, BLIK
Phase 7 — Cleanup and document

## Checklist
- [x] Code is unit tested

### DIFF
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
@@ -13,6 +13,8 @@ import com.adyen.checkout.core.components.internal.ui.state.model.PayButtonViewS
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputViewState
 
 internal data class CardViewState(
+    // The fields to render, in the order they appear on screen.
+    val fieldOrder: List<CardFieldId>,
     val cardNumber: TextInputViewState?,
     val expiryDate: TextInputViewState?,
     val securityCode: TextInputViewState?,
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateProducer.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateProducer.kt
@@ -15,9 +15,13 @@ import com.adyen.checkout.card.internal.ui.model.SecurityCodeTrailingIcon
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
 import com.adyen.checkout.core.components.data.model.Amount
 import com.adyen.checkout.core.components.internal.ui.state.ViewStateProducer
+import com.adyen.checkout.core.components.internal.ui.state.form.keyboardActionFor
 import com.adyen.checkout.core.components.internal.ui.state.model.PayButtonViewState
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputComponentState
+import com.adyen.checkout.core.components.internal.ui.state.model.TextInputViewState
+import com.adyen.checkout.core.components.internal.ui.state.model.TrailingIcon
 import com.adyen.checkout.core.components.internal.ui.state.model.toViewState
+import com.adyen.checkout.ui.internal.element.input.FocusRequestToken
 
 internal class CardViewStateProducer(
     private val amount: Amount?,
@@ -50,30 +54,33 @@ internal class CardViewStateProducer(
             null
         }
 
-        val supportedCardBrandsViewState = SupportedCardBrandsViewState(
-            supportedCardBrands = state.supportedCardBrands.filterNot {
-                isHiddenCardType(it.txVariant)
-            },
-            isVisible = isSupportedCardBrandsShown,
-        )
-
         return CardViewState(
-            cardNumber = state.cardNumber.copy(description = cardNumberInputDescription).toViewState(
+            fieldOrder = state.form.order,
+            cardNumber = state.fieldViewState(
+                id = CardFieldId.CARD_NUMBER,
+                field = state.cardNumber.copy(description = cardNumberInputDescription),
                 customTrailingIcon = getCardNumberTrailingIcon(isCardScanButtonVisible),
             ),
-            expiryDate = state.expiryDate.toViewState(
-                customTrailingIcon = getExpiryDateTrailingIcon(state.expiryDate),
+            expiryDate = state.fieldViewState(
+                CardFieldId.EXPIRY_DATE,
+                state.expiryDate,
+                getExpiryDateTrailingIcon(state.expiryDate),
             ),
-            securityCode = state.securityCode.toViewState(
-                customTrailingIcon = getSecurityCodeTrailingIcon(state.securityCode, cardNumberFormat),
+            securityCode = state.fieldViewState(
+                CardFieldId.SECURITY_CODE,
+                state.securityCode,
+                getSecurityCodeTrailingIcon(state.securityCode, cardNumberFormat),
             ),
-            holderName = state.holderName.toViewState(),
-            socialSecurityNumber = state.socialSecurityNumber.toViewState(),
-            kcpBirthDateOrTaxNumber = state.kcpBirthDateOrTaxNumber.toViewState(),
-            kcpCardPassword = state.kcpCardPassword.toViewState(),
-            postalCode = state.postalCode.toViewState(),
+            holderName = state.fieldViewState(CardFieldId.HOLDER_NAME, state.holderName),
+            socialSecurityNumber = state.fieldViewState(CardFieldId.SOCIAL_SECURITY_NUMBER, state.socialSecurityNumber),
+            kcpBirthDateOrTaxNumber = state.fieldViewState(
+                CardFieldId.KCP_BIRTH_DATE_OR_TAX_NUMBER,
+                state.kcpBirthDateOrTaxNumber,
+            ),
+            kcpCardPassword = state.fieldViewState(CardFieldId.KCP_CARD_PASSWORD, state.kcpCardPassword),
+            postalCode = state.fieldViewState(CardFieldId.POSTAL_CODE, state.postalCode),
             storePaymentViewState = storePaymentViewState,
-            supportedCardBrandsViewState = supportedCardBrandsViewState,
+            supportedCardBrandsViewState = getSupportedCardBrandsViewState(state, isSupportedCardBrandsShown),
             cardBrandViewState = cardBrandViewState,
             cardNumberFormat = cardNumberFormat,
             isLoading = state.isLoading,
@@ -82,6 +89,28 @@ internal class CardViewStateProducer(
             payButtonViewState = if (showSubmitButton) PayButtonViewState(amount, state.isLoading) else null,
         )
     }
+
+    private fun getSupportedCardBrandsViewState(
+        state: CardComponentState,
+        isVisible: Boolean,
+    ) = SupportedCardBrandsViewState(
+        supportedCardBrands = state.supportedCardBrands.filterNot { isHiddenCardType(it.txVariant) },
+        isVisible = isVisible,
+    )
+
+    /**
+     * Builds the view state of one field, adding the two things only the form as a whole can answer: which action key
+     * the field shows, and whether it is the one being asked to take focus.
+     */
+    private fun CardComponentState.fieldViewState(
+        id: CardFieldId,
+        field: TextInputComponentState,
+        customTrailingIcon: TrailingIcon? = null,
+    ): TextInputViewState? = field.toViewState(
+        customTrailingIcon = customTrailingIcon,
+        keyboardAction = form.keyboardActionFor(id),
+        focusRequest = form.focusRequest?.takeIf { it.id == id }?.let { FocusRequestToken(it) },
+    )
 
     private fun getCardNumberInputDescription(cardBrandState: CardBrandState): CheckoutLocalizationKey? {
         if (cardBrandState is CardBrandState.DualBrandWithShopperSelection) {
--- a/card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
+++ b/card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
@@ -29,6 +29,7 @@ import com.adyen.checkout.card.internal.ui.state.CardNumberFormat
 import com.adyen.checkout.core.common.internal.properties.SecurityCodeProperties.SECURITY_CODE_MAX_LENGTH_AMEX
 import com.adyen.checkout.core.common.internal.properties.SecurityCodeProperties.SECURITY_CODE_MAX_LENGTH_DEFAULT
 import com.adyen.checkout.core.common.internal.ui.CheckoutTextFieldTrailingIcon
+import com.adyen.checkout.core.common.internal.ui.toImeAction
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
 import com.adyen.checkout.core.common.localization.internal.helper.resolveString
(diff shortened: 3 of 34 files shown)


core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/form/FormState.kt
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param order The fields the form currently shows, in the order the shopper sees them. Being in the list means the
+ * field is visible, and the position in the list is the position on screen.
+ * @param focusRequest A focus move the state layer is asking the UI to make, or null if there is none pending. The UI
+ * reports back once it has acted on it, which clears it.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+data class FormState<Id : FormFieldId>(
+    val order: List<Id>,
+    val focusRequest: FocusRequest<Id>? = null,
```
araratthehero, line 28: Does the `focusRequest` need to be inside of the `FormState`? I see that you read it only in `CardViewStateProducer.kt:112` and there you could read directly from the `CardComponentState`.

araratthehero: If the `focusRequest` gets removed, then there will only be one field in the `FormState`, which makes the `FormState` redundant. Instead of the `FormState`, a simple `List<Id>` could be used?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
```diff
@@ -87,6 +89,11 @@ internal fun SecurityCodeField(
         supportingText = supportingTextSecurityCode,
         onValueChange = onValueChange,
         inputTransformation = inputTransformation,
+        focusRequest = securityCodeState.focusRequest,
+        onFocusRequestConsumed = onFocusRequestConsumed,
+        imeAction = securityCodeState.keyboardAction.toImeAction(),
```
araratthehero, line 94: Security code field for stored card component shows the `Next` button instead of `Done`.

Image

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/FocusRequestToken.kt
```diff
@@ -0,0 +1,23 @@
...
+ * A pending request to move focus to a field, as far as the field itself is concerned.
+ *
+ * The state layer decides which field should receive focus and why, but it lives in a module this one cannot see, so
+ * the request arrives here as an opaque value. All a field needs from it is to tell one request from the next, which
+ * is why the wrapped value is not readable: wrap whatever the state layer uses to describe the request, and make sure
+ * it is comparable by equality.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@JvmInline
+value class FocusRequestToken(private val request: Any)
```
araratthehero, line 29: Since the `request` has a type of `Any`, this can cause the fields to not skip recomposition. Would it make sense to make this `@Immutable`, like for State objects like `InstallmentViewState`, `CardBrandViewState`..?

araratthehero: Also does the type need to be `Any`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputViewState.kt
```diff
@@ -39,6 +48,8 @@ data class TextInputViewState(
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 fun TextInputComponentState.toViewState(
     customTrailingIcon: TrailingIcon? = null,
+    keyboardAction: KeyboardAction = KeyboardAction.NEXT,
+    focusRequest: FocusRequestToken? = null,
 ): TextInputViewState? {
     if (requirementPolicy == RequirementPolicy.Hidden) return null
```
araratthehero: The same visibility check also happens in the `CardFieldOrder`, is this something we can centralize? This way it looks like we have two sources of truth for visibility checks.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateProducer.kt
```diff
@@ -50,30 +54,33 @@ internal class CardViewStateProducer(
             null
         }
```
araratthehero, line 55: Can this visibility check be centralized too?

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextField.kt
```diff
@@ -159,13 +177,29 @@ fun CheckoutTextField(
...
+    LaunchedEffect(focusRequest) {
+        if (focusRequest != null) {
+            // Throws if the requester is not attached to a focusable node, which can happen when the field leaves
+            // composition between composition and this effect running. Losing the focus is preferable to crashing.
+            runCatching {
+                focusRequester.requestFocus()
+                // Taking focus scrolls the field into view on its own, but a field that already has focus is told
+                // nothing and so would stay off screen. Asking directly covers both.
+                bringIntoViewRequester.bringIntoView()
+            }
```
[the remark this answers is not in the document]
araratthehero: A comment from my agent which can be useful.

> When the cancellation is swallowed, the coroutine keeps going and still calls onFocusRequestConsumed(). FocusRequestConsumed matches on field id only, so a report from a dead request can clear a newer request for the same field and lose keepErrorHighlight — the error gets hidden again, which is the bug this PR fixes.
> 
> Gemini's patch cures that by accident (rethrowing means the callback never runs). Could you make it explicit instead — only wrap requestFocus(), leave bringIntoView() bare since it only throws CancellationException, and give the request an identity so a stale report can't match a new one?
> 
> Separate thing to check: on the happy path, gaining focus clears focusRequest, which changes the LaunchedEffect key and cancels bringIntoView() after a frame or two. Does the scroll ever actually complete?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
jreij: Stored cards is not implemented in this POC yet, only normal card component

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/CheckoutTextField.kt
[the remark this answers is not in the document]
jreij: @gemini-code-assist Fixed, with an adaptation: `runCatching` now wraps only `requestFocus()`, and `bringIntoView()` moved out of the effect into `rememberCoroutineScope()` instead of getting its own try/catch.

That was needed anyway — acting on the request clears it, which changes the `LaunchedEffect` key, so the scroll was being cancelled a frame or two in. Scoping it to the composition fixes that and closes the window where a cancelled effect could still report consumption.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateProducer.kt
```diff
@@ -83,6 +90,28 @@ internal class CardViewStateProducer(
...
+     */
+    private fun CardComponentState.fieldViewState(
+        id: CardFieldId,
+        field: TextInputComponentState,
+        customTrailingIcon: TrailingIcon? = null,
+    ): TextInputViewState? = field.toViewState(
+        customTrailingIcon = customTrailingIcon,
+        keyboardAction = form.keyboardActionFor(id),
+        focusRequest = form.focusRequest?.takeIf { it.id == id }?.let { FocusRequestToken(it) },
+    )
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist Fixed, adapted: `CardComponentState.form` is memoised with `lazy(PUBLICATION)` rather than passed in as a parameter, so the reducer benefits too — it also reads `form` several times. `PUBLICATION` because the merchant owns the coroutine scope, so the same state can be read from more than one thread.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/form/FormState.kt
jreij: The plan is to move `focusRequest` and the form state to `core` where we can handle it centrally. And later after this whole refactor, the next refactor would make `FormState` technically the whole view state (a list of fields + their data). So right now it looks like it can be moved to Card state but this refactor will achieve the opposite by extracting common things to `core`. WDYT?

card/src/main/java/com/adyen/checkout/card/internal/ui/view/SecurityCodeField.kt
jreij: Actually make a temp fix by changing the default keyboard action to `DONE` in the mapper. But also added a TODO to remove the default once the refactor is done, since we only need it now to not break the rest of the components in the POC.

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/FocusRequestToken.kt
jreij: Fixed, annotated with `@Immutable`.

As for `Any`, it has to stay since `FocusRequestToken` is in `ui` and `core` imports `ui` not the other way around. `FocusRequestToken` is only used as a param inside `CheckoutTextField` to make the `LaunchedEffect` work, and `LaunchedEffect` just needs a key of type `Any`.

So in order to not have a type `Any` being passed around in all text fields as well as not creating so much core logic inside `ui`, `FocusRequestToken` is a simple `value class` that makes sure we don't pass random stuff while keeping `ui` focused on UI. WDYT?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputViewState.kt
jreij: Extracted `TextInputComponentState.isVisible` and used it in both this mapping and `CardFieldOrder`.

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewStateProducer.kt
jreij: I agree that the form should be the source of truth for visibility but I think it's a bit overkill for now. The current structure is fragmented already, not every field has all its data in one place. And this refactor makes it a bit worse by adding another independent list.

Which is why we decided together that the next step is to refactor this further and create one big list of ordered fields + their data. So we won't even need these checks anymore. The view state producer will loop over the form and map each id and its data, so a field that isn't in the form is never mapped anyway. Adding checks now would require deleting them later anyway. WDYT?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/form/FormState.kt
araratthehero: Makes sense

card/src/main/java/com/adyen/checkout/card/internal/ui/state/CardViewState.kt
```diff
@@ -13,6 +13,8 @@ import com.adyen.checkout.core.components.internal.ui.state.model.PayButtonViewS
 import com.adyen.checkout.core.components.internal.ui.state.model.TextInputViewState
 
 internal data class CardViewState(
```
araratthehero, line 15: Since we have the `fieldOrder: List<CardFieldId>` here, the `CardViewState` becomes unstable too. I would add `@Immutable` here too.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/model/TextInputViewState.kt
```diff
@@ -10,19 +10,28 @@ package com.adyen.checkout.core.components.internal.ui.state.model
 
 import androidx.annotation.RestrictTo
 import com.adyen.checkout.core.common.localization.CheckoutLocalizationKey
+import com.adyen.checkout.core.components.internal.ui.state.form.KeyboardAction
+import com.adyen.checkout.ui.internal.element.input.FocusRequestToken
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 data class TextInputViewState(
```
araratthehero, line 17: It looks like this class is also unstable because of the `TrailingIcon`. I would either add `@Immutable` to the `TrailingIcon` or to the `TextInputViewState`. Could you verify?

Not sure what is the best approach to not forget to add the `@Immutable` to unstable fields in the future. Should we perhaps create a skill for this?

jreij: Actually I already looked into this and wrote a plan for later. There's a few things we can do such as turning on the compose compiler reports, creating a check on every PR that warns us when we introduce a compose stability regression and enabling a Compose detekt ruleset that catches some obvious errors.

Since this can be done at any point to fix all compose issues in one go, I kept this task for later and I'm focusing on this form refactor so we can pick up the address work when we need to. Let me know if this is better done now already.

araratthehero: Good idea. If you did not yet, could you create a ticket with your findings and ideas, so we can pick it up at a later stage?

jreij:
Agreed on the appraoch, POC PR not needed anymore

PR #2949, opened 2026-08-17
v6 - Agent docs - Split AGENTS.md into skills

araratthehero:
## Description
AGENTS.md is loaded into context on every agent turn, and had grown to 325 lines / 13.3 KB.
Roughly 70% of it was conditional procedure that only applies to a minority of tasks, while the
orientation an agent cannot cheaply infer was missing.

This continues the extraction pattern already used for `android-check`, `android-commit`,
`android-branch-create` and `android-pr-create`.

**Extracted into skills:**
- `android-public-api-change` — internal-by-default, `@RestrictTo`, breaking changes, sealed vs abstract, `apiDump`
- `android-ui-resources` — style hierarchy, theme-safe attributes, shopper-locale strings
- `android-add-module` — publishing setup, `compileOnly` + `dontwarn` + `runCompileOnly` for optional external SDKs

**AGENTS.md now carries** only what applies to every task: branch model (v6 alpha on `main`), the
working agreement, testing conventions, always-on rules, and a routing table to the skills.
325 → 68 lines (13.3 KB → 4.4 KB).

### DIFF
--- a/.agents/skills/android-add-module/SKILL.md
+++ b/.agents/skills/android-add-module/SKILL.md
@@ -20,7 +20,14 @@ Invoke this skill when creating a new module (for example a new payment method o
 
 Copy the Gradle setup from the existing module closest to what you are adding, then diff the two files and account for every difference. Add the module to `settings.gradle`, keeping the list alphabetical.
 
-### 2. Handle optional external SDKs with compileOnly
+### 2. Register a published module elsewhere
+
+A published module is not done when its own `build.gradle` is right — two lists outside it have to know about it:
+
+- **Documentation.** Add `dokka(project(':<module>'))` to the `dependencies` block at the bottom of the root `build.gradle`. Every published module has an entry there, so a missing one means the module is silently absent from the generated API documentation.
+- **Drop-in.** If merchants reach the payment method through Drop-in, add `api project(':<module>')` to `drop-in/build.gradle` so it ships with Drop-in. Shared infrastructure such as `core`, `ui-core` and `components-core`, and action modules such as `redirect` and `await`, arrive transitively and are not listed.
+
+### 3. Handle optional external SDKs with compileOnly
 
 Some modules wrap a third-party SDK: 3DS2, Twint, WeChat Pay, Cash App Pay, Google Pay, card scanning. Merchants who use standalone components should not be forced to pull these in, so the dependent module declares them as `compileOnly` and tolerates their absence at runtime.
 
@@ -37,7 +44,7 @@ Because the classes are absent at runtime, R8 warns about the missing references
 -dontwarn com.adyen.checkout.adyen3ds2.**
 ```
 
-### 3. Guard every call into an optional SDK
+### 4. Guard every call into an optional SDK
 
 Never touch a `compileOnly` type directly. Wrap the call so a missing class degrades gracefully instead of throwing `NoClassDefFoundError`:
 
@@ -51,7 +58,7 @@ The optional type must not appear outside the `runCompileOnly` block — not in 
 
 For a boolean availability check, use `checkCompileOnly`.
 
-### 4. Verify
+### 5. Verify
 
 Run the `android-check` skill (`.agents/skills/android-check/SKILL.md`) for the new module and for every module that depends on it.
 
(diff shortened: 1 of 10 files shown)


AGENTS.md
```diff
@@ -1,325 +1,68 @@
...
-    val result = validator.validate(input)
-    
-    // THEN
-    assertFalse(result.isValid)
-}
-```
+- **Write the test first**, then make it pass. Before refactoring, make sure the affected code is already covered — if it is not, write those tests first to capture current behavior.
+- New tests use JUnit 5 with Mockito-Kotlin, and Turbine for flows. A handful of older tests still use JUnit 4 or Robolectric; do not copy that as the default.
+- Name tests `when <condition> then <expectation>` and structure the body as GIVEN / WHEN / THEN, following neighboring tests.
+- Run checks through the `android-check` skill rather than calling Gradle directly, and run them as you go rather than accumulating unverified work.
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist Not applying this. The repo pins `compose-bom = "2025.05.01"`, not 2026.05.01, so `androidx.compose.ui.test.junit4.v2.createComposeRule` does not exist on the version we build against. There are also no `createComposeRule` usages anywhere in the repo and `compose-ui-test-junit4` is not wired into any module, so there is nothing for the guidance to apply to yet.

Beyond accuracy: AGENTS.md is loaded into context on every agent turn, and this PR exists to move conditional detail out of it. When we do add Compose UI tests, this belongs in a testing skill rather than in the always-on file.

.agents/skills/android-add-module/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+Invoke this skill when creating a new module (for example a new payment method or an internal helper module), or when an existing module needs to depend on an optional external SDK.
+
+## Steps
+
+### 1. Decide whether the module is published
+
+- **Published** — anything merchants integrate, such as a new payment method. It uses the `checkout.android.library` convention plugin plus a `checkoutPublishing { id, name, description }` block, and gets an `api/<module>.api` dump file. Compare with `card/build.gradle`.
+- **Not published** — internal tooling and test support. These skip both: `test-core` uses the plain `android.library` plugin, and `lint` is a Kotlin JVM module.
+
+Copy the Gradle setup from the existing module closest to what you are adding, then diff the two files and account for every difference. Add the module to `settings.gradle`, keeping the list alphabetical.
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist Not applying this. The bullet it targets tells an agent which Gradle plugin a new module applies; cast strategy inside build-logic is a separate concern and would read as a non-sequitur there.

`PublishConventionPlugin.kt` also contains no such safe cast today, so there is no invariant in this repo for the rule to protect. If we want guidance about failing loudly in publishing configuration, it should live with the build-logic conventions rather than in the add-a-module checklist.

atmamont (dismissed):

.agents/skills/android-add-module/SKILL.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Usage
+
+Invoke this skill when creating a new module (for example a new payment method or an internal helper module), or when an existing module needs to depend on an optional external SDK.
+
+## Steps
+
+### 1. Decide whether the module is published
+
+- **Published** — anything merchants integrate, such as a new payment method. It uses the `checkout.android.library` convention plugin plus a `checkoutPublishing { id, name, description }` block, and gets an `api/<module>.api` dump file. Compare with `card/build.gradle`.
```
jreij, line 18: In the root `build.gradle`, all published modules should have a `dokka(project(':MODULE_NAME'))` added to the `dependencies` block at the bottom.

.agents/skills/android-add-module/SKILL.md
jreij, line 21: For most modules, `drop-in` should include them explicitly in its `build.gradle`

araratthehero: @jreij Added, with one nuance worth checking: I scoped it to payment methods merchants reach through Drop-in, since 19 of the 55 published modules are not listed in `drop-in/build.gradle` — shared infrastructure (`core`, `ui-core`, `components-core`), action modules that arrive transitively (`redirect`, `await`), and `card-scanning`, which I read as deliberately excluded so it stays opt-in.

If the intent is that anything new should default to being listed unless there is a reason not to, I can flip the wording around.

jreij (approved):

PR #2950, opened 2026-08-18
v6 - Core - Unit tests for model, localization and session params helpers

araratthehero:
## Description

Adds unit tests for four `core` helpers that had no coverage.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-546

### DIFF
--- a/core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
+++ b/core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
@@ -9,9 +9,11 @@
 package com.adyen.checkout.core.common.internal.model
 
 import org.json.JSONArray
+import org.json.JSONException
 import org.json.JSONObject
 import org.junit.jupiter.api.Assertions.assertEquals
 import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.assertThrows
 import org.junit.jupiter.api.Assertions.assertTrue
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
@@ -85,6 +87,13 @@ internal class JsonUtilsTest {
         fun `when the key does not exist, then null is returned`() {
             assertNull(JSONObject().getIntOrNull("key"))
         }
+
+        @Test
+        fun `when the value is a numeric string, then it is coerced to an int`() {
+            val jsonObject = JSONObject("""{"key":"1"}""")
+
+            assertEquals(1, jsonObject.getIntOrNull("key"))
+        }
     }
 
     @Nested
@@ -125,6 +134,21 @@ internal class JsonUtilsTest {
             val jsonObject = JSONObject("""{"array":["a","b"]}""")
 
             assertEquals(listOf("a", "b"), jsonObject.getStringList("array"))
+        }
+
+        // Unlike optStringList, getStringList is not lenient about a missing key.
+        @Test
+        fun `when the string list key does not exist, then a JSONException is thrown`() {
+            assertThrows(JSONException::class.java) {
+                JSONObject().getStringList("array")
+            }
+        }
+
+        @Test
+        fun `when the optional string list holds a null entry, then that entry is skipped`() {
+            val jsonArray = JSONArray("""["a",null]""")
+
+            assertEquals(listOf("a"), JsonUtils.parseOptStringList(jsonArray))
         }
 
         @Test
--- a/core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
+++ b/core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
@@ -1,0 +1,288 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ararat on 17/8/2026.
+ */
+
+package com.adyen.checkout.core.common.internal.model
+
+import org.json.JSONArray
+import org.json.JSONObject
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+internal class JsonUtilsTest {
+
+    @Nested
+    inner class GetStringOrNullTest {
+
+        @Test
+        fun `when the value is a string, then that string is returned`() {
+            val jsonObject = JSONObject("""{"key":"value"}""")
+
+            assertEquals("value", jsonObject.getStringOrNull("key"))
+        }
+
+        @Test
+        fun `when the value is null, then null is returned`() {
+            val jsonObject = JSONObject("""{"key":null}""")
+
+            assertNull(jsonObject.getStringOrNull("key"))
+        }
+
+        @Test
+        fun `when the key does not exist, then null is returned`() {
+            assertNull(JSONObject().getStringOrNull("key"))
+        }
+    }
+
+    @Nested
+    inner class GetBooleanOrNullTest {
+
+        @Test
+        fun `when the value is a boolean, then that boolean is returned`() {
+            val jsonObject = JSONObject("""{"key":true}""")
+
+            assertEquals(true, jsonObject.getBooleanOrNull("key"))
+        }
+
+        @Test
+        fun `when the value is null, then null is returned`() {
+            val jsonObject = JSONObject("""{"key":null}""")
+
+            assertNull(jsonObject.getBooleanOrNull("key"))
+        }
+
+        @Test
+        fun `when the key does not exist, then null is returned`() {
+            assertNull(JSONObject().getBooleanOrNull("key"))
+        }
+    }
+
+    @Nested
+    inner class GetIntOrNullTest {
+
+        @Test
+        fun `when the value is an int, then that int is returned`() {
+            val jsonObject = JSONObject("""{"key":1}""")
+
+            assertEquals(1, jsonObject.getIntOrNull("key"))
+        }
+
+        @Test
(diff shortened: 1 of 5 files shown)


core/src/test/java/com/adyen/checkout/core/common/localization/internal/LocalizationResolverTest.kt
```diff
@@ -0,0 +1,120 @@
...
+    fun `when format args are passed, then the string is formatted with them`() {
+        val result = getLocalizedString(
+            localizationProvider = FakeLocalizationProvider("Pay %s"),
+            key = CheckoutLocalizationKey.PAY_BUTTON_WITH_AMOUNT,
+            formatArgs = arrayOf("EUR 10.00"),
+        )
+
+        assertEquals("Pay EUR 10.00", result)
+    }
+
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist Good catch — this was a real bug, so it is fixed rather than just re-tested.

The production fix lives in the PR this one is now stacked on, #2951: `string.format(*formatArgs)` becomes `string.format(locale, *formatArgs)`.

Worth noting on reachability: it is not observable through the SDK's own strings, since every key that takes format arguments uses `%s`, and amounts are pre-formatted via `Amount.format(locale)` before being passed in. It *is* reachable through `CheckoutLocalizationProvider`, which is public API, whenever a merchant-supplied string uses `%d`, `%,d` or `%.2f`.

The test here is adjusted, but not exactly as suggested. Asserting only `Locale.FRANCE` produces `"Pay 10,00"` would still pass against the buggy code on any machine whose default locale uses a comma decimal separator (fr, nl, de, es, it, pt), because `String.format` would fall back to that same default — so the test would silently prove nothing for a large share of developers. It is now a `@ParameterizedTest` over two locales:

- `Locale.US` produces `"Pay 10.00"`
- `Locale.FRANCE` produces `"Pay 10,00"`

The buggy implementation returns the *same* string for both, so at least one case fails regardless of the machine's default locale. Verified by reverting the production line locally: case `[2] fr_FR` failed with `Pay 10.00`, and both cases pass with the fix in place.

Two smaller deviations from the suggestion: the formatted value is kept small to avoid French grouping separators (U+202F), whose representation has changed between JDK versions and would make the assertion toolchain-dependent; and the backtick test naming used throughout this module is kept rather than the camelCase name in the snippet.

core/src/test/java/com/adyen/checkout/core/sessions/internal/model/SessionParamsFactoryTest.kt
```diff
@@ -0,0 +1,240 @@
...
+
+        @Test
+        fun `when the shopper locale is malformed, then it falls back to the root locale`() {
+            // Locale.forLanguageTag does not throw for malformed tags, it returns the root locale. The
+            // runCatching fallback in SessionParamsFactory is therefore never reached.
+            val session = createSession(shopperLocale = "not a language tag")
+
+            val result = SessionParamsFactory.create(session)
+
+            assertEquals(Locale.ROOT, result.shopperLocale)
```
OscarSpruit: As the comment says, this documents dead code: `LocaleUtil.fromLanguageTag` just delegates to `Locale.forLanguageTag`, which never throws, so the `runCatching`/error-log fallback in `SessionParamsFactory.getShopperLocale` is unreachable and a malformed locale coming from the API silently becomes `Locale.ROOT` instead of `null`.

The test name reads like `Locale.ROOT` is a deliberate fallback. Could we rename it to something neutral (e.g. `when the shopper locale is malformed, then the root locale is returned`) and create a follow-up ticket to either validate the tag with `LocaleUtil.validateLocale` or drop the dead `runCatching`?

core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        @Test
+        fun `when the key does not exist, then null is returned`() {
+            assertNull(JSONObject().getLongOrNull("key"))
+        }
+    }
+
+    @Nested
+    inner class StringListTest {
```
OscarSpruit, line 123: `JsonUtils.parseOptStringList` is only reached indirectly through `optStringList`, so its filtering behaviour is not covered. The int equivalent gets a dedicated "non int entry is skipped" test, a string counterpart (e.g. `["a", null]`, where `optString(i, null)` decides what is kept) would be a nice mirror.

Same nested class: the missing-key path of `getStringList` (which throws `JSONException`, unlike `optStringList`) is unasserted. Worth pinning down if that difference is intentional.

core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
```diff
@@ -0,0 +1,288 @@
...
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+internal class JsonUtilsTest {
+
+    @Nested
+    inner class GetStringOrNullTest {
```
OscarSpruit, line 24: Optional: the happy paths here are the predictable ones. The surprising `org.json` behaviour is the type coercion (`getStringOrNull` on `{"key":1}` returns `"1"`, `getIntOrNull` on `{"key":"1"}` returns `1`). Since these helpers are used for parsing API responses, locking that in would add more value than the plain string/int cases.

core/src/test/java/com/adyen/checkout/core/common/internal/model/ModelUtilsTest.kt
```diff
@@ -0,0 +1,187 @@
...
+        @Test
+        fun `when the json object is null, then the result is null`() {
+            val result = deserializeOpt(null, TestModelObject.SERIALIZER)
+
+            assertNull(result)
+        }
+    }
+
+    @Nested
+    inner class DeserializeOptListTest {
```
OscarSpruit, line 48: `deserializeOptList` and `deserializeOptMap` wrap their result in `Collections.unmodifiable*`. If that immutability is part of the contract, one assertion per helper (mutating the result throws) would protect it, since nothing here would catch its removal.

core/src/test/java/com/adyen/checkout/core/common/localization/internal/LocalizationResolverTest.kt
```diff
@@ -0,0 +1,136 @@
...
+            key = CheckoutLocalizationKey.PAY_BUTTON_WITH_AMOUNT,
+            locale = locale,
+            formatArgs = arrayOf(10.0),
+        )
+
+        assertEquals(expected, result)
+    }
+
+    @Test
+    fun `when no format args are passed, then the string is returned unformatted`() {
```
OscarSpruit, line 105: We cover provider string + format args, and default localization without args, but not default localization *with* format args. That last combination is the common production path (pay button with amount), so it would be good to have it here as well.

OscarSpruit (commented):
Nice addition, the nested grouping and the explanatory comments make these easy to follow. I ran the suite locally and everything is green. A couple of remarks: one about a test that locks in dead code, the rest are optional coverage gaps.

core/src/test/java/com/adyen/checkout/core/sessions/internal/model/SessionParamsFactoryTest.kt
araratthehero: Renamed to `when the shopper locale is malformed, then the shopper locale is null`. Rather than filing a follow-up we fixed it: #2954, stacked on this PR.

Two things came out of digging into it:

The `runCatching` is unreachable rather than merely unlikely. `Locale.forLanguageTag` only throws for a `null` argument, and both call sites already null-check before calling it. Probing 15 malformed inputs (`""`, `en_US`, `not a tag`, `-en`, `abcdefghi`, emoji, control characters) returns `Locale.ROOT` every time and never throws.

The same dead pattern is copy-pasted in `CheckoutParamsFactory`, which is the copy that actually decides `CheckoutParams.shopperLocale`. Because it returned a non-null `ROOT`, the `?: deviceLocaleProvider()` fallback was skipped and the entire checkout rendered in the root locale. The realistic trigger is `nl_NL` instead of `nl-NL`, which is the form `Locale.toString()` produces.

Both call sites are fixed in #2954, and the test here flips from `Locale.ROOT` to `null` there.

One note on the suggested `LocaleUtil.validateLocale` route: it does not catch this, because `Locale.ROOT` is itself a valid locale and passes validation. The fix uses `Locale.forLanguageTag(tag).takeIf { it.language.isNotEmpty() }` instead, which targets exactly the case where parsing yielded nothing while leaving leniently parsed tags such as `en-` and `en--US` working as they do today.

core/src/test/java/com/adyen/checkout/core/common/internal/model/JsonUtilsTest.kt
araratthehero: Added the int half: `getIntOrNull` on `{"key":"1"}` returns `1`. That holds on both `org.json` implementations, so it is safe to pin.

Left the string half out, because it does not hold in unit tests. I wrote exactly the suggested assertion and it failed:

```
JsonUtilsTest > when the value is a number, then it is coerced to a string FAILED
    org.json.JSONException at JsonUtilsTest.kt:49
```

Unit tests compile against `org.json:json:20250107` (`testImplementation libs.json`), whose `getString` throws unless the value is already a `String`. Android's `org.json` coerces through `JSON.toString()` and returns `"1"`. So asserting `"1"` fails the build, while asserting "throws" would pin behaviour that is false in production and could give false confidence.

Leaving this thread open rather than resolving it, because the general point is worth a wider decision: any unit test asserting `org.json` semantics can diverge from what runs on device. Pinning the coercion properly would need an instrumentation test — happy to raise that separately if you think it is worth it.

OscarSpruit (approved):

PR #2955, opened 2026-08-21
v6 - Drop-in - Custom theme

OscarSpruit:
## Description
Drop-in can now be styled with a `CheckoutTheme`, matching what the Compose components already support.

`CheckoutTheme` and `CheckoutColors` are now `Parcelable`, so the theme can travel to `DropInActivity` on the existing intent extra. `@Parcelize` does not support the `CheckoutColor` value class, so an internal `CheckoutColorParceler` writes the underlying `Long`. Because the theme rides the intent, it also survives process death.

The theme is passed to `DropIn.start(...)` and flows through `DropInLauncher` and `DropInResultContract.Input` into `DropInActivity`, replacing the hardcoded `CheckoutTheme()`. It is also forwarded to `CheckoutPaymentFlow` in `PaymentMethodScreen` — that composable re-provides the theme through its own `CheckoutCompositionLocalProvider`, so without an explicit argument the component subtree would fall back to the default palette. That screen serves both regular and stored payment methods.

The theme is fixed when drop-in starts: if the app switches between light and dark mode while drop-in is open, drop-in keeps the theme it was started with. Handling that is a follow-up.

The example app passes its light/dark palette to both v6 drop-in entry points, and `docs/v6/theme.md` documents the drop-in usage.

## Checklist
- [ ] If applicable, make sure Breaking change label is added.
- [ ] Code is unit tested
- [ ] Aligned public API changes with other platforms (if applicable)

## Ticket Number
COSDK-1486

### New
- Drop-in can now be customized with a `CheckoutTheme`, passed to `DropIn.start(...)`.


ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutColors.kt
```diff
@@ -8,18 +8,23 @@
...
+import com.adyen.checkout.ui.theme.CheckoutColors.Companion.dark
+import com.adyen.checkout.ui.theme.CheckoutColors.Companion.light
+import kotlinx.parcelize.Parcelize
 
 /**
  * Represents a color value used by the Checkout UI.
  *
  * @param value The color value encoded as an ARGB [Long]. For example, `0xFF000000` represents black.
  */
+@Parcelize
```
araratthehero, line 24: Is the `@Parcelize` necessary for a value class?

OscarSpruit: Yes unfortunately it is

araratthehero (approved):

PR #2957, opened 2026-08-24
v6 - Components - Submit button in every component

jreij:
## Description

Every payment component now renders a submit button by default, including the ones that previously rendered no UI at all. Previously a component that collected no shopper input rendered nothing, and could only be submitted by merchant-written code:

```kotlin
if (!controller.requiresUserInteraction()) {
    controller.submit()
}
```

If that code was missing, the shopper was left on a blank screen with no way to pay and no error. Now every payment method works the same way with no custom integration code: the shopper presses the button and the payment is submitted.

`submit()` and `requiresUserInteraction()` are unchanged and remain available. They are only needed by merchants who hide our button and drive submission themselves.

### Per component

| Component | Renders by default | `requiresUserInteraction()` |
|---|---|---|
| Generic / instant | Button | `false` |
| Stored Blik | Button | `false` |
| Stored card, CVC shown | CVC field and button | `true` |
| Stored card, CVC hidden | Button | `false` |
| Card, MB Way, Blik | Form and button | `true` |
| Google Pay | Google Pay button | `true` |

The stored card component deliberately keeps its own UI and state instead of adopting the generic scaffolding, so that both flows work: `requiresUserInteraction()` follows the security code requirement policy.

Google Pay stays `requiresUserInteraction() == true` even though it is button-only, because its own sheet is the interaction. This is intentional.

### Internal

The generic and stored Blik components now have the same scaffolding as every other component — state, intent, factory, reducer, validator and view state producer — so a component with no input fields is structurally the same as one with fields, just with no content above the button.

### Example app

Both v6 flows now only fall back to a custom button when the submit button is explicitly hidden in the configuration, instead of also doing it for components that used to render nothing.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1487

### Changed
- All components now have a submit button by default. For payment methods that do not require any shopper input, `CheckoutController.requiresUserInteraction()` will return false to indicate that you can skip the button click and submit the component automatically using `CheckoutController.submit()`.

### DIFF
--- a/blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikFactory.kt
+++ b/blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikFactory.kt
@@ -20,6 +20,10 @@ import com.adyen.checkout.core.components.data.model.paymentmethod.StoredPayment
 import com.adyen.checkout.core.components.internal.PaymentComponentFactory
 import com.adyen.checkout.core.components.internal.StoredPaymentComponentFactory
 import com.adyen.checkout.core.components.internal.data.provider.SdkDataProvider
+import com.adyen.checkout.core.components.internal.ui.state.GenericComponentStateFactory
+import com.adyen.checkout.core.components.internal.ui.state.GenericComponentStateReducer
+import com.adyen.checkout.core.components.internal.ui.state.GenericComponentStateValidator
+import com.adyen.checkout.core.components.internal.ui.state.GenericViewStateProducer
 import kotlinx.coroutines.CoroutineScope
 
 internal class BlikFactory :
@@ -56,6 +60,11 @@ internal class BlikFactory :
             storedPaymentMethod = storedPaymentMethod,
             analyticsManager = analyticsManager,
             sdkDataProvider = sdkDataProvider,
+            componentStateValidator = GenericComponentStateValidator(),
+            componentStateFactory = GenericComponentStateFactory(),
+            componentStateReducer = GenericComponentStateReducer(),
+            viewStateProducer = GenericViewStateProducer(params.amount, params.showSubmitButton),
+            coroutineScope = coroutineScope,
         )
     }
 }
--- a/blik/src/main/java/com/adyen/checkout/blik/internal/ui/StoredBlikComponent.kt
+++ b/blik/src/main/java/com/adyen/checkout/blik/internal/ui/StoredBlikComponent.kt
@@ -18,19 +18,42 @@ import com.adyen.checkout.core.components.data.PaymentComponentData
 import com.adyen.checkout.core.components.data.model.paymentmethod.StoredPaymentMethod
 import com.adyen.checkout.core.components.internal.PaymentComponentEvent
 import com.adyen.checkout.core.components.internal.data.provider.SdkDataProvider
+import com.adyen.checkout.core.components.internal.ui.GenericContent
 import com.adyen.checkout.core.components.internal.ui.PaymentComponent
+import com.adyen.checkout.core.components.internal.ui.state.ComponentStateFlow
+import com.adyen.checkout.core.components.internal.ui.state.GenericComponentStateFactory
+import com.adyen.checkout.core.components.internal.ui.state.GenericComponentStateReducer
+import com.adyen.checkout.core.components.internal.ui.state.GenericComponentStateValidator
+import com.adyen.checkout.core.components.internal.ui.state.GenericIntent
+import com.adyen.checkout.core.components.internal.ui.state.GenericViewStateProducer
+import com.adyen.checkout.core.components.internal.ui.state.viewState
 import com.adyen.checkout.core.components.paymentmethod.BlikDetails
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.receiveAsFlow
 
+@Suppress("LongParameterList")
 internal class StoredBlikComponent(
     private val storedPaymentMethod: StoredPaymentMethod,
     private val analyticsManager: AnalyticsManager,
     private val sdkDataProvider: SdkDataProvider,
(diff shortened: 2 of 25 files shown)


card/src/main/java/com/adyen/checkout/card/internal/ui/view/StoredCardContent.kt
```diff
@@ -55,26 +55,27 @@ private fun StoredCardContent(
...
+                        securityCodeState = securityCodeState,
+                        cardNumberFormat = viewState.cardNumberFormat,
+                        onValueChange = { onIntent(StoredCardIntent.UpdateSecurityCode(it)) },
+                        onFocusChange = { onIntent(StoredCardIntent.UpdateSecurityCodeFocus(it)) },
+                    )
+                }
             }
-        }
-    }
+        },
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist that's correct, it's redundant. But I find that it makes it clear and more readable that this is a lambda. Without the annotation it reads like an extra unnecessary bracket or some weird formatting.

blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikFactory.kt
```diff
@@ -20,6 +20,10 @@ import com.adyen.checkout.core.components.data.model.paymentmethod.StoredPayment
 import com.adyen.checkout.core.components.internal.PaymentComponentFactory
```
OscarSpruit, line 20: Because this PR changes behaviour of the released alpha.1 I think we should add release notes about this change

blik/src/main/java/com/adyen/checkout/blik/internal/ui/BlikFactory.kt
```diff
@@ -20,6 +20,10 @@ import com.adyen.checkout.core.components.data.model.paymentmethod.StoredPayment
...
 import com.adyen.checkout.core.components.internal.StoredPaymentComponentFactory
```
OscarSpruit, line 21: The KDocs of `CheckoutController` should be updated to reflect to new behaviour

card/src/main/java/com/adyen/checkout/card/internal/ui/StoredCardComponent.kt
```diff
@@ -159,8 +159,13 @@ constructor(
         )
     }
 
-    override fun requiresUserInteraction(): Boolean =
-        componentState.value.securityCode.requirementPolicy == RequirementPolicy.Required
+    override fun requiresUserInteraction(): Boolean {
+        return when (componentState.value.securityCode.requirementPolicy) {
+            RequirementPolicy.Hidden -> false
+            RequirementPolicy.Optional -> true
```
OscarSpruit, line 165: There seems to be no test capturing the behaviour of the optional case. Could you add one?

blik/src/main/java/com/adyen/checkout/blik/internal/ui/StoredBlikComponent.kt
```diff
@@ -18,20 +18,43 @@ import com.adyen.checkout.core.components.data.PaymentComponentData
...
+import com.adyen.checkout.core.components.internal.ui.state.GenericIntent
+import com.adyen.checkout.core.components.internal.ui.state.GenericViewStateProducer
+import com.adyen.checkout.core.components.internal.ui.state.viewState
 import com.adyen.checkout.core.components.paymentmethod.BlikDetails
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.receiveAsFlow
 
+@Suppress("LongParameterList")
 internal class StoredBlikComponent(
```
OscarSpruit, line 36: This component and in the future probably more components are identical to `GenericPaymentComponent`, except for the creation of the payment method details. Not in the scope of this PR maybe, but we should figure out of we can somehow reuse `GenericPaymentComponent`

jreij: I definitely see a good use of having a Generic stored component since all of them do the same thing except for cards. But I can't think of a use case where we would need to reuse generic logic in normal components, anything comes to mind? In all cases we can deal with that later as you mentioned.

card/src/main/java/com/adyen/checkout/card/internal/ui/StoredCardComponent.kt
jreij: At the moment it's not possible to have this use case, since the only variable is `StoredCVCVisibility` and it can either be `SHOW` or `HIDE`. We can probably find a way to force this in the tests but I'm concerned it would be an ugly solution. What do you think?

blik/src/main/java/com/adyen/checkout/blik/internal/ui/StoredBlikComponent.kt
jreij: Unless you mean that already at the factory level, we can decide if we need UI or not, and we return a different component altogether from the start. That would work for stored cards now and for Pay by Bank later since all the info we need is in the payment method object or the configuration. But I feel like it makes the logic fragmented and someone looking at the stored card or Pay by Bank components can't easily known that there is another flow, because it's hidden in the factory.

card/src/main/java/com/adyen/checkout/card/internal/ui/StoredCardComponent.kt
OscarSpruit: It's okay like this then

blik/src/main/java/com/adyen/checkout/blik/internal/ui/StoredBlikComponent.kt
OscarSpruit: For components with an optional form this doesn't sound feasible. I think this is only applicable to components that either only show a button or nothing

OscarSpruit (approved):

PR #2963, opened 2026-08-26
v6 - Drop-in - Checkout controller provider

araratthehero:
## Description

- `CheckoutController` creation moves out of `PaymentMethodViewModel` into a new `DropInControllerProvider`, so the
  view model receives a provider instead of building a controller from `CheckoutContext` and `DropInServiceManager`
  itself. This is the seam later phases need to share one controller across screens.
- `DropInViewModel` gets `controllerProvider` and `dropInServiceManager` injected from its factory, matching how
  `DropInNavigator` was already built there.
- New `TestCheckoutContext` fixture in `core/src/testFixtures`, since `CheckoutContext` subtypes have `internal`
  constructors and can't be faked. Not a public API change — `testFixtures` isn't in the API dump.

### Progress

✅ Phase 0 — Lifecycle 2.11.0 and compile SDK 37 (#2962)
➡️ **Phase 2 — Controller ownership (this PR)**
Phase 3a — View model store per nav entry
Phase 3b — Action screen loading UI
Phase 3c — Action destination
Phase 4 — Payment method screen states
Phase 5 — Action screen design
Phase 6 — Secondary content

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1420

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInControllerProvider.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInControllerProvider.kt
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ararat on 26/8/2026.
+ */
+
+package com.adyen.checkout.dropin.internal.ui
+
+import com.adyen.checkout.core.action.data.ActionComponentData
+import com.adyen.checkout.core.common.CheckoutContext
+import com.adyen.checkout.core.components.AdditionalDetailsResult
+import com.adyen.checkout.core.components.AdvancedCheckoutCallbacks
+import com.adyen.checkout.core.components.BeforeSubmitResult
+import com.adyen.checkout.core.components.CheckoutController
+import com.adyen.checkout.core.components.CheckoutTarget
+import com.adyen.checkout.core.components.SessionCheckoutCallbacks
+import com.adyen.checkout.core.components.SessionCheckoutResult
+import com.adyen.checkout.core.components.SubmitResult
+import com.adyen.checkout.core.components.data.BeforeSubmitData
+import com.adyen.checkout.core.components.data.PaymentComponentData
+import com.adyen.checkout.core.error.CheckoutError
+import com.adyen.checkout.dropin.internal.service.DropInServiceManager
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+/**
+ * Creates the [CheckoutController] that drives a single Drop-in checkout flow.
+ */
+internal fun interface DropInControllerProvider {
+
+    /**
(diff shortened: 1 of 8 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInControllerProvider.kt
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    private fun onComplete(coroutineScope: CoroutineScope, result: SessionCheckoutResult) {
+        // TODO - Implement after signature of onFinished is updated
+        coroutineScope.launch {
+            dropInServiceManager.onPaymentCompleted(result.resultCode)
+        }
+    }
+}
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist Good catch, but I'd like to leave this out of this PR.

This behaviour is pre-existing — `PaymentMethodViewModel` on `main` already did exactly the same:

```kotlin
private fun onFailure(error: CheckoutError) {
    viewModelScope.launch {
        dropInServiceManager.onFailure(error)
    }
}
```

This PR moved it verbatim into the provider, so nothing here introduced it. I'd rather not fold a behavioural fix into a refactor that is otherwise no-behaviour-change.

Your concern is actually a bit sharper than described: `_paymentResultFlow` is a `MutableSharedFlow()` with no buffer, so `emit` suspends until a collector is ready, and a suspended emit is exactly what cancellation catches. The realistic trigger is the shopper going back mid-payment — which cancels the payment call too, since the controller runs on that same scope — so there is usually nothing left to report. But the hole is real.

I don't want to take the suggested fix, though. A nullable `var` plus `setDropInScope()` is temporal coupling, and it cannot be wired cleanly: the provider is constructed in `DropInViewModel.Factory.create()`, before the view model exists, so its `viewModelScope` is not available there. It would have to be set after construction.

Doing this properly is a separate change — passing a `() -> CoroutineScope` for the terminal callbacks, or revisiting whether the whole payment call should be screen-scoped at all. Leaving this thread open so we can agree on a follow-up ticket.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModel.kt
```diff
@@ -31,19 +31,17 @@ import kotlin.reflect.KClass
 
 internal class DropInViewModel(
     private val input: DropInResultContract.Input,
-    val navigator: DropInNavigator
+    val navigator: DropInNavigator,
+    val controllerProvider: DropInControllerProvider,
```
jreij, line 35: Should these properties be marked only visible for testing? Or will we need to access them from the UI?

OscarSpruit (approved):

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/DropInViewModel.kt
araratthehero: They will be used by the `NavigationStack` in future PRs.

PR #2966, opened 2026-08-26
v6 - Cards - Refactor payment component state creation

jreij:
## Description
- Move encryption out of the component state mapping and into `CardComponent`, so that nothing is submitted when encryption or the public key lookup fails. Previously an `Error` event was followed by a `Submit` event carrying a null payment method.
- Replace `CardComponentStateExt` with `CardPaymentComponentStateFactory`.
- Cover `CardComponent.submit()` and the new factory with unit tests.

## Checklist 
- [x] Code is unit tested


card/src/test/java/com/adyen/checkout/card/internal/ui/model/CardComponentParamsMapperTest.kt
```diff
@@ -486,22 +486,6 @@ internal class CardComponentParamsMapperTest {
         assertEquals(listOf(2, 3, 6), params.installmentParams?.defaultOptions?.values)
     }
 
-    @Test
-    fun `when funding source is null then installmentConfiguration is used`() {
```
OscarSpruit, line 490: Is this test removed on purpose?

jreij: Yes, the same behaviour is tested elsewhere

OscarSpruit: Cool, I couldn't directly find the equivalent

OscarSpruit (approved):

PR #2967, opened 2026-08-27
v6 - Drop-in - Payment progress content

araratthehero:
## Description

Adds `PaymentProgressContent`, the screen Drop-in shows while a payment is under way and the shopper has nothing left to do — an action that needs no interaction, or a payment method with no UI of its own. Nothing renders it yet.

- Logo, payment method name, description, then the status block.
- `description` is a parameter because it is the one line that differs per case — a redirect explains where the shopper is being sent, a payment method without UI names the amount being charged.
- `status` is a `PaymentProgressStatus` enum so the indicator and the copy that goes with it cannot be paired wrongly. Only `Processing` exists today; the design also has a `Success` state (checkmark and "You've paid!") which lands with phase 5.
- `footer` is an optional slot pinned to the bottom for whatever way out a case offers — "Cancel without result" on a redirect, "Finish" once the payment is done. The progress block centres in what is left, so it reads the same with and without a footer.

### Screenshots

Without a footer:
Screenshot 2026-08-27 at 10 19 45

With a footer:
Screenshot 2026-08-27 at 10 19 54

### Progress

✅ Phase 0 — [Dependencies and compile SDK](https://github.com/Adyen/adyen-android/pull/2962)
✅ Phase 2 — [Controller ownership](https://github.com/Adyen/adyen-android/pull/2963)
✅ Phase 3a — [A view model store per nav entry](https://github.com/Adyen/adyen-android/pull/2964)
➡️ **Phase 3b — Payment progress content (this PR)**
Phase 3c — Action destination
Phase 4 — Payment method screen
Phase 6 — Secondary content

## Checklist

## Ticket Number
COSDK-1420

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/progress/PaymentProgressContent.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/progress/PaymentProgressContent.kt
@@ -1,0 +1,176 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by ararat on 27/8/2026.
+ */
+
+// The footer slot is not the content of this component, which builds its own. Renaming it to "content" would describe
+// it wrongly at every call site.
+@file:Suppress("ComposableLambdaParameterNaming")
+
+package com.adyen.checkout.dropin.internal.ui.progress
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import com.adyen.checkout.core.common.internal.ui.CheckoutNetworkLogo
+import com.adyen.checkout.ui.internal.element.ProgressBar
+import com.adyen.checkout.ui.internal.element.button.SecondaryButton
+import com.adyen.checkout.ui.internal.helper.CheckoutThemePreviewWrapper
+import com.adyen.checkout.ui.internal.helper.ThemePreviewParameterProvider
+import com.adyen.checkout.ui.internal.text.Body
+import com.adyen.checkout.ui.internal.text.Title
+import com.adyen.checkout.ui.internal.theme.CheckoutThemeProvider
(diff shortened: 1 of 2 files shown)


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/progress/PaymentProgressContent.kt
```diff
@@ -0,0 +1,174 @@
...
+    Column(
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.Center,
+        modifier = modifier.fillMaxWidth(),
+    ) {
+        CheckoutNetworkLogo(
+            txVariant = logoTxVariant,
+            contentDescription = null,
+            modifier = Modifier.size(Dimensions.LogoSize.large),
+        )
```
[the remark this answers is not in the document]
araratthehero: Keeping `contentDescription = null` here. The logo sits directly above a `Title` that renders the same `paymentMethodName`, so setting it as the content description would make TalkBack announce the payment method twice. `null` marks the logo as decorative, which is correct when the adjacent text already carries the information.

This also matches `PreselectedPaymentMethodScreen`, which has the identical logo-then-title shape and passes `null` for the same reason.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/progress/PaymentProgressStatus.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package com.adyen.checkout.dropin.internal.ui.progress
+
+/**
+ * How far along the payment is, as reported by [PaymentProgressContent].
+ *
+ * The status owns both the indicator and the copy that goes with it, so a caller cannot pair one with the other.
+ */
+internal enum class PaymentProgressStatus {
+    Processing,
```
jreij: I think we use all caps for enum entries in other places

araratthehero: Good catch, fixed!

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/progress/PaymentProgressStatus.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+package com.adyen.checkout.dropin.internal.ui.progress
+
+/**
+ * How far along the payment is, as reported by [PaymentProgressContent].
+ *
+ * The status owns both the indicator and the copy that goes with it, so a caller cannot pair one with the other.
+ */
+internal enum class PaymentProgressStatus {
```
OscarSpruit: What is the idea for this? If it is for the result page, maybe we can leave it out for now and we will implement it once we get to that

araratthehero: Indeed, this is for showing different icons/messages for each state, processing, failed, succeeded. I will remove it then from this PR, and show the progress icon for now.

OscarSpruit (approved):

PR #2971, opened 2026-08-28
v6 - Core - Improve image loading

OscarSpruit:
## Description

Reworks the internal image loader behind `CheckoutNetworkLogo` so logo loading is non-blocking, deduplicated and resilient to failures.

- Replaced the callback-shaped `load(url, onSuccess, onError)` with `suspend fun load(url: String): Result<Bitmap>`, which removes all six `withContext(DispatcherProvider.Main)` hops — `NetworkImage` already resumes on the composition's dispatcher. The `ImageLoader` interface is gone, since there was one implementation and one consumer.
- Replaced the blocking `call.execute()` with `enqueue` wrapped in `suspendCancellableCoroutine`, so a pending request no longer occupies a `Dispatchers.IO` thread and requests are subject to OkHttp's dispatcher limits instead of bypassing them.
- Deduplicated in-flight requests, so a URL rendered twice on one screen (for example `mc` as both a stored-card icon and a Cards brand-row entry) results in a single HTTP call. Only requests that are still running are joined, so a completed failure is never replayed to a later caller.
- Added negative caching for `4xx` responses, so a permanently missing logo is no longer re-requested on every recomposition and every screen entry. `IOException` and `5xx` stay retryable.
- An `IOException` while loading now surfaces as a failure and renders the error fallback, instead of propagating out of the composition.
- Moved the image classes from `ui` into `core` under `com.adyen.checkout.core.internal.image`, narrowed them to `internal`, and migrated `NetworkImage` to `produceState`.
- Called `prepareToDraw()` after decoding, so the GPU texture upload happens off the first draw rather than mid slide-in on `PaymentMethodListScreen`.

## Checklist
- [x] Code is unit tested

## Ticket Number
N/A


core/src/main/java/com/adyen/checkout/core/internal/image/DefaultImageLoader.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    private suspend fun evict(url: String, deferred: Deferred<Result<Bitmap>>) {
+        inFlightMutex.withLock {
+            // Only evict this request, it might already have been replaced by a newer one.
+            if (inFlight[url] === deferred) {
+                @Suppress("DeferredResultUnused")
+                inFlight.remove(url)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
OscarSpruit: Thanks — the coroutine-per-eviction point is fair, and the snippet would work (`synchronized` is reentrant, so the immediate-completion path through `invokeOnCompletion` is safe rather than a deadlock).

I'd still rather keep the `Mutex`, because of where `load` is called from. `NetworkImage` invokes it inside `produceState`, whose block runs on the composition's dispatcher — so the synchronous prologue of `load` (the two cache lookups plus acquiring the lock) executes on the main thread. `Mutex.withLock` suspends under contention, whereas `synchronized` blocks. `evict` takes the same lock from OkHttp's callback threads, so contention with the main thread isn't hypothetical when a screen loads ~20 logos at once.

The saving is one coroutine launch per completed load, so roughly 20 per screen. That doesn't seem worth putting a blocking lock on the main thread, even with a critical section this small.

There's also a consistency angle: this PR deletes the old hand-rolled double-checked locking on `Context.imageLoader`, so reintroducing JVM monitors into coroutine code would pull against that.

core/src/main/java/com/adyen/checkout/core/internal/image/DefaultImageLoader.kt
```diff
@@ -0,0 +1,126 @@
...
+        if (!response.isSuccessful) {
+            val error = HttpError(response.code, response.message, null)
+            if (error.code in HTTP_4XX_RANGE) {
+                failureCache.put(url, error)
+            }
+            return Result.failure(error)
+        }
+
+        val bytes = response.body?.bytes() ?: ByteArray(0)
+        val bitmap = BitmapFactory.decodeByteArray(bytes, 0, bytes.size)
```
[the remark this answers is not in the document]
OscarSpruit: Applied, with a small deviation from the suggestion.

`handleResponse` is already called inside `response.use { }` in `fetch`, so closing the `Response` closes the body and its stream. That makes the extra `?.use { }` redundant, so I went with:

```kotlin
val stream = response.body?.byteStream()
val bitmap = BitmapFactory.decodeStream(stream)
```

`decodeStream` accepts a nullable stream and returns `null` for it, so a missing body still falls through to the existing decode-failure branch rather than needing its own path.

One behavioural note for the record: `bytes()` throws `IOException` on a truncated response, whereas `decodeStream` swallows stream `IOException`s internally and returns `null`. Both end up as a retryable failure, so there's no functional change, but a truncated download will now be logged as "Failed to decode bitmap." instead of the underlying network error. That seems an acceptable trade for dropping the intermediate array.

This also required updating the new unit tests, which stubbed `BitmapFactory.decodeByteArray` — they now stub `decodeStream`. Worth noting they caught the change immediately rather than passing silently.

core/src/main/java/com/adyen/checkout/core/internal/image/DefaultImageLoader.kt
```diff
@@ -0,0 +1,126 @@
...
+
+                // Finished requests have to be evicted, because a completed [Deferred] holds on to its result, and
+                // therefore to the bitmap, keeping it alive even after the in memory cache has evicted it.
+                new.invokeOnCompletion {
+                    scope.launch { evict(url, new) }
+                }
+            }
+        }
+
+        return deferred.await()
```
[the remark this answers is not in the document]
OscarSpruit: Confirmed, this is deliberate — it's the main trade-off that made deduplication correct.

The fetch runs in a loader-owned scope rather than the caller's precisely so that one subscriber going away doesn't cancel the shared request for everyone still waiting on it. Caller cancellation is still honoured: the awaiting coroutine unwinds with `CancellationException` while the shared `Deferred` continues and warms the cache.

The alternative — reference-counting awaiters and cancelling at zero — is materially more code and more edge cases, and the payoff is limited in this context: it's a short payment-method list rather than an infinite feed, the payloads are a few KB, and letting the request finish means a scroll-back is a cache hit. Since the request is enqueued rather than executed blocking, a pending one also holds no thread while it completes.

Happy to revisit if the profiling pass shows logo traffic actually contending with API calls.

core/src/main/java/com/adyen/checkout/core/internal/image/DefaultImageLoader.kt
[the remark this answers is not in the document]
OscarSpruit: Correcting my previous reply — I've backed this out after actually measuring it.

`decodeStream` doesn't decode without a buffer, it allocates its own. From `BitmapFactory.decodeStreamInternal`:

```java
byte [] tempStorage = null;
if (opts != null) tempStorage = opts.inTempStorage;
if (tempStorage == null) tempStorage = new byte[DECODE_BUFFER_SIZE];
```

`DECODE_BUFFER_SIZE` is `16 * 1024`, and we pass no `Options`, so every decode allocates a fresh 16 KB array. `nativeDecodeByteArray` takes no temp storage at all. So the trade is a fixed 16 KB buffer in exchange for avoiding a `byte[N]`, which only pays off once N exceeds 16 KB.

Actual logo sizes from the CDN, using the path format from `buildLogoPath`:

| Asset | Size |
|---|---|
| `small/visa` (mdpi) | 0.5 KB |
| `small/visa-xxxhdpi` | 1.4 KB |
| `small/mc-xxxhdpi` | 1.2 KB |
| `medium/visa-xxxhdpi` | 2.6 KB |
| `large/visa-xxxhdpi` | 5.2 KB |
| `large/paypal-xxxhdpi` | 6.8 KB |

Everything is well under the 16 KB break-even, and the common case on the payment method list is `small` at around 1.2–1.4 KB. For a list of roughly 20 logos that's about 28 KB of transient allocation with `decodeByteArray` versus about 320 KB with `decodeStream` — an order of magnitude more GC pressure, rather than less.

The reasoning in the original suggestion is sound for large images, it just inverts below 16 KB, and this loader only ever handles CDN logos: `CheckoutNetworkLogo` already requests a density- and `LogoSize`-appropriate asset, so bitmaps arrive at roughly display size by design.

The one genuine upside of streaming is that decoding overlaps with the download, since OkHttp resumes at headers. At 1.4 KB on a warm connection that's sub-millisecond, so it doesn't outweigh the above.

Reverted to `bytes()` + `decodeByteArray`, which also keeps the real `IOException` message on a truncated response instead of it surfacing as a decode failure.

core/src/main/java/com/adyen/checkout/core/internal/image/LogoSize.kt
jreij: Is it better if we map this explicitly and remove the proguard rule?

OscarSpruit: Good one, I'll improve it

core/src/main/java/com/adyen/checkout/core/internal/image/DefaultImageLoader.kt
```diff
@@ -0,0 +1,126 @@
...
+import kotlin.coroutines.resumeWithException
+
+internal class DefaultImageLoader(
+    private val cache: InMemoryCache,
+    private val failureCache: LruCache<String, HttpError>,
+    private val okHttpClient: OkHttpClient = OkHttpClient(),
+    dispatcher: CoroutineDispatcher = DispatcherProvider.IO,
+) {
+
+    private val scope = CoroutineScope(SupervisorJob() + dispatcher)
```
araratthehero, line 41: Since `SupervisorJob() + dispatcher` is used here, the scope can outlive its caller and `continuation.invokeOnCancellation` might never be called

OscarSpruit: That's technically true, but it's intentional: the `Deferred` is shared between all callers of the same URL, so tying it to one caller would mean the first composable leaving the composition kills the download for everyone else still waiting. Letting it finish also warms the shared `InMemoryCache` for the next recomposition. Cost is bounded by the OkHttp timeouts, and evict drops the entry on completion so the bitmap isn't pinned.

Doing this "properly" would need reference counting (cancel the shared job only when the last awaiter leaves), which I'd rather not add for logo loading without a measured problem. Keeping `invokeOnCancellation` since it's the correct thing to write in a `suspendCancellableCoroutine` and costs nothing.

The related question that is worth deciding: the loader in `ImageLoaderProvider` is process-wide and never cancelled. Are we fine with that as the contract, or do we want it closeable?

araratthehero: I agree with the trade-off to have a shared `Deferred`. But reading the class gives a wrong impression that the `invokeOnCancellation` actually happens, while it is currently unreachable. I would at least add a short comment explaining that.

Regarding the loader in `ImageLoaderProvider`, I would keep it process-wide, since there is no proper owner of the lifecycle.

jreij (approved):

PR #2978, opened 2026-09-01
v6 - Drop-in - Use LazyColumn for the payment method list

OscarSpruit:
## Description

Renders the payment method list lazily, so only the items that are actually visible get composed. Previously every payment method — including its logo — was composed and laid out on entering the screen, which made logo loading noticeably lag the slide-in animation on longer lists.

The list description, the section headers and each payment method are now emitted as separate lazy items instead of being nested in scrollable columns. Item spacing moved to per-item padding, and the explicit spacer between the stored and regular sections was dropped, since the section header's own vertical padding provides the separation.

Visually the list is unchanged.

## Checklist
- [ ] Code is unit tested

## Ticket Number
N/A

### DIFF
--- a/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
+++ b/drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
@@ -10,16 +10,16 @@ package com.adyen.checkout.dropin.internal.ui
 
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyListScope
+import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.CircleShape
-import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Close
 import androidx.compose.material3.Icon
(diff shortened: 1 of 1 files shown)


araratthehero (dismissed):

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListScreen.kt
```diff
@@ -127,24 +126,30 @@ private fun PaymentMethodListContent(
...
             onActionClick = onActionClick,
         )
+    }
 
-        PaymentMethodItemList(
-            paymentMethodItems = items,
-            onItemClick = onPaymentMethodClick,
+    items(items) { item ->
+        PaymentMethodListItem(
+            item = item,
```
araratthehero, line 146: Wouldn't adding key to the item help with performance on recomposition?

OscarSpruit: We are using the txVariant as id for the item and this in not unique, so I choose to not make it more complicated and set no key. When no key is provided the position is used as key. I thought this is acceptable, because items don't really change position

araratthehero: I think it might be a problem when you go to the favorites page, remove a stored payment method and come back to the payment methods list screen. The list will be mutated and the items will start changing positions. Is this something that we should be worried about?

araratthehero: Also I noticed that in `PaymentMethodViewModel`, we do `paymentMethods.first { it.type == txVariant }` to get the payment method. When we implement Giftcards, this might fail. When tapping on the second giftcard on the list will open the first Giftcard. I will make a note about this on Giftcard tickets to test/fix this when implementing.

OscarSpruit: For now it won't be an issue. If removing a stored payment method will prove to be an issue, we can try to fix it then

araratthehero: Okay, I will add a note to test this case in the ticket.

araratthehero (approved):

PR #2988, opened 2026-09-01
ADR-0002

OscarSpruit:
## Description
Drafts an ADR that documents the activity host acquisition for components.

### DIFF
--- a/ADR/ADR-0002-activity-host-acquisition.md
+++ b/ADR/ADR-0002-activity-host-acquisition.md
@@ -1,0 +1,328 @@
+# Activity host acquisition for components
+
+| Field         | Value        |
+|---------------|--------------|
+| Author(s)     | Oscar Spruit |
+| Status        | Accepted     |
+| Creation date | 2026-09-01   |
+
+## Abstract
+
+Some payment and action components need an Android `Activity` to do their work: launching Custom Tabs, registering an
+`ActivityResultLauncher`, or handing an `Activity` to a third-party SDK. In v6 these components obtain that `Activity`
+from the Compose tree, inside their `Content` composable. This ADR evaluates four ways of acquiring an activity host and
+proposes keeping the current composition-based acquisition, while making its failure mode explicit. Adopting SDK-owned
+activities is proposed per payment method, only when a concrete case requires it, rather than as an SDK-wide
+architecture. Letting merchants hide the contents of our composables was considered and is deferred, pending a separate
+decision on whether every action must always show UI.
+
+## Motivation
+
+Five capabilities in the SDK cannot be served by an application `Context` alone:
+
+| Capability                  | Consumers                     | Requires                                       |
+|-----------------------------|-------------------------------|------------------------------------------------|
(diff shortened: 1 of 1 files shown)


ADR/ADR-0002-activity-host-acquisition.md
```diff
@@ -0,0 +1,236 @@
...
+
+Every capability that needs an `Activity` is moved inside an activity the SDK declares and owns. The five capabilities
+collapse into one: start our activity. Results return either through an `ActivityResultContract` or through a
+process-global bridge.
+
+**Pros:**
+
+- Removes the pre-`STARTED` problem entirely. An activity we own registers in its own `onCreate`, which is trivially
+  before `STARTED`. This is the only option that solves Twint without a headless fragment.
+- The SDK can own the redirect `<intent-filter>`, so merchants no longer declare one, override `onNewIntent`, or call
```
jreij, line 131: I don't think this is feasible since we need this intent filter to match the merchant's return URL and we want to move away from the default `adyencheckout` custom scheme

ADR/ADR-0002-activity-host-acquisition.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Concerns and follow-up actions
+
+- **Twint's v6 migration is the most likely trigger for revisiting D.** The v5 implementation relies on
+  `Twint(this, ::onTwintResult)` being a field initializer of `TwintActionFragment`, which works because the pre-`STARTED`
+  constraint is relative to the fragment's own lifecycle. If v6 has no place for a fragment, Twint needs either a
+  headless fragment of its own or a Twint-owned activity.
+- **WeChat Pay's `callbackClassName`** should point at an SDK-declared activity rather than the merchant's class. This is
+  a manifest concern and can be solved locally without adopting D.
+- **A v6 return URL helper is still missing.** v5 exposes `RedirectComponent.getReturnUrl(context)` and
```
jreij: Related to the comment above, I think this helper won't be needed anymore since we will move away from default return URLs / intent filters

ADR/ADR-0002-activity-host-acquisition.md
OscarSpruit: It would work similar to drop-in. Where we can have a default value like `adyencheckout://` and merchants can override the intent filter if they want to use something else

jreij: But the default value is fragile because it's a custom scheme and many payment methods are blocking those. Which is we discussed that we should completely remove it rather than keep it and risk flows breaking because of it. This is going to be even more critical with `/sessions` only allowing one return URL for the whole session - merchants can't patch one payment method only if it starts failing.

OscarSpruit: Fair enough. I have documented this

ADR/ADR-0002-activity-host-acquisition.md
OscarSpruit: I removed this

araratthehero (approved):

PR #3012, opened 2026-09-02
v6 - Drop-in - Always show the submit button

araratthehero:
## Description
Drop-in submits a payment through the submit button that the component renders itself, so a merchant setting
`showSubmitButton = false` would leave Drop-in with a card form and no way to pay. Drop-in now ignores that value
and always turns the button on, which is what `CheckoutConfiguration.showSubmitButton` already documents.

The rule lives entirely in Drop-in. Core only gains two restricted extensions in its `internal` packages:

- `CheckoutConfiguration.copy(showSubmitButton)` — carries over every other value, including the payment method
  configurations added through the configuration block, so the merchant configuration is never partially lost.
- `CheckoutContext.withCheckoutConfiguration(...)` — swaps the configuration while keeping the subtype, next to the
  existing `checkoutAttemptId` and `publicKey` extensions.

`DefaultDropInControllerProvider` applies both once, at the boundary, before it creates any controller. Both
extensions are `@RestrictTo(LIBRARY_GROUP)`, so there is no public API change and no `.api` diff.

A copy that lists fields silently drops any field added later, so `CheckoutConfigurationExtTest` pins the primary
constructor of `CheckoutConfiguration` by reflection and fails with an explicit message when it changes.

### Progress

✅ Phase 2 — Controller ownership in a flow-scoped view model ([#2963](https://github.com/Adyen/adyen-android/pull/2963))
✅ Phase 3a — A view model store per nav entry ([#2964](https://github.com/Adyen/adyen-android/pull/2964))
➡️ **Phase 4a — Always show the submit button in Drop-in (this PR)**
Phase 4b — `PaymentMethodScreen`: no-UI routing and payment progress
Phase 6 — Secondary content

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1420

### DIFF
--- a/core/src/main/java/com/adyen/checkout/core/common/internal/CheckoutContextExt.kt
+++ b/core/src/main/java/com/adyen/checkout/core/common/internal/CheckoutContextExt.kt
@@ -16,7 +16,10 @@
 
 package com.adyen.checkout.core.common.internal
 
+import androidx.annotation.RestrictTo
 import com.adyen.checkout.core.common.CheckoutContext
+import com.adyen.checkout.core.components.Checkout
+import com.adyen.checkout.core.components.CheckoutConfiguration
 
 internal val CheckoutContext.checkoutAttemptId: String
     get() = when (this) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/adyen/checkout/core/components/CheckoutConfiguration.kt
```diff
@@ -100,7 +100,15 @@ class CheckoutConfiguration(
...
         // method type as the key instead of the class name.
-        availableConfigurations[configuration::class.java.name] = configuration
+        addConfiguration(configuration::class.java.name, configuration)
+    }
+
+    /**
+     * Adds a configuration under an existing [key], instead of deriving the key from [configuration]. Only meant for
+     * carrying configurations over to another instance, which has to keep them under the very same keys.
+     */
+    internal fun addConfiguration(key: String, configuration: Configuration) {
```
OscarSpruit: Is this function truly needed? I think we should just reuse the existing `addConfiguration`

OscarSpruit (approved):

PR #3015, opened 2026-09-03
v6 - Build - Raise the Java target to 17

jreij:
## Description
JUnit 6 and Android Lint 32 both require a Java 17 baseline. Lint had already moved to 17 on its own, which left the rest of the project on 11 and blocked the JUnit 6 upgrade.

This raises every module to 17 and removes the duplication that allowed the drift:

- The version is declared once, in the version catalog.
- A new `checkout.java` convention plugin applies it to every subproject, covering both Android modules and plain Kotlin JVM modules like `:lint`. The root build only ever configured the Android plugins, so `:lint` had to repeat the setting, which is exactly why it drifted.
- `build-logic` was never pinned at all and compiled against whichever JDK ran Gradle. It now pins the same version.
- Detekt no longer sets a JVM target. It derives one from the toolchain, and this project never sets a `classpath` on the Detekt tasks, so the value only applied to type resolution that is not enabled.
- The `jvmToolchain` block for Kotlin Android is removed because it never ran: AGP 9 does not apply the `kotlin-android` plugin.

Verified against the emitted bytecode: `:card`, `:lint` and `build-logic` all produce class file major version 61. The branch was also validated against the JUnit 6 upgrade before those changes were taken back out of it.

## Checklist

### Breaking changes
- The SDK is now compiled with a Java 17 target. Building an integration requires JDK 17 or newer.

### DIFF
--- a/build-logic/convention/src/main/kotlin/JavaConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/JavaConventionPlugin.kt
@@ -1,0 +1,49 @@
+/*
+ * Copyright (c) 2026 Adyen N.V.
+ *
+ * This file is open source and available under the MIT license. See the LICENSE file for more info.
+ *
+ * Created by josephj on 3/9/2026.
+ */
+
+import com.adyen.checkout.libs
+import com.android.build.api.dsl.CommonExtension
+import com.android.build.gradle.BasePlugin
+import org.gradle.api.JavaVersion
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
+
+/**
+ * Applies the JVM target every module is built against, so that it is declared in a single place.
+ *
+ * The two kinds of module are pinned differently. Android modules only get a bytecode target, and are compiled by
+ * whichever JDK runs Gradle, which is what AGP expects. Plain Kotlin JVM modules get a toolchain, because nothing
+ * else would pin them at all.
+ */
+class JavaConventionPlugin : Plugin<Project> {
+
+    override fun apply(target: Project) {
+        with(target) {
+            val jvmTarget = JavaVersion.toVersion(libs.versions.jvm.target.get())
+
+            // The Kotlin support built into AGP derives its own JVM target from these values, so there is nothing to
+            // configure on the Kotlin side.
+            plugins.withType<BasePlugin> {
+                extensions.configure<CommonExtension> {
+                    compileOptions.sourceCompatibility = jvmTarget
+                    compileOptions.targetCompatibility = jvmTarget
+                }
+            }
+
+            plugins.withType<KotlinPluginWrapper> {
+                extensions.configure<KotlinJvmProjectExtension> {
+                    jvmToolchain(jvmTarget.majorVersion.toInt())
+                }
+            }
+        }
+    }
+}
--- a/build.gradle
+++ b/build.gradle
@@ -28,9 +28,9 @@ ext {
     checkoutRedirectScheme = "adyencheckout"
 }
 
-subprojects {
-    def javaVersion = 11
+def javaVersion = libs.versions.jvm.target.get().toInteger()
 
+subprojects {
     plugins.withType(com.android.build.gradle.BasePlugin).configureEach {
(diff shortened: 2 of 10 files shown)


gradle/libs.versions.toml
```diff
@@ -8,6 +8,7 @@ min-sdk = "23"
 version-code = "1"
 
 # Build script
+java = "17"
```
OscarSpruit: I'd rename this to `jvm-target` to be more precise. We do not want java in this project ;)

build.gradle
```diff
@@ -29,7 +29,7 @@ ext {
 }
 
 subprojects {
-    def javaVersion = 11
```
OscarSpruit, line 32: Applying a plugin here has some downsides: isolated Projects incompatibility, eager configuration of all subprojects and a module whose build file gives no hint that its Java version is set elsewhere. I think it would be a better idea to apply this plugin per specific module, so we could add it to `CheckoutAndroidLibraryPlugin` and then we should add it for some none library modules as well (lint).

Ultimately, we should get rid of this whole `subprojects` block

gradle/libs.versions.toml
jreij: Fixed!

build-logic/convention/src/main/kotlin/JavaConventionPlugin.kt
```diff
@@ -0,0 +1,52 @@
...
+ * Applies the Java version every module is built against, so that it is declared in a single place.
+ */
+class JavaConventionPlugin : Plugin<Project> {
+
+    override fun apply(target: Project) {
+        with(target) {
+            // Android modules take their Java version from the Android DSL. The Kotlin support built into AGP derives
+            // its own JVM target from these values, so there is nothing to configure on the Kotlin side.
+            plugins.withType<BasePlugin> {
+                extensions.configure<CommonExtension> {
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist Not applicable here: in AGP 9 `CommonExtension` has no type parameters at all. I originally wrote it as `CommonExtension<*, *, *, *, *, *>` and the compiler rejected it with `No type arguments expected for 'interface CommonExtension : ExtensionAware'`, which is why it ended up without them.

Keeping the reified form, since it matches the rest of build-logic (`CheckoutAndroidLibraryPlugin` uses `extensions.configure<LibraryExtension>`), and if a future AGP did reintroduce generics that would surface as a compile error rather than a silent problem.

build-logic/convention/src/main/kotlin/JavaConventionPlugin.kt
```diff
@@ -0,0 +1,52 @@
...
+                }
+            }
+
+            // Plain Kotlin JVM modules are not covered by AGP, so they would otherwise be built against whichever JDK
+            // happens to run Gradle.
+            plugins.withType<KotlinPluginWrapper> {
+                extensions.configure<KotlinJvmProjectExtension> {
+                    jvmToolchain(javaVersion.majorVersion.toInt())
+                }
+            }
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist `KotlinPluginWrapper` is not the base class of `KotlinAndroidPluginWrapper`, they are siblings:

```
KotlinPluginWrapper        extends AbstractKotlinPluginWrapper
KotlinAndroidPluginWrapper extends AbstractKotlinAndroidPluginWrapper
```

So the block only runs for `org.jetbrains.kotlin.jvm` and Android modules never reach it. A full `./gradlew check` configures all 58 modules without an `UnknownDomainObjectException`.

On top of that, AGP 9 does not apply the `kotlin-android` plugin at all. I verified that with a probe before writing this plugin, and it is the reason this PR deletes the `jvmToolchain` block that used to hang off `KotlinAndroidPluginWrapper` in the root build: it had silently stopped running.

build.gradle
jreij: Agreed, switched to composition.

`checkout.java` is now applied from `CheckoutAndroidLibraryPlugin`, which covers 56 modules, plus an explicit alias in the three that don't use it: `lint`, `test-core` (raw `android.library`) and `example-app` (`android.application`). The `subprojects` block no longer applies anything.

A nice side effect: the plugin is now applied during each module's own evaluation, so the version catalog extension already exists and the lazy `Project.javaVersion` accessor could go. That workaround only existed because applying from `subprojects` runs before the catalog is available on the project — it failed with `Extension of type 'LibrariesForLibs' does not exist` when I first tried it.

Not touching the rest of the `subprojects` block here, but agreed it should go.

OscarSpruit (approved):

PR #3016, opened 2026-09-03
v6 - Drop-in - Regular payment method screens

araratthehero:
## Description

Turns `PaymentMethodViewState` into a sealed interface and gives each case its own screen. This PR covers the two cases a payment method that is not stored can be in; the stored cases follow in the next layer.

| State | Screen |
|---|---|
| `RegularInput` | `RegularPaymentMethodInputScreen` — collapsing app bar title, description under it, the component's form and its own button |
| `Progress` | `PaymentMethodProgressScreen` — submitted on arrival, centred header plus progress status on the titleless scaffold from the layer below |

`PaymentMethodViewModel` picks between them on `CheckoutController.requiresUserInteraction()`, and submits immediately when there is nothing for the shopper to do. Both view states are now plain values rather than `StateFlow`s: all cases are decided when the flow starts and never change, and anything that does change during a payment is rendered by the component itself. `ActionScreen` reads its state directly as a result.

`DropInNavigator.isEmptyAfterCurrent()` finally gets tests — it already existed and was already used to turn the navigation icon into a close button, but was untested.

**Two known issues this PR does not fix**, both pre-existing on `main` and both carried deliberately so this layer stays reviewable:

### Screenshots

The input screen is unchanged from `main`. The progress screen is new.

image

### Progress

✅ Phase 2 — Controller ownership ([#2963](https://github.com/Adyen/adyen-android/pull/2963))
✅ Phase 3a — View model store per nav entry ([#2964](https://github.com/Adyen/adyen-android/pull/2964))
✅ Phase 4a — Always show the submit button ([#3012](https://github.com/Adyen/adyen-android/pull/3012))
✅ Phase 4b-i — Shared screen elements ([#3014](https://github.com/Adyen/adyen-android/pull/3014))
➡️ **Phase 4b-ii — Regular payment methods: input and progress (this PR)**
Phase 4b-iii — Stored payment methods: input and confirmation
Phase 4b-iv — The amount in the payment method copy
Phase 6 — Secondary content

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1420


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodViewModel.kt
```diff
@@ -53,19 +50,21 @@ internal class PaymentMethodViewModel(
...
 
     init {
         observeNavigation()
+        startPaymentIfNothingToConfirm()
+    }
+
+    private fun startPaymentIfNothingToConfirm() {
+        if (paymentMethodViewState !is PaymentMethodViewState.Progress) return
+        controller.submit()
     }
```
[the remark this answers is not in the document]
araratthehero: Added a TODO at the call site in fa383b95e1 so the risk is recorded where the submit happens, and picked up with the rest of process death handling.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodScreen.kt
```diff
@@ -37,60 +25,36 @@ internal fun PaymentMethodScreen(
     viewModel: PaymentMethodViewModel,
     theme: CheckoutTheme,
 ) {
-    val viewState by viewModel.paymentMethodViewState.collectAsStateWithLifecycle()
-    PaymentMethodScreenContent(navigator, viewState, viewModel.controller, theme)
+    PaymentMethodScreenContent(
+        navigator = navigator,
+        viewState = viewModel.paymentMethodViewState,
+        // TODO - Pass localization provider
+        content = { CheckoutPaymentMethod(controller = viewModel.controller, theme = theme) },
```
OscarSpruit: Why would we pass the content from here? It seems like it's only used in `RegularPaymentMethodInputScreen` - maybe we can keep it there

araratthehero: The `StoredPaymentMethodScreen` will use it too in the next PR

OscarSpruit: Hmm okay, still I think it's a bit confusing to pass a `content` param that might be used. Since `PaymentMethodScreenContent` is basically just a `when` I think we could simplify this by doing something like:
```kotlin
@Composable
internal fun PaymentMethodScreen(
    navigator: DropInNavigator,
    viewModel: PaymentMethodViewModel,
    theme: CheckoutTheme,
) {
    when (val viewState = viewModel.paymentMethodViewState) {
        is PaymentMethodViewState.RegularInput ->
            RegularPaymentMethodInputScreen(navigator, viewState) {
                // TODO - Pass localization provider
                CheckoutPaymentMethod(controller = viewModel.controller, theme = theme)
            }

        is PaymentMethodViewState.Progress ->
            PaymentMethodProgressScreen(navigator, viewState)
    }
}
```
This makes `PaymentMethodScreenContent` obsolete and prevents the situation where `content` usage is ambiguous

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodProgressScreen.kt
```diff
@@ -0,0 +1,80 @@
...
+import com.adyen.checkout.ui.internal.helper.ThemePreviewParameterProvider
+import com.adyen.checkout.ui.internal.theme.Dimensions
+import com.adyen.checkout.ui.theme.CheckoutTheme
+
+/**
+ * A payment method with nothing to enter and nothing to confirm: it was submitted on arrival, and this only reports
+ * that it is under way.
+ */
+@Composable
+internal fun PaymentMethodProgressScreen(
```
OscarSpruit: This and `RegularPaymentMethodInputScreen` are ending with `Screen` to me this is a bit confusing, because usually screen composables are opened from `NavigationStack`. What do you think about renaming these `Screen` suffixes to `Content`?

OscarSpruit (approved):

PR #3021, opened 2026-09-08
v6 - Drop-in - Google Pay on the payment method list

araratthehero:
## Description

Renders Google Pay above the payment method list with its own button, and filters it out of the list items so it cannot appear twice. Both Google Pay types are filtered, not only the one that gets a controller — otherwise offering both would render one as a button and the other as a list item.

`PaymentMethodListViewModel` owns this flow, because the button is on the list rather than behind a list item and its controller has to exist before the shopper picks anything. `ActionNavKey` therefore gains an `owner`, so the action screen knows which view model holds the controller to continue on, and `ActionScreen` takes plain parameters instead of a `PaymentMethodViewModel` so either owner can drive it.

The naming follows adyen-web's Drop-in, which already calls this an instant payment method.

### Screenshots
image

### Progress

✅ Phase 2 — Controller ownership ([#2963](https://github.com/Adyen/adyen-android/pull/2963))
✅ Phase 3a — View model store per nav entry ([#2964](https://github.com/Adyen/adyen-android/pull/2964))
✅ Phase 4a — Always show the submit button ([#3012](https://github.com/Adyen/adyen-android/pull/3012))
✅ Phase 4b-i — Shared screen elements ([#3014](https://github.com/Adyen/adyen-android/pull/3014))
✅ Phase 4b-ii — Regular payment methods ([#3016](https://github.com/Adyen/adyen-android/pull/3016))
✅ Phase 4b-iii — Stored payment methods ([#3017](https://github.com/Adyen/adyen-android/pull/3017))
✅ Phase 4b-iv — The amount in the payment method copy ([#3018](https://github.com/Adyen/adyen-android/pull/3018))
➡️ **Phase 7 — Google Pay on the payment method list (this PR)**
Phase 6 — Secondary content
Phase 10 — Stop discarding the theme Drop-in established
Phase 11 — Payment method list spacing

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1423


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/NavigationStack.kt
```diff
@@ -132,32 +132,80 @@ private fun paymentMethodNavEntry(
...
+): PaymentMethodViewModel = viewModel(
+    viewModelStoreOwner = viewModelStoreOwner,
+    factory = PaymentMethodViewModel.Factory(
+        paymentFlowType = paymentFlowType,
+        dropInParams = viewModel.dropInParams,
+        paymentMethodRepository = viewModel.paymentMethodRepository,
+        navigator = viewModel.navigator,
+        controllerProvider = viewModel.controllerProvider,
+    ),
+)
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist `LocalSharedViewModelStoreOwner.current` is not nullable, so this change is not needed.

It is declared non-null with a throwing default, in `SharedViewModelStoreNavEntryDecorator.kt`:

```kotlin
internal val LocalSharedViewModelStoreOwner = staticCompositionLocalOf<ViewModelStoreOwner> {
    error("No LocalSharedViewModelStoreOwner provided!")
}
```

androidx's `LocalViewModelStoreOwner` *is* `ViewModelStoreOwner?`, which is exactly why the default value of this parameter wraps it in `checkNotNull` — the two locals differ deliberately. `:drop-in:check` passes on this branch, which settles the "compilation errors" part.

There is also a reason not to make it nullable. `ViewModelStoreOwner? = null` would let a caller pass null and silently fall back to *this* entry's store instead of the parent's. The explicit parameter exists precisely so the action screen resolves the view model against the parent entry's store, and therefore continues on the controller that started the flow. A silent fallback would build a second controller and restart the payment instead.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/NavigationStack.kt
```diff
@@ -132,32 +132,80 @@ private fun paymentMethodNavEntry(
...
+    viewModelStoreOwner: ViewModelStoreOwner = checkNotNull(LocalViewModelStoreOwner.current),
+): PaymentMethodListViewModel = viewModel(
+    viewModelStoreOwner = viewModelStoreOwner,
+    factory = PaymentMethodListViewModel.Factory(
+        dropInParams = viewModel.dropInParams,
+        paymentMethodRepository = viewModel.paymentMethodRepository,
+        navigator = viewModel.navigator,
+        controllerProvider = viewModel.controllerProvider,
+    ),
 )
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist Same as the thread above: `LocalSharedViewModelStoreOwner` is declared `staticCompositionLocalOf<ViewModelStoreOwner>` — non-null, with a throwing default — in `SharedViewModelStoreNavEntryDecorator.kt`, so `parentOwner` is never null and the module compiles as it stands.

The fallback would do more damage here than in the other case. Silently resolving against the store of this entry would create a *second* `PaymentMethodListViewModel`, and its `init` builds a new Google Pay controller. The action screen would then render a flow with no payment in progress, so it would never complete — instead of continuing the one that was actually started.

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListViewModel.kt
```diff
@@ -13,33 +13,81 @@ import androidx.lifecycle.ViewModelProvider
...
+    private val navigator: DropInNavigator,
+    controllerProvider: DropInControllerProvider,
 ) : ViewModel() {
 
     val viewState: StateFlow<PaymentMethodListViewState> = paymentMethodRepository.storedPaymentMethods
         .map { createInitialViewState(it) }
         .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), createInitialViewState(emptyList()))
 
+    // Created eagerly rather than when the shopper taps it, because the button is drawn by the controller itself.
+    val instantPaymentMethod: InstantPaymentMethod? = createInstantPaymentMethod(controllerProvider)
```
OscarSpruit, line 54: Can you add a TODO here to do the availability check before creating it eagerly? I think that currently if Google Pay is not available it will fully close drop-in, because the component triggers onError

drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PaymentMethodListViewModel.kt
```diff
@@ -13,33 +13,81 @@ import androidx.lifecycle.ViewModelProvider
...
+
+    init {
+        observeInstantPaymentMethodNavigation()
+    }
+
+    fun findInstantPaymentMethod(paymentFlowType: DropInPaymentFlowType): InstantPaymentMethod? =
+        instantPaymentMethod?.takeIf { it.paymentFlowType == paymentFlowType }
+
+    private fun createInstantPaymentMethod(controllerProvider: DropInControllerProvider): InstantPaymentMethod? {
+        val paymentMethod = paymentMethodRepository.paymentMethods
```
OscarSpruit, line 64: What happens now visually when Google Pay is the only payment method in the list?

araratthehero: This is what we have

image

OscarSpruit: Oh perfect!

OscarSpruit (approved):

PR #3023, opened 2026-09-08
v6 - CI - Stop the stale workflow from processing pull requests

jreij:
## Description
`days-before-pr-stale` was unset, so pull requests fell back to the 60 day default of `days-before-stale`, and `any-of-labels` applies to them too. A labelled pull request would be picked up by a job that only holds `issues: write`, and the run would fail. Stale handling is only meant to apply to issues here.

Also renames the job from `close_stale_prs` to `close_stale_issues` to match what it does.

## Checklist

### DIFF
--- a/.github/workflows/stale_issues.yml
+++ b/.github/workflows/stale_issues.yml
@@ -13,7 +13,7 @@ permissions:
   issues: write
 
 jobs:
-  close_stale_prs:
+  close_stale_issues:
     runs-on: ubuntu-latest
     steps:
       - name: Close stale issues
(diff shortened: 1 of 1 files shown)


.github/workflows/stale_issues.yml
OscarSpruit, line 12: Shouldn't this permission be moved down to the job?

jreij: Yes, fixed in the other job

OscarSpruit (approved):

PR #3032, opened 2026-09-10
v6 - Core - Add form state, focus requests and post processing

jreij:
## Description

Introduce the structural form model that will own visible element order, focus decisions, and keyboard actions across payment components, together with the pipeline stage those decisions run in.

- `FormState` contains visible UI elements in shopper order
- `FormElementState` carries the validity fact needed by ordered focus decisions
- Pure helpers select the first invalid text input, derive `Next` or `Done`, and create focus requests
- `FocusRequest` carries the target and whether an existing error remains visible
- `updateErrorVisibility` centralizes the focus and error-visibility decision
- `ComponentStatePostProcessor` runs after the reducer and the validator, so it can read the errors the validator just set. `NoPostProcessing` is the default, so the stage changes nothing yet
- `ComponentStateFlow` gains that stage, and calls `processInitialState` on the validated initial state so opening decisions come from the same code as every later one
- `CheckoutTextField` receives the future focus and IME contract without changing behavior yet

A focus target is almost always "the first field that is still wrong", and that can only be answered once validation has run. A reducer runs before it and would read the previous pass's errors, which is why this stage exists rather than the reducer deciding.

The intentionally unused UI parameters and helpers have phase-owned TODOs explaining where they will first be used. Existing `shouldFocus`, `isFocused`, `updateFocus`, and direct validity behavior remain until the later adoption and cleanup phases.

### Reviewing

1. Form state and element model
2. Ordering, focus, and keyboard helpers
3. `ComponentStatePostProcessor` and its two entry points
4. Focus-change behavior
5. UI contract and temporary TODOs
6. Tests

### Progress

✅ Phase 1 — [Core: error type](https://github.com/Adyen/adyen-android/pull/3031)
➡️ **Phase 2 — Core: structural form unit, focus handshake, post processor, and IME plumbing (this PR)**
Phase 3 — Card: form state and canonical order
Phase 4 — Card: ordered elements and UI
Phase 5 — Card: focus requests and IME actions
Phase 6 — Card: form-driven validity
Phase 7 — Stored card, MB Way, and BLIK
Phase 8 — Cleanup and ADR

## Checklist
- [x] Code is unit tested


core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/form/FormStateExt.kt
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Returns the next text input after [id], or null if [id] is the last one or is not in the form. Pickers and switches
+ * are skipped, because this answers on a keyboard's behalf.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+fun <Id : FormElementId> FormState<Id>.nextTextInputAfter(id: Id): Id? {
+    val index = elements.indexOfFirst { it.id == id }
+    if (index == -1) return null
+    return elements.subList(index + 1, elements.size).firstOrNull { it.id.isTextInput }?.id
+}
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist I think your suggestion makes this less readable

core/src/main/java/com/adyen/checkout/core/components/internal/ui/state/form/FormElementState.kt
```diff
@@ -0,0 +1,42 @@
...
+    return if (isVisible) {
+        val isValid = error == null
+        FormElementState(
+            id = id,
+            isValid = isValid,
+        )
+    } else {
+        null
+    }
+}
```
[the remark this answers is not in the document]
jreij: @gemini-code-assist the local `isVisible` and `isValid` variables deliberately expose functionality that is only defined in this function:

- Visibility comes from `requirementPolicy`
- Validity comes from `error`

Removing those two would make the functionality implicit and would require adding a comment which is worse than what we have now.

ui/src/main/java/com/adyen/checkout/ui/internal/element/input/FocusRequestToken.kt
```diff
@@ -0,0 +1,20 @@
...
+import androidx.annotation.RestrictTo
+import androidx.compose.runtime.Immutable
+
+/**
+ * Identifies a pending focus request without exposing its form element ID to the UI module.
+ */
+@Immutable
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+@JvmInline
+value class FocusRequestToken(private val request: Any)
```
araratthehero: Shouldn't this `FocusRequestToken` be replaced to a simpler type as discussed before?

jreij: We didn't decide on it so I kept it for the review, what do you suggest? A simple type instead of `Any`?

araratthehero: Yes, I think it is unnecessary to have a custom type to run a LaunchedEffect. I would use a simple boolean for `requestFocus`.

jreij: Fixed!

araratthehero (approved):

PR #3051, opened 2026-09-15
v6 - Components - Secondary screen navigation

OscarSpruit:
## Description
Payment components now own the navigation to their own secondary screens.

`SecondaryScreenComponent` exposes a `navigation` flow of `SecondaryNavigationEvent.Open`/`Close` events. `CheckoutPaymentMethod` collects them and keeps a saveable back stack, rendering the top entry in a full screen dialog on top of the payment method UI. Nested secondary screens show a back arrow instead of a close icon.

`card` and `mbway` are moved onto this mechanism: the installment picker and the country code picker open through `navigation` and close themselves after a selection.

### Progress
➡️ **Layer 1 — Secondary screen navigation (this PR)**
Layer 2 — Remove the secondary screen route from the public API (#3052)
Layer 3 — Animate nested secondary screens (#3054)

## Checklist
- [ ] Code is unit tested

## Ticket Number
N/A

### Changed
- `CheckoutPaymentMethod` now displays the secondary screens of a payment method, such as the MB Way country code picker and the Card installment picker.


core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutFullScreenDialog.kt
```diff
@@ -32,6 +33,7 @@ import com.adyen.checkout.ui.internal.theme.Dimensions
 @Composable
 internal fun CheckoutFullScreenDialog(
     onDismissRequest: () -> Unit,
+    navigationIcon: ImageVector = Icons.Default.Close,
     content: @Composable () -> Unit,
 ) {
```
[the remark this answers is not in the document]
OscarSpruit: This will be addressed in a follow up PR

core/src/main/java/com/adyen/checkout/core/components/internal/CheckoutFullScreenDialog.kt
```diff
@@ -53,7 +55,7 @@ internal fun CheckoutFullScreenDialog(
         ) {
             IconButton(onClick = onDismissRequest) {
                 Icon(
-                    imageVector = Icons.Default.Close,
+                    imageVector = navigationIcon,
                     contentDescription = resolveString(CheckoutLocalizationKey.GENERAL_CLOSE),
                     tint = CheckoutThemeProvider.colors.primary,
                 )
```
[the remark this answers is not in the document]
OscarSpruit: This will be addressed in a follow up PR

core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentMethod.kt
```diff
@@ -43,9 +56,47 @@ fun CheckoutPaymentMethod(
...
+            CheckoutFullScreenDialog(
+                onDismissRequest = { backStack = backStack.dropLast(1) },
+                navigationIcon = navigationIcon,
+            ) {
+                component.SecondaryContent(key, Modifier)
+            }
+        }
+    }
+
+    LaunchedEffect(component) {
```
araratthehero, line 89: Since the component is being used in compose, I think the `PaymentComponent` and `SecondaryScreenComponent` should be `@Stable`.

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
```diff
@@ -139,8 +143,10 @@ constructor(
             modifier = modifier,
             identifier = identifier,
             viewState = viewState,
-            onIntent = ::onIntent,
-            onDismissRequest = { eventChannel.trySend(PaymentComponentEvent.CloseSecondaryScreen) },
+            onInstallmentClick = { installment ->
+                onIntent(CardIntent.UpdateInstallment(installment))
+                navigationChannel.trySend(SecondaryNavigationEvent.Close)
+            },
```
araratthehero, line 149: An AI agents says:

> Click lambdas recreated inside SecondaryContent CardComponent.SecondaryContent and MBWayComponent.SecondaryContent create onInstallmentClick / onItemClick lambdas inline:
> ```
> onInstallmentClick = { installment ->
>     onIntent(CardIntent.UpdateInstallment(installment))
>     navigationChannel.trySend(SecondaryNavigationEvent.Close)
> }
> ```
> 
> New lambda references on every recomposition will force CardSecondaryContent / MBWaySecondaryContent (and the pickers below them) to recompose even when nothing else changed. Since these lambdas do not depend on the current composition, they could be hoisted as val properties on the component class for a stable reference.
> 

Is this something we should follow up?

core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentMethod.kt
OscarSpruit: We're on Kotlin 2.4.10 and don't override any Compose compiler feature flags, so strong skipping is active (default since the 2.0.20 compiler). That already makes composables with unstable parameters skippable, comparing those parameters by instance, so `SecondaryScreenHost` skips as it is.

Annotating would switch that comparison from instance to `equals`, and since neither interface overrides `equals` that is identity — same outcome, no codegen difference. Happy to add it purely as documentation of intent if you prefer

card/src/main/java/com/adyen/checkout/card/internal/ui/CardComponent.kt
OscarSpruit: I don't think we need to follow up. That reasoning describes the behaviour before strong skipping, where a lambda capturing an unstable value wasn't memoized. Strong skipping extends lambda memoization to lambdas capturing unstable values and it's on by default from the 2.0.20 compiler — we're on Kotlin 2.4.10. So these lambdas are remembered across recompositions and the picker composables still skip. Hoisting them to `val` properties would add boilerplate to every component and picker without changing behaviour

core/src/main/java/com/adyen/checkout/core/components/CheckoutPaymentMethod.kt
araratthehero: In this case, shouldn't we fix other places where we added `@Stable`?

OscarSpruit: The outcome is the same, so I don't think we need to fix it anywhere. I'm just trying to say it doesn't matter if we add it and imo it would pollute the codebase, so I wouldn't add it if it's not necessary

araratthehero: It's okay to not add it, but I also think we should not leave code that is unnecessary either. I will create a cleanup task.

araratthehero (approved):

PR #3054, opened 2026-09-16
v6 - Components - Animate nested secondary screens

OscarSpruit:
## Description
Animates navigation between the secondary screens of a payment method.

`CheckoutPaymentMethod` now hosts a single full screen dialog for the whole secondary screen back stack, instead of one dialog per entry, so moving between screens can be animated with `AnimatedContent`. Pushing a screen slides the next one in horizontally, popping slides it back. Each screen carries its own navigation icon, a close icon at the root and a back arrow deeper in the stack.

The composables involved are split into their own files and moved to `components.internal.ui`, next to the component contracts they build on. No public API changes.

Depends on #3052.

https://github.com/user-attachments/assets/233c0731-67cb-4c9d-b7ab-66147d743f90

\* The nested secondary screen is only for testing purposes and is not actually added to our codebase.

### Progress
✅ Layer 1 — Secondary screen navigation (#3051)
✅ Layer 2 — Remove the secondary screen route from the public API (#3052)
➡️ **Layer 3 — Animate nested secondary screens (this PR)**

## Checklist
- [ ] Code is unit tested

## Ticket Number
N/A

### Improved
- Navigating between the secondary screens of a payment method, such as pickers, is now animated.


core/src/main/java/com/adyen/checkout/core/components/internal/ui/SecondaryScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+    content: @Composable () -> Unit,
+) {
+    Column {
+        val (navigationIcon, navigationDescription) = remember {
+            if (isNested) {
+                Icons.AutoMirrored.Filled.ArrowBack to CheckoutLocalizationKey.GENERAL_BACK
+            } else {
+                Icons.Default.Close to CheckoutLocalizationKey.GENERAL_CLOSE
+            }
+        }
```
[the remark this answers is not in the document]
OscarSpruit: This is on purpose, so the icon doesn't change during the lifecycle of a screen

core/src/main/java/com/adyen/checkout/core/components/internal/ui/SecondaryScreen.kt
```diff
@@ -0,0 +1,62 @@
...
+
+@Composable
+internal fun SecondaryScreen(
+    isNested: Boolean,
+    onNavigationClick: () -> Unit,
+    content: @Composable () -> Unit,
+) {
+    Column {
+        val (navigationIcon, navigationDescription) = remember {
+            if (isNested) {
```
araratthehero, line 38: Shouldn't remember be changed to `remember(isNested)`? In this case remember won't update if `isNested` is changed.

core/src/main/java/com/adyen/checkout/core/components/internal/ui/CheckoutFullScreenDialog.kt
```diff
@@ -0,0 +1,54 @@
...
+            dismissOnClickOutside = true,
+            usePlatformDefaultWidth = false,
+            decorFitsSystemWindows = false,
+        ),
+    ) {
+        val window = (LocalView.current.parent as? DialogWindowProvider)?.window
+        SideEffect(window) {
+            // Remove the dim to make the transition look better
+            window?.setDimAmount(0f)
+        }
```
araratthehero, line 42: Is this necessary to do in every recomposition? What if you put this inside of a `DisposableEffect`?

core/src/main/java/com/adyen/checkout/core/components/internal/ui/SecondaryScreen.kt
OscarSpruit: This is on purpose, so the icon doesn't change when you press it

core/src/main/java/com/adyen/checkout/core/components/internal/ui/CheckoutFullScreenDialog.kt
OscarSpruit: This is why the `SideEffect` has the window as key. Only if the window would change it is necessary to set again

araratthehero (approved):

PR #3055, opened 2026-09-16
v6 - Card - Unify CardBrand and CardType

araratthehero:
## Description

Combines the two types representing card schemes into a single public `CardBrand`.

- Predefined brands move to the `CardBrand` companion object (`CardBrand.VISA`, `CardBrand.MASTERCARD`, ...). Names are carried over from `CardType` unchanged.
- The regexes move into a new internal `CardBrandRegexes`, keyed by `CardBrand`. It keeps `java.util.regex.Pattern` rather than Kotlin `Regex` because estimation relies on `Matcher.hitEnd()` to match a card number that is still being typed.
- `CardType` and `CardType.getByBrandName(String)` are removed.
- `CardBrand.estimate(String)` becomes `@RestrictTo(LIBRARY_GROUP)`, matching iOS where the whole estimation surface (`pattern`, `matches`, `CardBrandDetector`) is `internal`/`package`.
- Merchants pass custom schemes via `CardBrand("txVariant")`, unchanged from v5.

Detection behavior is unchanged. The first commit pins the current `estimate()` results using only string `txVariant`s, and that test file is untouched by the following commits. All 33 regexes were verified byte-identical to the removed `CardType`.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1304

### Changed
- Removed `CardType`. Use the predefined brands on `CardBrand` instead, for example `CardBrand.VISA` instead of `CardBrand(CardType.VISA)`.
- Removed `CardType.getByBrandName(String)`. Create a `CardBrand` with its `txVariant` instead, for example `CardBrand("visa")`.


core/src/test/java/com/adyen/checkout/core/common/CardBrandTest.kt
```diff
@@ -0,0 +1,56 @@
...
+        val estimatedBrands = CardBrand.estimate("4571")
+
+        assertEquals(
+            listOf(
+                CardBrand(txVariant = "cartebancaire"),
+                CardBrand(txVariant = "visa"),
+                CardBrand(txVariant = "visadankort"),
+            ),
+            estimatedBrands,
+        )
```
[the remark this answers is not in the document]
araratthehero: @gemini-code-assist Thanks for the review. I'd like to keep this assertion order-dependent, because the order of `CardBrand.estimate()` is observable behavior here rather than an implementation detail.

`CardBrandIntentsHandler` depends on it directly when building the dual-brand state:

```kotlin
// select the first 2 brands and discard the rest (should only have 2 brands normally)
val firstTwoDetectedCardTypes = supportedDetectedCardTypes.take(2)
...
// auto select the first brand
cardBrandDataList.first()
```

So the estimation order decides which two brands are offered in the dual-brand picker, and which one is auto-selected — i.e. which brand the payment is ultimately submitted with. `StoredCardComponentStateValidator` similarly relies on `selectedOrFirstCardType`.

There is also precedent that this ordering is treated as product behavior rather than incidental: on `v5`, commit `0f6b09d3c5` ("Adjust regexes for Dankort and Bcmc to make sure we prioritize them over visa and mastercard when local brands are available") changed the regexes specifically to influence this priority.

Comparing sets would drop the only coverage that catches a brand-priority regression, which is something a shopper could observe. So the ordered assertion is deliberate: if the declaration order in `REGEXES` changes, this test *should* fail and make us look at it.

I'll leave this thread open in case you or a reviewer disagrees — happy to instead document the ordering contract explicitly on `REGEXES` if that reads better.

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
```diff
@@ -8,24 +8,61 @@
...
          * Estimate all potential [CardBrands][CardBrand] for a given card number.
          *
          * @param cardNumber The potential card number.
          * @return All matching [CardBrands][CardBrand] if the number was valid, otherwise an empty [List].
          */
-        fun estimate(cardNumber: String): List<CardBrand> {
-            return CardType.entries.filter { it.isEstimateFor(cardNumber) }.map { CardBrand(txVariant = it.txVariant) }
-        }
+        @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+        fun estimate(cardNumber: String): List<CardBrand> = CardBrandRegexes.estimate(cardNumber)
```
OscarSpruit: Should we remove this function from here? We can directly use `CardBrandRegexes.estimate` internally

core/src/main/java/com/adyen/checkout/core/common/internal/helper/CardBrandDetector.kt
```diff
@@ -0,0 +1,81 @@
...
+ * The regexes of the predefined [CardBrands][CardBrand], used to estimate the brand of a card number.
+ *
+ * [Pattern] is used instead of Kotlin's [Regex] because estimation relies on [java.util.regex.Matcher.hitEnd] to also
+ * match a card number that is still being typed.
+ */
+internal object CardBrandRegexes {
+
+    private val WHITESPACE_REGEX = "\\s".toRegex()
+
+    private val REGEXES: Map<CardBrand, Pattern> = mapOf(
```
OscarSpruit, line 31: Should we maybe add a test that assert if every card brand also has a regex tied to it? This would help us to not forget about a regex if we ever add a new brand

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
```diff
@@ -8,24 +8,61 @@
...
+        val DISCOVER = CardBrand("discover")
+        val ELO = CardBrand("elo")
+        val FORBRUGSFORENINGEN = CardBrand("forbrugsforeningen")
+        val VISAALPHABANKBONUS = CardBrand("visaalphabankbonus")
+        val MCALPHABANKBONUS = CardBrand("mcalphabankbonus")
+        val HIPER = CardBrand("hiper")
+        val HIPERCARD = CardBrand("hipercard")
+        val JCB = CardBrand("jcb")
+        val OASIS = CardBrand("oasis")
+        val KARENMILLER = CardBrand("karenmillen")
```
OscarSpruit, line 39:
```suggestion
        val KAREN_MILLEN = CardBrand("karenmillen")
```

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
```diff
@@ -8,24 +8,61 @@
...
+        val CARTEBANCAIRE = CardBrand("cartebancaire")
+        val CODENSA = CardBrand("codensa")
+        val CUP = CardBrand("cup")
+        val DANKORT = CardBrand("dankort")
+        val DINERS = CardBrand("diners")
+        val DISCOVER = CardBrand("discover")
+        val ELO = CardBrand("elo")
+        val FORBRUGSFORENINGEN = CardBrand("forbrugsforeningen")
+        val VISAALPHABANKBONUS = CardBrand("visaalphabankbonus")
+        val MCALPHABANKBONUS = CardBrand("mcalphabankbonus")
```
OscarSpruit: Should we use screaming snake case here as well to separate the words? And also for `VISADANKORT`?

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
```diff
@@ -8,24 +8,61 @@
...
+ * This class represents a card scheme. Use one of the predefined brands, or the constructor with a `txVariant` to
+ * create a [CardBrand] with a scheme that is not predefined. Can be used to configure the supported card schemes with
+ * `CardConfiguration.supportedCardBrands`.
  */
 @Parcelize
 data class CardBrand(val txVariant: String) : Parcelable {
+
     companion object {
+
+        val AMERICAN_EXPRESS = CardBrand("amex")
```
OscarSpruit, line 23: Nit: having the alphabetically order would make it more readable

card/src/main/java/com/adyen/checkout/card/internal/data/api/NetworkCardBrandDetectionService.kt
```diff
@@ -15,7 +15,6 @@ import com.adyen.checkout.card.internal.data.model.DetectedCardType
 import com.adyen.checkout.card.internal.helper.isHiddenCardType
```
OscarSpruit, line 15: in `card.md` the old way is still used - could you update it there?

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
araratthehero: This is fixed in the next PR

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
araratthehero: It is fixed in the next PR

core/src/main/java/com/adyen/checkout/core/common/CardBrand.kt
araratthehero: Done for the public constants. I've left the order in CardBrandDetector though, since estimate() returns brands in map declaration order and that acts as a priority. Reordering it would change which brand gets preselected for dual branded cards, so I've added a KDoc note on the map instead.

core/src/main/java/com/adyen/checkout/core/common/internal/helper/CardBrandDetector.kt
araratthehero: I added a test and it looks like the most viable one is using reflection. You can check the last commit.

OscarSpruit: Wouldn't it be better to use `internal` on this field and annotate it with `@VisibleForTesting`?

araratthehero: Since the list is not an enum, it is limiting how we can test. Could you check the last commit and we discuss if you have a better alternative?

OscarSpruit: Oh yes, I see what you mean. I think this is good.

OscarSpruit (approved):

PR #3059, opened 2026-09-17
v6 - Drop-in - Configuration and params split

OscarSpruit:
## Description
Phase 1 of adding a Drop-in configuration: introduce `DropInConfiguration` and split Drop-in's params the way card already does. `CheckoutParams` carries the values every integration shares (shopper locale, environment, amount), `DropInParams` carries only what comes from `DropInConfiguration`.

- Adds `DropInConfiguration` with `hideStoredPaymentMethods` and `startWithLastStoredPaymentMethod`, set through the `CheckoutConfiguration.dropIn()` extension.
- Exposes `CheckoutParamsFactory` and `CheckoutContext.publicKey` to the library group, so Drop-in can build a `CheckoutParams` instead of re-implementing the shopper locale, environment and amount resolution in `DropInParamsMapper`.
- `DropInViewModel` now creates the `CheckoutParams` and derives the `DropInParams` from it. The screen view models, `NavigationStack` and `DropInActivity` read the shared values from `CheckoutParams`.
- The amount is nullable now: a missing amount no longer fails Drop-in through `error("Amount cannot not be null")`, it formats to an empty string instead.

Nothing reads the two `DropInConfiguration` flags yet, so behavior is unchanged. Wiring them is phase 2.

### Progress
➡️ **Phase 1 — Add `DropInConfiguration` and split the params (this PR)**
Phase 2 — Wire the configurations into Drop-in behavior

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1398


drop-in/src/main/java/com/adyen/checkout/dropin/internal/ui/PreselectedPaymentMethodViewModel.kt
```diff
@@ -42,7 +43,7 @@ internal class PreselectedPaymentMethodViewModel(
     }
 
     private fun createInitialViewState(storedPaymentMethod: StoredPaymentMethod): PreselectedPaymentMethodViewState {
-        val formattedAmount = dropInParams.amount.format(dropInParams.shopperLocale)
+        val formattedAmount = checkoutParams.amount?.format(checkoutParams.shopperLocale).orEmpty()
```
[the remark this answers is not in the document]
OscarSpruit: This will be fixed as part of a different ticket

drop-in/src/main/java/com/adyen/checkout/dropin/DropInConfiguration.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.adyen.checkout.core.components.internal.Configuration
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+class DropInConfiguration internal constructor(
+    val hideStoredPaymentMethods: Boolean?,
+    val startWithLastStoredPaymentMethod: Boolean?,
+) : Configuration
+
+fun CheckoutConfiguration.dropIn(
```
jreij, line 21: Can we add kdocs here?

OscarSpruit: They are added in the the next PR

araratthehero (approved):

PR #3062, opened 2026-09-17
v6 - Core - Action callback

OscarSpruit:
## Description
Adds a required `onAction` callback to the v6 advanced and sessions flows, so merchants are notified when the SDK is about to handle an action.

- New public `ActionData(type)` in `com.adyen.checkout.core.action.data`, carrying the action type as returned by the Checkout API (`redirect`, `nativeRedirect`, `threeDS2`, `sdk`, `qrCode`, `await`, `voucher`).
- `onAction` is a required parameter on `AdvancedCheckoutCallbacks` and `SessionCheckoutCallbacks`. The action-only flow does not get it: it never receives actions from a payment, the merchant passes the action in themselves.
- `ActionHandler` dispatches the callback after the action component is created and before the action is handled, so it is not invoked for an unsupported action type.
- Drop-in passes a no-op for now; forwarding it to `DropInService` needs a separate callback there.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1556

### Breaking changes
- `AdvancedCheckoutCallbacks` and `SessionCheckoutCallbacks` now require an `onAction` callback.

### New
- Added the `onAction` callback, which is invoked with the `ActionData` of the action that the SDK is about to handle.


core/src/main/java/com/adyen/checkout/core/components/internal/ActionHandler.kt
```diff
@@ -76,6 +77,8 @@ internal class ActionHandler(
             }
             .launchIn(coroutineScope)
 
+        (componentRequestDispatcher as? SubmittableComponentRequestDispatcher)?.action(ActionData(action.type))
```
araratthehero: What do you think of moving this out of here, to remove the casting too.

You could add `onAction` to `ActionHandler`, like:
```
ActionHandler(
    ..., 
    onAction: (ActionData) -> Unit = {}
)
```

Later in the `CheckoutControllerFactory` it is already known the dispatcher type. In `createFullCheckoutController` you could call the `action()`, since the dispatcher is `SubmittableComponentRequestDispatcher` and for action only you could ignore.

OscarSpruit: Good suggestion! I will just remove the default value, so if we ever add a new flow we are required to pass it

araratthehero (approved):

PR #3064, opened 2026-09-18
v6 - Currencies - Align CheckoutCurrency with canonical Adyen list

nauaros:
## Description
Aligns Android's v6 `CheckoutCurrency` enum with the canonical Adyen currency list documented in `alignment/currencies/`, based on a cross-platform audit against iOS and Web.

- Audit doc: [adyen-checkout-sdk-meta#102](https://github.com/Adyen/adyen-checkout-sdk-meta/pull/102) (`PLATFORM_DISCREPANCIES.md`)
- Companion iOS PRs: [adyen-ios#2723](https://github.com/Adyen/adyen-ios/pull/2723) (v6), [adyen-ios#2722](https://github.com/Adyen/adyen-ios/pull/2722) (v5)

### Removed
Superseded ISO 4217 codes, or currencies not supported by Adyen for processing:

| Code | Reason |
| --- | --- |
| `ANG` | Replaced by `XCG` (Curaçao/Sint Maarten redenomination) |
| `HRK` | Croatia joined the euro; no replacement needed |
| `SLL` | Replaced by `SLE`, which Android already had alongside the old code |
| `BGN` | Active ISO 4217 currency, but not supported by Adyen for processing ([docs](https://docs.adyen.com/development-resources/currency-codes)) |
| `LBP` | Same as `BGN` — active currency, not Adyen-supported |

### Added
| Code | Digits | Notes |
| --- | --- | --- |
| `XCG` | 2 | Replacement for `ANG`, missing from Android |
| `CNH` | 2 | Real Adyen-supported currency, missing entirely — see bug fix below |

### Fixed
- **`CNH` silently formatted with 0 decimals instead of 2.** It was missing from `CheckoutCurrency`, and `java.util.Currency` doesn't recognize it either, so the fallback resolved to 0 fraction digits.
- **`Amount.format(locale)` defaulted straight to 0 fraction digits** for any currency not in `CheckoutCurrency`. It now reuses `AmountFormat`'s existing `java.util.Currency`-based fallback, so currencies the platform does recognize (e.g. `CLF` at 4 digits) format correctly instead of being assumed to have 0 decimals.

### Breaking change
`CheckoutCurrency` is public, so removing entries is a breaking change for any merchant code referencing them directly. Confirmed as acceptable since v6 is still in alpha.

### Remaining work
- This PR only covers v6 (`core`). The deprecated v5 `CheckoutCurrency` in `components-core` has the same drift and needs the same fix — tracked as follow-up work, out of scope here.
- `UYW` (not recognized by `java.util.Currency` at all) is not handled by the new fallback and remains a known gap — also out of scope here.

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1060

### Breaking changes
- Removed `ANG`, `BGN`, `HRK`, `LBP` and `SLL` from `CheckoutCurrency`. `ANG` is replaced by `XCG`; `BGN` and `LBP` are not supported by Adyen for processing; `HRK` and `SLL` are retired ISO codes (`HRK` was replaced by `EUR`, `SLL` by `SLE`, which was already present).

### New
- Added `XCG` and `CNH` to `CheckoutCurrency`.

### Fixed
- Fixed `CNH` amounts silently formatting with 0 decimal digits instead of 2.
- Fixed `Amount.format(locale)` defaulting to 0 fraction digits for any currency not in `CheckoutCurrency`, instead of falling back to the platform's currency data like `AmountFormat` already does.


core/src/test/java/com/adyen/checkout/core/components/data/model/AmountTest.kt
```diff
@@ -205,6 +205,17 @@ internal class AmountTest {
...
+        @Test
+        fun `when currency is not an Adyen currency but is recognized by the platform then platform fraction digits are used`() {
+            // CLF is not an Adyen-supported currency, but the JVM recognizes it with 4 fraction digits.
+            // Falling back to 0 fraction digits here would be off by two orders of magnitude.
+            val amount = Amount(currency = "CLF", value = 1337L)
+
+            val result = amount.format(Locale.US)
+
+            assertTrue(result.contains("0.1337"), "Expected '0.1337' in '$result'")
+        }
```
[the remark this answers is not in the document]
nauaros: Thanks — parameterized the CLF fallback test across `en-US` and `fr-FR` as suggested.

Didn't add the `CNH` locale-formatting test though: `Amount.format(locale)` calls `Currency.getInstance(currencyCode)` unconditionally to build the `DecimalFormat`, and the JVM's `java.util.Currency` doesn't recognize `CNH` at all (confirmed on OpenJDK 21 — throws `IllegalArgumentException`). So `amount.format(Locale.US)` for `CNH` still falls into the `runCatching` catch-all and returns `""`, regardless of this fix. That's a separate, pre-existing limitation of `Amount.format` (unrelated to the `CheckoutCurrency`/`AmountFormat` fallback fixed here) — added a test for it in the internal `AmountFormat.toBigDecimal` path instead (`AmountFormatTest.kt`), which doesn't depend on `java.util.Currency` recognizing the code.

core/src/test/java/com/adyen/checkout/core/common/internal/helper/AmountFormatTest.kt
```diff
@@ -0,0 +1,55 @@
...
+        // platform fallback, otherwise it silently falls back to zero fraction digits.
+        val amount = Amount(currency = "CNH", value = 1337L)
+
+        val result = AmountFormat.toBigDecimal(amount)
+
+        assertEquals(BigDecimal.valueOf(1337L, 2), result)
+    }
+
+    @Test
+    fun `when currency is not an Adyen currency but is recognized by the platform then platform fraction digits are used`() {
```
jreij, line 38: Is this the behavior we want? Web falls back to `2` and never relies on the system if I'm not mistaken

nauaros: As discussed during our session

core/src/main/java/com/adyen/checkout/core/common/internal/helper/CheckoutCurrency.kt
jreij, line 171: @OscarSpruit we should probably remove this completely

OscarSpruit: Are you sure? I think this is still useful for advanced flow, so we fail early if they use an unsupported currency

jreij: Ah I meant when we go sessions only

OscarSpruit: We agreed to remove this once we remove advanced flow

OscarSpruit (approved):

PR #3071, opened 2026-09-18
v6 - Core - Remove the navigation API

OscarSpruit:
## Description
Removes the checkout navigation API. With the `onAction` callback in place, nothing in the SDK needs a route flow
anymore: the merchant learns about an action from the callback and renders `CheckoutAction` themselves, exactly like
Drop-in already does.

- Deleted `CheckoutPaymentFlow`, `CheckoutRoute` and `CheckoutController.navigation`, plus the internal
  `CheckoutFlow.navigation` and its three implementations.
- `CheckoutPaymentMethod` and `CheckoutAction`, both already public, are now the rendering entry points. Their KDoc
  describes when to switch from one to the other.
- The example app switches on an `isHandlingAction` flag in `V6UiState.Component`, set from `onAction` and reset when
  another payment method is selected.
- `MIGRATION.md`, `docs/v6/README.md`, `card.md`, `card-advanced-flow.md`, `card-session-flow.md` and `theme.md` moved to
  this pattern. `docs/v6/README.md` now states the contract explicitly: an action does not progress until
  `CheckoutAction` is rendered.
- `ADR-0002` referenced `CheckoutRoute.Action` when describing the implicit host contract; it now refers to `onAction`.

The `core.api` diff is 19 deletions and no additions.

### Known gap
`CheckoutPaymentFlow` restored the action screen across configuration changes, seeded from its internal
`actionComponent`. Merchants cannot reproduce that seeding, because `CheckoutController.actionComponent` is internal and
`CheckoutAction` renders nothing when a recreated controller has no action component. Keeping the flag in a `ViewModel`,
as the example app does, covers configuration changes but not process death. Process death is already an open TODO in
`PaymentMethodViewModel`, so this is worth folding into that work rather than solving here.

### Progress

Phase 1 — `onAction` callback in core (#3062, merged)
Phase 2 — Drop-in navigates from `onAction` (#3070)
**Phase 3 — remove the navigation API (this PR)**

## Checklist
- [x] Code is unit tested

## Ticket Number
COSDK-1556

### Breaking changes
- Removed `CheckoutPaymentFlow`. Render `CheckoutPaymentMethod` while the shopper provides input, and `CheckoutAction`
  once the `onAction` callback reports that an action is being handled.
- Removed `CheckoutController.navigation` and `CheckoutRoute`. Use the `onAction` callback to decide which composable to
  render.


docs/v6/card-session-flow.md
```diff
@@ -33,8 +33,8 @@ lifecycleScope.launch {
                 target = CheckoutTarget.PaymentMethod(PaymentMethodTypes.SCHEME),
                 context = result.checkoutContext,
                 callbacks = SessionCheckoutCallbacks(
-                    onAction = { actionData ->
-                        displayAction(actionData.type)
+                    onAction = {
+                        isHandlingAction = true
```
jreij, line 37: If this is just an example of how they could _save the state_ of the flow, can we add a comment to clarify that this is just sample code for their convenience?

